PR #30, opened 2021-12-17
Sound GUI

lazynessmind:
### Done:

- New setting button that opens the SoundGUI with an option to turn on or off the music.

PS: I tried to implement a slider with the music gain in mind, but the current GUI API doesn't allow float numbers, and the current music implementation doesn't have much flexibility to hack on, so I made a decision to don't mess with it since I didn't know if it would break the fade in and fade-out effects.

image

image


IntegratedQuantum (approved):
Seems good.
I don't know how to work with the gain either. For now your attempt seems like a good alternative.
@zenith391 made the music stuff, so he can probably explain how that works or make the gain more accessible.

PR #35, opened 2021-12-22
ConsoleGUI improvements.

lazynessmind:
## Done:

Quality of Life improvements:

- Close when Enter is pressed.
- Simulate left-click on TextInput when focusing the TextInput. (this makes the cursor to appear when you open the ConsoleGUI)

wfr


IntegratedQuantum (approved):
Seems good.

Although if we have a chat in the future, closing by pressing Enter might be a bad idea. But I think we should make a unique GUI for that anyways.

PR #36, opened 2021-12-23
Prevent multi-button highlight

LordDeatHunter:
This Un-highlights clicked buttons when moved away.
Before, if you clicked on a button (without releasing the left mouse button), and moved the cursor away from the button, it would stay highlighted. Going over multiple buttons would highlight all of them as pressed.


IntegratedQuantum (approved):
Looks good. I like that you simplified the logic while also fixing the bug.

PR #38, opened 2021-12-23
Russian language support

UltraQbik:


IntegratedQuantum (changes requested):
Your file is named `en_RU.lang`.
The first part of the name stands for the language while the second part stands for the accent (like `en_US` and `en_UK` are both english, but different accents)
So the correct file name for russian would be `ru_RU.lang`.

PR #43, opened 2022-01-04
GUI: Center GUI's on screen.

lazynessmind:
### Done:

- Move all the inventory screens to the middle of the screen.

### Example: (before and after)
image

image


IntegratedQuantum (approved):
Good idea. That also ~~hides~~ fixes the tooltip bugs in the bottom-most row.

Seeing that `GeneralInventory` uses a lot of `Container` functionality it might be useful to consider making it a `Container` to give mods some flexibility in positioning inventories.

PR #45, opened 2022-01-11
Fix: Custom ore .lang file.

lazynessmind:
### Done:

- Implementation of the Discord idea on the translation of custom ores.
image

### How it works:

- When the world is created a language file is created with all the world custom ore translations. This file is then used as a fallback when the translation key isn't found in the main language in use. 

#### Example:
```properties
#Contains all the translated names for the generated ores.
block.cubyz.Wyartite_ore.name=Wyartite Ore
block.cubyz.Ekanksite_ore.name=Ekanksite Ore
block.cubyz.Vivinella_ore.name=Vivinella Ore
block.cubyz.Derssulan_ore.name=Derssulan Ore
block.cubyz.Yeliotil_ore.name=Yeliotil Ore
block.cubyz.Evesuvite_ore.name=Evesuvite Ore
block.cubyz.Kamackiter_ore.name=Kamackiter Ore
block.cubyz.Geridymium_ore.name=Geridymium Ore
```

### DIFF
--- a/src/cubyz/api/CurrentWorldRegistries.java
+++ b/src/cubyz/api/CurrentWorldRegistries.java
@@ -53,11 +53,20 @@ public class CurrentWorldRegistries {
 		assets.mkdirs();
 		new File(assets, "blocks/textures").mkdirs();
 		new File(assets, "items/textures").mkdirs();
+		new File(assets, "lang").mkdirs();
+		Properties oreLang = new Properties();
 		Random rand = new Random(world.getSeed());
 		int randomAmount = 9 + rand.nextInt(3); // TODO
 		int i = 0;
 		for(i = 0; i < randomAmount; i++) {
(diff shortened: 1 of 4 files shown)


IntegratedQuantum (changes requested):
Looks good.
But I don't like the name. You are refering to it as `oreLang` and `en_ore.lang`.
In the future there might be other per-world assets, like for example server specific assets made by the server owner or just other procedural content.
That's why I'd suggest calling it `fallbackLang` and `fallback.lang` to make the actual use of that file more clear.

lazynessmind:
So the comment on the file should be removed as well or changed?

image

IntegratedQuantum:
That can stay. Right now it is correct and later new comments can be added to split the file into sections.

lazynessmind:
Done in e755df6

PR #46, opened 2022-01-12
Button Text: Current implementation.

lazynessmind:
## Idea:

- Clip the text that is bigger than the button width and make it scroll on the X-axis to make it readable.

This PR is mostly a Draft to resolve the text being bigger than the button width, if you have a suggestion feel free to speak about it. 

This could be solved by increasing the button width, but that can break the rest of the current GUI.

**PS:** Code isn't final and it's hacky for now. The clean-up is the last process.

## The current implementations are:

1. Reset X to origin X
2. Reset X to origin X + Button Width
3. -X if the most left part isn't visible, +X if the most right side isn't visible.

v1
v2
v3

### DIFF
--- a/src/cubyz/gui/components/Button.java
+++ b/src/cubyz/gui/components/Button.java
@@ -1,10 +1,12 @@
 package cubyz.gui.components;
 
-import cubyz.utils.Logger;
 import cubyz.gui.input.Mouse;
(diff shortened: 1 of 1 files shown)


lazynessmind:
Changed the `v3` to go slower and have some delay between x-scroll.
v4

PR #340, opened 2024-05-01
New coloured blocks palette, Chalk and glass

careeoki:
Replaces the current "colored blocks" palette with a new, more balanced and versatile one. There are now a total of 20 colors:

color palette
Glass has been updated with these new colours.

A new opaque coloured block, Chalk, has been added. It has a subtle, low-contrast texture and is designed for building. It can be chiseled.
Chalk could be found in warm oceans and a chalk cliffs biome.
kirby
Kirby made of Pink, Magenta, and Black chalk, sitting on top of "default" White chalk

(i'm sure there's a better way to delete files but i do not know what it is. i believe that i formatted everything correctly, apologies if i missed something


IntegratedQuantum:
The new building blocks are here!

> i believe that i formatted everything correctly, apologies if i missed something

I've setup an automated system recently. Github should automatically send you an e-mail if the formatting is wrong. But as you can see at the bottom of the PR "All checks have passed", so everything is correctly formatted.

> i'm sure there's a better way to delete files but i do not know what it is.

Well the best way would be to learn how to use git on your desktop instead of using the web interface.
You could for example try the github desktop app https://desktop.github.com/.

careeoki:
Committed the new chalk colours.
chalk desat

PR #432, opened 2024-06-02
Add ability to add stripes to biomes in the .json format

OneAvargeCoder193:


OneAvargeCoder193:
Some examples:
Screenshot 2024-06-02 005424
Screenshot 2024-06-02 093647

src/server/terrain/biomes.zig
```diff
@@ -50,6 +52,72 @@ const StructureModel = struct {
...
+		if (parameters.JsonObject.get("width")) |width| {
+			minwidth = width.as(f64, 0);
+			maxwidth = width.as(f64, 0);
+		} else {
+			minwidth = parameters.get(f64, "minWidth", 0);
+			maxwidth = parameters.get(f64, "maxWidth", 0);
+		}
+
+		return Stripe {
+			// .{dx, dy, dz},
```
IntegratedQuantum, line 109: This comment looks like a leftover from previous code. Please remove it.

src/server/terrain/biomes.zig
```diff
@@ -50,6 +52,72 @@ const StructureModel = struct {
...
+	direction: ?Vec3d,
+	block: u16,
+	mindistance: f64,
+	maxdistance: f64,
+	minoffset: f64,
+	maxoffset: f64,
+	minwidth: f64,
+	maxwidth: f64,
+
+	pub fn initStripe(parameters: JsonElement) ?Stripe {
```
IntegratedQuantum, line 65: This function never returns null, so maybe it should not return an optional.

Also the zig convention is to call these functions just `init`.

src/server/terrain/biomes.zig
```diff
@@ -153,12 +222,23 @@ pub const Biome = struct {
...
+
+		const stripes = json.getChild("stripes");
+		var stripeList = main.ListUnmanaged(Stripe){};
+		defer stripeList.deinit(main.stackAllocator);
+		for (stripes.toSlice()) |elem| {
+			if (Stripe.initStripe(elem)) |stripe| {
+				stripeList.append(main.stackAllocator, stripe);
+			}
+		}
+		self.stripes = main.globalAllocator.dupe(Stripe, stripeList.items);
```
IntegratedQuantum, line 234: Instead of using a list, you can intialize the slice directly with `stripes.toSlice().len`
This is possible since Stripe.initStripe() never returns null.

IntegratedQuantum (changes requested):
Thanks for making this, I think this a great addition to the generation system.

There are a couple of cosmetic changes and some performance improvements that I would like you to make before I merge this.

IntegratedQuantum:
Looks good, thank you!

PR #435, opened 2024-06-03
Terrain Updates

OneAvargeCoder193:
# Stalagmite Caves:
Screenshot 2024-06-02 202809

# Lakes:
Screenshot 2024-06-02 185920

# Upgraded Trees:
Screenshot 2024-06-02 194119

# Cliffs:
Screenshot 2024-06-02 134031
Screenshot 2024-06-02 152721
Screenshot 2024-06-02 195431


src/server/terrain/mapgen/MapGenV1.zig
```diff
@@ -67,7 +67,7 @@ pub fn generateMapFragment(map: *MapFragment, worldSeed: u64) void {
 	// A ridgid noise map to generate interesting mountains.
 	const mountainMap = Array2D(f32).init(main.stackAllocator, scaledSize, scaledSize);
 	defer mountainMap.deinit(main.stackAllocator);
-	RandomlyWeightedFractalNoise.generateSparseFractalTerrain(map.pos.wx, map.pos.wy, 64, worldSeed ^ 6758947592930535, mountainMap, map.pos.voxelSize);
+	RandomlyWeightedFractalNoise.generateSparseFractalTerrain(map.pos.wx, map.pos.wy, 128, worldSeed ^ 6758947592930535, mountainMap, map.pos.voxelSize);
```
IntegratedQuantum, line 70: This change, as well as the new biomes, change the terrain generation.
This breaks my current world in a bad way:
Screenshot at 2024-06-03 11-32-48
Screenshot at 2024-06-03 11-33-45
To prevent this, we will have to wait on #360 #350 and #349 before merging this pull request.

src/server/terrain/structures/Lake.zig
```diff
@@ -0,0 +1,88 @@
+const std = @import("std");
```
IntegratedQuantum, line 1: This lake has a bunch of bugs:
Caves can be seen through the bottom of a lake:
Screenshot at 2024-06-03 11-31-56
The lake interacts badly with other structures:
Screenshot at 2024-06-03 11-55-34
Screenshot at 2024-06-03 11-55-49
Screenshot at 2024-06-03 11-56-00
The lake itself sometimes cuts off at a chunk border (note that this is not #203, since it breaks close to the center of the lake):

Screenshot at 2024-06-03 11-57-04


I think it might be better to do structures like this during map generation. This would require some more work though, like introducing a new structure system for MapGen structures.

IntegratedQuantum (commented):
I like that your are trying to add so many new terrain generation features. However there are a couple of problems here (not only in your code).

First up: This Pull Request is doing too many different things at once.
Some parts of this Pull Request cannot be merged before some other conditions are met. So I suggest splitting this PR into smaller ones, one for each structure/biome. Then we can easily merge the changes that don't have other requirements.

careeoki:
I wouldn't name the block for stalagmites "Stalagmite"
I suggest Limestone.

updated texture:
limestone

PR #445, opened 2024-06-04
Show task performance

archbirdplus:
image

Notes:
- Are the categories properly assigned?
- Rename TaskType to TaskClass?
- Which time units to use?
- Numbers fluctuate a lot -> #438 
- "Wouldn't this make adding new tasks more difficult?" -> just put `.misc` if the task doesn't take a significant amount of time
- Mesh generation is insanely slow -> #444


archbirdplus:
Looks like my "meshgen" was just more lighting under the hood. Updated to reflect that.

IntegratedQuantum:
It crashes with division by zero error when I try it:
```
thread 6678 panic: division by zero
/home/mint/Downloads/Cubyz/src/gui/windows/debug.zig:52:97: 0x1330e34 in render (Cubyzig)
   draw.print("    " ++ name ++ " time: {} ms ({} µs/task)", .{@divFloor(perf.utime[i], 1000), @divFloor(perf.utime[i], perf.tasks[i])}, 0, y, 8, .left);
                                                                                                ^
```

archbirdplus:
Is my divFloor different from yours?

IntegratedQuantum:
I don't know. Maybe you didn't test it in debug mode?

archbirdplus:
Building in debug mode takes extra steps because zig ld crashes.

archbirdplus:
Does it crash in release mode?

IntegratedQuantum:
> Building in debug mode takes extra steps because zig ld crashes.

Could you send me the issue link? Then we can add it to #308

IntegratedQuantum:
It also crashes in release, looking at gdb it seems to be the same issue:
```gdb
Thread 1 "Cubyzig" received signal SIGFPE, Arithmetic exception.
0x000000000110878c in windows.debug.render () at /home/mint/Downloads/Cubyz/src/gui/windows/debug.zig:52
52				draw.print("    " ++ name ++ " time: {} ms ({} µs/task)", .{@divFloor(perf.utime[i], 1000), @divFloor(perf.utime[i], perf.tasks[i])}, 0, y, 8, .left);
```

archbirdplus:
Resolved both complaints.
Zig issue: https://github.com/ziglang/zig/issues/17424

IntegratedQuantum:
It's really hard to read anything here. The numbers seem to get reset every frame:

https://github.com/PixelGuys/Cubyz/assets/43880493/8fa1abdf-758a-40bb-9b45-14812d76296f

src/gui/windows/debug.zig
```diff
@@ -38,6 +39,19 @@ pub fn render() void {
 		y += 8;
 		draw.print("Queue size: {}", .{main.threadPool.queueSize()}, 0, y, 8, .left);
 		y += 8;
+		const perf = main.threadPool.getPerformance();
+		const values = comptime std.enums.values(TaskType);
+		inline for(values) |t| {
+			const name = switch (t) {
```
IntegratedQuantum, line 45: Why not just use `@tagName`?

archbirdplus:
In my defense, it is quite legible at 10 fps.

src/utils.zig
```diff
@@ -985,6 +985,12 @@ pub fn BlockingMaxHeap(comptime T: type) type {
 }
 
 pub const ThreadPool = struct {
+	pub const TaskType = enum(usize) {
+		chunkgen,
+		lighting,
+		misc,
+		taskTypes,
```
IntegratedQuantum, line 992: Instead of having a C-style enum length field, I think it would be better to use `std.enums.directEnumArrayLen` to determine the length instead.
This would also make it possible to use `@tagName` above.

src/utils.zig
```diff
@@ -999,6 +1005,23 @@ pub const ThreadPool = struct {
 		isStillNeeded: *const fn(*anyopaque, milliTime: i64) bool,
 		run: *const fn(*anyopaque) void,
 		clean: *const fn(*anyopaque) void,
+		taskType: TaskType = .misc,
+	};
+	pub const Performance = struct {
+		mutex: std.Thread.Mutex = .{},
```
IntegratedQuantum, line 1011: I think this would be a great place to use atomics with monotonic ordering instead of locking.

IntegratedQuantum (changes requested):
A couple of ideas how to improve it. Also zig reported a memory leak:
```
memory address 0x7fa9aed53040 leaked: 
/home/mint/Downloads/Cubyz/src/utils.zig:627:31: 0x13250c2 in create__anon_13927 (Cubyzig)
  return self.allocator.create(T) catch unreachable;
                              ^
/home/mint/Downloads/Cubyz/src/utils.zig:1040:35: 0x13152e5 in init (Cubyzig)
   .performance = allocator.create(Performance),
                                  ^
/home/mint/Downloads/Cubyz/src/main.zig:356:36: 0x1313028 in main (Cubyzig)
 threadPool = utils.ThreadPool.init(globalAllocator, @max(1, (std.Thread.getCpuCount() catch 4) -| 1));
                                   ^
/home/mint/Cubyz/compiler/zig/lib/std/start.zig:501:22: 0x1312b49 in main (Cubyzig)
            root.main();
                     ^
```

src/gui/windows/debug.zig
archbirdplus: To separate source code and rendering options. Suppose we added localization?

src/utils.zig
archbirdplus: Sometimes I wonder if Zig devs have a naming scheme.

src/gui/windows/debug.zig
IntegratedQuantum: Debug info does not need to be translated, since it shouldn't be used by the majority of players.

archbirdplus: I suppose that will be a secondary concern after we get any localization in place.

src/renderer/chunk_meshing.zig
```diff
@@ -633,6 +633,7 @@ pub const ChunkMesh = struct {
 			.isStillNeeded = @ptrCast(&isStillNeeded),
 			.run = @ptrCast(&run),
 			.clean = @ptrCast(&clean),
+			.taskType = .lighting,
```
IntegratedQuantum, line 636: Also I just noticed you have three .lighting tasks, but they do completely different things.
Having multiple tasks with the same tag can hide performance problems. For example, this LightRefreshTask is relatively fast, which can skew the average mesh generation time.

archbirdplus: It wouldn't be hard to change. Is there a better classification? I literally just classified it based on my impressions of the performance problems.

IntegratedQuantum: I would just put the LightRefreshTaks and the LightMapLoadTask under .misc, they shouldn't take much performance.

IntegratedQuantum:
This PR has been stale for 4 months now. Do you still intend to finish this?
I think this would still be a useful addition for profiling things more quickly.

archbirdplus:
Didn't realize it was this old already. Sure, I'll finish it up.

archbirdplus:
There are actually some controversial changes here.
I added a global `frameCount` to only update the performance on screen every 10 frames.
I'm also not very experienced with atomics, and I don't think my `readAndReset` is particularly sound. Maybe I should read the whole struct atomically by packing it with `@Vector`s? Maybe I should have an array of `{count, utime}` tuples instead?

src/gui/windows/debug.zig
```diff
@@ -61,6 +62,16 @@ pub fn render() void {
...
 		draw.print("Queue size: {}", .{main.threadPool.queueSize()}, 0, y, 8, .left);
 		y += 8;
+		if (main.frameCount % main.perfUpdateFrequency == 0)
+			main.lastPerformance = main.threadPool.performance.readAndReset();
+		const perf = main.lastPerformance;
+		const values = comptime std.enums.values(TaskType);
+		inline for(values) |t| {
+			const name = @tagName(t);
+			const i = @intFromEnum(t);
+			draw.print("    " ++ name ++ " time: {} ms/frame ({} µs/task)", .{@divFloor(perf.utime[i], 1000*main.perfUpdateFrequency), @divFloor(perf.utime[i], @max(1, perf.tasks[i]))}, 0, y, 8, .left);
```
IntegratedQuantum, line 72: Better use `draw.print("    {s} time: ...", .{name, ...})` to reduce the comptime code generation (format strings are passed at comptime)

src/gui/windows/debug.zig
```diff
@@ -61,6 +62,16 @@ pub fn render() void {
 		y += 8;
 		draw.print("Queue size: {}", .{main.threadPool.queueSize()}, 0, y, 8, .left);
 		y += 8;
+		if (main.frameCount % main.perfUpdateFrequency == 0)
+			main.lastPerformance = main.threadPool.performance.readAndReset();
```
IntegratedQuantum, line 66: Is there any reason not to just use the moving average and never resetting?
I think with a moving average it should be fine to update the value every frame even.

src/gui/windows/debug.zig
```diff
@@ -61,6 +62,16 @@ pub fn render() void {
...
+		const perf = main.lastPerformance;
+		const values = comptime std.enums.values(TaskType);
+		inline for(values) |t| {
+			const name = @tagName(t);
+			const i = @intFromEnum(t);
+			draw.print("    " ++ name ++ " time: {} ms/frame ({} µs/task)", .{@divFloor(perf.utime[i], 1000*main.perfUpdateFrequency), @divFloor(perf.utime[i], @max(1, perf.tasks[i]))}, 0, y, 8, .left);
+		y += 8;
```
IntegratedQuantum, line 73: incorrect indentation

src/main.zig
```diff
@@ -648,6 +656,7 @@ pub fn main() void { // MARK: main()
 			gui.openWindow("main");
 			audio.setMusic("cubyz:cubyz");
 		}
+
```
IntegratedQuantum, line 659: unintended newline?

src/renderer/mesh_storage.zig
```diff
@@ -885,6 +885,7 @@ pub const MeshGenerationTask = struct { // MARK: MeshGenerationTask
 		.isStillNeeded = @ptrCast(&isStillNeeded),
 		.run = @ptrCast(&run),
 		.clean = @ptrCast(&clean),
+		.taskType = .lighting,
```
IntegratedQuantum, line 888: I think "meshgenAndLighting" would be more fitting.

src/utils.zig
```diff
@@ -1024,6 +1030,39 @@ pub const ThreadPool = struct { // MARK: ThreadPool
 		isStillNeeded: *const fn(*anyopaque, milliTime: i64) bool,
 		run: *const fn(*anyopaque) void,
 		clean: *const fn(*anyopaque) void,
+		taskType: TaskType = .misc,
+	};
+	pub const Performance = struct {
+		tasks: [taskTypes]u32,
+		utime: [taskTypes]i64,
```
IntegratedQuantum, line 1037: Use the `Atomic` wrapper type from std to denote atomic variables.

src/utils/.vimrc
```diff
@@ -0,0 +1 @@
+set noexpandtab
```
IntegratedQuantum, line 1: Huh?

src/utils/list.zig
```diff
@@ -433,7 +433,7 @@ pub fn ListUnmanaged(comptime T: type) type {
 	};
 }
 
-const pageSize = 4096;
+const pageSize = 4 << 10;
```
IntegratedQuantum, line 436: unintended/unrelated change?

IntegratedQuantum (changes requested):
> I'm also not very experienced with atomics, and I don't think my readAndReset is particularly sound. Maybe I should read the whole struct atomically by packing it with @Vectors? Maybe I should have an array of {count, utime} tuples instead?

Yeah, your atomics are definitely incorrect here.

Sadly bigger atomics don't seem to work very well, see https://github.com/ziglang/zig/issues/14235 (although it might be worth checking if this still happens)
The solution in this case would probably just be to use a Mutex instead of atomics.

Or I guess just don't reset the values.

src/gui/windows/debug.zig
archbirdplus: Are we certain performance will not vary over time? I can imagine different biomes, especially colorado, being significantly different.

src/renderer/mesh_storage.zig
archbirdplus: Is it just MeshGenerationTask that needs to be added to meshgenAndLighting?

src/utils/.vimrc
archbirdplus: Not sure why it's committed. But it tells vim to use tab characters.

src/gui/windows/debug.zig
IntegratedQuantum: Hmm, that is probably true.
The reason I prefer not having it reset is because averaging over 10000s of requests can yield more comparable performance numbers.

So maybe a manual reset would be best. Could either add a separate button, or just tie it to the debug window visibility.

src/renderer/mesh_storage.zig
IntegratedQuantum: I would say that no two tasks should have the same type (except from misc), otherwise the information becomes less useful.

src/gui/windows/debug.zig
```diff
@@ -61,6 +66,14 @@ pub fn render() void {
 		y += 8;
 		draw.print("Queue size: {}", .{main.threadPool.queueSize()}, 0, y, 8, .left);
 		y += 8;
+		const perf = main.threadPool.performance.read();
+		const values = comptime std.enums.values(TaskType);
+		for(values) |t| {
+			const name = @tagName(t);
+			const i = @intFromEnum(t);
+			draw.print("    {s} time: {} ms/frame ({} µs/task)", .{name, @divFloor(perf.utime[i], 1000), @divFloor(perf.utime[i], @max(1, perf.tasks[i]))}, 0, y, 8, .left);
```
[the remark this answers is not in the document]
IntegratedQuantum: On that note, is this is even a useful metric? The time per tasks seems rather unrelated to the time per frame. Coupling the two seems weird.

src/utils.zig
```diff
@@ -1024,6 +1030,46 @@ pub const ThreadPool = struct { // MARK: ThreadPool
...
+
+		pub fn read(self: *Performance) Performance {
+			var copy = Performance{.tasks = undefined, .utime = undefined};
+			self.mutex.lock();
+			defer self.mutex.unlock();
+			for(0..taskTypes) |i| {
+				copy.tasks[i] = self.tasks[i];
+				copy.utime[i] = self.utime[i];
+			}
+			return copy;
```
IntegratedQuantum, line 1071: Why not just `return self.*;`? That should automatically copy the entire thing.

archbirdplus: Oh, thanks.

src/gui/windows/debug.zig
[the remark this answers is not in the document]
archbirdplus: Both are useful- in determining the effect on the user versus benchmarking the task in isolation. Maybe there is a clearer way to format that?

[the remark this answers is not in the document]
IntegratedQuantum: But the framerate could fluctuate. Like if you look in a different direction, then the ms/frame metric will be different, even though the actual task performance is the same.
Maybe milliseconds per second would be a more stable unit? But even then the number will get continuously smaller if you don't have any tasks generating. So it still seems kind of useless to me for doing actual measurements.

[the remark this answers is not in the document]
archbirdplus: Well, it does at least tell you where to look. If some tasks are very expensive but very rare, it might be worth checking how it is overall.
Maybe we should use %time spent on this task rather than ms/s? As a proxy to cpu usage? In any case I believe it _should_ approach zero when the tasks decrease, to signal that the developer need not optimize it further for the current scene.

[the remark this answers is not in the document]
IntegratedQuantum: Oh, then I think percentage of total task time (sum over the times of all task types) would be best. Then the percentages would always add up to 100%, giving us a clearer picture than the total system time would.

src/gui/windows/debug.zig
```diff
@@ -61,6 +67,21 @@ pub fn render() void {
 		y += 8;
 		draw.print("Queue size: {}", .{main.threadPool.queueSize()}, 0, y, 8, .left);
 		y += 8;
+		const perf = main.threadPool.performance.read();
+		const values = comptime std.enums.values(TaskType);
+		var totalUtime: i64 = 0;
+		for(values) |task|
+			totalUtime += perf.utime[@intFromEnum(task)];
+		draw.print("Total task time: {} µs", .{totalUtime}, 0, y, 8, .left);
```
IntegratedQuantum, line 75: To me printing the total time seems kind of useless. It's just kind of a random number.

IntegratedQuantum (changes requested):
When I run this I get a memory leak:
```
memory address 0x756fcc252040 leaked: 
/home/mint/Downloads/archbirdplus/src/utils.zig:627:31: 0x13ab6a2 in create__anon_15345 (Cubyzig)
  return self.allocator.create(T) catch unreachable;
                              ^
/home/mint/Downloads/archbirdplus/src/utils.zig:1058:33: 0x13499d6 in init (Cubyzig)
   const self = allocator.create(Performance);
                                ^
/home/mint/Downloads/archbirdplus/src/utils.zig:1083:35: 0x1332672 in init (Cubyzig)
   .performance = Performance.init(allocator),
                                  ^
/home/mint/Downloads/archbirdplus/src/main.zig:519:36: 0x132c8e2 in main (Cubyzig)
 threadPool = utils.ThreadPool.init(globalAllocator, settings.cpuThreads orelse @max(1, (std.Thread.getCpuCount() catch 4) -| 1));
                                   ^
/home/mint/Cubyz/compiler/zig/lib/std/start.zig:608:22: 0x132bfa9 in main (Cubyzig)
            root.main();
                     ^

Memory leak
```

PR #527, opened 2024-06-26
Mahogany trees, updated jungle (somewhat)

careeoki:
Adds mahogany trees, for the jungle!
The jungle is still "(TODO)" since there's only so much I can do with the current biome parameters (specifically it needs #279, #519, and its own grass colour, which we haven't decided how to go about doing yet)
image
image
image


IntegratedQuantum:
To be honest the jungle looks pretty similar to the regular forest, due to using similar colors for the log and the leaves.
Could you maybe make the leaves more saturated and give the stem a mossy/overgrown look?

careeoki:
> To be honest the jungle looks pretty similar to the regular forest, due to using similar colors for the log and the leaves. Could you maybe make the leaves more saturated and give the stem a mossy/overgrown look?

I can try that.

careeoki:
> I personally think the trees should be taller, but there is no way to do that without also making giant leaf orbs that lag heavily.

Yes, they absolutely should be, but it's not possible at the moment.

careeoki:
New colours:
image
The log has small amounts of green in it. I found that adding more obvious moss makes it tile poorly and generally look ugly.
image
image

PR #550, opened 2024-07-07
Update biome structure chance, Ice and Frost

careeoki:
Trees in forests and jungles generate the correct amount again. Fallen logs no longer use log_tops, which i think sells the fallen log look much better.
Mahogany trees are taller, and jungles have bushy trees on the ground.
image
Default cave biome has a smooth stone floor (instead of gravel-stone floor), which i think looks much nicer and is easier on the eyes. Gravel is still present as ground_patches.
Marble caves are less common.
image
The current ice texture has been moved to a new block, Frost, to accommodate a new transparent ice block.
image
Glacier using both icy blocks.
image-1
Peak with little ice spikes.
image


IntegratedQuantum:
While it does look nice to have a thin layer of ice around caves like this, I also think that it breaks in too many cases because you can just look through the entire mountain.
Screenshot at 2024-07-08 12-04-27
I guess I should try to make a generation option for thin layers next to open air.

careeoki:
I reverted it to use frost for now.

assets/cubyz/biomes/forest.json
```diff
@@ -22,7 +22,7 @@
 			"leaves" : "cubyz:oak_leaves",
 			"log" : "cubyz:oak_log",
 			"top" : "cubyz:oak_top",
-			"chance" : 0.03,
+			"chance" : 0.5,
```
IntegratedQuantum, line 25: Is it intentional that you move from 3:1 oak:birch to 10:1?

careeoki: No, I just didn't do the math. fixed it.

PR #665, opened 2024-09-03
Add View Bobbing to Camera

AlexDavies8:
View bobbing can make movement feel more natural. I've added view bobbing to the camera while walking on the ground. The strength of it can be changed in the graphics section of the settings menu.

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -295,6 +295,9 @@ pub const Player = struct { // MARK: Player
 	pub var eyePos: Vec3d = .{0, 0, 0};
 	pub var eyeVel: Vec3d = .{0, 0, 0};
 	pub var eyeStep: @Vector(3, bool) = .{false, false, false};
+    pub var bobTime: f64 = 0;
+    pub var bobVel: f64 = 0;
+    pub var bobVec: Vec3d = .{0, 0, 0};
 	pub var id: u32 = 0;
 	pub var isFlying: Atomic(bool) = Atomic(bool).init(false);
 	pub var isGhost: Atomic(bool) = Atomic(bool).init(false);
@@ -342,7 +345,7 @@ pub const Player = struct { // MARK: Player
 	pub fn getEyePosBlocking() Vec3d {
 		mutex.lock();
 		defer mutex.unlock();
-		return eyePos + super.pos + desiredEyePos;
+		return eyePos + super.pos + desiredEyePos + bobVec;
 	}
 
 	pub fn getVelBlocking() Vec3d {
@@ -665,6 +668,22 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 			Player.selectedSlot = @intCast(@mod(newSlot, 12));
 			main.Window.scrollOffset = 0;
 		}
+        
+        // View Bobbing
+        if (Player.onGround) { // No view bobbing in the air
+            // Smooth lerping of bobTime with framerate independent damping
+            const fac: f64 = 1 - std.math.exp(-10 * deltaTime);
+            var targetBobVel: f64 = 0;
+            if (movementSpeed > 0) {
+                targetBobVel = vec.dot(movementDir * @as(Vec3d, @splat(std.math.log2(movementSpeed / 4 + 1))), forward);
+            }
+            Player.bobVel = Player.bobVel * (1 - fac) + targetBobVel * fac;
+            Player.bobTime += std.math.pow(f64, Player.bobVel, 0.7) * 8 * deltaTime;
+        }
+        const bobStrength: f64 = Player.bobVel * 0.04 * settings.viewBobStrength;
+        const xBob = std.math.cos(Player.bobTime); // Horizontal Component
+        const zBob = std.math.pow(f64, std.math.sin(Player.bobTime), 2) * 1.5; // Vertical Component
+        Player.bobVec = vec.rotateZ(Vec3d{ xBob * bobStrength, 0, zBob * bobStrength }, -camera.rotation[2]);
 
 		// This our model for movement on a single frame:
 		// dv/dt = a - λ·v
--- a/src/gui/windows/graphics.zig
+++ b/src/gui/windows/graphics.zig
@@ -98,6 +98,15 @@ fn resolutionScaleCallback(newValue: u16) void {
 	main.Window.GLFWCallbacks.framebufferSize(null, main.Window.width, main.Window.height);
 }
 
+fn viewBobbingCallback(newValue: f32) void {
+	settings.viewBobStrength = newValue;
+	settings.save();
+}
+
+fn viewBobbingFormatter(allocator: main.utils.NeverFailingAllocator, value: f32) []const u8 {
+	return std.fmt.allocPrint(allocator.allocator, "#ffffffView Bobbing Strength: {d:.2}", .{value}) catch unreachable;
+}
+
 pub fn onOpen() void {
 	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
 	list.add(ContinuousSlider.init(.{0, 0}, 128, 10.0, 154.0, @floatFromInt(settings.fpsCap orelse 144), &fpsCapCallback, &fpsCapFormatter));
@@ -108,6 +117,7 @@ pub fn onOpen() void {
 	list.add(CheckBox.init(.{0, 0}, 128, "Vertical Synchronization", settings.vsync, &vsyncCallback));
 	list.add(DiscreteSlider.init(.{0, 0}, 128, "#ffffffAnisotropic Filtering: ", "{}x", &anisotropy, switch(settings.anisotropicFiltering) {1 => 0, 2 => 1, 4 => 2, 8 => 3, 16 => 4, else => 2}, &anisotropicFilteringCallback));
 	list.add(DiscreteSlider.init(.{0, 0}, 128, "#ffffffResolution Scale: ", "{}%", &resolutions, @as(u16, @intFromFloat(@log2(settings.resolutionScale) + 2.0)), &resolutionScaleCallback));
+	list.add(ContinuousSlider.init(.{0, 0}, 128, 0.0, 1.0, settings.viewBobStrength, &viewBobbingCallback, &viewBobbingFormatter));
 	list.finish(.center);
 	window.rootComponent = list.toComponent();
 	window.contentSize = window.rootComponent.?.pos() + window.rootComponent.?.size() + @as(Vec2f, @splat(padding));
--- a/src/settings.zig
+++ b/src/settings.zig
@@ -44,6 +44,8 @@ pub var musicVolume: f32 = 1;
 
 pub var leavesQuality: u16 = 2;
 
+pub var viewBobStrength: f32 = 0.5;
+
 
 pub var storageTime: i64 = 5000;
 


IntegratedQuantum:
I think it feels kind of weird, but I might be biased since I've played this game without any view bobbing for a while. Maybe someone else could take a look.

Also make sure you fix the formatting.

ITR13:
> The teleport is caused by walking backwards.
Doesn't happen for me after 2ced078, so I think this is fixed.

https://github.com/user-attachments/assets/8a83f2d8-bcdb-4d05-87f3-fb2bac8116d4
If the FPS is set to unlimited and vsync is off, you kinda slowly sink into the ground even when you're standing still.

AlexDavies8:
Thanks for the feedback. To address each of the original points:
1. How would this work? The only thing I can think that wouldn't change the sightline would be to change the roll of the camera or to move it along the local z axis (so forwards/backwards), both of which would feel very strange. I've taken a look at minecraft's view bobbing and it also affects the player's sightline. I've updated the view bobbing to be more natural based off minecraft's implementation.
2. Not sure what you mean by affecting the player's position. I'm adding an extra position term to the return of Player.getEyePosBlocking() which seems to only be used for clientside rendering stuff and some calculations around which chunks should be visible.
3. Fixed now. I was accidentally taking a fractional power of a negative number, so everything became 'nan'.

AlexDavies8:
Also, the sinking isn't caused by my changes. It also happens on commit before any of my changes. It also seems to only happen if you're looking below the horizon.

careeoki:
I'm not a fan of view bobbing in any game. I don't think it adds to immersion, and at worst it can cause motion sickness.
I bet most players would turn it off as soon as they find the option. 

Regardless, setting View Bobbing to 0 doesn't seem to actually turn it off.
If you're walking directly into a wall or corner, your view still bobs despite the player not actually moving.

AlexDavies8:
Fixed the issue with view bobbing 0 not turning it off. I super disagree with the view bobbing. For me personally it feels massively jarring without it, but there's always the option of just defaulting it to be off if most people don't like it.

IntegratedQuantum:
I agree with @ikabod-kee that the focus point should be stable, and not move. Our eyes do this as well when we walk around. This should also be better for people with motion sickness.

As for how this could be implemented:
Instead of just moving the camera it also needs to be rotated such that it still points at the same position in 3d space as it would without bobbing. This position can be determined with a raycast from the camera, just like it's already done for block selection.

AlexDavies8:
I've changed it so that the selected block isn't affected by the bobbing. Also, we no longer have view bobbing if you are stuck on a wall (view only bobs when you are actually moving, instead of it being based on input). The magnitude of the motion also increases slightly when moving faster as the bobbing motion felt weaker when sprinting compared to walking (guessing some kind of optical illusion or weird way our brains see motion).

ITR13:
> The bobbing effect seems very different now. It's as if you're walking with a limp now instead of step by step, which doesn't feel very satisfying. The bug fixes are very good though!

Also agree that the current implementation feels weird.

AlexDavies8:
Yeah, sorry, was away for the weekend so couldn't work on it for a few days. I'm planning on taking a better look at reworking the movement to feel more natural tomorrow. :)

AlexDavies8:
Ended up doing it now instead of tomorrow. This should feel more like walking - it's much less lopsided than before.

https://github.com/user-attachments/assets/f2515e52-635f-4491-9acc-6557b1337158

Sorry for the bad video quality, github has file limits, and the lack of antialiasing isn't good for compression.

AlexDavies8:
Made changes you mentioned. The motion is now a 'u' shape instead of an 'n' shape. Not sure how to feel about it - in real life it would be an 'n' shape and minecraft also uses an 'n' shape, but the 'u' shape doesn't feel particularly different to me, so I guess whatever people prefer.

https://github.com/user-attachments/assets/342a2d33-5bd9-4f40-b562-0e5d32b8e4e1

AlexDavies8:
Nevermind, just realised vertical stuff seems to get flipped when transformed to player's rotation, so it was actually 'u' shape before.

AlexDavies8:
Personally, I prefer the n shape, but I think both feel pretty ok now.

AlexDavies8:
I've tried out the equations linked in the commit message. It feels a bit weird if it's too strong but I quite like it at around 0.5 strength.

AlexDavies8:
I'm not quite sure what you mean by the bobbing being inconsistent. It could be the rotation part maybe, as that is on a slightly different speed to the movement? I've changed them both to have the same period of repetition. Also, the view bobbing is now only dependent on the raw xy velocity of the player.

AlexDavies8:
Should all be good now. Turns out there was also a minor bug where the camera wouldn't recentre properly when in the air, so that should be fixed too.

src/game.zig
```diff
@@ -665,6 +685,20 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
+		// Smooth lerping of bobTime with framerate independent damping
+		const fac: f64 = 1 - std.math.exp(-15 * deltaTime);
+		var targetBobVel: f64 = 0;
+		const horizontalMovementSpeed = vec.length(vec.xy(Player.getVelBlocking()));
+		if (horizontalMovementSpeed > 0.01 and Player.onGround) {
+			targetBobVel = std.math.pow(f64, horizontalMovementSpeed / 4, 0.5);
+		}
+		Player.bobVel = Player.bobVel * (1 - fac) + targetBobVel * fac;
+		if (Player.onGround) { // No view bobbing in the air
+			Player.bobTime += std.math.pow(f64, Player.bobVel, 0.9) * 8 * deltaTime;
```
IntegratedQuantum, line 699: What is the motivation/purpose of the `pow(vel, 0.9)` is?
Could the same be achieved without a power function?

src/renderer.zig
```diff
@@ -556,7 +569,7 @@ pub const MenuBackGround = struct {
 			// Draw to frame buffer.
 			buffer.bind();
 			c.glClear(c.GL_DEPTH_BUFFER_BIT | c.GL_STENCIL_BUFFER_BIT | c.GL_COLOR_BUFFER_BIT);
-			main.renderer.render(game.Player.getEyePosBlocking());
+			main.renderer.render();
```
IntegratedQuantum, line 572: It seems that your method will now always apply bobbing to any render call.
This call however requires a stable camera, otherwise the recorded background image will contain seems.

IntegratedQuantum (changes requested):
While reviewing I found that many of your math expressions are unnecessarily complicated and could be simplified by applying some trigonometric identities.

Additionally one thing I would like you to think about: You have set a limit of 1.2 to bobMag.
This seems like it would cause further problems down the line. In the game I think it would be nice to have variable player hitboxes. Because of that I would like to have a generic limit to bobMag that guarantees that the camera won't ever go outside of the players bounding box.

IntegratedQuantum:
Oh and also please fix the merge conflict.

src/game.zig
AlexDavies8: This was done so that the speed of the view bobbing would fall off slightly faster than the magnitude, as otherwise it felt quite a lot too fast at high speeds. This could maybe be done with a piecewise linear function, possibly combined with a hard cap.

src/renderer.zig
AlexDavies8: I'm not quite sure what you mean here in terms of background seams, but yes, view bobbing will apply to all render calls that also render the world, although shouldn't happen on the title screen as that's a different code path. The only thing I can think that could be done differently would be to let the camera store its own position, so no player-related data needs to be passed to the renderer, and then handle the special case of the block selection somehow as we don't want to have view bobbing affect it.

AlexDavies8:
> Additionally one thing I would like you to think about: You have set a limit of 1.2 to bobMag. This seems like it would cause further problems down the line. In the game I think it would be nice to have variable player hitboxes. Because of that I would like to have a generic limit to bobMag that guarantees that the camera won't ever go outside of the players bounding box.
It's clamped to be equal (roughly) to the magnitude caused by sprinting, so going faster won't cause the camera to freak out. However, even at max view bobbing strength, it would only risk exiting the bounds if it was clamped at something pretty high like 4. I could probably enforce the limit earlier on, though, so it's clearer why it exists. It shouldn't be hard to scale it by the size of the player's hitbox either - the height would probably be best.

src/renderer.zig
AlexDavies8: Actually, decoupling the player and the camera position might be a good idea generally, as it would allow for third person/spectator and other camera effects to be handled in some update loop instead of directly in the rendering code.

src/game.zig
AlexDavies8: Nvm, after trying without it's not noticably different unless you turn movement speed up to crazy high levels.

src/game.zig
```diff
@@ -345,6 +348,23 @@ pub const Player = struct { // MARK: Player
 		return eyePos + super.pos + desiredEyePos;
 	}
 
+	pub fn applyViewBobbingPosBlocking(pos: Vec3d) Vec3d {
+		mutex.lock();
+		defer mutex.unlock();
+		const xBob = (std.math.atan(0.25 * @sin(bobTime)) / std.math.atan(@as(f64, 0.25))) * 0.05; // Horizontal Component
+		const zBob = (std.math.atan(0.5 * @cos(2 * bobTime + std.math.pi * 0.5)) / std.math.atan(@as(f64, 0.5)) + 0.5 * std.math.pow(f64, @cos(bobTime + std.math.pi * 0.5), 2)) * 0.8 * 0.05; // Vertical Component
```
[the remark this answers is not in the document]
AlexDavies8: sin²(x) shouldn't be any slower than it's identity as long as I use an intermediary variable to store sin(x).

AlexDavies8:
Addressed feedback as much as I can without decoupling the camera position from player position.

src/game.zig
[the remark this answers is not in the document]
IntegratedQuantum: That is true, but I think it's more readable, when both operands to zBob have the same frequency.

src/renderer.zig
```diff
@@ -199,9 +212,9 @@ pub fn renderWorld(world: *World, ambientLight: Vec3f, skyColor: Vec3f, playerPo
 	const meshes = mesh_storage.updateAndGetRenderChunks(world.conn, playerPos, settings.renderDistance);
 
 	gpu_performance_measuring.startQuery(.chunk_rendering_preparation);
-	const direction = crosshairDirection(game.camera.viewMatrix, lastFov, lastWidth, lastHeight);
+	const direction = crosshairDirection(baseViewMatrix, lastFov, lastWidth, lastHeight);
```
AlexDavies8, line 215: Is there a reason we are using `crosshairDirection `instead of just `game.camera.direction`? `game.camera.direction` seems to do the same without having to extract the direction from the view matrix.

AlexDavies8:
Finally got around to decoupling the camera from the player. It turned out to not be that much work. I've also moved the camera to it's own file and put the viewbobbing in there instead of inside of the Player struct to keep it cleaner. In the future, the logic for third person/first person camera control, or camera shake (such as on taking damage or maybe is there's some kind of explosion) could also go in this file.

To me this feels a lot easier to work with. It might also be good to redo some of the block selection stuff, as it seems like we are selecting the block during rendering and then reusing that result to place/remove blocks. To me, that is game logic, and should be done before rendering inside of the update loop, and then the renderer doesn't need to update the selected block itself. I haven't done that here, but I have swapped over to using `camera.direction` instead of `crosshairDirection` as it removes the need for quite a few temporary variables, and we calculate the look direction of the camera when updating the matrix anyway, so we are saving a tiny bit of computation each frame as a bonus.

src/game.zig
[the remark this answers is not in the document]
AlexDavies8: Done 👍

src/game.zig
```diff
@@ -351,9 +351,13 @@ pub const Player = struct { // MARK: Player
 	pub fn applyViewBobbingPosBlocking(pos: Vec3d) Vec3d {
 		mutex.lock();
 		defer mutex.unlock();
-		const xBob = (std.math.atan(0.25 * @sin(bobTime)) / std.math.atan(@as(f64, 0.25))) * 0.05; // Horizontal Component
-		const zBob = (std.math.atan(0.5 * @cos(2 * bobTime + std.math.pi * 0.5)) / std.math.atan(@as(f64, 0.5)) + 0.5 * std.math.pow(f64, @cos(bobTime + std.math.pi * 0.5), 2)) * 0.8 * 0.05; // Vertical Component
-		const bobVec = vec.rotateZ(Vec3d{ xBob * bobMag, 0, zBob * bobMag }, -camera.rotation[2]);
+		// Horizontal Component - atan(0.25 * sin(bobTime)) / atan(0.25)
+		const xBob = 1.0204970377 * @sin(bobTime);
```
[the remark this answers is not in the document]
AlexDavies8: Just realised I'd misread the function when calculating constants. In both cases of atan, a constant isn't needed. This is now fixed.

src/camera.zig
```diff
@@ -0,0 +1,85 @@
...
+const Vec3f = vec.Vec3f;
+const Vec4f = vec.Vec4f;
+const Vec3d = vec.Vec3d;
+const Mat4f = vec.Mat4f;
+const game = @import("game.zig");
+const settings = @import("settings.zig");
+
+// MARK: camera
+// A zig file is implicitly a struct, so when importing we can have camera.position, camera.rotation, etc.
+// instead of camera.camera.position, camera.camera.rotation, etc.
```
IntegratedQuantum, line 13: This should be basic zig knowledge, no need to reiterate it here.

src/game.zig
```diff
@@ -63,55 +42,55 @@ pub const collision = struct {
 		const f2 = triangle[0] - triangle[2];
 
 		// Test axis a00
-		const a00 = Vec3d{0, -f0[Z], f0[Y]};
+		const a00 = Vec3d{ 0, -f0[Z], f0[Y] };
```
IntegratedQuantum, line 45: Oh no, auto-formatting strikes again :​(
Please all the unnecessary formatting changes in this file.

src/renderer.zig
```diff
@@ -95,7 +96,7 @@ fn initReflectionCubeMap() void {
 	fakeReflectionShader.bind();
 	c.glUniform1f(fakeReflectionUniforms.frequency, 1);
 	c.glUniform1f(fakeReflectionUniforms.reflectionMapSize, reflectionCubeMapSize);
-	for(0..6) |face| {
+	for (0..6) |face| {
```
IntegratedQuantum, line 99: same in this file

IntegratedQuantum (changes requested):
It's difficult to review this PR due to all the unintentional formatting changes. Please revert them.
Also again please fix the merge conflicts.

src/game.zig
AlexDavies8: I think something got messed up and formatted while I was merging. I've reverted and re-merged and it should be fixed now.

src/game.zig
```diff
@@ -781,8 +781,8 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 			if (Player.onGround) { // No view bobbing in the air
 				Player.bobTime += Player.bobVel * 8 * deltaTime;
 			}
-			// Maximum magnitude when sprinting (2x walking speed)
-			Player.bobMag = @sqrt(@min(Player.bobVel, 2)) * settings.viewBobStrength;
+			// Maximum magnitude when sprinting (2x walking speed). Magnitude is scaled by player bounding box height
+			Player.bobMag = @sqrt(@min(Player.bobVel, 2)) * settings.viewBobStrength * Player.outerBoundingBoxExtent[2];
```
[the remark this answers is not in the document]
AlexDavies8: I'm still confused about what you mean about the eyeBox in relation to view bobbing. Scaling by the bounding box height should scale the view bobbing by player height, so scaling the player up or down should cause the view bobbing to appear the same relative to the player's size. Is there a plan to have characters with tiny legs and a huge hat or something, where the eyes and view bobbing should be ow but the bounding box tall?

src/game.zig
```diff
@@ -385,6 +388,27 @@ pub const Player = struct { // MARK: Player
...
 
+	pub fn applyViewBobbingPosBlocking(pos: Vec3d) Vec3d {
+		mutex.lock();
+		defer mutex.unlock();
+		// Horizontal Component - atan(0.25 * sin(bobTime)) / atan(0.25)
+		const xBob = 1.0204970377 * @sin(bobTime);
+		// Vertical Component (*0.8) - atan(0.5 * -sin(2 * bobTime)) / atan(0.5) + 0.5 * (-@sin(bobTime))^2
+		const a = 0.5 * -@sin(2 * bobTime);
+		const b = -@sin(bobTime);
+		const zBob = ((a - a * a * a / 3) / 0.4636476090 + 0.5 * b * b) * 0.8;
```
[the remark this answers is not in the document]
AlexDavies8: I've updated it to just multiply by 2 instead of dividing by a constant. Not much visual difference and if anything looks a bit better to me, and keeps it simple.

AlexDavies8:
Fixed the formatting stuff, think I did it by accident while merging. The background image isn't affected by view bobbing as the camera now has the view bobbing applied in the renderer. At the moment this is essentially hard-coded to the renderer, but if there are more effects in the future such as camera shaking such as on taking damage or explosions etc, it might be better to move this to some kind of function on the camera such as `applyRenderTransforms` and `removeRenderTransforms`.
It might also make sense to switch the camera physics stuff to apply as one of these render-only transformations if we don't want them to affect the currently selected block.

src/renderer.zig
archbirdplus: #323 
Try moving the crosshair in menu mode. Check if block selection is still accurate after removing this code.

src/game.zig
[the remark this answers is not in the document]
IntegratedQuantum: Scaling the bobbing by player height is fine, but
Let's say you sprint jump through a narrow tunnel.
I need you to guarantee that under no circumstances the camera will clip into the terrain.
The camera can be anywhere inside the eyeBox The eyeBox should always be smaller than the collision box. But your code adds a value to that cameraPosition, so you might push it outside the eyeBox.
Please ensure that the resulting position is never outside the collision box, which could make it clip into the terrain. It probably shouldn't even get too close to the collision, because OpenGL doesn't render triangles at depth 0.

Now you have probably tested this scenario and everything works fine, but all these parameters, like the hitboxes and how often the camera actually gets close to the bounds of the eyeBox, may change.

src/renderer.zig
IntegratedQuantum: This is still broken. If I record a background image while moving then it contains seams between the recorded images. Here take a look:
client_1146

IntegratedQuantum:
The background image is still broken, and there's also ensuring that the camera stays inside the bounding box.
The rest is looking good, so we are almost done here.

> It might also make sense to switch the camera physics stuff to apply as one of these render-only transformations if we don't want them to affect the currently selected block.

We definitely want the camera physics to affect block selection, otherwise it would be completely broken when climbing stairs, the camera physics is important for smoothing out movement.

AlexDavies8:
Finally got around to fixing the last little bits. The view bobbing shouldn't be able to escape the eyeBox anymore. It will scale down the view bobbing if the camera is already near the edge (such as due to camera physics), and will clamp the final view bobbing to always remain inside eyeBox to catch the case where the view bobbing alone would be outside the eyeBox.
The seams on the background image should be fixed too. It might be worth disabling the block selection when taking the image as well in the future.

IntegratedQuantum:
It appears to be broken:

https://github.com/user-attachments/assets/93d33971-4226-4828-9474-2a63cbede7fb

AlexDavies8:
Oops 😅, forgot to reset view bobbing strength after testing out the eyeBox clamping. Should be fixed now.

IntegratedQuantum:
Please do more thorough testing in the future. Stairs and stepping in general is broken:

https://github.com/user-attachments/assets/63a4ed91-2778-4c0a-a9b6-89a54ed2be5b

PR #717, opened 2024-09-12
Game controller support

catmeow72:
This pull request adds controller support to the game.
TODO:
 - [x] Add controller input
 - [x] Read controller mappings
   - [x] Controller mappings from file in binary directory
   - [x] Controller mappings from environment variable
   - [x] Controller mappings from config
 - [x] Download controller mappings
 - [x] Add cursor sprite for controller cursor


src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+		.response_storage = .{ .dynamic = &list }
+	}) catch null) |_| {
+		if (files.openDir(".") catch null) |dir| {
+			dir.write("gamecontrollerdb.txt", list.items) catch unreachable;
+			const timeStampStr = std.fmt.allocPrint(main.globalAllocator.allocator, "{d}", .{curTimestamp}) catch unreachable;
+			dir.write("gamecontrollerdb.stamp", timeStampStr) catch unreachable;
+			main.globalAllocator.free(timeStampStr);
+		}
+	}
+	list.deinit();
```
IntegratedQuantum, line 617: Please use defer

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+	return !controllerMappingsDownloaded and downloadControllerMappingsThread != null;
+}
+fn downloadControllerMappingsThreadFunc(curTimestamp: i128) void {
+	var client: std.http.Client = .{.allocator = main.globalAllocator.allocator};
+	var list = std.ArrayList(u8).init(main.globalAllocator.allocator);
+	if (client.fetch(.{
+		.method = .GET,
+		.location = .{.url = "https://raw.githubusercontent.com/mdqinc/SDL_GameControllerDB/master/gamecontrollerdb.txt"},
+		.response_storage = .{ .dynamic = &list }
+	}) catch null) |_| {
```
IntegratedQuantum, line 609: It would be nice to log the error, to make it easier to debug if someone complains about their controller not working.

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+		if (files.openDir(".") catch null) |dir| {
```
IntegratedQuantum, line 610: same here

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+			dir.write("gamecontrollerdb.txt", list.items) catch unreachable;
```
IntegratedQuantum, line 611: This error can be reached, if e.g. `gamecontrollerdb.txt` already exists but the user doesn't have write permissions for whatever reason. Please properly handle errors like this.

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+			const timeStampStr = std.fmt.allocPrint(main.globalAllocator.allocator, "{d}", .{curTimestamp}) catch unreachable;
+			dir.write("gamecontrollerdb.stamp", timeStampStr) catch unreachable;
```
IntegratedQuantum, line 613: same here

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+	var client: std.http.Client = .{.allocator = main.globalAllocator.allocator};
+	var list = std.ArrayList(u8).init(main.globalAllocator.allocator);
+	if (client.fetch(.{
+		.method = .GET,
+		.location = .{.url = "https://raw.githubusercontent.com/mdqinc/SDL_GameControllerDB/master/gamecontrollerdb.txt"},
+		.response_storage = .{ .dynamic = &list }
+	}) catch null) |_| {
+		if (files.openDir(".") catch null) |dir| {
+			dir.write("gamecontrollerdb.txt", list.items) catch unreachable;
+			const timeStampStr = std.fmt.allocPrint(main.globalAllocator.allocator, "{d}", .{curTimestamp}) catch unreachable;
```
IntegratedQuantum, line 612: should use the stackAllocator

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+				needDownload = true;
+				break;
+			}
+		}
+	}
+	controllerMappingsDownloaded = !needDownload;
+	if (needDownload) {
+		downloadControllerMappingsThread = std.Thread.spawn(.{}, downloadControllerMappingsThreadFunc, .{curTimestamp}) catch null;
+		if (downloadControllerMappingsThread == null) {
+			std.log.err("Couldn't spawn controller mapping download thread!", .{});
```
IntegratedQuantum, line 655: Please log the error name as well with `catch |err|`.

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+		if (files.openDir(".") catch null) |dir| {
+			dir.write("gamecontrollerdb.txt", list.items) catch unreachable;
+			const timeStampStr = std.fmt.allocPrint(main.globalAllocator.allocator, "{d}", .{curTimestamp}) catch unreachable;
+			dir.write("gamecontrollerdb.stamp", timeStampStr) catch unreachable;
+			main.globalAllocator.free(timeStampStr);
+		}
+	}
+	list.deinit();
+	client.deinit();
+	controllerMappingsDownloaded = true;
```
IntegratedQuantum, line 619: Since you do this from a separate thread, it should probably be an atomic.

src/graphics/Window.zig
IntegratedQuantum: Also maybe consider using the threadpool instead of spawning new threads (which is rather expensive)

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+	if (needDownload) {
+		downloadControllerMappingsThread = std.Thread.spawn(.{}, downloadControllerMappingsThreadFunc, .{curTimestamp}) catch null;
+		if (downloadControllerMappingsThread == null) {
+			std.log.err("Couldn't spawn controller mapping download thread!", .{});
+			controllerMappingsDownloaded = true;
+		}
+	}
+}
+pub fn updateControllerMappings() void {
+	if (std.fs.selfExeDirPathAlloc(main.globalAllocator.allocator) catch null) |selfExeDirPath| {
```
IntegratedQuantum, line 661: log the error, stackAllocator

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+		if (std.fs.path.join(main.globalAllocator.allocator, &.{selfExeDirPath, "gamecontrollerdb.txt"}) catch null) |mappings_path| {
```
IntegratedQuantum, line 662: What errors can happen here? If it's just OutOfMemory, then you can use catch unreachable, otherwise please log the error.
also stackAllocator

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+			if (std.fs.openFileAbsolute(mappings_path, .{.mode = .read_only}) catch null) |file| {
+				const data = main.globalAllocator.alloc(u8, file.getEndPos() catch 0);
+				const size = file.readAll(data) catch 0;
```
IntegratedQuantum, line 665: Better use `main.files.read` instead of doing all this boilerplate to read the file.
This is less error prone (like for example here you forgot to close the file).
Also stackAllocator

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+				if (size > 0) {
+					_ = main.globalAllocator.resize(data, size);
+					_ = c.glfwUpdateGamepadMappings(@ptrCast(data));
+				}
+				main.globalAllocator.free(data);
```
IntegratedQuantum, line 670: defer

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+			}
+			main.globalAllocator.free(mappings_path);
```
IntegratedQuantum, line 672: defer

src/graphics/Window.zig
```diff
@@ -349,17 +583,174 @@ pub fn init() void { // MARK: init()
...
+		}
+		main.globalAllocator.free(selfExeDirPath);
```
IntegratedQuantum, line 674: defer

src/graphics/Window.zig
IntegratedQuantum: Also you can use `main.files.write()` to avoid all this boilerplate

IntegratedQuantum (changes requested):
I haven't looked through the entire code yet, but about half-way through I discovered that I keep repeating myself, so please apply my comments to the remaining code.

I mainly highlighted some syntax changes to simplify the code or make it less error prone.
Most notably, please use `defer`.

Also I would suggest to use the stackAllocator for all local allocations. It's much faster than using the global allocator.

And finally: Only OutOfMemory errors are unreachable. Other errors, like for file and network code can be reached by e.g. missing network connection or incorrect file permission. These errors must be handled and printed, otherwise we end up in hard to debug situations where the game just crashes and we have no idea why.

src/graphics/Window.zig
catmeow72: I converted it into a task for the threadpool.

src/graphics/Window.zig
```diff
@@ -1,25 +1,381 @@
...
+	};
+	pub fn downloadControllerMappings(self: *Gamepad) void {
+		var needDownload: bool = false;
+		const curTimestamp = std.time.nanoTimestamp();
+		if (settings.downloadControllerMappings) {
+			var timestamp: i128 = 0;
+			if (files.openDir(".") catch null) |dir| {
+				var _dir: files.Dir =dir;
+				if (_dir.hasFile("gamecontrollerdb.stamp")) {
+					const stamp = _dir.read(main.globalAllocator, "gamecontrollerdb.stamp") catch unreachable;
```
IntegratedQuantum, line 282: This error is not unreachable! The file might for example have incorrect permissions.
also: stackAllocator

src/graphics/Window.zig
```diff
@@ -1,25 +1,381 @@
...
+					defer main.globalAllocator.free(stamp);
+					if (std.fmt.parseInt(i128, stamp, 16) catch null) |newTimestamp| {
+						timestamp = newTimestamp;
+					}
+				}
+			}
```
IntegratedQuantum, line 288: Here a block would be more readable in my opinion:
```zig
const timestamp: i128 = blk: {
	var dir = files.openDir(".") catch break :blk 0;
	if(!dir.hasFile(...)) break :blk 0;
	const stamp = _dir.read(main.stackAllocator, "gamecontrollerdb.stamp") catch break :blk 0;
	defer main.stackAllocator.free(stamp);
	break :blk std.fmt.parseInt(i128, stamp, 16) catch 0;
};
```

src/graphics/Window.zig
```diff
@@ -1,25 +1,381 @@
...
+			return;
+		};
+		defer main.stackAllocator.free(data);
+		_ = c.glfwUpdateGamepadMappings(@ptrCast(data));
+	}
+
+	pub fn init() *Gamepad {
+		const self: *Gamepad = main.globalAllocator.create(Gamepad);
+		if (!settings.askToDownloadControllerMappings) {
+			self.*.downloadControllerMappings();
```
IntegratedQuantum, line 346: You don't need to do .* when accessing members of a pointer.
This is Zig, not C/C++.

src/graphics/Window.zig
```diff
@@ -1,25 +1,381 @@
...
+		}
+		self.*.updateControllerMappings();
+		self.*.gamepadState = std.AutoHashMap(c_int, *c.GLFWgamepadstate).init(main.globalAllocator.allocator);
```
IntegratedQuantum, line 349: As of the new zig version that I updated to recently, you can just use
```zig
self.*.gamepadState = .init(main.globalAllocator.allocator);
```

IntegratedQuantum (changes requested):
Sorry for taking so long to review.
I still think you could do some improvements here.
Could also explain the intentions behind this new cursor you added?

src/graphics/Window.zig
```diff
@@ -1,25 +1,381 @@
...
+	}
+	pub fn isControllerConnected(self: *Gamepad) bool {
+		return self.gamepadState.count() > 0;
+	}
+	pub fn controllerMappingsDownloading(self: *Gamepad) bool {
+		return !self.*.controllerMappingsDownloaded;
+	}
+	const ControllerMappingDownloadTask = struct { // MARK: ControllerMappingDownloadTask
+		curTimestamp: i128,
+		controllerMappingsDownloaded: *bool,
```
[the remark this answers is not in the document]
catmeow72: I already had it so that it doesn't schedule multiple tasks, but in a different function. I'll add this to the schedule function as well.

catmeow72:
I just rebased the branch. I'm now about to check the changes.

catmeow72:
Also, I had to fix a bunch of merge conflicts when rebaseing, which is why I wanted to check the changes.

src/gui/gui.zig
```diff
@@ -559,6 +562,9 @@ pub fn updateAndRenderGui() void {
 		}
 		inventory.render(mousePos);
 	}
+	const oldScale = draw.setScale(scale);
+	Cursor.render();
+	defer draw.restoreScale(oldScale);
```
IntegratedQuantum, line 567: Always put the defer after the line that changed the value.

src/gui/windows/_windowlist.zig
```diff
@@ -27,3 +27,5 @@ pub const save_selection = @import("save_selection.zig");
 pub const settings = @import("settings.zig");
 pub const sound = @import("sound.zig");
 pub const workbench = @import("workbench.zig");
+pub const download_controller_mappings = @import("download_controller_mappings.zig");
+pub const controller_mappings_settings = @import("controller_mappings_settings.zig");
```
IntegratedQuantum, line 31: Please keep it sorted alphabetically.

src/gui/windows/controller_mappings_settings.zig
```diff
@@ -0,0 +1,56 @@
...
+	gui.closeWindowFromRef(&window);
+	gui.openWindow("download_controller_mappings");
+}
+pub fn onOpen() void {
+	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
+	list.add(Label.init(.{0, 0}, 256, "Controller Mapping Settings", .center));
+	list.add(CheckBox.init(.{0, 0}, 256, "Don't ask at startup", !settings.askToDownloadControllerMappings, &askCallback));
+	list.add(CheckBox.init(.{0, 0}, 256, "Download controller mappings periodically", settings.downloadControllerMappings, &periodicCallback));
+	list.add(CheckBox.init(.{0, 0}, 256, "Download controller mappings when unrecognized controller plugged in", settings.downloadControllerMappingsWhenUnrecognized, &unrecognizedCallback));
+	list.add(Button.initText(.{0, 0}, 256, "Save and Close", .{.callback = &onSave}));
```
IntegratedQuantum, line 45: If there is a save and close, that implies that the window close button will not save it, but that's not true. The changes are still stored in settings.zig and will get stored to disk on close or when another option is set.

src/gui/windows/controls.zig
```diff
@@ -80,4 +126,4 @@ pub fn render() void {
 		onOpen();
 		window.rootComponent.?.verticalList.scrollBar.currentState = oldScroll;
 	}
-}
\ No newline at end of file
+}
```
IntegratedQuantum, line 129: Screenshot at 2024-09-30 19-05-48
At default scale the window is too high for the default screen (it's cut off at the bottom). Please make the window smaller to ensure that it works for all resolutions.

src/gui/windows/download_controller_mappings.zig
```diff
@@ -0,0 +1,42 @@
...
+const HorizontalList = @import("../components/HorizontalList.zig");
+
+pub var window = GuiWindow {
+	.contentSize = Vec2f{256, 256},
+	.closeable = false
+};
+
+const padding: f32 = 8;
+pub fn update() void {
+	if (!main.Window.gamepad.controllerMappingsDownloading()) {
```
IntegratedQuantum, line 24: This is a double negatation (one here, the other inside the function) and I find it a bit confusing. Better change the function name to `controllerMappingsDownloaded` and remove the two negations.

src/gui/windows/settings.zig
```diff
@@ -22,6 +22,7 @@ pub fn onOpen() void {
 	list.add(Button.initText(.{0, 0}, 128, "Controls", gui.openWindowCallback("controls")));
 	list.add(Button.initText(.{0, 0}, 128, "Advanced Controls", gui.openWindowCallback("advanced_controls")));
 	list.add(Button.initText(.{0, 0}, 128, "Change Name", gui.openWindowCallback("change_name")));
+	list.add(Button.initText(.{0, 0}, 128, "Controller mappings", gui.openWindowCallback("controller_mappings_settings")));
```
IntegratedQuantum, line 25: I think this is cluttering the settings menu. How about putting this button inside the controls window, when the gamepad controls are edited.

src/gui/windows/controls.zig
```diff
@@ -18,14 +18,25 @@ pub var window = GuiWindow {
 
 const padding: f32 = 8;
 var selectedKey: ?*main.Window.Key = null;
+var kbd: bool = true;
```
IntegratedQuantum, line 21: Please avoid abbreviations unless they are well known.

src/main.zig
```diff
@@ -297,6 +297,13 @@ fn toggleNetworkDebugOverlay() void {
 fn toggleAdvancedNetworkDebugOverlay() void {
 	gui.toggleWindow("debug_network_advanced");
 }
+fn cycleHotbarSlot(i: comptime_int) *const fn() void {
+	return &struct {
+		fn set() void {
+			game.Player.selectedSlot = @as(u32, @intCast(@mod(@as(i33, game.Player.selectedSlot) + i, 12)));
```
IntegratedQuantum, line 303: The outermost `@as(u32, ...)` is not necessary.

src/main.zig
```diff
@@ -309,31 +316,37 @@ pub const KeyBoard = struct { // MARK: KeyBoard
...
+		.{.name = "forward", .key = c.GLFW_KEY_W, .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_LEFT_Y, .positive = false}},
+		.{.name = "left", .key = c.GLFW_KEY_A, .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_LEFT_X, .positive = false}},
+		.{.name = "backward", .key = c.GLFW_KEY_S, .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_LEFT_Y, .positive = true}},
+		.{.name = "right", .key = c.GLFW_KEY_D, .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_LEFT_X, .positive = true}},
+		.{.name = "scrollUp", .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_RIGHT_Y, .positive = false}},
+		.{.name = "scrollDown", .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_RIGHT_Y, .positive = true}},
+		.{.name = "uiUp", .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_LEFT_Y, .positive = false}},
+		.{.name = "uiLeft", .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_LEFT_X, .positive = false}},
+		.{.name = "uiDown",  .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_LEFT_Y, .positive = true}},
+		.{.name = "uiRight", .gamepadAxis = .{.axis = c.GLFW_GAMEPAD_AXIS_LEFT_X, .positive = true}},
```
IntegratedQuantum, line 328: ui buttons should be further down in the Gui section of the key bindings. Otherwise they clutter the important stuff.
Maybe even make a uniquue section for all the new controller control stuff at the bottom (above the debug section)

src/main.zig
```diff
@@ -402,6 +420,27 @@ fn isValidIdentifierName(str: []const u8) bool { // TODO: Remove after #480
 	}
 	return true;
 }
+var startup_finished: bool = false;
+var controller_mappings_window_shown: bool = false;
```
IntegratedQuantum, line 424: Better use an enum with `showController`, `showNamePrompt`, `finished`.
Then you can also increment by one each time, making it easier to change order or add new startup menus.

src/main.zig
```diff
@@ -445,6 +484,12 @@ pub fn convertJsonToZon(jsonPath: []const u8) void { // TODO: Remove after #480
 				zonString.append(c);
 			},
 		}
+		controller_mappings_window_shown = true;
+	} else if(settings.playerName.len == 0) {
+		gui.openWindow("change_name");
+	} else {
+		startup_finished = true;
+		gui.openWindow("main");
```
IntegratedQuantum, line 492: This appears to be in the wrong function.

src/main.zig
```diff
@@ -569,7 +610,6 @@ pub fn main() void { // MARK: main()
 	}
 
 	audio.setMusic("cubyz:cubyz");
-
```
IntegratedQuantum, line 572: unintended line change

src/settings.zig
```diff
@@ -56,6 +57,17 @@ pub var developerAutoEnterWorld: []const u8 = "";
...
+pub var askToDownloadControllerMappings: bool = true;
+
+pub var downloadControllerMappings: bool = false;
+
+pub const downloadControllerMappingsInterval: i128 = std.time.ns_per_week;
+
+pub var downloadControllerMappingsWhenUnrecognized: bool = false;
+pub var controllerAxisDeadzone: f32 = 0.0;
+
+pub const minTimeToDownloadControllerMappingsWhenUnrecognized: i128 = std.time.ns_per_day;
```
IntegratedQuantum, line 69: unused

src/settings.zig
```diff
@@ -56,6 +57,17 @@ pub var developerAutoEnterWorld: []const u8 = "";
 
 pub var developerGPUInfiniteLoopDetection: bool = false;
 
+pub var askToDownloadControllerMappings: bool = true;
+
+pub var downloadControllerMappings: bool = false;
+
+pub const downloadControllerMappingsInterval: i128 = std.time.ns_per_week;
```
IntegratedQuantum, line 64: unused

IntegratedQuantum (changes requested):
I still found a bunch of things that can be improved, but overall I think it has already greatly improved.

catmeow72:
Just so you know, the reason why there's a custom cursor implemented is because Wayland on Linux doesn't support moving the mouse at all, and you need root to work around that.

src/graphics/Window.zig
```diff
@@ -1,25 +1,308 @@
...
+		const curTimestamp = std.time.nanoTimestamp();
+		if (settings.downloadControllerMappings) {
+			const timestamp: i128 = blk: {
+				var dir = files.openDir(".") catch break :blk 0;
+				if (!dir.hasFile("gamecontrollerdb.stamp")) break :blk 0;
+				const stamp = dir.read(main.stackAllocator, "gamecontrollerdb.stamp") catch break :blk 0;
+				defer main.stackAllocator.free(stamp);
+				break :blk std.fmt.parseInt(i128, stamp, 16) catch 0;
+			};
+			const delta = curTimestamp - timestamp;
```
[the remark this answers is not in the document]
catmeow72: I decided to use |-, since I'd rather it doesn't wrap around, as that would screw up the needsDownload calculation

src/graphics/Window.zig
```diff
@@ -1,25 +1,308 @@
...
+			}
+		}
+		const selfExeDirPath = std.fs.selfExeDirPathAlloc(main.stackAllocator.allocator) catch |err| {
+			std.log.err("Error getting executable directory{s}", .{@errorName(err)});
+			return;
+		};
+		defer main.stackAllocator.free(selfExeDirPath);
+		const mappings_path = std.fs.path.join(main.stackAllocator.allocator, &.{selfExeDirPath, "gamecontrollerdb.txt"}) catch unreachable;
+		defer main.stackAllocator.free(mappings_path);
+		const data =  main.files.read(main.stackAllocator, mappings_path) catch |err| {
```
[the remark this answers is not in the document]
catmeow72: I had it there initially, and didn't think about it again.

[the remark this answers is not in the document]
catmeow72: I'm going to remove it in my next commit

src/gui/cursor.zig
```diff
@@ -0,0 +1,31 @@
...
+	texture.deinit();
+}
+
+pub fn render() void {
+	if (main.Window.lastUsedMouse or main.Window.grabbed) return;
+	texture.bindTo(0);
+	graphics.draw.setColor(0xffffffff);
+	const mousePos = main.Window.getMousePosition();
+	graphics.draw.boundImage(@as(Vec2f, @splat(-size/2.0)) + (mousePos/@as(Vec2f, @splat(gui.scale))), .{size, size});
+}
```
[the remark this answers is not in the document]
catmeow72: It is basically a mouse cursor, but for a gamepad. I'll rename it to gamepadCursor.

src/gui/windows/download_controller_mappings.zig
catmeow72: Changed it to wereControllerMappingsDownloaded to prevent a name collision with the value it references.

src/main.zig
catmeow72: Probably didn't notice this when fixing merge conflicts while rebasing

src/graphics/Window.zig
```diff
@@ -10,16 +12,278 @@ pub const c = @cImport ({
 });
 
 var isFullscreen: bool = false;
+pub var lastUsedMouse: bool = true;
 pub var width: u31 = 1280;
 pub var height: u31 = 720;
 pub var window: *c.GLFWwindow = undefined;
 pub var grabbed: bool = false;
+
```
IntegratedQuantum, line 20: unintended change

src/graphics/Window.zig
```diff
@@ -10,16 +12,278 @@ pub const c = @cImport ({
...
 pub var scrollOffset: f32 = 0;
+pub const Gamepad = struct {
```
IntegratedQuantum, line 22: Please add an empty line before the struct declaration.

src/graphics/Window.zig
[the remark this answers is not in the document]
IntegratedQuantum: Well, `-|` is just incorrect for calculating the time difference. When the time wraps around, `-%` gives the actual difference between timestamps.

src/gui/gamepad_cursor.zig
```diff
@@ -0,0 +1,31 @@
...
+const gui = @import("gui.zig");
+const GuiWindow = gui.GuiWindow;
+const GuiComponent = gui.GuiComponent;
+
+const size: f32 = 16;
+
+var texture: Texture = undefined;
+
+pub fn init() void {
+	texture = Texture.initFromFile("assets/cubyz/ui/cursor.png");
```
IntegratedQuantum, line 18: Please also rename the image file

src/gui/windows/controller_mappings_settings.zig
```diff
@@ -0,0 +1,56 @@
...
+	gui.closeWindowFromRef(&window);
+	gui.openWindow("download_controller_mappings");
+}
+pub fn onOpen() void {
+	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
+	list.add(Label.init(.{0, 0}, 256, "Controller Mapping Settings", .center));
+	list.add(CheckBox.init(.{0, 0}, 256, "Don't ask at startup", !settings.askToDownloadControllerMappings, &askCallback));
+	list.add(CheckBox.init(.{0, 0}, 256, "Download controller mappings periodically", settings.downloadControllerMappings, &periodicCallback));
+	list.add(CheckBox.init(.{0, 0}, 256, "Download controller mappings when unrecognized controller plugged in", settings.downloadControllerMappingsWhenUnrecognized, &unrecognizedCallback));
+	list.add(Button.initText(.{0, 0}, 256, "Close", .{.callback = &onSave}));
```
IntegratedQuantum, line 45: None of the other menus have a separate close button. Why should this one have it?

You can still put all your behavior into the `onClose()` function.

src/main.zig
```diff
@@ -626,6 +678,7 @@ pub fn main() void { // MARK: main()
 		if(game.world != null) { // Update the game
 			game.update(deltaTime);
 		}
+		Window.Gamepad.update(deltaTime);
```
IntegratedQuantum, line 681: Should be moved into `Window.handleEvents`. Here it's just cluttering the main loop.

IntegratedQuantum (changes requested):
A few more small things, also the game doesn't compile for me anymore. Please always test if everything works, even if you just make seemingly small changes.

catmeow72:
It seems std.math.clamp doesn't support vectors. I must have forgotten to check if it still compiled when making that change

src/gui/windows/controller_mappings_settings.zig
catmeow72: I'll see if this works. When I last tried that, opening a window in onClose would fail, for unknown reasons.

catmeow72: It still does, which is a bug. I think I'll temporarily keep the close button until that is fixed. I might fix it in another PR.

catmeow72: I also can't get the window to open in onClose. I need the button to open the controller mapping download window, when it's needed.

IntegratedQuantum: I guess another question would be: Does this window even make sense? Is the fact that we download the controller mappings even useful information to the average player?

catmeow72: Sorry, your replies didn't show up until I reloaded the window. The controller mapping download window makes sense as when the user has extremely slow download speeds (like me), it might take a few seconds or longer. I'm not entirely sure about the controller mapping download settings window.

catmeow72: I guess it could be used to allow the user to supply custom mappings in the case that their controller isn't in the download.

catmeow72: The controller mappings need to be downloaded as early as possible, otherwise the controller can't be used at all. It affects startup times. The thread is used to prevent the UI from freezing when downloading.

catmeow72: That way there can be an indication that something is actually happening, and the game isn't broken

catmeow72: Perhaps we should download controller mappings without asking, in case the user only wants to use a controller with the game.

catmeow72: It may make sense to download in the background, then. I think I'll have an indication that it's happening just in case, but it'll be kept out of the way.

src/graphics/Window.zig
```diff
@@ -207,32 +214,30 @@ pub const Gamepad = struct {
...
 		std.log.info("Game controller mappings {s}need downloading.", .{if (needsDownload) "" else "do not "});
 		if (needsDownload) {
 			ControllerMappingDownloadTask.schedule(curTimestamp);
 		} else {
 			controllerMappingsDownloaded.store(true, .monotonic);
+                        updateControllerMappings();
 		}
 	}
 	pub fn updateControllerMappings() void {
+		std.log.debug("Updating controller mappings in-memory...", .{});
```
IntegratedQuantum, line 240: Should be log.info, since this data might be useful information for the player.

src/graphics/Window.zig
```diff
@@ -251,12 +256,11 @@ pub const Gamepad = struct {
 		defer main.stackAllocator.free(newData);
 		newData[data.len - 1] = 0;
 		_ = c.glfwUpdateGamepadMappings(newData.ptr);
+		std.log.debug("Controller mappings updated!", .{});
```
IntegratedQuantum, line 259: same here

src/main.zig
```diff
@@ -642,6 +633,7 @@ pub fn main() void { // MARK: main()
 	audio.setMusic("cubyz:cubyz");
 
 	while(c.glfwWindowShouldClose(Window.window) == 0) {
+		openNextStartupWindow();
```
IntegratedQuantum, line 636: Could you explain why you moved this into the loop?
Do we really need to re-check this every frame?
Did you maybe just intend to open both windows at startup?

IntegratedQuantum (changes requested):

src/main.zig
catmeow72: I can't seem to open two windows at once. One will always take priority unless it's in a different frame.

catmeow72: Might have to test on another machine, that doesn't consistently have memory errors

IntegratedQuantum: Well, the last one will be on top, if that's what mean by priority.

src/graphics/Window.zig
```diff
@@ -10,16 +12,283 @@ pub const c = @cImport ({
...
+				.curTimestamp = curTimestamp,
+			};
+			main.threadPool.addTask(task, &vtable);
+		}
+
+		pub fn getPriority(_: *ControllerMappingDownloadTask) f32 {
+			return std.math.inf(f32);
+		}
+
+		pub fn isStillNeeded(_: *ControllerMappingDownloadTask, _: i64) bool {
```
IntegratedQuantum, line 175: The second parameter was just removed in e863e9f311a785f8c98eeb35a0e664de8512dd2e

src/main.zig
catmeow72: If I open one window first, it won't show up. If I open it second, it will.

catmeow72: Even when in entirely different positions

IntegratedQuantum: That is weird, the HUD windows are also opened in a single frame, maybe check what the difference between the code that opens the HUD and your code is.

catmeow72: Seems it's working now. My best guess is memory errors on my machine.

src/main.zig
```diff
@@ -402,6 +421,36 @@ fn isValidIdentifierName(str: []const u8) bool { // TODO: Remove after #480
...
+				openNextStartupWindow();
+			}
+		},
+		.showMainWindow => {
+			startup_status = .finished;
+			gui.openWindow("main");
+		},
+		.finished => { }
+	}
+}
```
IntegratedQuantum, line 452: Do we even need this now? If you open both windows anyways, then I'd prefer to just go back to the previous solution which didn't require this extra function.

catmeow72: I undid that and merged.

IntegratedQuantum:
Ok, I just tested it and one issue I have is that it downloads the controller db, even though I certainly don't have a controller connected.

If I don't have a controller, then I expect the game to not waste resources on downloading the controller mappings.

catmeow72:
I think I'll make it download controller mappings when you plug in a controller, in case you plug one in mid-game, or when the game starts if one is already plugged in.

src/main.zig
```diff
@@ -643,6 +642,10 @@ pub fn main() void { // MARK: main()
 		lastBeginRendering = begin;
 
 		Window.handleEvents(deltaTime);
+		if (Window.Gamepad.areControllerMappingsDownloading() and !gui.isWindowOpen("download_controller_mappings")) {
```
IntegratedQuantum, line 645: Why not just open the window when you start downloading?

catmeow72: I didn't want to add a dependency on the GUI to the Window.Gamepad

catmeow72: I didn't realize it already had access to the GUI, through an existing import

src/graphics/Window.zig
```diff
@@ -260,7 +268,10 @@ pub const Gamepad = struct {
 	}
 
 	pub fn init() void {
-		downloadControllerMappings();
+		controllerConnectedPreviously = isControllerConnected();
+		if (controllerConnectedPreviously) {
+			downloadControllerMappings();
+		}
```
IntegratedQuantum, line 278: Why even run this in the init, if you check it on every update anyways? This should allow you to fix the other problem more smoothly as well.

src/gui/windows/download_controller_mappings.zig
```diff
@@ -0,0 +1,45 @@
...
+const GuiWindow = gui.GuiWindow;
+const Button = @import("../components/Button.zig");
+const CheckBox = @import("../components/CheckBox.zig");
+const Label = @import("../components/Label.zig");
+const VerticalList = @import("../components/VerticalList.zig");
+const HorizontalList = @import("../components/HorizontalList.zig");
+
+pub var window = GuiWindow {
+	.contentSize = Vec2f{128, 64},
+	.pos = Vec2f{0, 0},
```
IntegratedQuantum, line 19: .pos is intended for internal usage only and is overwritten frequently, instead you should init the .relativePosition.

src/gui/windows/download_controller_mappings.zig
```diff
@@ -0,0 +1,45 @@
...
+	.showTitleBar = false,
```
IntegratedQuantum, line 20: This is mainly intended for GUI components. An actual unmovable window, even if it's a status message, seems like it would be rather annoying for the player to deal with.

IntegratedQuantum (changes requested):

src/graphics/Window.zig
```diff
@@ -10,16 +12,294 @@ pub const c = @cImport ({
...
+			const joystickFound = c.glfwJoystickPresent(jid) != 0 and c.glfwJoystickIsGamepad(jid) != 0;
+			if (joystickFound) {
+				if (!gamepadState.contains(jid)) {
+					gamepadState.put(jid, main.globalAllocator.create(c.GLFWgamepadstate)) catch unreachable;
+				}
+				_ = c.glfwGetGamepadState(jid, gamepadState.get(jid).?);
+			} else {
+				if (gamepadState.contains(jid)) {
+					if (c.glfwJoystickPresent(jid) != 0) {
+						std.log.warn("Joystick {d} lost controller mapping!", .{jid});
```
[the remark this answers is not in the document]
catmeow72: It was mostly for debugging. I was just making sure I wasn't losing the controller mappings accidentally. I'll remove it now.

src/gui/windows/download_controller_mappings.zig
```diff
@@ -0,0 +1,45 @@
...
+	.hasBackground = true,
+	.closeable = false
+};
+
+const padding: f32 = 8;
+pub fn update() void {
+	window.relativePosition[0].ratio = 1.0;
+	window.relativePosition[1].ratio = 1.0;
```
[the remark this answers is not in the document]
catmeow72: It opens for me, I see it for a few seconds when controller mappings have to be downloaded.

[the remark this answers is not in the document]
catmeow72: But I've implemented the changes, which will be in the next commit

src/gui/windows/download_controller_mappings.zig
```diff
@@ -16,16 +16,21 @@ const HorizontalList = @import("../components/HorizontalList.zig");
...
-	.closeable = false
+	.closeable = false,
+	.relativePosition = .{
+		.{
+			.ratio = 1.0,
+		},
+		.{
+			.ratio = 1.0
+		}
+	}
```
IntegratedQuantum, line 29: The intended way to move things into one corner would actually be
```
	.relativePosition = .{
		.{ .attachedToFrame = .{.selfAttachmentPoint = .upper, .otherAttachmentPoint = .upper} },
		.{ .attachedToFrame = .{.selfAttachmentPoint = .upper, .otherAttachmentPoint = .upper} },
	},
```
Also please add the trailing commas in multi-line initializations, this makes it easier to move around or add more attributes below.

src/gui/windows/download_controller_mappings.zig
```diff
@@ -16,16 +16,21 @@ const HorizontalList = @import("../components/HorizontalList.zig");
 
 pub var window = GuiWindow {
 	.contentSize = Vec2f{128, 64},
-	.pos = Vec2f{0, 0},
-	.showTitleBar = false,
+	.showTitleBar = true,
```
IntegratedQuantum, line 19: There is no need to add default attributes.

IntegratedQuantum (changes requested):
two more things, and then I think we are finally done here.

IntegratedQuantum (approved):
Alright it's all good now (well there is one more thing which I'll extract into an issue instead of letting this go through another review cycle.)

I need to be honest here: While I am thankful for you contributing and testing controller support, reviewing this PR was not pleasant. For the future I'd suggest you either choose smaller topics, or split up your PR into smaller steps. You could have for example split it like this:
- change keys to support gamepad and add default gamepad keybindings
- handle gamepad input
- download gamepad controller mappings
- add the keybindings menu for gamepad
- controlling the mouse with a gamepad

PR #774, opened 2024-11-13
New water animation

careeoki:
A new and improved water animation, finally. My laptop can't run Cubyz and OBS at the same time so you'll have to look at it ingame yourself. Let me know what you think.
image
Ideally, it could be smoother, but I think it's good enough for something hand animated.

I'll animate lava next, after an animation break.


IntegratedQuantum (approved):
I agree with Ikabod's assessment here. And I'll try to prioritize #509.
Also if I'm not mistaken, this is the last of the old textures 🥳. Great work @careeoki!

PR #783, opened 2024-11-19
Add cloth

careeoki:
image
Cloth is a coloured block using the new carpet model & rotation mode. In the future, it could be used as a crafting material as well.
I'm excited for the building possibilities this new shape allows!
It has item textures too:
image
Also, the palettes of chalk have been made more consistent.

**Known Issues:**

- It's nearly impossible to place more than 2 cloths in a block. The selection box becomes a full cube, and you can't select any more faces inside the block. Also, the huge selection box is obtrusive.
image
-  You can't place cloth on the side of fences, and it also breaks the fence.
- The camera spring overshoots when stepping onto cloth
- You can place cloth vertically on the side of another cloth block, which makes it very difficult to place cloth quickly. Also looks kinda silly.
image
- The texture is rotated in some directions (the texture should always be horizontal) (this also affects blocks with Log rotationmode)
image
- Cloth z-fights where it meets in corners
image
- You should be able to place cloth on the side of itself in the air. So the player could make flags, banners, etc. Like this: (i hid blocks away from the camera for this example)
image


IntegratedQuantum:
Looks nice, I also like the choice of not adding a stupid wool block.

As for the issues, could you please open issues about them? I don't want them to get lost after the PR is merged.

> The camera spring overshoots when stepping onto cloth

About this: Should we even have collision on cloth? I think this would add some nice opportunities for secret rooms and traps, especially if, like you said, they should be able to exist without sticking to a support block.

careeoki:
Will do. 

I think having collision on cloth is nice, I like being able to make short tables or stools with it. 
We could add other blocks more specifically for traps and such.

IntegratedQuantum:
Could you show me how such a table or stool would look?

careeoki:
image

IntegratedQuantum:
Honestly I think carpet should not be rigid, to me a wall carpets would be the perfect place for a secret room.
Maybe we should add stone and wood "carpets" (or whatever you would call them) instead for the purpose of tables?

careeoki:
Alright, done

IntegratedQuantum (approved):
Thank you!

PR #818, opened 2024-12-08
New player model

careeoki:
A new default player that matches the player's hitbox better.
It's some sort of weird gastropod thing.
snailoid
image
image


somanshurath:
Can the effective center of mass of the model be brought a bit lower? That would allow the eyes to be above water in the ocean...

PR #822, opened 2024-12-09
upd: kick notification on client side & fix: no kick option for local user

somanshurath:
fix: dont show kick for local user


somanshurath:
Preview:
Screenshot from 2024-12-09 16-50-17

src/network.zig
```diff
@@ -861,6 +861,17 @@ pub const Protocols = struct {
 			conn.sendUnimportant(id, &noData);
 		}
 	};
+	pub const kick = struct {
```
IntegratedQuantum, line 864: I would prefer to have this as part of the disconnect protocol, you could for example just transfer a one-byte enum with the disconnect reason.

src/network.zig
```diff
@@ -861,6 +861,17 @@ pub const Protocols = struct {
...
+		pub const id: u8 = 21;
```
IntegratedQuantum, line 865: Please don't just use random numbers as protocol ids.

somanshurath:
@IntegratedQuantum as of now we have three possible "disconnect" reasons right?
- kicked
- exit
- connectionError

IntegratedQuantum:
Yes. We could have more in the future though.

src/main.zig
```diff
@@ -417,8 +417,10 @@ pub var lastFrameTime = std.atomic.Value(f64).init(0);
 pub var lastDeltaTime = std.atomic.Value(f64).init(0);
 
 var shouldExitToMenu = std.atomic.Value(bool).init(false);
-pub fn exitToMenu(_: usize) void {
+var exitNotification = std.atomic.Value(network.DisconnectType).init(.exit);
+pub fn exitToMenu(exitData: network.DisconnectType) void {
 	shouldExitToMenu.store(true, .monotonic);
+	exitNotification.store(exitData, .monotonic);
```
IntegratedQuantum, line 423: Having two atomics like this is a problem.
The main thread could see the new value of `shouldExitToMenu`, but still see the old value of `exitNotification`, which causes a small chances of seeing an incorrect exit value.

Could you maybe just use one attomic of `?network.DisconnectType` to cover both in one?

src/network.zig
```diff
@@ -21,6 +21,26 @@ const Vec3f = vec.Vec3f;
 const Vec3i = vec.Vec3i;
 const NeverFailingAllocator = main.utils.NeverFailingAllocator;
 
+pub const DisconnectType = enum(u8) {
```
IntegratedQuantum, line 24: I don't particularly like that this is just declared randomly at the beginning of the file.
Wouldn't it make more sense to move it into the `disconnect` protocol?

src/network.zig
```diff
@@ -1718,18 +1738,18 @@ pub const Connection = struct { // MARK: Connection
...
-		Protocols.disconnect.disconnect(self);
+		Protocols.disconnect.disconnect(self, disconnectType);
 		std.time.sleep(10000000);
-		Protocols.disconnect.disconnect(self);
+		Protocols.disconnect.disconnect(self, disconnectType);
 		self.disconnected.store(true, .unordered);
 		self.manager.removeConnection(self);
 		std.log.info("Disconnected", .{});
-	}
+		}
```
IntegratedQuantum, line 1752: Incorrect indentation

IntegratedQuantum (changes requested):

IntegratedQuantum:
Do you still intend to work on this? It's been a few months since your last change.

IntegratedQuantum:
Closing due to inactivity. If/when you intend to work on this again please fix the merge conflicts and reopen.

PR #825, opened 2024-12-15
Stone Walls

careeoki:
A new block using fence rotation, available in all stone types. Ferrock and Voidstone now have tile blocks.
image
image
image
Unlike Minecraft's wall blocks, these are flat, which makes them more versatile in builds.

**Known Issues**

Light can pass through walls.
image
Either I'm missing something or this is a limitation.


IntegratedQuantum:
The model you have here has 21 faces, which is needlessly inefficient. Ideally, the model should be composed like the fences, with one cuboid for the central post, and two hollow cuboids, one for each side, so 14 faces.
Could you fix this, please?

careeoki:
I have the wall model set up exactly like the fence. I don't want to fix the z-fighting by changing the textures because it would look really bad.
image
The fence has this same z-fighting, you just can't see it. Why can't we just have the UV use the other part of the texture?
image

IntegratedQuantum:
> Why can't we just have the UV use the other part of the texture?

I'll try if I can find a solution for this.

IntegratedQuantum:
Ok, I found a solution, but it obviously breaks the fence model. It would nice if you could include the patch in this PR and also fix the fence model, which is broken with this change.
Here is the patch:
```diff
--- a/src/rotation.zig
+++ b/src/rotation.zig
@@ -219,11 +219,19 @@ pub const RotationModes = struct {
                                }
                                if(!data.isConnectedNegY and corner[1] == 0) {
                                        corner[1] = 0.5;
-                                       cornerUV[0] = 0.5;
+                                       if(@abs(quad.normal[2]) > 0.7) {
+                                               cornerUV[1] = 0.5;
+                                       } else {
+                                               cornerUV[0] = 0.5;
+                                       }
                                }
                                if(!data.isConnectedPosY and corner[1] == 1) {
                                        corner[1] = 0.5;
-                                       cornerUV[0] = 0.5;
+                                       if(@abs(quad.normal[2]) > 0.7) {
+                                               cornerUV[1] = 0.5;
+                                       } else {
+                                               cornerUV[0] = 0.5;
+                                       }
                                }
                        }
                }
```
I have tested this with the previous stone wall model you sent on discord, and there appears to be no distortions or z-fighting.

careeoki:
I've fixed the fence model but I don't know how to apply the patch properly.

IntegratedQuantum (approved):
I inserted it for you. Thank you for fixing the fence model for me!

PR #832, opened 2024-12-17
Added a void to separate the cave world from the root world.

ikabod-kee:
The Void starts at negative 10k blocks, the Root starts at negative 15k blocks. I had to adjust the other cave biomes to ensure they wouldn't spawn below this point.

You might need to make sure the multiplayer world doesn't mess up when adding this change.


careeoki:
I think you should keep void roots at -50000.

careeoki:
When doing a new PR you should make a new branch on your fork. There’s a bunch of old commits here from your music and recipie changes.

careeoki:
Don't worry, it took me awhile to figure out how github worked.

PR #833, opened 2024-12-18
Added a void to separate the cave world from the root world.

### LINKED ISSUE #828: Add an abyss to separate the caves and Root World
image

Abyss starts at -1000
Root World starts at -5000

The presence of the Abyss could pose as a challenge to the player who wants to get to the Root World (and out.) The player will need the right equipment to survive those depths. You could include enemies or platforms if it's too easy to get down there.

It's also a good discovery moment if the player doesn't know the Root World exists.

ikabod-kee:
The Void starts at negative 10k blocks, the Root starts at negative 15k blocks. I had to adjust the other cave biomes to ensure they wouldn't spawn below this point.

You might need to make sure the multiplayer world doesn't mess up when adding this change.

fixes #828


careeoki:
Make sure to put "fixes #828" in the first comment so github can automatically close the issue if it's merged.

careeoki:
I also still think you should keep void roots at -50000, even if just to keep the multiplayer world intact for now. We won't able to balance it correctly until we actually have gameplay to test it with.

assets/cubyz/biomes/cave/void_cavern.zig.zon
```diff
@@ -1,9 +1,9 @@
 .{
 	.isCave = true,
-	.maxHeight = -1024,
-
-	.chance = 0.01,
-
+	.maxHeight = -44750,
```
IntegratedQuantum, line 3: Is it intentional that these cause floaters?
Screenshot at 2024-12-18 19-23-02

assets/cubyz/biomes/cave/void.zig.zon
```diff
@@ -0,0 +1,14 @@
+.{
+	.isCave = true,
+	.maxHeight = -45000,
+	.minHeight = -50000,
+	.stoneBlock = "cubyz:air",
+	.fogDensity = 10,
+	.fogColor = 0x090511,
```
IntegratedQuantum, line 7: I think the fog color should be bit brighter. With this you cannot really see a whole lot.

assets/cubyz/biomes/cave/void_roots.zig.zon
```diff
@@ -1,15 +1,22 @@
...
 	.music = "cubyz:heart-of-the-beast",
 
 	.structures = .{
+		.{
+			.id = "cubyz:simple_vegetation",
+			.chance = 0.001,
+			.block = "cubyz:torch",
+			.height = 1,
+			.height_variation = 0,
+		},
```
IntegratedQuantum, line 19: Wow, what a nice surprise!
You really got me there for a moment, wondering why I had gone to the root in this random world from your fork.

But honestly I think we should make it a bit rarer still.

Also another idea: You could use a 2-block tall tree with a workbench for the logs, a torch for the top and no leaves to create a crafting table with a torch on top.

assets/cubyz/biomes/cave/void_roots.zig.zon
```diff
@@ -19,6 +26,15 @@
...
 		},
+		.{
+			.id = "cubyz:ground_patch",
+			.block = "cubyz:air",
+			.chance = 0.005,
+			.width = 10,
+			.variation = 5,
+			.depth = 5,
+			.smoothness = 0.1,
+		},
```
IntegratedQuantum, line 37: Could you share how these ones look? Sadly I couldn't find any.

assets/cubyz/biomes/cave/crystal.zig.zon
```diff
@@ -1,6 +1,7 @@
 .{
 	.isCave = true,
 	.maxHeight = -512,
+	.minHeight = -44750,
```
IntegratedQuantum, line 4: Honestly I kind of enjoyed the occasional crystal biomes down in the root world.
Would mind adding a copy of it specialized to be in the roots?

IntegratedQuantum (changes requested):
A couple ideas and question.

Also since this new zone of the underworld is getting more and more populated, would you mind adding a new folder for it?
I really would like to see a bunch of mega-rares down there, and I think that would crowd all the regular cave biomes.

assets/cubyz/biomes/cave/void.zig.zon
careeoki: We can make it a bit brighter.

IntegratedQuantum:
Ok, I found a solution, but it obviously breaks the fence model. It would nice if you could include the patch in this PR and also fix the fence model, which is broken with this change.
Here is the patch:
```diff
--- a/src/rotation.zig
+++ b/src/rotation.zig
@@ -219,11 +219,19 @@ pub const RotationModes = struct {
                                }
                                if(!data.isConnectedNegY and corner[1] == 0) {
                                        corner[1] = 0.5;
-                                       cornerUV[0] = 0.5;
+                                       if(@abs(quad.normal[2]) > 0.7) {
+                                               cornerUV[1] = 0.5;
+                                       } else {
+                                               cornerUV[0] = 0.5;
+                                       }
                                }
                                if(!data.isConnectedPosY and corner[1] == 1) {
                                        corner[1] = 0.5;
-                                       cornerUV[0] = 0.5;
+                                       if(@abs(quad.normal[2]) > 0.7) {
+                                               cornerUV[1] = 0.5;
+                                       } else {
+                                               cornerUV[0] = 0.5;
+                                       }
                                }
                        }
                }
```
I have tested this with the previous stone wall model you sent on discord, and there appears to be no distortions or z-fighting.

assets/cubyz/biomes/cave/void_roots.zig.zon
IntegratedQuantum: I don't know, maybe 10× rarer?

assets/cubyz/biomes/cave/void.zig.zon
IntegratedQuantum: no more purple?

IntegratedQuantum: I did kind of like the idea of making it purple. Have you tried a brighter purple? Mabye 0x553c64?
Screenshot at 2024-12-19 17-20-32

assets/cubyz/biomes/cave/void/void_roots.zig.zon
```diff
@@ -12,11 +12,18 @@
 	.structures = .{
 		.{
 			.id = "cubyz:simple_vegetation",
-			.chance = 0.001,
+			.chance = 0.00001,
```
IntegratedQuantum, line 15: I think you added a zero too much here. I flew around for ~30000 blocks and only encountered 2 torches.

IntegratedQuantum (changes requested):
Another thing that I noticed: 5000 blocks is kind of a lot.
Your intention was to make this gap as a thing the player falls into by accident, but 50 seconds of falling is too much. That's enough time to write a bug report.
I'd suggest to make it so that you can clearly see the bottom through the fog within at most 10 seconds.

assets/cubyz/biomes/cave/void.zig.zon
IntegratedQuantum: Hmm, I'm not sure, maybe if you made it a little darker?

IntegratedQuantum: Yeah, I think that's going in the right direction. Maybe a little more saturation?
Also could you share a screenshot with a torch as a reference, (our eyes are too good at filtering out the color)

IntegratedQuantum: I think the second one is good. @careeoki what do you think?

careeoki: Second looks good.

assets/cubyz/biomes/cave/void/void_cavern.zig.zon
```diff
@@ -1,9 +1,9 @@
 .{
 	.isCave = true,
-	.maxHeight = -1024,
-
-	.chance = 0.01,
-
+	.maxHeight = -50000,
+	.chance = 0.1,
+	.fogDensity = 10,
+	.fogColor = 0x1B1333,
```
IntegratedQuantum, line 6: I think you used the wrong fog color in some of the biomes.

IntegratedQuantum:
Also it it intentional that the top of the void roots has this mostly solid layer before the roots begin?

IntegratedQuantum:
It's caused by the noise caves.
The void biome has the default `.caves` parameter which controls the density of the noise caves.
When it meets the top of the void roots, it interpolates the `.caves` parameter between the two.

Now for the void biome you set the stoneBlock to air, which is why we don't see the noise-cave terrain, but right below, at the top of the void roots there is a small section of regular stone where the cave density is still big enough.

An alternative way to implement would be to set the `.caves` parameter to some higher value. This might also give you a less weird transition at the top.

IntegratedQuantum:
I think the artifacts might come from the fact that you are still setting the stoneBlock to air.

IntegratedQuantum:
Setting `.caves = 1` also generates a void without setting the stoneBlock.

IntegratedQuantum:
Yeah, I also noticed.

PR #894, opened 2025-01-07
Update glow crystals to new coloured blocks palette

careeoki:
"New" as of... (checks calendar) 9 months ago.
image
image
image
They now have reflectivity textures which enhances the crystal look a lot.

Reminder to how the colours have been renamed:
gray -> grey
dark_gray -> grey
light_blue -> aqua
dark_red -> crimson
dark_green -> viridian


IntegratedQuantum:
I'm not sure if reflectivity is actually a good fit here, it kind of makes things brighter on the sides, and I'm not so happy about this.
Screenshot at 2025-01-07 23-18-48
But maybe that would need a fix in the actual reflectivity calculation, like this:
Screenshot at 2025-01-07 23-25-39

IntegratedQuantum:
Alright I have decided that this should not block this issue should be resolved by changing the responsible logic and I made an issue for it: #895

assets/cubyz/blocks/glass/indigo.zig.zon
```diff
@@ -6,7 +6,7 @@
 	},
 	.transparent = true,
 	.hasBackFace = true,
-	.absorbedLight = 0xd0c16d,
+	.absorbedLight = 0x2f3e92,
```
IntegratedQuantum, line 9: I don't think this is correct, this makes it absorb more blue light, turning it yellow.

assets/cubyz/blocks/glass/yellow.zig.zon
```diff
@@ -6,7 +6,7 @@
 	},
 	.transparent = true,
 	.hasBackFace = true,
-	.absorbedLight = 0x0d36f1,
+	.absorbedLight = 0xf2c90e,
```
IntegratedQuantum, line 9: same here

IntegratedQuantum (changes requested):
I don't think the glass changes are correct.

assets/cubyz/blocks/glass/yellow.zig.zon
careeoki: I must have accidentally committed these.

careeoki: Okay, reverted the glass

PR #903, opened 2025-01-08
Hot, Dry Biomes (Terrain Revamp 1.0)

ikabod-kee:
Hopefully this is good enough!


assets/cubyz/biomes/ferrock_mountains.zig.zon
```diff
@@ -6,13 +6,23 @@
 	},
 	.minHeight = 80,
 	.maxHeight = 256,
-	
+	.minRadius = 100,
+	.maxRadius = 160,
```
IntegratedQuantum, line 10: Making the radius so small, makes the transitions really harsh:
Screenshot at 2025-01-08 23-20-56
Furthermore it makes the biome selection more random when multiple biomes share the same climate parameters.

assets/cubyz/biomes/savannah.zig.zon
```diff
@@ -0,0 +1,84 @@
...
+		},
+		.{
+			.id = "cubyz:trans/sparse_bushlands",
+			.chance = 1,
+			.width = 2,
+			.keepOriginalTerrain = 1,
+			.properties = .{.neitherWetNorDry, .inland, .land},
+		},
+		.{
+			.id = "cubyz:trans/savannah_widow",
```
IntegratedQuantum, line 72: In my opinion it looks kind of wrong to have a zone without trees there.
Do we really want transition biomes for everything?
Screenshot at 2025-01-08 23-26-40

assets/cubyz/biomes/savannah.zig.zon
```diff
@@ -0,0 +1,84 @@
+.{
+	.properties = .{
+		.dry,
```
IntegratedQuantum, line 3: It doesn't look that dry to me, there are many trees and grass after all.
Also it probably should be hot, no?

assets/cubyz/biomes/prairie.zig.zon
```diff
@@ -0,0 +1,128 @@
+.{
+	.properties = .{
+		.dry
```
IntegratedQuantum, line 3: It doesn't look that dry to me, there are many trees and grass after all.

assets/cubyz/biomes/prairie.zig.zon
```diff
@@ -0,0 +1,128 @@
...
+	.minHeight = 40,
+	.maxHeight = 44,
+	.minRadius = 100,
+	.maxRadius = 160,
+	.roughness = 1,
+	.hills = 3,
+	
+	.music = "cubyz:sunrise",
+
+	.validPlayerSpawn = false,
```
IntegratedQuantum, line 14: It has trees, so I think it should be a valid spawn point.
This variable is to prevent soft-locking the player in an area with insufficient resources to get started.
I think the player should be allowed to spawn in as many biomes as possible, so there is more variety when creating a new world.

assets/cubyz/biomes/prairie.zig.zon
```diff
@@ -0,0 +1,128 @@
...
+	.transitionBiomes = .{
+		.{
+			.id = "cubyz:trans/dry_grassy_coast",
+			.chance = 1,
+			.width = 1,
+			.keepOriginalTerrain = 0.5,
+			.properties = .{.ocean},
+		},
+		.{
+			.id = "cubyz:trans/prairie_widow",
```
IntegratedQuantum, line 28: see comment below

IntegratedQuantum (changes requested):
I still only looked around for a bit looking for obvious problems. This is far from a thorough review.

assets/cubyz/biomes/ferrock_mountains.zig.zon
IntegratedQuantum: It still doesn't seem to be fixed. I still see many places where the "mountain" is just a flat wall:
Screenshot at 2025-01-09 17-54-22

assets/cubyz/biomes/desert.zig.zon
```diff
@@ -25,5 +25,28 @@
...
+	.transitionBiomes = .{
+		.{
+			.id = "cubyz:trans/sandy_coast",
+			.chance = 1,
+			.width = 1,
+			.keepOriginalTerrain = 0.5,
+			.properties = .{.ocean, .wet},
+		},
+		.{
+			.id = "cubyz:trans/drylands",
```
IntegratedQuantum, line 37: I'm also not sold on this biome. It does seem to cause some trouble:
Screenshot at 2025-01-09 18-08-48
Screenshot at 2025-01-09 18-09-12
Screenshot at 2025-01-09 18-10-31

assets/cubyz/biomes/trans/beach.zig.zon
```diff
@@ -0,0 +1,33 @@
+.{
+	.properties = .{
+		.ocean,
+		.wet,
+	},
+	.radius = 128,
```
IntegratedQuantum, line 6: Transition biomes don't need a radius, only sub and normal biomes do, let's not add extra noise with these unnecessary parameters.

assets/cubyz/biomes/trans/beach.zig.zon
```diff
@@ -0,0 +1,33 @@
...
+		.structures = .{
+		.{
+			.id = "cubyz:ground_patch",
+			.block = "cubyz:sand",
+			.chance = 1,
+			.width = 4,
+			.variation = 8,
+			.depth = 4,
+			.smoothness = 0.8,
+		},
```
IntegratedQuantum, line 27: Could you explain why you want to have ground patches that fill the entire biome?

IntegratedQuantum: Yeah, but then we get trees that grow on sand and such.
The biome transition should probably be improved in the code instead.

assets/cubyz/biomes/desert.zig.zon
IntegratedQuantum: Then how about you exclude these for now from this PR, and add them in a future PR, after all the other biome changes have been made?

assets/cubyz/biomes/ferrock_mountains.zig.zon
IntegratedQuantum: The interpolation depends on the biome size. If the biomes are too small, and a radius of 100 is too small, then it can only interpolate on a small range.
If I change the biome size of the desert and the ferrock mountains to 500 then the transition is much smoother:
Screenshot at 2025-01-09 22-12-07

assets/cubyz/biomes/desert.zig.zon
IntegratedQuantum: Could you also exclude the related files from this PR, I don't want to have to review these just now.

assets/cubyz/biomes/savannah.zig.zon
IntegratedQuantum: Well the point is that for this PR you'd only need to touch those biomes to start with.
Here you could probably give the desert and ferrock mountains a barren property, and the savannah could take their spot in the hot and dry and section.
And they really should be hot, to avoid situations like this:
Screenshot at 2025-01-10 19-19-17

IntegratedQuantum (changes requested):
> If there's no more issues, consider merging so I can get back to work adjusting all my biomes for the new lushness parameter.

Of course there is more, what did you expect from such a big PR?
And I still haven't even looked at some of the files.

assets/cubyz/biomes/desert.zig.zon
```diff
@@ -2,16 +2,16 @@
...
-	
+	.minRadius = 150,
+	.maxRadius = 200,
 	.hills = 20,
-	
+	.validPlayerSpawn = false,
 	.music = "cubyz:EasternThought",
-
-	.fogDensity = 20,
-	.fogColor = 0xe0dcc3,
```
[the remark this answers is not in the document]
IntegratedQuantum: Do you have any image of this?

assets/cubyz/biomes/desert.zig.zon
```diff
@@ -25,5 +25,14 @@
 			.height_variation = 3,
 		},
 	},
+	.transitionBiomes = .{
+		.{
+			.id = "cubyz:trans/sandy_coast",
```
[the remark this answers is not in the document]
IntegratedQuantum: Please also remove it here then.

IntegratedQuantum (changes requested):

IntegratedQuantum:
I also feel like I should explain some more why I am so critical of the transition biomes.
As you already discovered transition biomes are rather fragile. In order to make a good transition, you need to know the block composition of the other biome. You seem to have tried to solve this by putting the biomes into mostly distinct climate zones. That certainly works, until we have addons that add new biomes, like take the savannah coast for example. It works well when the beaches are made of sand, but if some addon adds an evil black beach biome, we are in trouble, because now the sand patches aren't fitting anymore:
Screenshot at 2025-01-11 16-39-47
In conclusion, I think transition biomes should only be used under the assumption that there is a different climate (e.g. transitioning to a barren biome could reduce the plant density a little bit), or when a perimeter is explicitly desired (e.g. beaches, or maybe adding a mountain ring around the biome or whatever)

IntegratedQuantum:
> Could be fixed with some sort of custom tag system.

Yeah, that might be possible, but it would make generation times worse proportional to the total number of tags.

Also I still don't think transition biomes really are the right solution. Especially with more tags, it encourages making one transition biome per possible biome tag combination, which is rather tedious. Instead I think we should probably just find a way to make the biome transitions better in general.

assets/cubyz/biomes/sub/butte.zig.zon
```diff
@@ -0,0 +1,38 @@
...
+			.distance= 12,
+			.offset= 6,
+			.width= 1,
+		},
+		.{
+			.direction = .{1, 1, 5},
+			.block= "cubyz:sandstone",
+			.distance= 10,
+			.offset= 6,
+			.width= 1,
```
IntegratedQuantum, line 28: The formatting is also bad in this file.

IntegratedQuantum: No space before `=`

assets/cubyz/biomes/savannah.zig.zon
```diff
@@ -0,0 +1,81 @@
...
+			.width = 1,
+			.keepOriginalTerrain = 0.5,
+			.properties = .{.ocean},
+		},
+		.{
+			.id = "cubyz:trans/savannah_widow",
+			.chance = 1,
+			.width = 2,
+			.keepOriginalTerrain = 1,
+			.properties = .{.dry, .hot, .inland, .land},
```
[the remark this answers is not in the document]
IntegratedQuantum: It still says savannah_widow everywhere throughout the biome:
Screenshot at 2025-01-11 17-53-04

assets/cubyz/biomes/trans/savannah_widow.zig.zon
```diff
@@ -0,0 +1,55 @@
...
+			.chance = 0.05,
+			.type = .round,
+			.height = 1,
+			.height_variation = 7,
+			.leafRadius = 2,
+			.leafRadius_variation = 1,
+		},
+		.{
+			.id = "cubyz:simple_vegetation",
+			.block = "cubyz:cactus",
```
IntegratedQuantum, line 49: Why does the savannah widow have cacti, but not the base biome? I think the cacti look kind of nice in the biome I think it deserves to be in the base biome as well.

assets/cubyz/biomes/savannah.zig.zon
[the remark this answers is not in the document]
IntegratedQuantum: Maybe you forgot to commit it?

assets/cubyz/biomes/trans/savannah_widow.zig.zon
IntegratedQuantum: Argh, you reordered them in the files.

assets/cubyz/biomes/savannah.zig.zon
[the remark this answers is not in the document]
IntegratedQuantum: I tried pulled again, and it's definitely still happening to me.

IntegratedQuantum:
In general I'm not really sold on the widow biomes yet.
In my opinion they just make the biomes here a little more boring at the edges to the desert without changing much visually.
I think this type of biome only makes sense for dense forests, like e.g. the taiga, where you can really see that change in density.

assets/cubyz/biomes/savannah.zig.zon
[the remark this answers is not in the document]
IntegratedQuantum:
```zig
.{
	.properties = .{
		.hot,
		.dry,
	},
	.minHeight = 30,
	.maxHeight = 40,
	.minRadius = 100,
	.maxRadius = 160,
	.hills = 6,
	.validPlayerSpawn = true,
	.music = "cubyz:EasternThought",
	.ground_structure = .{
		"7 to 8 cubyz:limestone",
	},
	.structures = .{
		.{
			.id = "cubyz:ground_patch",
			.block = "cubyz:dirt",
			.chance = 0.33,
			.width = 5,
			.variation = 4,
			.depth = 1,
			.smoothness = 0.1,
		},
		.{
			.id = "cubyz:ground_patch",
			.block = "cubyz:dry_grass",
			.chance = 0.33,
			.width = 5,
			.variation = 4,
			.depth = 1,
			.smoothness = 0.1,
		},
		.{
			.id = "cubyz:simple_tree",
			.leaves = "cubyz:baobab_leaves",
			.log = "cubyz:baobab_log",
			.top = "cubyz:baobab_top",
			.chance = 0.075,
			.type = .round,
			.height = 1,
			.height_variation = 7,
			.leafRadius = 2,
			.leafRadius_variation = 1,
		},
		.{
			.id = "cubyz:flower_patch",
			.block = "cubyz:dry_grass_vegetation",
			.chance = 0.07,
			.width = 5,
			.variation = 8,
			.density = 0.2,
			.priority = 0.1,
		},
		.{
			.id = "cubyz:simple_vegetation",
			.block = "cubyz:cactus",
			.chance = 0.02,
			.height = 1,
			.height_variation = 3,
		},
	},
	.transitionBiomes = .{
		.{
			.id = "cubyz:trans/savannah_coast",
			.chance = 1,
			.width = 1,
			.keepOriginalTerrain = 0.5,
			.properties = .{.ocean},
		},
		.{
			.id = "cubyz:trans/savannah_widow",
			.chance = 1,
			.width = 2,
			.keepOriginalTerrain = 1,
			.properties = .{.dry, .hot, .barren, .lowTerrain, .inland, .land},
		},
	},
	.stoneBlock = "cubyz:sandstone",
}
```

IntegratedQuantum:
What about the prairie widow then?

IntegratedQuantum:
Ok next up, I would like to rethink the file structure.
I think with #857 it would be best to have related biomes in one folder, so e.g. the savannah should be in one folder with all of its sub and transition biomes.

IntegratedQuantum:
As for the coast biomes, am I correct in the assumption that their only purpose is to avoid having plants and vegetation on the cliff thingy?

IntegratedQuantum:
Ok, but then why do all of them look like cliffs?

IntegratedQuantum:
> because by golly is it heart-wrecking having to cut stuff out.

Yeah, I get it. And I'm sorry that we ended up cutting so much out.

But I think it's important to remind yourself that the most valuable stuff still is in this PR:
You added two cool new biomes, with some really nice subbiome terrain features in the new savannah.

IntegratedQuantum:
yes, please

IntegratedQuantum:
It looks like prairie and savannah are now in separate climates. Is that intentional?
Screenshot at 2025-01-11 20-18-53

assets/cubyz/biomes/savannah/savannah.zig.zon
```diff
@@ -0,0 +1,65 @@
+.{
```
IntegratedQuantum, line 1: The file name is now redundant `savannah/savannah` and `savannah/elevated_savannah`.
I'd prefer something like `savannah/base` and `savannah/elevated`

assets/cubyz/biomes/prairie.zig.zon
```diff
@@ -0,0 +1,121 @@
...
+
+	.validPlayerSpawn = true,
+	
+	.ground_structure = .{
+		"1 cubyz:dirt",
+		"3 to 4 cubyz:limestone",
+	},
+	.transitionBiomes = .{
+		.{
+			.id = "cubyz:trans/dry_grassy_coast",
```
IntegratedQuantum, line 22: This biome was removed and now causes a warning.

assets/cubyz/biomes/prairie.zig.zon
```diff
@@ -0,0 +1,112 @@
...
+			.block = "cubyz:limestone",
+			.size = 3,
+			.size_variance = 7,
+		},
+		.{
+			.id = "cubyz:boulder",
+			.chance = 0.0125,
+			.block = "cubyz:limestone",
+			.size = 4,
+			.size_variance = 6,
```
IntegratedQuantum, line 75: (not a change request)
Wow, you put quite some work into changing the random number distribution of the rock size here.
Maybe we should allow parametrizing that somehow?

assets/cubyz/biomes/savannah/elevated.zig.zon
```diff
@@ -0,0 +1,72 @@
+.{
```
IntegratedQuantum, line 1: Maybe this should also be a valid player spawn? It got trees after all.

IntegratedQuantum (approved):
Alright I can't find anymore problems.
Thanks for bearing through the tough review process. And thanks for adding these great new biomes!

PR #907, opened 2025-01-09
Give plants such as daisies some thickness

OneAvargeCoder193:
Im creating this as a draft as im not sure a model with 65 faces is read to be added


careeoki:
From an artstyle standpoint this is too many faces. These will look overly detailed next to full blocks.

OneAvargeCoder193:
> From an artstyle standpoint this is too many faces. These will look overly detailed next to full blocks.

The reason there are so many faces is because there is no other way to get it to look like a voxel mesh without it. In game it just looks like each pixel is a voxel
Screenshot 2025-01-09 164931

careeoki:
> The reason there are so many faces is because there is no other way to get it to look like a voxel mesh without it. In game it just looks like each pixel is a voxel

I don't want these to be made of voxels. They should be flat like they are currently.

OneAvargeCoder193:
There is another option where you can make it just show stems underneath

careeoki:
For stems I would just do small cross models under the flowers.

IntegratedQuantum:
It wouldn't.

careeoki:
> Would drawing lines be an options? The stems are supposed to be rather small anyways.

If the lines were 1 "texture pixel" (not screen pixels) wide that would be good.

PR #926, opened 2025-01-14
Added crouching and block based friction

OneAvargeCoder193:


assets/cubyz/blocks/ice.zig.zon
```diff
@@ -10,4 +10,5 @@
 	.model = "cubyz:cube",
 	.texture = "cubyz:ice",
 	.blockEntity = "cubyz.modding.base.MeltableBlockEntity",
+	.friction = 2,
```
IntegratedQuantum, line 13: I would like to see lower friction, it still seems to stop way too soon.

src/game.zig
```diff
@@ -277,6 +278,43 @@ pub const collision = struct {
...
+			var y: i32 = minY;
+			while (y <= maxY) : (y += 1) {
+				const _block = if(side == .client) main.renderer.mesh_storage.getBlock(x, y, z)
+				else main.server.world.?.getBlock(x, y, z);
+				const min = @max(Vec2d{@floatFromInt(x), @floatFromInt(y)}, vec.xy(boundingBox.min));
+				const max = @min(Vec2d{@floatFromInt(x + 1), @floatFromInt(y + 1)}, vec.xy(boundingBox.max));
+				const area = (max[0] - min[0]) * (max[1] - min[1]);
+				
+				if (_block) |block| {
+					friction += @as(f32, @floatCast(area / areaBottom)) * (1 / block.friction());
```
IntegratedQuantum, line 308: Friction should be averaged over just the contact area, not the total area.
If you stand on a ledge of a block it should give the same friction as if you stand on a full block.

src/game.zig
```diff
@@ -277,6 +278,43 @@ pub const collision = struct {
...
+		
+		const z: i32 = @intFromFloat(@floor(boundingBox.min[2] - 0.01));
+
+		var friction: f32 = 0;
+
+		var x: i32 = minX;
+		while (x <= maxX) : (x += 1) {
+			var y: i32 = minY;
+			while (y <= maxY) : (y += 1) {
+				const _block = if(side == .client) main.renderer.mesh_storage.getBlock(x, y, z)
```
IntegratedQuantum, line 301: Friction should be ignored for non-solid blocks.

IntegratedQuantum (changes requested):
I think you should average the friction differently.

Also @ikabod-kee could you take a look at how these changes feel in-game?

careeoki:
You should be able to step up slabs while crouching, especially since you can still walk down stairs while crouching.

src/game.zig
```diff
@@ -277,6 +278,43 @@ pub const collision = struct {
...
+				else main.server.world.?.getBlock(x, y, z);
+				const min = @max(Vec2d{@floatFromInt(x), @floatFromInt(y)}, vec.xy(boundingBox.min));
+				const max = @min(Vec2d{@floatFromInt(x + 1), @floatFromInt(y + 1)}, vec.xy(boundingBox.max));
```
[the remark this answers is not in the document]
OneAvargeCoder193: I would but im not sure how to calculate it with the triangle.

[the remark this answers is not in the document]
IntegratedQuantum: Hmm, yeah I see the problem.

I think using triangles for collision might have been a bad idea.
I think maybe we should have just used with bounding boxes configured by the artist.

src/game.zig
```diff
@@ -277,6 +278,66 @@ pub const collision = struct {
...
+				while (y <= maxY) : (y += 1) {
+					const _block = if(side == .client) main.renderer.mesh_storage.getBlock(x, y, z)
+					else main.server.world.?.getBlock(x, y, z);
+					const min = @max(Vec2d{@floatFromInt(x), @floatFromInt(y)}, vec.xy(boundingBox.min));
+					const max = @min(Vec2d{@floatFromInt(x + 1), @floatFromInt(y + 1)}, vec.xy(boundingBox.max));
+					const area = (max[0] - min[0]) * (max[1] - min[1]);
+
+					if (_block) |block| {
+						if (block.collide()) {
+							totalArea += area;
```
IntegratedQuantum, line 309: Please merge this with the loop below.
Dividing friction by totalArea after the loop should be enough to do this.

src/game.zig
```diff
@@ -779,6 +872,44 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
+				.min = -Player.outerBoundingBoxExtent,
+				.max = Player.outerBoundingBoxExtent,
+			};
+			Player.eyeBox = .{
+				.min = -Vec3d{Player.outerBoundingBoxExtent[0]*0.2, Player.outerBoundingBoxExtent[1]*0.2, Player.outerBoundingBoxExtent[2] - 0.2},
+				.max = Vec3d{Player.outerBoundingBoxExtent[0]*0.2, Player.outerBoundingBoxExtent[1]*0.2, Player.outerBoundingBoxExtent[2] - 0.05},
+			};
+			Player.desiredEyePos = (Vec3d{0, 0, 1.3 - Player.crouchingBoundingBoxExtent[2]} - Vec3d{0, 0, 1.7 - Player.standingBoundingBoxExtent[2]}) * @as(Vec3f, @splat(smoothPerc)) + Vec3d{0, 0, 1.7 - Player.standingBoundingBoxExtent[2]};
+		}
+		// }
```
IntegratedQuantum, line 911: Another piece of commented code

IntegratedQuantum (changes requested):

IntegratedQuantum (changes requested):
My only complaint is that crouching when jumping feels really weird, and it might give the player the impression that they jump further.

I think the player's crouchPerc should only change when on the ground to fix this.

assets/cubyz/blocks/ice.zig.zon
IntegratedQuantum, line 13: I would still like to have more slippery ice.

src/game.zig
```diff
@@ -911,10 +1024,25 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 		}
 		steppingHeight = @min(steppingHeight, Player.eyePos[2] - Player.eyeBox.min[2]);
 
-		const xMovement = collision.collideOrStep(.client, .x, move[0], Player.super.pos, hitBox, steppingHeight);
+		const slipLimit = 0.00175 * Player.currentFriction;
+		
+		var xMovement = collision.collideOrStep(.client, .x, move[0], Player.super.pos, hitBox, steppingHeight);
 		Player.super.pos += xMovement;
-		const yMovement = collision.collideOrStep(.client, .y, move[1], Player.super.pos, hitBox, steppingHeight);
+		if (KeyBoard.key("crouch").pressed and Player.onGround and @abs(xMovement[0]) < slipLimit) {
```
IntegratedQuantum, line 1031: xMovement is the raw movement done in the current frame. Naturally this is fps dependent.
Better compare slipLimit against the velocity to avoid problems.

src/game.zig
```diff
@@ -911,10 +1024,25 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
+			if (collision.collides(.client, .x, 0, Player.super.pos - Vec3d{0, 0, 1}, hitBox) == null) {
+				Player.super.pos -= xMovement;
+				xMovement[0] = 0;
+			}
+		}
+
+		var yMovement = collision.collideOrStep(.client, .y, move[1], Player.super.pos, hitBox, steppingHeight);
 		Player.super.pos += yMovement;
+		if (KeyBoard.key("crouch").pressed and Player.onGround and @abs(yMovement[1]) < slipLimit) {
```
IntegratedQuantum, line 1040: same here

src/game.zig
```diff
@@ -911,10 +1024,25 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
-		const xMovement = collision.collideOrStep(.client, .x, move[0], Player.super.pos, hitBox, steppingHeight);
+		const slipLimit = 0.00175 * Player.currentFriction;
+		
+		var xMovement = collision.collideOrStep(.client, .x, move[0], Player.super.pos, hitBox, steppingHeight);
 		Player.super.pos += xMovement;
-		const yMovement = collision.collideOrStep(.client, .y, move[1], Player.super.pos, hitBox, steppingHeight);
+		if (KeyBoard.key("crouch").pressed and Player.onGround and @abs(xMovement[0]) < slipLimit) {
+			if (collision.collides(.client, .x, 0, Player.super.pos - Vec3d{0, 0, 1}, hitBox) == null) {
+				Player.super.pos -= xMovement;
+				xMovement[0] = 0;
```
IntegratedQuantum, line 1034: This isn't needed? Also xMovement can be const without this.

src/game.zig
```diff
@@ -911,10 +1024,25 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
+			}
+		}
+
+		var yMovement = collision.collideOrStep(.client, .y, move[1], Player.super.pos, hitBox, steppingHeight);
 		Player.super.pos += yMovement;
+		if (KeyBoard.key("crouch").pressed and Player.onGround and @abs(yMovement[1]) < slipLimit) {
+			if (collision.collides(.client, .y, 0, Player.super.pos - Vec3d{0, 0, 1}, hitBox) == null) {
+				Player.super.pos -= yMovement;
+				yMovement[1] = 0;
```
IntegratedQuantum, line 1043: same here

src/game.zig
```diff
@@ -779,6 +853,45 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
+				if (Player.crouching) {
+					Player.crouchPerc += @floatCast(deltaTime * 10);
+					if (Player.crouchPerc > 1) {
+						Player.crouchPerc = 1;
+					}
+				} else {
+					Player.crouchPerc -= @floatCast(deltaTime * 10);
+					if (Player.crouchPerc < 0) {
+						Player.crouchPerc = 0;
+					}
```
IntegratedQuantum, line 872: Please use min/max or clamp to avoid having these ifs cluttering the code here.

src/game.zig
```diff
@@ -277,6 +278,47 @@ pub const collision = struct {
...
+				const _block = if(side == .client) main.renderer.mesh_storage.getBlock(x, y, z)
+				else main.server.world.?.getBlock(x, y, z);
+				const min = @max(Vec2d{@floatFromInt(x), @floatFromInt(y)}, vec.xy(boundingBox.min));
+				const max = @min(Vec2d{@floatFromInt(x + 1), @floatFromInt(y + 1)}, vec.xy(boundingBox.max));
+				const area = (max[0] - min[0]) * (max[1] - min[1]);
+				
+				if (_block) |block| {
+					if (block.collide()) {
+						totalArea += area;
+						friction += area * @as(f64, @floatCast(1 / block.friction()));
```
IntegratedQuantum, line 309: Could you also explain your rationale behind using an average of inverses?

src/game.zig
```diff
@@ -341,23 +384,29 @@ pub const Player = struct { // MARK: Player
...
+
+	pub var outerBoundingBoxExtent: Vec3d = standingBoundingBoxExtent;
+	pub var outerBoundingBox: collision.Box = .{
+		.min = -standingBoundingBoxExtent,
+		.max = standingBoundingBoxExtent,
 	};
-	const eyeBox: collision.Box = .{
-		.min = -Vec3d{outerBoundingBoxExtent[0]*0.2, outerBoundingBoxExtent[1]*0.2, 0.6},
-		.max = Vec3d{outerBoundingBoxExtent[0]*0.2, outerBoundingBoxExtent[1]*0.2, 0.9 - 0.05},
+	pub var eyeBox: collision.Box = .{
```
IntegratedQuantum, line 405: Why does this need to be public?

src/game.zig
```diff
@@ -341,23 +384,29 @@ pub const Player = struct { // MARK: Player
...
 	pub var jumpCooldown: f64 = 0;
 	const jumpCooldownConstant = 0.3;
 
-	pub const outerBoundingBoxExtent: Vec3d = .{0.3, 0.3, 0.9};
-	pub const outerBoundingBox: collision.Box = .{
-		.min = -outerBoundingBoxExtent,
-		.max = outerBoundingBoxExtent,
+	pub const standingBoundingBoxExtent: Vec3d = .{0.3, 0.3, 0.9};
+	pub const crouchingBoundingBoxExtent: Vec3d = .{0.3, 0.3, 0.725};
+	pub var crouchPerc: f32 = 0;
```
IntegratedQuantum, line 398: Why does this need to be public?

src/game.zig
```diff
@@ -697,9 +747,18 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 		const right = Vec3d{-forward[1], forward[0], 0};
 		var movementDir: Vec3d = .{0, 0, 0};
 		var movementSpeed: f64 = 0;
+		
+		var crouch: bool = false;
```
IntegratedQuantum, line 751: This variable is assigned once and used once, so why not just insert the `KeyBoard.key("crouch").pressed` into the current 1 use case of this variable?

src/game.zig
```diff
@@ -716,16 +775,31 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
 			if(KeyBoard.key("right").value > 0.0) {
-				movementSpeed = @max(movementSpeed, 4*KeyBoard.key("right").value);
-				movementDir += right*@as(Vec3d, @splat(-4*KeyBoard.key("right").value));
+				if (Player.crouching) {
+					movementSpeed = @max(movementSpeed, 2)*KeyBoard.key("right").value;
+					movementDir += right*@as(Vec3d, @splat(-2*KeyBoard.key("right").value));
+				} else {
+					movementSpeed = @max(movementSpeed, 4)*KeyBoard.key("right").value;
+					movementDir += right*@as(Vec3d, @splat(-4*KeyBoard.key("right").value));
+				}
```
IntegratedQuantum, line 802: Instead of duplicating all of the movement logic, why not just introduce a parameter for the one single value you are actually changing.

IntegratedQuantum (changes requested):

src/game.zig
OneAvargeCoder193: It feels smoother. If i don't do it like this it would just abruptly cut off when you walk from ice to grass.

src/game.zig
OneAvargeCoder193: i set the eyebox in the update function when you crouch

src/game.zig
IntegratedQuantum: Yeah, that part is clear to me, you of course need to average it somehow.
But there must be some other reason why it's specifically totalArea/Σ(area/friction) and not (1/totalArea)×Σ(area·friction)

IntegratedQuantum: I just want to understand the math behind it.

OneAvargeCoder193: im pretty sure the friction isn't linear, so i thought it would be the inverse based on what i saw when i tweaked the values

src/game.zig
IntegratedQuantum: It's in the same file, so it doesn't need to be public.

src/game.zig
IntegratedQuantum: Well, the equation I used is pretty linear:
`dv/dt = a - λ·v`
So I think linear averaging would make more sense mathematically.

IntegratedQuantum (changes requested):
It seems that I am slipping off of all blocks now.

src/game.zig
```diff
@@ -911,10 +997,25 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 		}
 		steppingHeight = @min(steppingHeight, Player.eyePos[2] - Player.eyeBox.min[2]);
 
+		const slipLimit = 0.25 * Player.currentFriction;
```
IntegratedQuantum, line 1000: Could you make it so that sprinting up to an edge and then pressing shift still catches you?

OneAvargeCoder193: so just remove the slipping?

IntegratedQuantum: No, just make it so it doesn't affect me as much when I'm not on the ice.

OneAvargeCoder193: it seems you can only do this when going hyperspeed on not slippery blocks though so is it really that big of a concern?
it shouldn't really affect anyone since you can't go hyperspeed in survival and in creative you can just fly back up

IntegratedQuantum: Hmm, yeah, maybe it's just me hitting shift too late before the edge anyways, so I guess it's probably fine.

src/game.zig
```diff
@@ -355,21 +398,30 @@ pub const Player = struct { // MARK: Player
 	pub var inventory: Inventory = undefined;
 	pub var selectedSlot: u32 = 0;
 
+	pub var currentFriction: f32 = 0;
+
+	pub var maxHealth: f32 = 8;
+	pub var health: f32 = 4.5;
```
IntegratedQuantum, line 404: unintended change from conflicts with the health PR?

src/game.zig
```diff
@@ -277,6 +278,47 @@ pub const collision = struct {
...
+		var totalArea: f64 = 0;
+
+		var x = minX;
+		while (x <= maxX) : (x += 1) {
+			var y = minY;
+			while (y <= maxY) : (y += 1) {
+				const _block = if(side == .client) main.renderer.mesh_storage.getBlock(x, y, z)
+				else main.server.world.?.getBlock(x, y, z);
+				const min = @max(Vec2d{@floatFromInt(x), @floatFromInt(y)}, vec.xy(boundingBox.min));
+				const max = @min(Vec2d{@floatFromInt(x + 1), @floatFromInt(y + 1)}, vec.xy(boundingBox.max));
```
IntegratedQuantum, line 303: Instead of assuming that the bounding box is always (0, 0), (1, 1), you should at least get the bounding box of the block model. This would still be less accurate than using the triangle model, but at least it gives us something slightly more accurate.
And while you are at it you could also check if the z coordinate even intersects the bounding box at all, then we'd at least get accurate friction for slabs.

IntegratedQuantum (changes requested):

src/game.zig
```diff
@@ -299,11 +299,23 @@ pub const collision = struct {
...
+						.max = blockPos + @as(Vec3d, @floatCast(main.models.models.items[block.mode().model(block)].max)),
+					};
+
+					if (boundingBox.min[0] > blockBox.max[0] or boundingBox.max[0] < blockBox.min[0] or
+						boundingBox.min[1] > blockBox.max[1] or boundingBox.max[1] < blockBox.min[1] or
+						boundingBox.min[2] > blockBox.max[2] or boundingBox.max[2] < blockBox.min[2]) {
+						continue;
+					}
+
+					const area = (blockBox.max[0] - blockBox.min[0]) * (blockBox.max[1] - blockBox.min[1]);
```
IntegratedQuantum, line 317: You should still clamp the values, just like before, to calculate the true overlapping area.

src/game.zig
```diff
@@ -299,11 +299,23 @@ pub const collision = struct {
...
 				if (_block) |block| {
+					const blockPos: Vec3d = .{@floatFromInt(x), @floatFromInt(y), @floatFromInt(z)};
+
+					const blockBox: Box = .{
+						.min = blockPos + @as(Vec3d, @floatCast(main.models.models.items[block.mode().model(block)].min)),
+						.max = blockPos + @as(Vec3d, @floatCast(main.models.models.items[block.mode().model(block)].max)),
+					};
+
+					if (boundingBox.min[0] > blockBox.max[0] or boundingBox.max[0] < blockBox.min[0] or
+						boundingBox.min[1] > blockBox.max[1] or boundingBox.max[1] < blockBox.min[1] or
```
IntegratedQuantum, line 312: The first two checks aren't needed if you clamp the values anyways.

IntegratedQuantum (changes requested):

src/game.zig
```diff
@@ -308,13 +308,11 @@ pub const collision = struct {
...
 
-					if (boundingBox.min[0] > blockBox.max[0] or boundingBox.max[0] < blockBox.min[0] or
-						boundingBox.min[1] > blockBox.max[1] or boundingBox.max[1] < blockBox.min[1] or
-						boundingBox.min[2] > blockBox.max[2] or boundingBox.max[2] < blockBox.min[2]) {
+					if (boundingBox.min[2] > blockBox.max[2] or boundingBox.max[2] < blockBox.min[2]) {
 						continue;
 					}
 
-					const area = (blockBox.max[0] - blockBox.min[0]) * (blockBox.max[1] - blockBox.min[1]);
+					const area = @max((blockBox.max[0] - blockBox.min[0]) * (blockBox.max[1] - blockBox.min[1]), 0);
```
IntegratedQuantum: You need to clamp the values individually. Just copy your old code for it.

OneAvargeCoder193: i fixed it, accidently clamped the wrong part

src/game.zig
```diff
@@ -277,6 +278,60 @@ pub const collision = struct {
...
+						.min = blockPos + @as(Vec3d, @floatCast(main.models.models.items[block.mode().model(block)].min)),
+						.max = blockPos + @as(Vec3d, @floatCast(main.models.models.items[block.mode().model(block)].max)),
+					};
+
+					if (boundingBox.min[2] > blockBox.max[2] or boundingBox.max[2] < blockBox.min[2]) {
+						continue;
+					}
+
+					const min = @max(vec.xy(blockBox.min), vec.xy(boundingBox.min));
+					const max = @min(vec.xy(blockBox.max), vec.xy(boundingBox.max));
```
IntegratedQuantum, line 316: Actually I think this needs to be a full clamp, otherwise we may end up with max being smaller than min.

IntegratedQuantum (approved):
Looks good now. Thanks for adding this!

Next time I'd prefer it if you would split PRs like this, I'd prefer to see a separate PR for each feature introduced, and here the features crouching and friction are pretty distinct.

PR #950, opened 2025-01-20
Add a skybox

OneAvargeCoder193:
Its not complete yet as the fog now no longer affects transparent blocks far away


Wbjitscool:
ohh what shape is the skybox going to look like

Wbjitscool:
nvm it looks cool

Wbjitscool:
I am just wondering if a moon will be added or something with a little moonlight shining down adding a little bit of light at the night when full moon it adds a bit of light at no moon there's next to almost no light

assets/cubyz/shaders/skybox/fragment.fs
```diff
@@ -0,0 +1,179 @@
...
+	x = vec3(
+		dot(x,vec3(127.1,311.7, 74.7)),
+		dot(x,vec3(269.5,183.3,246.1)),
+		dot(x,vec3(113.5,271.9,124.6))
+	);
+
+	return fract(sin(x)*43758.5453123);
+}
+
+vec3 voronoi_sphere( in vec3 x )
```
IntegratedQuantum, line 105: This function produces rather boring stars, since they are all uniformly distributed on a sphere.
(technically they are not even uniform since I can see artifacts everywhere)

Also I think that rasterization is the better solution here:
Instead of looking for every pixel, which star is closest, you can just draw each star individually.
This allows for arbitrary precomputed star patterns (how about a kind of volumetric disk, similar to the milky way?)

assets/cubyz/shaders/skybox/fragment.fs
```diff
@@ -0,0 +1,179 @@
...
+		0.758
+	);
+
+	color += smoothstep(0.998, 0.999, dot(rayDir, sunDir)) * 20;
+
+	fragColor = vec4(color, 1);
+
+	float brightness = dot(color, vec3(0.2126, 0.7152, 0.0722));
+	if (brightness < 0.02) {
+		vec3 starColor = vec3(smoothstep(0.07, 0.05, voronoi_sphere((invLightDir * vec4(rayDir, 1)).xyz * 50).x));
```
IntegratedQuantum, line 173: It's boring if all stars have the same color. Stars in real life are often slightly red or slightly blue:


Furthermore it would be nice if some stars could have a brightness larger than 1 to get some bloom.

src/renderer.zig
```diff
@@ -606,6 +611,90 @@ pub const MenuBackGround = struct {
...
+		c.glUniformMatrix4fv(uniforms.invLightDir, 1, c.GL_TRUE, @ptrCast(&invLightMatrix));
+		
+		c.glUniformMatrix4fv(uniforms.viewMatrix, 1, c.GL_TRUE, @ptrCast(&viewMatrix));
+		c.glUniformMatrix4fv(uniforms.projectionMatrix, 1, c.GL_TRUE, @ptrCast(&game.projectionMatrix));
+
+		c.glBindVertexArray(vao);
+		c.glDrawElements(c.GL_TRIANGLES, 36, c.GL_UNSIGNED_INT, null);
+
+		c.glEnable(c.GL_CULL_FACE);
+		c.glEnable(c.GL_DEPTH_TEST);
```
IntegratedQuantum, line 694: Please encapsulate this function in `gpu_performance_measuring.start/stopQuery` of a new type, so it shows up in the GPU perfomance debug menu (F5).

assets/cubyz/shaders/skybox/fragment.fs
```diff
@@ -0,0 +1,179 @@
...
+	float c = dot(r0, r0) - (sr * sr);
+	float d = (b*b) - 4.0*a*c;
+	if (d < 0.0) return vec2(1e5,-1e5);
+	return vec2(
+		(-b - sqrt(d))/(2.0*a),
+		(-b + sqrt(d))/(2.0*a)
+	);
+}
+
+vec3 atmosphere(vec3 r, vec3 r0, vec3 pSun, float iSun, float rPlanet, float rAtmos, vec3 kRlh, float kMie, float shRlh, float shMie, float g) {
```
IntegratedQuantum, line 28: What are those names? What is a `shMie`?
And no, copying code is no excuse for shitty naming.

IntegratedQuantum (changes requested):
I hope I don't sound too harsh in the following review, but I do not like the shader code at all.

It looks nicer than what we have right now, but the performance cost is insane.
When looking at the sky my frame time is ~18 ms at day and ~22ms at night.
This is unacceptable.
(And don't forget to leave a dislike on the shadertoys that you copied from)

Also I think we can do some improvements to the stars as well.

Lastly, this shader code is unreadable, thanks to shitty variable names, different formatting everywhere and random unnamed constants to make sure no one knows what's actually going on. (not to mention that the one named constant isn't even used)

PR #967, opened 2025-01-26
The Great Ocean Revamp

ikabod-kee:
Oceans now dominate the sea!
Edit: .... yeah duh

### DIFF
--- a/assets/cubyz/biomes/ocean/shallows.zig.zon
+++ b/assets/cubyz/biomes/ocean/shallows.zig.zon
@@ -30,7 +30,7 @@
 			.id = "cubyz:ocean/cliffs",
 			.chance = 0.075,
 			.width = 2,
-			.properties = .{.antiMountain, .ocean},
+			.properties = .{.antiMountain},
 		},
 		.{
 			.id = "cubyz:ocean/beach",


assets/cubyz/biomes/ocean/base.zig.zon
```diff
@@ -38,14 +37,9 @@
 	},
 	.transitionBiomes = .{
 		.{
-			.id = "cubyz:beach",
-			.chance = 1,
-			.width = 1,
-			.properties = .{.land, .inland},
-		},
-		.{
-			.id = "cubyz:ocean_shelf",
```
IntegratedQuantum, line 47: Could your explain your reasoning for removing the ocean shelf? It does seem to make the transition a lot more extreme, Removing it seems to make the transition pretty harsh in some cases:
Screenshot at 2025-01-26 15-58-27
keepOriginalTerrain seems to help a little bit in some cases, but it doesn't appear to reliably solve the issue.

assets/cubyz/biomes/ocean/beach.zig.zon
```diff
@@ -0,0 +1,27 @@
...
+		.structures = .{
+		.{
+			.id = "cubyz:ground_patch",
+			.block = "cubyz:sand",
+			.chance = 1,
+			.width = 4,
+			.variation = 8,
+			.depth = 4,
+			.smoothness = 0.8,
+		},
```
IntegratedQuantum, line 22: Once again, could you avoid the ground patch spam in favor of making a better biome transition in the first place?

Using structures for this is really not cheap.

assets/cubyz/biomes/ocean/cliffs.zig.zon
```diff
@@ -0,0 +1,38 @@
...
+	.properties = .{
+		.ocean,
+		.wet,
+		.mountain,
+	},
+	.radius = 128,
+	.minHeight = -1,
+	.maxHeight = 5,
+	.chance = 0.25,
+	.mountains = 50,
```
IntegratedQuantum, line 11: The cliffs that stick out of the water do appear rather flat/round. Maybe more mountain noise would be good?

IntegratedQuantum (changes requested):
The climate parameters don't appear well thought out.
The base ocean is `antiMountain`, the shallow ocean is `flatland` and the cliffs are `mountain`
Based on how the climate generator tries to separate these three tags, we end up always seeing cliffs next to shallow ocean. This seems unintentional.

Also why did you remove the islands?

assets/cubyz/biomes/ocean/base.zig.zon
IntegratedQuantum: It made the transition smoother by introducing an extra biome slightly below the water surface.

IntegratedQuantum:
> With this new ocean system, islands are automatically generated as individual biomes. I might add them back at some point as little rare biomes.

Do you have an example of this, because I don't think it really works without subbiomes.

assets/cubyz/biomes/ocean/shelf.zig.zon
```diff
@@ -1,18 +1,15 @@
 .{
 	.properties = .{
 		.ocean,
+		.wet,
 	},
-	.radius = 500,
-	.minHeight = -4,
+	.minHeight = -2,
 	.maxHeight = 0,
```
IntegratedQuantum, line 7: Maybe also set maxHeight to -2 for a more consistent transition?

assets/cubyz/biomes/ocean/shallows.zig.zon
```diff
@@ -1,53 +1,49 @@
...
+			.id = "cubyz:ocean/beach",
 			.chance = 1,
 			.width = 1,
 			.properties = .{.land, .inland},
 		},
 		.{
-			.id = "cubyz:ocean_shelf",
+			.id = "cubyz:ocean/shelf",
 			.chance = 1,
+			.keepOriginalTerrain = 0.1,
```
IntegratedQuantum, line 44: I think this might break the transition in some cases. But you should test it out if that really is the cause.

assets/cubyz/biomes/ocean/shallows.zig.zon
```diff
@@ -1,53 +1,49 @@
...
-	.minHeight = -100,
-	.maxHeight = -22,
-	.radius = 500,
-	
-	.roughness = 20,
-	.hills = 10,
-
+	.radius = 128,
+	.minHeight = -64,
+	.maxHeight = -32,
```
IntegratedQuantum, line 8: The shallow biome is kind of difficult to see in the rest of the ocean, maybe it should be shallower?

IntegratedQuantum (changes requested):
I think even with the shelf, there are still some cases where the beaches are kind of messy, or falling off too quickly.

PR #979, opened 2025-01-29
Update ores in prep for #611 

careeoki:
Ore textures have been redesigned for better contrast against most stone types, and now are a transparent overlay.
Generally, the ore chunks are larger, and have strong reflectivity textures, which improves legibility a lot.
image
It doesn't always look perfect, but I think this is the best approach since it allows us to easily add new stones or ores without having to create an exponential amount of unique textures.
image

I added temporary "stone ore" textures which can be used until #611 is done. After that, they can be deleted.

Raw iron is an item dropped from iron ore. This makes iron ore consistent with coal ore and diamond ore (drops an item instead of the ore block), as well as saving us the trouble of worrying what ore blocks look like in the inventory.
The iron ingot recipe is now coal+raw iron
image
Also coal and iron ingot have been reshaded. 
Iron blocks match the colour of the ingot better and are less reflective


assets/cubyz/items/raw_iron.zig.zon
```diff
@@ -0,0 +1,3 @@
+.{
+	.texture = "raw_iron.png",
+}
```
IntegratedQuantum, line 3: I would like to allow you to put the ore back into the stone by right clicking it with the item.
To do that it needs to be a block item. You can still overwrite the texture of it of course.

IntegratedQuantum (changes requested):
If you don't mind I would like to try and fix #611 before merging. It shouldn't take too long.

assets/cubyz/items/raw_iron.zig.zon
careeoki: That's a good idea, I'll do that.

careeoki:
> If you don't mind I would like to try and fix #611 before merging. It shouldn't take too long.

Should i delete the temporary textures then?

IntegratedQuantum:
> Should i delete the temporary textures then?

Depends on how fast I can get it done.

assets/cubyz/items/raw_iron.zig.zon
careeoki: Should coal and diamond be made into block items aswell?

IntegratedQuantum:
I also want to point out that, since ores are rare, it would be totally viable to add more complex shaders, so if you've got any ideas how I can help you add more details to the surrounding stone with new blending modes or palettization stuff, then I'd be ready to add a new shader for it.

careeoki:
Some way of adding shading to the stone would look way better and improve legibility a lot. Mainly I would want to add a shadow to show how the ore chunks are embedded into the stone

Hard Light lets me add some shading on different stones.
image
Though, it looks kinda messy sometimes like on the marble here. Maybe some way to choose what the shade color is for different stones? I don't know.

careeoki:
> I do have some criticisms about the iron ore texture. It seems far too abundant, if that makes any sense. All the other textures look awesome.

I intentionally made it more dense so it was more visible with the current implementation. If we get a new shader for ores, I would revisit it.

> Also, having a shader to handle the depth of the ore is a cool idea. 

I don't mean any actual depth, just lighting baked into the texture.

careeoki:
> Also, you could consider making iron ore its _own_ type of stone, like how it is in real life (think hematite and etc.)

A distinct "ore" block makes it more clear to the player. If it's just a generic looking stone, how does the player know it's something important?
(Also, we have already have ferrock, which is ferrous)

assets/cubyz/items/raw_iron.zig.zon
careeoki: Raw iron is a block item now. I dummied it out to just have an item texture since I'm not sure how you want to set it up.

IntegratedQuantum:
I now fixed #611, maybe it helps being able to test it in-game.

careeoki:
> I now fixed #611, maybe it helps being able to test the new textures in-game. (They don#t support transparency at the moment)

That's awesome, thanks.
They're already designed around not having transparency, so that's fine

IntegratedQuantum:
And you can remove those temporary textures now too.

careeoki:
> And you can remove those temporary textures now too.

Done. Is there anything I should do about raw iron being placeable? Or should I just leave it for now?

assets/cubyz/blocks/raw_iron.zig.zon
```diff
@@ -0,0 +1,5 @@
+.{
+	.item = .{
+		.texture = "raw_iron.png",
+	},
+}
```
IntegratedQuantum, line 5: Just put this into iron_ore.zig, so that it's associated with the ore block.
We don't need this to be a separate thing.

IntegratedQuantum (changes requested):

assets/cubyz/blocks/raw_iron.zig.zon
careeoki: Okay, done.

IntegratedQuantum (approved):
Looks good now.
Thanks for reminding me of #611 so we could get this done before the beginning the survival tests.

PR #986, opened 2025-01-31
Raise default stackSize to 120

careeoki:
I saw Quantum suggest this number a few weeks ago. 
Obviously, a larger stack size lets you carry more stuff. It also leads to less slots of the same item (ex. 4 slots of stone -> 2 slots of stone)

I tried it in survival for a little while and it felt nice. I was able to keep my torches in one slot, for example.
image


careeoki:
> 120 might be nice, since it's divisible by 2, 3, 4, 5, 6, 8 and 10.

IntegratedQuantum (approved):
Alright let's try it

PR #1009, opened 2025-02-03
Jungle Revamp

ikabod-kee:
(Work in progress. I don't know how to add Ivy to the trees, and it could use some more details like boulders.)

### DIFF
--- a/assets/cubyz/biomes/jungle/base.zig.zon
+++ b/assets/cubyz/biomes/jungle/base.zig.zon
@@ -17,6 +17,27 @@
 	},
 	.structures = .{
 		.{
+			.id = "cubyz:boulder",
+			.chance = 0.025,
+			.block = "cubyz:basalt",
+			.size = 3,
+			.size_variance = 7,
+		},
+		.{
+			.id = "cubyz:boulder",
+			.chance = 0.05,
+			.block = "cubyz:cobblestone",
+			.size = 2,
+			.size_variance = 9,
+		},
+		.{
+			.id = "cubyz:boulder",
+			.chance = 0.1,
+			.block = "cubyz:mossy_cobblestone",
+			.size = 3,
+			.size_variance = 5,
+		},
+		.{
 			.id = "cubyz:flower_patch",
 			.block = "cubyz:lush_grass_vegetation",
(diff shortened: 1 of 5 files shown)


careeoki:
Wow, these canopies are very impressive.
I don't think there's any way to add ivy to trees.

There's some areas like this that are dark (which is cool), but if the sun is being blocked, grass shouldn't be growing here. Maybe you could add mud patches, or seperate the jungle into sparse (with smaller trees) and dense (with larger, shadier trees) biomes?
Mud on the ground would be a good opportunity to add flower patches of Ivy as ground cover.
image


I think a thicker layer of mud would be nice.
Any reason it's a small stripe of basalt instead of it being the biome's stoneBlock?
image

careeoki:
Oh yeah, that's awesome! Just like what I was imagining.
I do think it could be a bit brighter though. It's completely pitch-dark in places, which again, is cool, but I think would be better suited for another biome. Some kind of spooky, evil forest.
You also could add a few lush grass plants as well, to add some variation.

careeoki:
By making the leaves smaller i was able to have some light shining through without making it too bright.

image

image
[had to make this a txt to upload](https://github.com/user-attachments/files/18649594/thicket.zig.txt)

careeoki:
> I'll insert your version soon. I'm doing things currently.

Actually, I'll just do a pull request on your branch, then you can just click the button.

careeoki:
Did the PR, check it out when you have the chance.

IntegratedQuantum (changes requested):
Before we go any further:
1. Please put all the remaining file moves (including this one) into a separate PR. I don't want to end up fixing my worlds once every week for the next months. And github also doesn't seem to detect it properly if you make too many changes afterwards, making it harder to review.
2. You should consider using _defaults.zig.zon. This is especially useful here where many structures and parameters are shared between all 3 files.

IntegratedQuantum:
> Unsure what this means

In every PR there is one file you move into a folder and rename to base.
This changes the ID of the biome which breaks older worlds that used the old ID. The effect of this is usually that the biome turns into stone, and you should also get the new error popup every time after joining.

PR #1026, opened 2025-02-10
Wetlands and Swamp Revamp 2!

ikabod-kee:


careeoki:
I think they could be distinct enough, I imagine the wetlands as primarly a pond while the Swamp has some dry ground and denser plant life.
You should try adding clay to the wetlands also.

careeoki:
Yeah, that looks better. They feel very distinct.
A lot of the swamp plants are leaking into the pond sub-biome. Not much you can do about that right now though.

I think the wetlands should be sparser, with less "thicket" areas, a bit more like how the wetlands looked before this PR.
The density of wetlands/base is where I would want it.

IntegratedQuantum (changes requested):
Please rebase onto the latest master version and fix the resulting formatting errors.

PR #1028, opened 2025-02-10
Add Ice Caves!

ikabod-kee:


careeoki:
> I tried this too! It caused a lot of lag for some reason, 

You could try to reduce the amount faces by the ground patches having no roughness and only having a depth of one.

IntegratedQuantum (approved):
Wow, these caves are truly challenging to traverse.

PR #1043, opened 2025-02-14
Add stars and a simple skybox

OneAvargeCoder193:


careeoki:
This looks really cool.

The stars are really tiny. I think it would fit the rest of the game's art style better if they were a bit bigger. When i put the resolution scale at 50% they feel about the right size.

Are you able to make the day sky more of a natural sky blue? It still has the weird desaturated indigo colour currently in the base game.

src/renderer.zig
```diff
@@ -606,6 +610,312 @@ pub const MenuBackGround = struct {
...
+	var skyUniforms: struct {
+		viewMatrix: c_int,
+		projectionMatrix: c_int,
+		skyColor: c_int,
+	} = undefined;
+
+	var skyVao: c_uint = undefined;
+	var skyVbos: [2]c_uint = undefined;
+
+	const NUM_STARS = 100000;
```
IntegratedQuantum, line 634: constants should have the same case as all other variables.

assets/cubyz/shaders/skybox/star.vs
```diff
@@ -0,0 +1,16 @@
+#version 430
+
+in vec3 vPos;
+in vec3 vColor;
+
+uniform mat4 projectionMatrix;
+uniform mat4 viewMatrix;
+uniform mat4 modelMatrix;
```
IntegratedQuantum, line 8: Isn't it just a position offset? We shouldn't do another matrix multiplication, if an addition is enough.

assets/cubyz/shaders/skybox/star.vs
```diff
@@ -0,0 +1,16 @@
...
+
+out vec3 color;
+
+void main() {
+	gl_Position = projectionMatrix*viewMatrix*modelMatrix*vec4(vPos, 1);
```
IntegratedQuantum, line 13: Also, please precompute `projectionMatrix*viewMatrix` on the CPU, to avoid repeating this calculation 100000 times.

src/renderer.zig
```diff
@@ -182,6 +184,8 @@ pub fn renderWorld(world: *World, ambientLight: Vec3f, skyColor: Vec3f, playerPo
 
 	const time: u32 = @intCast(std.time.milliTimestamp() & std.math.maxInt(u32));
 
+	Skybox.render();
```
IntegratedQuantum, line 187: Please add a `gpu_performance_measuring` query for that, so we can see its performance impact in the F5 debug menu.

src/renderer.zig
```diff
@@ -606,6 +610,312 @@ pub const MenuBackGround = struct {
...
+		return rgb;
+	}
+
+	fn init() void {
+		starShader = Shader.initAndGetUniforms("assets/cubyz/shaders/skybox/star.vs", "assets/cubyz/shaders/skybox/star.fs", "", &starUniforms);
+		starShader.bind();
+
+		var starData: [NUM_STARS * 6]f32 = undefined;
+
+		var starRandom = std.Random.DefaultPrng.init(0);
```
IntegratedQuantum, line 804: Please use main.random, instead of Zig's random number interface.
main.random should be faster, I think we should use the same random number solution everywhere.

src/renderer.zig
```diff
@@ -606,6 +610,312 @@ pub const MenuBackGround = struct {
...
+
+			const radius: f64 = starRandom.random().floatExp(f64) * 4 + 0.2;
+
+			const temperature: f64 = (@abs(starRandom.random().floatNorm(f64) * 3000.0 + 5000.0) + 1000.0) / 5772.0;
+
+			const luminosity = 4.0 * std.math.pi * radius * radius * temperature * temperature * temperature * temperature;
+
+			const flux = luminosity / (4.0 * std.math.pi * 1.36e+7 * vec.length(pos) * vec.length(pos));
+
+			const magnitude = -2.5 * @log10(flux) - 26.83;
```
IntegratedQuantum, line 817: Do we really need to drag around all this physics, if we could just choose a random brightness to begin with?

src/renderer.zig
```diff
@@ -606,6 +610,312 @@ pub const MenuBackGround = struct {
...
+		const b = @as(Vec3d, @splat(12.92));
+		const gamma = 1.0 / 2.4;
+
+		const mask = rgb > @as(Vec3d, @splat(threshold));
+		const corrected = a * Vec3d {
+			std.math.pow(f64, rgb[0], gamma),
+			std.math.pow(f64, rgb[1], gamma),
+			std.math.pow(f64, rgb[2], gamma),
+		} - @as(Vec3d, @splat(0.055));
+		rgb = @select(f64, mask, corrected, rgb * b);
```
IntegratedQuantum, line 784: Instead of doing all these complex calculations, I think it would be best to just read the temperature→color conversion map from a texture instead.
This would allow us to use physically inaccurate colors (green, purple) and might be nice for texture packs and events too (imagine a terraria-like blood-moon with red stars).

IntegratedQuantum (changes requested):
I also agree with carrie's statement:

> The stars are really tiny. I think it would fit the rest of the game's art style better if they were a bit bigger. When i put the resolution scale at 50% they feel about the right size.

Furthermore I think it would be good if we could have the same total brightness independently of the screen resolution.

Also I think we shouldn't draw stars as single distinct pixels, because it creates weird aliasing on movement(stars jitter around a bit in a weird way).

assets/cubyz/shaders/skybox/star.vs
OneAvargeCoder193: the model matrix is a rotation

src/renderer.zig
OneAvargeCoder193: I can't find any ways to use some of the functions i need, mainly floatExp and floatNorm

assets/cubyz/shaders/skybox/star.vs
IntegratedQuantum: But why?

src/renderer.zig
IntegratedQuantum: Then add those functions

IntegratedQuantum:
Please rebase the PR, and also fix all the formatting issues that will likely follow.

src/renderer.zig
```diff
@@ -752,18 +752,9 @@ pub const Skybox = struct {
...
 
-		const rawData = [_]f32 {
-			-1, -1, -1,
-			1, -1, -1,
-			1, 1, -1,
-			-1, 1, -1,
-			-1, -1, 1,
-			1, -1, 1,
-			1, 1, 1,
-			-1, 1, 1
```
IntegratedQuantum: I think it should be kept in separate lines, you can tell the formatter to keep it if you add a trailing comma after the last element.

assets/mymod/mod/main.zig
```diff
@@ -0,0 +1,5 @@
+const std = @import("std");
+
+export fn add(a: i32, b: i32) i32 {
+    return a + b;
+}
```
IntegratedQuantum, line 5: What is this doing here?

assets/cubyz/items/textures/torch.zig.zon
```diff
@@ -0,0 +1,4 @@
+.{
+	.frames = 4,
+	.time = 100,
+}
```
IntegratedQuantum, line 4: And this too

Argmaster:
@OneAvargeCoder193 what is the status of the PR? We could really use some more convincing sky...

assets/cubyz/shaders/skybox/star.fs
```diff
@@ -0,0 +1,16 @@
...
+
+uniform float starOpacity;
+
+layout (location = 0, index = 0) out vec4 fragColor;
+
+void main() {
+	if (distance(pos, starPos) > 1.0/3.0)
+		discard;
+
+	fragColor = vec4(color * starOpacity, 1);
```
IntegratedQuantum, line 15: This could be done in the vertex shader.

assets/cubyz/shaders/skybox/star.fs
```diff
@@ -0,0 +1,16 @@
...
+in vec3 pos;
+in vec3 starPos;
+in vec3 color;
+
+uniform float starOpacity;
+
+layout (location = 0, index = 0) out vec4 fragColor;
+
+void main() {
+	if (distance(pos, starPos) > 1.0/3.0)
```
IntegratedQuantum, line 12: Please use distance square to avoid the square root.

IntegratedQuantum: Also where does the hardcoded 1/3 come from?

assets/cubyz/shaders/skybox/star.vs
```diff
@@ -0,0 +1,27 @@
+#version 430
+
+in vec3 vPos;
+in vec3 vColor;
+
+struct star {
+	vec4 pos;
+	vec4 col;
```
IntegratedQuantum, line 8: Why not use vec3 to save some bandwidth?

assets/cubyz/star.png
IntegratedQuantum, line 1: Is it just me, or is the first pixel white?

main.dll.obj
IntegratedQuantum, line 1: What is this random object file?

assets/cubyz/shaders/skybox/star.vs
```diff
@@ -0,0 +1,27 @@
+#version 430
+
+in vec3 vPos;
+in vec3 vColor;
```
IntegratedQuantum, line 4: Also why do you have these twice, once in the VAO, and once in the star buffer?

IntegratedQuantum (changes requested):
Also, why not have stars of varying radius?

assets/cubyz/shaders/skybox/star.fs
OneAvargeCoder193: its the radius of an equilateral triangle's incircle i think

assets/cubyz/shaders/skybox/star.vs
OneAvargeCoder193: it broke due to memory alignment issues

assets/cubyz/star.png
OneAvargeCoder193: ?

OneAvargeCoder193: ah, that was due to an issue when calculating, but it never is actually accessed in the code

assets/cubyz/shaders/skybox/star.fs
IntegratedQuantum: [That does not seem correct](https://en.wikipedia.org/wiki/Equilateral_triangle#Relationship_with_circles)

assets/cubyz/shaders/skybox/star.vs
IntegratedQuantum: Then you could use manual fields instead of `@Vector`

assets/cubyz/star.png
IntegratedQuantum: Well in my opinion all pixels of the texture should be accessed.

assets/cubyz/shaders/skybox/star.vs
IntegratedQuantum: on the cpu side

IntegratedQuantum (changes requested):

assets/cubyz/shaders/skybox/star.vs
```diff
@@ -0,0 +1,29 @@
+#version 430
+
+in vec3 vPos;
+
+struct star {
+	vec3 pos;
+	float pad1;
```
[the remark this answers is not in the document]
OneAvargeCoder193: it doesn't work without the padding.

src/renderer.zig
```diff
@@ -605,6 +611,245 @@ pub const MenuBackGround = struct {
...
 
+pub const Skybox = struct {
+	var starShader: Shader = undefined;
+	var starUniforms: struct {
+		mvp: c_int,
+		starOpacity: c_int,
+	} = undefined;
+
+	var starVao: c_uint = undefined;
+	var starVbo: c_uint = undefined;
```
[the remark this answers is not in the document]
OneAvargeCoder193: ?

src/renderer.zig
```diff
@@ -605,6 +611,245 @@ pub const MenuBackGround = struct {
...
+		const triVertC = Vec3f{0.0, starDist, 1.0 - off};
+
+		var seed: u64 = 0;
+
+		for(0..numStars) |i| {
+			var pos = getStarPos(&seed);
+
+			var radius: f64 = @floatCast(main.random.nextFloatExp(&seed)*4 + 0.2);
+
+			var temperature: f64 = @floatCast((@abs(main.random.nextFloatGauss(&seed)*3000.0 + 5000.0) + 1000.0)/5772.0);
```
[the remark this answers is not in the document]
OneAvargeCoder193: Its the temperature of the sun, so when i multiply it 0 would be 0, 1 would be the sun temperature, 2 would be 2 sun temperatures

src/renderer.zig
```diff
@@ -605,6 +611,245 @@ pub const MenuBackGround = struct {
...
+
+			var light = (3966.91*radius*radius*temperature*temperature*temperature*temperature)/(vec.dot(pos, pos));
```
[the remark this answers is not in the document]
OneAvargeCoder193: Its the value i got when i simplified the math

src/renderer.zig
```diff
@@ -605,6 +611,245 @@ pub const MenuBackGround = struct {
...
+		const rgbCol = image.getRGB(@intFromFloat(std.math.clamp(temperature, 1000, 14999)), 0);
+		var rgb: Vec3d = @floatFromInt(Vec3i{rgbCol.r, rgbCol.g, rgbCol.b});
+		rgb /= @splat(255.0);
+
+		rgb *= @as(Vec3d, @splat(light));
+
+		const m = @reduce(.Max, rgb);
+		if(m > 1.0) {
+			rgb /= @as(Vec3d, @splat(m));
+		}
```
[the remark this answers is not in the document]
OneAvargeCoder193: the reason i did this is because the bloom flickers a lot for the stars

assets/cubyz/shaders/skybox/star.vs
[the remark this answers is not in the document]
IntegratedQuantum: Well, of course you have to remove it on the CPU side too.

src/renderer.zig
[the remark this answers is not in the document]
IntegratedQuantum: Instead of having a vertex buffer at all, you'd just put a vec3[3] into the ssbo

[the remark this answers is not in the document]
OneAvargeCoder193: how would i render the mesh then?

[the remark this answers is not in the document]
IntegratedQuantum: Just bind an empty vao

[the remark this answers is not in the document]
OneAvargeCoder193: im confused on how this would work, how could i bind an empty vao? where would it get the vertices from if not from the vertex attributes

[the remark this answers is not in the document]
IntegratedQuantum: It gets the number of vertices from the drawArrays call, and the vertex data itself is all inside the ssbo, which you then manually fetch in the vertex shader.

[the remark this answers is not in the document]
OneAvargeCoder193: how would i access it in the vertex shader?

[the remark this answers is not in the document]
OneAvargeCoder193: would i do something like starData[gl_VertexID/3].vPos[gl_VertexID%3]?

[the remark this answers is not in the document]
IntegratedQuantum: yes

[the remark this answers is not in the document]
OneAvargeCoder193: done

Argmaster:
```
Error Linking Shader program:
Link info
---------
error: interpolation modifier mismatch for varying parameter (named starPos) between shader stages
error: interpolation modifier mismatch for varying parameter (named color) between shader stages


OpenGL API error: GL_INVALID_OPERATION error generated. <program> object is not successfully linked, or is not a program object.
OpenGL API error: GL_INVALID_OPERATION error generated. <program> object is not successfully linked, or is not a program object.
OpenGL API error: GL_INVALID_OPERATION error generated. <program> has not been linked, or is not a program object.
OpenGL API error: GL_INVALID_OPERATION error generated. <program> object is not successfully linked, or is not a program object.
```

assets/cubyz/shaders/skybox/star.fs
```diff
@@ -0,0 +1,14 @@
+#version 430
+
+in vec3 pos;
+in vec3 starPos;
+in vec3 color;
```
Argmaster, line 5:
```suggestion
in flat vec3 starPos;
in flat vec3 color;
```

IntegratedQuantum (changes requested):
Also I'm getting the same compiler error as @Argmaster, so please fix that.

src/renderer.zig
```diff
@@ -611,6 +617,222 @@ pub const MenuBackGround = struct {
...
+			var light: f32 = 0;
+
+			while(light < 0.1) {
+				pos = getStarPos(&seed);
+
+				radius = @floatCast(main.random.nextFloatExp(&seed)*4 + 0.2);
+
+				temperature = @floatCast((@abs(main.random.nextFloatGauss(&seed)*3000.0 + 5000.0) + 1000.0)/5772.0);
+
+				light = (3966.91*radius*radius*temperature*temperature*temperature*temperature)/(vec.dot(pos, pos));
```
IntegratedQuantum, line 706: You still got some magic constants in here. What's 3966.91? What's 5772.0? Where do these numbers come from?

src/renderer.zig
```diff
@@ -611,6 +617,222 @@ pub const MenuBackGround = struct {
...
+		const time = game.world.?.gameTime.load(.monotonic);
+
+		var starOpacity: f32 = 0;
+		const dayTime = @abs(@mod(time, game.World.dayCycle) -% game.World.dayCycle/2);
+		if(dayTime < game.World.dayCycle/4 - game.World.dayCycle/16) {
+			starOpacity = 1;
+		} else if(dayTime > game.World.dayCycle/4 + game.World.dayCycle/16) {
+			starOpacity = 0;
+		} else {
+			starOpacity = @as(f32, @floatFromInt(dayTime - (game.World.dayCycle/4 - game.World.dayCycle/16)))/@as(f32, @floatFromInt(game.World.dayCycle/8));
```
IntegratedQuantum, line 798: This logic is inverted at dawn, causing the stars to suddenly disappear, then slowly reappear, then disappear again.

assets/cubyz/shaders/skybox/sky.fs
```diff
@@ -0,0 +1,11 @@
...
+
+layout (location=0) out vec4 fragColor;
+
+in vec3 pos;
+
+uniform vec3 skyColor;
+
+void main() {
+	fragColor = vec4(skyColor, 1);
+}
```
IntegratedQuantum, line 11: Do we even need this shader anymore? It seems to just do what `glClear` is already doing, setting everything to one color, except it costs an additional 1.2 ms per frame for me.

IntegratedQuantum (changes requested):

src/renderer.zig
```diff
@@ -611,6 +616,171 @@ pub const MenuBackGround = struct {
...
+			var temperature: f32 = undefined;
+
+			var light: f32 = 0;
+
+			while(light < 0.1) {
+				pos = getStarPos(&seed);
+
+				radius = @floatCast(main.random.nextFloatExp(&seed)*4 + 0.2);
+
+				// 5772 is the temperature of the sun, so I divide by it to get the temperature in solar temperature
```
IntegratedQuantum, line 693: I don't see a reason not to just keep it in Kelvin.
Since the `4000` below now is just an empirical constant, you can just readjust it bases on (1/5572)⁴
This would allow also removing the `*5772.0` below

src/renderer.zig
```diff
@@ -611,6 +616,171 @@ pub const MenuBackGround = struct {
...
+		c.glDeleteVertexArrays(1, &starVao);
+	}
+
+	pub fn getSkyColor() Vec3f {
+		return game.fog.skyColor*@as(Vec3f, @splat(@reduce(.Add, game.fog.skyColor)/3.0));
+	}
+
+	pub fn render() void {
+		c.glDisable(c.GL_CULL_FACE);
+		c.glDisable(c.GL_DEPTH_TEST);
```
IntegratedQuantum, line 742: Please put these into the `starOpacity != 0` case.

src/renderer.zig
```diff
@@ -611,6 +616,171 @@ pub const MenuBackGround = struct {
...
+			c.glBindVertexArray(starVao);
+			c.glDrawArrays(c.GL_TRIANGLES, 0, numStars*9);
+
+			c.glBindBuffer(c.GL_SHADER_STORAGE_BUFFER, 0);
+
+			c.glBlendFunc(c.GL_SRC_ALPHA, c.GL_ONE_MINUS_SRC_ALPHA);
+		}
+
+		c.glEnable(c.GL_CULL_FACE);
+		c.glEnable(c.GL_DEPTH_TEST);
```
IntegratedQuantum, line 780: Please put these into the `starOpacity != 0` case.

src/renderer.zig
```diff
@@ -611,6 +616,171 @@ pub const MenuBackGround = struct {
...
+
+			const starMatrix = game.projectionMatrix.mul(viewMatrix.mul(Mat4f.rotationX(@as(f32, @floatFromInt(time))/@as(f32, @floatFromInt(main.game.World.dayCycle)))));
+
+			starSsbo.bind(12);
+
+			c.glUniform1f(starUniforms.starOpacity, starOpacity);
+			c.glUniformMatrix4fv(starUniforms.mvp, 1, c.GL_TRUE, @ptrCast(&starMatrix));
+
+			c.glBindVertexArray(starVao);
+			c.glDrawArrays(c.GL_TRIANGLES, 0, numStars*9);
```
IntegratedQuantum, line 772: Shouldn't it be *3?

IntegratedQuantum (changes requested):
Something appears to be broken. I see artifacts in the sky and no stars anymore.
Screenshot at 2025-04-27 11-51-59

OneAvargeCoder193:
> Something appears to be broken. I see artifacts in the sky and no stars anymore. Screenshot at 2025-04-27 11-51-59

thats weird, I don't see anything wrong

Argmaster:
Hmm I also can't reproduce this issue

IntegratedQuantum:
It's fixed now. I think it was just reading bogus data because you drew 3 times as many triangles.

src/renderer.zig
```diff
@@ -611,6 +616,171 @@ pub const MenuBackGround = struct {
...
+
+				// 3.6e-12 can be modified to change the brightness of the stars
+				light = (3.6e-12*radius*radius*temperature*temperature*temperature*temperature)/(vec.dot(pos, pos));
+			}
+
+			pos = vec.normalize(pos)*@as(Vec3f, @splat(starDist));
+
+			const normPos = vec.normalize(pos);
+
+			const col = getStarColor(temperature, light, starColorImage);
```
IntegratedQuantum, line 703:
```suggestion
			const color = getStarColor(temperature, light, starColorImage);
```

src/renderer.zig
```diff
@@ -611,6 +616,171 @@ pub const MenuBackGround = struct {
...
+			std.debug.print("{d}\n", .{col});
```
IntegratedQuantum, line 704: leftover debug print
```suggestion
```

src/renderer.zig
```diff
@@ -611,6 +616,171 @@ pub const MenuBackGround = struct {
...
+
+			const lat: f32 = @floatCast(std.math.asin(normPos[2]));
+			const lon: f32 = @floatCast(std.math.atan2(-normPos[0], normPos[1]));
```
IntegratedQuantum, line 707:
```suggestion
			const latitude: f32 = @floatCast(std.math.asin(normPos[2]));
			const longitude: f32 = @floatCast(std.math.atan2(-normPos[0], normPos[1]));
```

IntegratedQuantum (changes requested):

OneAvargeCoder193:
done

IntegratedQuantum (approved):
Alright let's do this!

PR #1053, opened 2025-02-15
Add coyote time

RanPix:
This should make it feel better when jumping, because before it felt like i couldn't jump on the block. Maybe the time could be smaller.

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -476,12 +476,6 @@ pub const Player = struct { // MARK: Player
 		return eyeCoyote;
 	}
 
-	pub fn getJumpBlocking() f64 {
-		mutex.lock();
-		defer mutex.unlock();
-		return jumpCoyote;
-	}
-
 	pub fn setGamemode(newGamemode: Gamemode) void {
 		gamemode.store(newGamemode, .monotonic);
 
--- a/src/game.zig
+++ b/src/game.zig
@@ -1080,7 +1080,7 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 			Player.eyePos[2] -= stepAmount;
 			move[2] = -0.01;
 			Player.onGround = true;
-			Player.jumpCoyote = jumpCoyoteTimeConstant;
+			Player.jumpCoyote = Player.jumpCoyoteTimeConstant;
 		}
 
 		const wasOnGround = Player.onGround;
@@ -1095,7 +1095,7 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 				Player.onGround = true;
 				Player.super.pos[2] = box.max[2] - hitBox.min[2];
 				Player.eyeCoyote = 0;
-				Player.jumpCoyote = jumpCoyoteTimeConstant;
+				Player.jumpCoyote = Player.jumpCoyoteTimeConstant;
 			} else {
 				Player.super.pos[2] = box.min[2] - hitBox.max[2];
 			}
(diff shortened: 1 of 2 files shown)


RanPix:
I'll try looking into it

RanPix:
@ikabod-kee can you try that? It feels good to me.

RanPix:
nevermind

RanPix:
Yeah, im sorry i thought it could be fixed but it wasn't.

RanPix:
Hey sorry i was just doing other things because i got stuck, looking at it again i noticed that every time that there is an input delay there is a part of code that tells that it might be uncertain if the player is actually falling or not. I think that might be causing the issue in some way.
https://github.com/PixelGuys/Cubyz/blob/efd9ce81413fa17f0f133274324e9a100e199d2f/src/game.zig#L1109

IntegratedQuantum:
That piece of code should only affect the eye position, which is independent of the physics position.

RanPix:
I think i found the issue, the problem was that it only looked like there was an input delay, in fact what happened is because there was that ledge test it sets eye coyote time to a high number which makes the eye wait until it can follow the actual player position (sorry for messy prints, i forgot to put newlines on the last time i tested, and i realized what was the issue), setting eye coyote time to 0 when you jump seems to fix the issue.
image

RanPix:
Ok wait, testing i tried it for a bit again and i also see that there is a delay sometimes, let me look at it again

RanPix:
Ok, i made a paranoid check to see if i need to reset the eye coyote, but the condition was flawed, after i fixed it i think its better

RanPix:
Thanks!

src/game.zig
```diff
@@ -1110,6 +1119,7 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 			// This calculates how long the player has to fall until we know they're not walking over a small gap.
 			// We add deltaTime because we subtract deltaTime at the bottom of update
 			Player.eyeCoyote = @sqrt(2 * Player.steppingHeight()[2] / gravity) + deltaTime;
+			Player.jumpCoyote = Player.jumpCoyoteTimeConstant;
```
IntegratedQuantum, line 1122: Setting it here should be enough. I don't think you need to reset the jumpCoyote anywhere else.

src/game.zig
```diff
@@ -1090,6 +1098,7 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 				Player.onGround = true;
 				Player.super.pos[2] = box.max[2] - hitBox.min[2];
 				Player.eyeCoyote = 0;
+				Player.jumpCoyote = Player.jumpCoyoteTimeConstant;
```
IntegratedQuantum, line 1101: (see other comment)

src/game.zig
```diff
@@ -1076,6 +1083,7 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 			Player.eyePos[2] -= stepAmount;
 			move[2] = -0.01;
 			Player.onGround = true;
+			Player.jumpCoyote = Player.jumpCoyoteTimeConstant;
```
IntegratedQuantum, line 1086: (see other comment)

IntegratedQuantum (changes requested):
It indeed feels quite nice.

Also could you add this to the F3 debug menu?

IntegratedQuantum (approved):
Thank you!

PR #1059, opened 2025-02-16
Coal and Diamond blocks

careeoki:
image
Crafted and uncrafted as you would expect.
image
Diamond blocks are not nearly as reflective as I would like, but #895 makes it look super blown out if I increase it further.


IntegratedQuantum:
Honestly I'm not sure about the diamond block texture.
In my mind a diamond block should be rougher, and maybe also be transparent? Not sure if you can pull that off without making it look like ordinary glass.

careeoki:
Are you imagining a block of crystals, similar to the glow crystal texture?

careeoki:
Not sure if this is what you were looking for, but I made it rougher and much more reflective (at the cost of it being darker than I would like, but I think the tradeoff is worth it)
image
I'd rather not make it transparent, I think it's more fun to have a solid cube of diamond.

IntegratedQuantum:
I think I would just like something that more resembles the textures and roughness of the raw ore and the item. I think it's missing more detailed shapes.
The new texture is better, but maybe a different base shape with a more detailed polygon like an octagon would be better?

IntegratedQuantum:
Yeah, that seems like a reasonable compromise. Terraria for example also has bricks of all the ores I think.

careeoki:
image
Diamond block, and cut diamond block.

IntegratedQuantum:
Yeah, that's nice!

PR #1073, opened 2025-02-19
Add Pebbles

### LINKED ISSUE #1047: Pebbles
Like raw iron ore but on the floor :) 

Image

careeoki:
Fixes #1047 
image
image
These are a ton of fun to play with and provide a lot of interesting building opportunities. 
image
You can use gravel to craft 4 pebbles.


IntegratedQuantum (approved):
This is great!

PR #1101, opened 2025-02-23
Unique Baobab leaves and leafElongation in savannah

careeoki:
New texture for baobab leaves (no longer a recolour) and a more distinct shape for baobab trees
image


IntegratedQuantum (approved):
looks good

PR #1109, opened 2025-02-25
Add player item display

RanPix:
Shows currently selected item in player's hand.
image
image
Probably related to #953.
Next steps in improving this feature -
- [X] React to player movements (e.g. making the tool jumpy when you jump and land)
- Bobbing
- Animations
- Overrides for different values like - tool rotation, position, animation and etc.
- [X] React to light levels


RanPix:
I think things that i listed should be moved to a separate issue once im done with this PR

RanPix:
I think it would be fun to have a separate setting to disable ui and item display, what if you want to play without ui but you want to see your item selected?

RanPix:
Im not sure which keybind we could use for that, f2 is unused but i feel like there are other more important things that could be used by f2, im thinking we could use ctrl+f1 to hide the tool separately, what do you think?

RanPix:
Feels a bit clunky but i think this should be the easiest way to do it. I like the first solution

RanPix:
True, got it.

src/itemdisplay.zig
```diff
@@ -0,0 +1,141 @@
...
+
+// Going to handle item animations and other things like - bobbing, interpolation
+pub const PlayerItemDisplayManager = struct {
+	pub var cameraFollow: Vec3f = .{0, 0, 0};
+	var lastTime: i16 = 0;
+	var timeDifference: utils.TimeDifference = .{};
+
+	pub fn init() void {
+		cameraFollow = game.camera.rotation;
+	}
```
IntegratedQuantum, line 30: This is pointless, at the point the init function is called, the camera rotation is 0 anyways.

src/itemdisplay.zig
```diff
@@ -0,0 +1,141 @@
...
+		texture_sampler: c_int,
+		emissionSampler: c_int,
+		reflectivityAndAbsorptionSampler: c_int,
+		reflectionMap: c_int,
+		reflectionMapSize: c_int,
+		contrast: c_int,
+	} = undefined;
+
+	pub fn init() void {
+		itemShader = graphics.Shader.initAndGetUniforms("assets/cubyz/shaders/item_drop.vs", "assets/cubyz/shaders/item_drop.fs", "", &itemUniforms);
```
IntegratedQuantum, line 64: Please use the shader and uniforms from the ItemDropRenderer.
Keeping the same uniforms in two different places just makes it needlessly difficult to change parameters.

src/itemdisplay.zig
```diff
@@ -0,0 +1,141 @@
...
+		const deltaTime = @as(f32, @floatFromInt(time -% lastTime))/1000;
+
+		const blend: f32 = deltaTime * 19;
+		cameraFollow = vec.lerp(cameraFollow, game.camera.rotation, blend);
+		lastTime = time;
+	}
+};
+
+pub const PlayerItemDisplay = struct { // MARK: PlayerItemDisplay
+	// stolen from itemdrop.zig with love!!!
```
IntegratedQuantum, line 44: Maybe instead of duplicating all the code, we could just call into the ItemDropRenderer from here?

src/itemdrop.zig
```diff
@@ -542,8 +542,70 @@ pub const ItemDropRenderer = struct { // MARK: ItemDropRenderer
...
+				modelMatrix = modelMatrix.mul(Mat4f.scale(@splat(scale)));
+				modelMatrix = modelMatrix.mul(Mat4f.translation(@splat(-0.5)));
+				c.glUniformMatrix4fv(itemUniforms.modelMatrix, 1, c.GL_TRUE, @ptrCast(&modelMatrix));
+
+				c.glBindVertexArray(main.renderer.chunk_meshing.vao);
+				c.glDrawElements(c.GL_TRIANGLES, vertices, c.GL_UNSIGNED_INT, null);
+			}
+		}
+	}
+};
```
IntegratedQuantum, line 605: If you want to refactor things, I'd appreciate it if you could do it in a separate PR, together with a justification why you think it should be structured that way.

src/renderer.zig
```diff
@@ -122,6 +123,7 @@ pub fn updateViewport(width: u31, height: u31, fov: f32) void {
 	lastHeight = @intFromFloat(@as(f32, @floatFromInt(height))*main.settings.resolutionScale);
 	lastFov = fov;
 	game.projectionMatrix = Mat4f.perspective(std.math.degreesToRadians(fov), @as(f32, @floatFromInt(lastWidth))/@as(f32, @floatFromInt(lastHeight)), zNear, zFar);
+	game.playerProjectionMatrix = Mat4f.perspective(std.math.degreesToRadians(65), @as(f32, @floatFromInt(lastWidth))/@as(f32, @floatFromInt(lastHeight)), zNear, zFar);
```
IntegratedQuantum, line 126: This variable should be exclusively defined and assigned to in the itemdisplay.zig file.
Everything else just makes it harder to reason about where it came from.

src/vec.zig
```diff
@@ -38,6 +38,10 @@ pub fn normalize(self: anytype) @TypeOf(self) {
 	return self/@as(@TypeOf(self), @splat(length(self)));
 }
 
+pub fn lerp(self: anytype, other: @TypeOf(self), t: @typeInfo(@TypeOf(self)).vector.child) @TypeOf(self) {
+	return self * (@as(@TypeOf(self), @splat(1)) - @as(@TypeOf(self), @splat(t))) + (other * @as(@TypeOf(self), @splat(t)));
+}
```
IntegratedQuantum, line 43: Does `std.math.lerp` not work?

IntegratedQuantum (changes requested):

src/itemdisplay.zig
RanPix: Yeah, i realized that a bit later

src/vec.zig
RanPix: I didn't know it existed there

src/itemdisplay.zig
RanPix: I could rename the "ItemDropRenderer" to "ItemRenderer" and create separate functions for rendering display item and dropped items

IntegratedQuantum: What are the differences exactly? Could you just make things like the projection matrix a parameter?

RanPix: Also, i need to render thee display item on top of everything so the render call should be done separately

src/itemdisplay.zig
```diff
@@ -0,0 +1,30 @@
...
+	pub fn update() void {
+		var time = @as(i16, @truncate(std.time.milliTimestamp()));
+		time -%= timeDifference.difference.load(.monotonic);
+		const deltaTime = @as(f32, @floatFromInt(time -% lastTime))/1000;
+
+		const blend: f32 = deltaTime * 21;
+		cameraFollow = std.math.lerp(cameraFollow, game.camera.rotation, @as(Vec3f, @splat(blend)));
+		lastTime = time;
+	}
+};
```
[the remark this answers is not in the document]
RanPix: Yeah i also thought about that but then if we move it to itemdrop.zig should it be called "itemdrop"? because it does other things. At this point just move everything to items.zig

src/renderer.zig
```diff
@@ -122,6 +123,7 @@ pub fn updateViewport(width: u31, height: u31, fov: f32) void {
 	lastHeight = @intFromFloat(@as(f32, @floatFromInt(height))*main.settings.resolutionScale);
 	lastFov = fov;
 	game.projectionMatrix = Mat4f.perspective(std.math.degreesToRadians(fov), @as(f32, @floatFromInt(lastWidth))/@as(f32, @floatFromInt(lastHeight)), zNear, zFar);
+	itemdisplay.itemDisplayProjectionMatrix = Mat4f.perspective(std.math.degreesToRadians(65), @as(f32, @floatFromInt(lastWidth))/@as(f32, @floatFromInt(lastHeight)), zNear, zFar);
```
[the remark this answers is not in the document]
RanPix: I don't understand what you mean here, do you want me to calculate item display proj matrix every frame?

src/itemdisplay.zig
```diff
@@ -0,0 +1,30 @@
...
+// Going to handle item animations and other things like - bobbing, interpolation
+pub const PlayerItemDisplayManager = struct {
+	pub var cameraFollow: Vec3f = .{0, 0, 0};
+	var lastTime: i16 = 0;
+	var timeDifference: utils.TimeDifference = .{};
+
+	pub fn update() void {
+		var time = @as(i16, @truncate(std.time.milliTimestamp()));
+		time -%= timeDifference.difference.load(.monotonic);
+		const deltaTime = @as(f32, @floatFromInt(time -% lastTime))/1000;
```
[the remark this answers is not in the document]
RanPix: Didn't notice, sorry

RanPix:
> I'd like the interpolation to be more subtle, as it's very distracting.

Yeah, it needs more polish, thanks for the feedback!

src/renderer.zig
```diff
@@ -299,6 +302,10 @@ pub fn renderWorld(world: *World, ambientLight: Vec3f, skyColor: Vec3f, playerPo
 	c.glBindFramebuffer(c.GL_FRAMEBUFFER, 0);
 
 	entity.ClientEntityManager.renderNames(game.projectionMatrix, playerPos);
+
+	c.glEnable(c.GL_DEPTH_TEST);
+	c.glEnable(c.GL_CULL_FACE);
+	itemdrop.ItemDropRenderer.renderPlayerDisplayItem(itemdisplay.itemDisplayProjectionMatrix, ambientLight, playerPos, time);
```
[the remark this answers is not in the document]
RanPix: If i move it back next to the item drop render the display items are being overdrawn by other things, i could try drawing it to a separate framebuffer and then blitting it before post processing. Or is there a better solution?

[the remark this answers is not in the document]
RanPix: It looks like i manage to solve this issue by drawing the item with glDepthFunc(GL_ALWAYS) after all transparent chunk meshes are drawn. Found no issues with this way of doing it so far.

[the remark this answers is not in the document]
RanPix: But this might become problematic again if we are going to have multiple display items 😕

RanPix:
The last 2 things im thinking to do is finalize the item interpolation so its not jittery and _maybe_ add player hand display.

src/renderer.zig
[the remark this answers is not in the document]
IntegratedQuantum: Yes, the cost is rather small, and the other projectionMatrix will also be recomputed each frame after #812

src/itemdisplay.zig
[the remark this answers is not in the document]
IntegratedQuantum: Yeah, I agree, but I don't like having an extra file in the root, that effectively only has one small function.

src/renderer.zig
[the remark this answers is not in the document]
RanPix: The thing is that the idem display projection is not getting modified by an fov setting, it always stays the same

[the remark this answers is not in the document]
IntegratedQuantum: I still prefer having it locally next to where it's used. We are drawing thousands of pixels with this. Surely we can afford to compute one projection matrix.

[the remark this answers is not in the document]
RanPix: Ok then

RanPix:
I accidentally force pushed the commit 🔥

RanPix:
Ok, thank you for your feedback!

IntegratedQuantum:
https://github.com/user-attachments/assets/70790f92-7857-42e7-a78d-b9380e35001c

When I move my mouse slowly, the item appears to jitter around which feels disorienting.

RanPix:
> When I move my mouse slowly, the item appears to jitter around which feels disorienting.

Yeah, not sure how to fix that yet, thanks for the feedback!

RanPix:
I dont think this will help much because is jittering also happening when you are just moving the camera around, i need a better way to smooth out the position or input..

RanPix:
Thanks 😅

IntegratedQuantum:
Just intuitively I'd say that your spring is too stiff, which makes it overreact to the small jitter of mouse movement.

RanPix:
But at the same time i need it to be fast so its not flying around the screen

IntegratedQuantum:
What if you use a non-linear spring then: Increase acceleration depending on the distance from the target position.

IntegratedQuantum:

src/itemdrop.zig
```diff
@@ -700,10 +700,7 @@ pub const ItemDropRenderer = struct { // MARK: ItemDropRenderer
 	}
 
 	fn bindLightUniform(light: u32, ambientLight: Vec3f) void {
-		c.glUniform3fv(itemUniforms.ambientLight, 1, @ptrCast(&@max(
-			ambientLight*@as(Vec3f, @splat(@as(f32, @floatFromInt(light >> 24))/255)),
-			Vec3f{@floatFromInt(light >> 16 & 255), @floatFromInt(light >> 8 & 255), @floatFromInt(light & 255)}/@as(Vec3f, @splat(255))
-		)));
+		c.glUniform3fv(itemUniforms.ambientLight, 1, @ptrCast(&@max(ambientLight*@as(Vec3f, @splat(@as(f32, @floatFromInt(light >> 24))/255)), Vec3f{@floatFromInt(light >> 16 & 255), @floatFromInt(light >> 8 & 255), @floatFromInt(light & 255)}/@as(Vec3f, @splat(255)))));
```
IntegratedQuantum, line 703: If you want to keep it in separate lines, you can add a trailing comma after the last argument, as a hint for the formatter.

RanPix:
So, lighting interpolation feels good enough, the item reacting to player velocity thing should behave the same way as before, kind of tilting the item, i think this looks better, but tell me if you have other ideas

src/renderer.zig
```diff
@@ -126,7 +126,7 @@ pub fn updateViewport(width: u31, height: u31, fov: f32) void {
 	worldFrameBuffer.unbind();
 }
 
-pub fn render(playerPosition: Vec3d) void {
+pub fn render(playerPosition: Vec3d, deltaTime: f64, isCapturingFrame: bool) void {
```
IntegratedQuantum, line 129: `isCapturingFrame` should be irrelevant, since `deltaTime` is 0 in that case.

src/renderer.zig
```diff
@@ -250,6 +253,13 @@ pub fn renderWorld(world: *World, ambientLight: Vec3f, skyColor: Vec3f, playerPo
 	c.glDepthMask(c.GL_TRUE);
 	c.glDepthFunc(c.GL_LESS);
 	c.glBlendFunc(c.GL_SRC_ALPHA, c.GL_ONE_MINUS_SRC_ALPHA);
+
+	if(!isCapturingFrame) {
```
IntegratedQuantum, line 257: Instead of being a new parameter, this should use the same mechanism, that hides the GUI if you press F1.

src/itemdrop.zig
```diff
@@ -674,46 +696,176 @@ pub const ItemDropRenderer = struct { // MARK: ItemDropRenderer
...
+						}
+					}
+				}
+			}
+
+			localBlockPos[0] -= if(quadrant[0] == -1) 0 else 0.5;
+			localBlockPos[1] -= if(quadrant[1] == -1) 0 else 0.5;
+			localBlockPos[2] -= if(quadrant[2] == -1) 0 else 0.5;
+
+			// TODO: fix jagged light, do better interpolation
```
IntegratedQuantum, line 811: yes, please use trilinear interpolation. I cannot check what was wrong with that code, if I can't see it.

src/itemdrop.zig
```diff
@@ -674,46 +696,176 @@ pub const ItemDropRenderer = struct { // MARK: ItemDropRenderer
...
+		if(selectedItem) |item| {
+			var pos: Vec3d = Vec3d{0, 0, 0};
+			const rot: Vec3f = ItemDisplayManager.cameraFollow;
+			const blockPos: Vec3i = @intFromFloat(@floor(playerPos));
+			var localBlockPos = playerPos - @as(Vec3d, @floatFromInt(blockPos));
+			const quadrant: Vec3i = .{
+				if(localBlockPos[0] - 0.5 < 0) -1 else 1,
+				if(localBlockPos[1] - 0.5 < 0) -1 else 1,
+				if(localBlockPos[2] - 0.5 < 0) -1 else 1,
+			};
```
IntegratedQuantum, line 789: If you subtract 0.5 from the player pos before flooring, then we can always use positive offsets here.

IntegratedQuantum (changes requested):

RanPix:
I switched to my trilinear interpolation here (which also doesnt work properly)

src/itemdrop.zig
```diff
@@ -522,6 +523,29 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
 	}
 };
 
+// Going to handle item animations and other things like - bobbing, interpolation, movement reactions
+pub const ItemDisplayManager = struct { // MARK: ItemDisplayManager
+	pub var showItem: bool = true;
+	pub var cameraFollow: Vec3f = @splat(0);
```
IntegratedQuantum, line 529: I don't think these things need to be public anymore.

src/itemdrop.zig
```diff
@@ -674,46 +696,164 @@ pub const ItemDropRenderer = struct { // MARK: ItemDropRenderer
...
 				const light: [6]u8 = main.renderer.mesh_storage.getLight(blockPos[0], blockPos[1], blockPos[2]) orelse .{0} ** 6;
-				c.glUniform3fv(itemUniforms.ambientLight, 1, @ptrCast(&@max(
-					ambientLight*@as(Vec3f, @as(Vec3f, @floatFromInt(Vec3i{light[0], light[1], light[2]}))/@as(Vec3f, @splat(255))),
-					@as(Vec3f, @floatFromInt(Vec3i{light[3], light[4], light[5]}))/@as(Vec3f, @splat(255)),
-				)));
+				bindLightUniform(light, ambientLight);
 				pos -= playerPos;
 
 				const model = getModel(item);
 				c.glUniform1i(itemUniforms.modelIndex, model.index);
```
IntegratedQuantum, line 736: No longer necessary, since you bind it in the function below.

src/itemdrop.zig
```diff
@@ -674,46 +696,164 @@ pub const ItemDropRenderer = struct { // MARK: ItemDropRenderer
...
+			const lightPos = @as(Vec3f, @floatCast(playerPos)) - @as(Vec3f, @splat(0.5));
+			const blockPos: Vec3i = @intFromFloat(@floor(lightPos));
+			const localBlockPos = lightPos - @as(Vec3f, @floatFromInt(blockPos));
+
+			var samples: [8][6]f32 = @splat(@splat(0));
+			inline for(0..2) |z| {
+				inline for(0..2) |y| {
+					inline for(0..2) |x| {
+						const light: [6]u8 = main.renderer.mesh_storage.getLight(
+							blockPos[0] + @as(i32, @intCast(x)),
```
IntegratedQuantum, line 793: should use `+%` for when we are at the world border.

src/itemdrop.zig
```diff
@@ -674,46 +696,164 @@ pub const ItemDropRenderer = struct { // MARK: ItemDropRenderer
...
 				modelMatrix = modelMatrix.mul(Mat4f.translation(@splat(-0.5)));
-				c.glUniformMatrix4fv(itemUniforms.modelMatrix, 1, c.GL_TRUE, @ptrCast(&modelMatrix));
+				drawItem(vertices, modelMatrix);
+			}
+		}
+	}
 
-				c.glBindVertexArray(main.renderer.chunk_meshing.vao);
-				c.glDrawElements(c.GL_TRIANGLES, vertices, c.GL_UNSIGNED_INT, null);
+	inline fn getPos(x: u8, y: u8, z: u8) u32 {
```
IntegratedQuantum, line 760: Should be called `getIndex`, since that's what it does.

src/renderer.zig
```diff
@@ -250,6 +251,11 @@ pub fn renderWorld(world: *World, ambientLight: Vec3f, skyColor: Vec3f, playerPo
 	c.glDepthMask(c.GL_TRUE);
 	c.glDepthFunc(c.GL_LESS);
 	c.glBlendFunc(c.GL_SRC_ALPHA, c.GL_ONE_MINUS_SRC_ALPHA);
+
+	c.glDepthFunc(c.GL_ALWAYS);
```
IntegratedQuantum, line 255: I'm skeptical that this produces the correct results for non-convex item models.
Could you try removing e.g. the custom item image from a fence, I doubt that it would render correctly.

I'm not sure what the best solution for this would be. Maybe we'll just have to leave it for the future.

Pushing zFar for the model matrix might help maybe, to avoid the original issue, of having items stick into blocks. Maybe you could set it up, so it never takes more of the depth buffer than the distance between the eyes and the bounding box?

IntegratedQuantum (changes requested):

src/itemdrop.zig
RanPix: showItem is modified externally so it should be public

src/renderer.zig
RanPix: >Pushing zFar for the model matrix might help maybe, to avoid the original issue, of having items stick into blocks. Maybe you could set it up, so it never takes more of the depth buffer than the distance between the eyes and the bounding box?

To be honest, im not sure what you mean here. I tried making zFar for the projection matrix higher but i didnt see any changes, but scaling down the model matrix seems to work, tho if there is something that is able to get into the player then there will be issues

IntegratedQuantum: > I tried making zFar for the projection matrix higher but i didnt see any changes

Well it should make item intersections less likely, allowing you to turn on the glAlways which would lead to correct rendering of arbitrary block models.

> tho if there is something that is able to get into the player then there will be issues

I think it wouldn't be too bad for grass and such to be partially on top of the item.

src/itemdrop.zig
```diff
@@ -522,6 +523,29 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
...
+	var cameraFollow: Vec3f = @splat(0);
+	var cameraFollowVel: Vec3f = @splat(0);
+	const damping: Vec3f = @splat(130);
+
+	pub fn update(deltaTime: f64) void {
+		if(deltaTime == 0) return;
+		const dt: f32 = @floatCast(deltaTime);
+
+		var playerVel: Vec3f = .{@floatCast((game.Player.super.vel[2]*0.009 + game.Player.eyeVel[2]*0.0075)), 0, 0};
+		playerVel = vec.clampMag(playerVel, 0.32);
```
IntegratedQuantum, line 538: Instead of stopping it abruptly, have you considered using a logistic growth function?
This is still linear for small velocities, but for larger velocities it smoothly plateaus.

src/main.zig
```diff
@@ -292,10 +292,21 @@ fn openCommand() void {
...
+	if(itemdrop.ItemDisplayManager.showItem and !gui.hideGui) {
+		gui.hideGui = true;
+	} else if(itemdrop.ItemDisplayManager.showItem and gui.hideGui) {
+		itemdrop.ItemDisplayManager.showItem = false;
+	} else if(!itemdrop.ItemDisplayManager.showItem and gui.hideGui) {
+		gui.hideGui = false;
+	} else {
+		itemdrop.ItemDisplayManager.showItem = true;
+	}
 }
```
IntegratedQuantum, line 310: I don't think we should have the inventory but no hand item option. Cycling through so many options is kind of annoying.

src/itemdrop.zig
RanPix: Never heard of that

src/renderer.zig
RanPix: >Well it should make item intersections less likely, allowing you to turn on the glAlways which would lead to correct rendering of arbitrary block models.

You meant "turn off"? Because it still happens

image

IntegratedQuantum: yes I meant turn off

RanPix: Disabling it also breaks models

image

IntegratedQuantum: Why would it draw it twice now? That doesn't make any sense to me.

RanPix: The big flower is placed in world look at the small flower, sorry

IntegratedQuantum: Well, then you probable made zNear too small, which pushes the values in the depth buffer further away.

IntegratedQuantum: If that doesn't work, you could also use `glDepthRange` to map the held item into the start of the depth range.

RanPix: but then the item is going to clip into other blocks

IntegratedQuantum: > but then the item is going to clip into other blocks
That's the opposite of how zNear works.

IntegratedQuantum: higher zNear → more clipping with the camera
lower zNear → more clipping with other geometry (which uses a larger zNear) because now the depth value that gets written into the depth buffer is higher.

src/main.zig
IntegratedQuantum: How about shift+F1? F2 is used in screenshots for minecraft, so I would like to preserve it for the same purpose.

IntegratedQuantum (approved):
Let's get this in the game

PR #1118, opened 2025-02-26
Add branch block type

Argmaster:
This pull request adds a branch block. Branch blocks work similarly to fences but connect in all 3 directions:

image

Image show a 3D cross made out of automatically connecting branch blocks.

image

Image shows a base form of branch block when not connected to any block, compared to solid oak log block.

image

Tree with detailed branches made out of oak branch block.


careeoki:
I would like this block implemented with a new rotation mode that connects to itself as proposed in #194. That would make it more versatile and unique.
Also, it should be called "oak branch"

Argmaster:
Do anyone know what creates those bush-like trees with one block trunks? I can't really trace that down 😄 

image

src/rotation.zig
```diff
@@ -320,6 +320,115 @@ pub const RotationModes = struct {
...
+		}
+
+		fn deinit() void {
+			branchModels.deinit();
+		}
+
+		fn branchTransform(quad: *main.models.QuadInfo, data: BranchData) void {
+			for(&quad.corners, &quad.cornerUV) |*corner, *cornerUV| {
+				// X
+				if(!data.isConnectedNegX and corner[0] == 0) {
```
Argmaster, line 347: Honestly, I have no idea what I am doing, but while paired with UV coordinates from `branch.obj` it seems to work, so IMO profit.

Argmaster:
Let me ping you guys 😄 
@ikabod-kee  @careeoki @IntegratedQuantum 

I consider the texture temporary, It definitely needs love I can't give it no matter how hard I try.

careeoki:
This is great! I can start working on the textures for all the branch types, as well as their item textures. I would add those in a followup pull request so you don't have to wait on me.

One thought though:
Would it be possible for branches to not always connect in all directions, to allow for some more detailed shapes (without it turning into a blob of branch)
**current:**
image
**ideally:**
image

I'm imagining a mix between the log rotation and the branch rotation. I don't know how doable this is but I think it would make branches a lot more useful
image

IntegratedQuantum (changes requested):
It would also be great if you could figure out Carrie's suggestion. That would probably require working with the `generateData` function as well.

Argmaster:
@careeoki I don't see a trivial way to implement it without propagating some more information here and there and this might be disruptive, I would prefer it as separate PR.  I will look into that after this PR gets merged.

With current state of branches you might be able to use them in worldgen if placing blocks with wordgen does not cause block updates, not sure about that one.

Argmaster:
The directional thing is tricky, because you would have to stop a block from being updated in a situation like this:

image

This is currently not possible because you don't know the source of block update. If I understand correctly generateData applies only to block that is being placed, hence if we create `u` and place a block on top of it we will cause unconditional block update to on existing blocks converting `u` to `b` shape, even though we set the block data correctly previously. To avoid that we would have to know where this block update comes from, so to limit what updates are allowed, I don't think it is possible right now.

Logs work because existing logs don't have to be updated to connect to new logs, since they are full blocks and they are automatically connected to all neighbor blocks.

IntegratedQuantum:
> The directional thing is tricky, because you would have to stop a block from being updated in a situation like this:

Consider the following approach:
Untitled

Argmaster:
Yee this is why you are the maintainer and I am the contributor 😄

Argmaster:
For documentation purposes here is the result:

image

careeoki:
Branches are connecting to non-full blocks:
image

Argmaster:
Fixed:

image

src/rotation.zig
```diff
@@ -320,6 +320,193 @@ pub const RotationModes = struct {
...
+					degenerateQuad(quad);
+					break;
+				}
+				if (!data.isConnectedNegZ and corner[2] == 0) {
+					degenerateQuad(quad);
+					break;
+				}
+				if (!data.isConnectedPosZ and corner[2] == 1) {
+					degenerateQuad(quad);
+					break;
```
IntegratedQuantum, line 368: You could also say `return degenerateQuat(quad)` to save some lines here.
(This works because both return values are void)

IntegratedQuantum (changes requested):
Some proposals to simplify the code:
Instead of defining your own packed struct, you could use just a `u6` and `Neighbor.bitMask` to avoid those big, space-consuming switches. And it would also avoid some of the bitcasting.

To change a direction value:
```zig
currentData &= ~bitMask;
if(condition) currentData |= bitMask;
```
To check if a direction is set:
```zig
if(data & Neighbor.dirNegX.bitMask() != 0)
```

IntegratedQuantum:
I don't think the logs should connect to any `viewThrough` blocks like glasss or leaves:
Screenshot at 2025-03-01 10-16-18

Also I think it might be better to not automatically connect them to all neighboring solid blocks.
Instead I would propose to add a way to manually add more connections. e.g. in the following image, right clicking should place a connection between the two branches:
Screenshot at 2025-03-01 10-17-08

Argmaster:
At first I did not agree, I think I might be mildly inconvenient default behavior when placing a lot of branches. However, I am considering that this branch logic can double down for pipe/cable logic either for base game or for addons, and for that purpose this behavior is ideal, so I am all in.

Could you point me where I can find right click detection for workbench? Is that done exclusively for workbench or for all blocks?

IntegratedQuantum:
Check out the carpet and torch code, they handle changes to the block data of the block itself on placement.

The workbench right click is a different mechanic, it is hardcoded for opening a GUI window.

Argmaster:
Okay, so here are the changes:
- branches now connect to the block they are placed on if it is sold or also branch
- branches do not automatically connect to other blocks around
- you can connect branch to another branch or solid block by right clicking it with a branch block
- you can disconnect branch block from whatever it is connected to by hitting the connection
- you can destroy branch by destroying its middle part

When destroying branch from below the hitbox is funky, but I can't figure out why, Im talking about situation like this:
image

When (1) I break the connection on top of bottom block, for some reason whole branch gets destroyed as if i were aiming at the center of it. When (2) I destroy connection from the center block, everything works fine.

1:

image

2:

image

src/rotation.zig
```diff
@@ -321,6 +321,193 @@ pub const RotationModes = struct {
...
+			if(result == block.data) return false;
+
+			block.data = result;
+			return true;
+		}
+
+		fn closestRay(
+			comptime typ: enum{bit, intersection},
+			block: Block,
+			_: ?main.items.Item,
```
IntegratedQuantum, line 453: Please remove unused parameters.

src/rotation.zig
```diff
@@ -321,6 +321,193 @@ pub const RotationModes = struct {
...
+			// opposite direction than they were placed from, effectively connecting
+			// to the block they were placed at.
+			if(blockBaseModel == neighborBaseModel) {
+				const neighborData: BranchData = @bitCast(@as(u6, @truncate(neighborBlock.data)));
+				currentData.setConnection(neighbor, neighborData.isConnected(neighbor.reverse()));
+			} else if(!neighborBlock.solid()) {
+				currentData.setConnection(neighbor, false);
+			}
+
+			const result: u16 = @as(u6, @bitCast(currentData));
```
IntegratedQuantum, line 443: Instead of all this bitCasting, how about adding a `.init` function to BranchData, and use it's field in the other direction?

src/rotation.zig
```diff
@@ -321,6 +321,193 @@ pub const RotationModes = struct {
...
+			const blockBaseModel = blocks.meshes.modelIndexStart(currentBlock.*);
+			const neighborBaseModel = blocks.meshes.modelIndexStart(neighborBlock);
+
+			if(
+				blockPlacing
+				or (blockBaseModel == neighborBaseModel)
+				or neighborBlock.solid()
+			) {
+				const neighborModel = blocks.meshes.model(neighborBlock);
+				const targetVal = (neighborBlock.solid() and  ((blockBaseModel == neighborBaseModel) or main.models.models.items[neighborModel].isNeighborOccluded[neighbor.?.reverse().toInt()]));
```
IntegratedQuantum, line 412: Please remove the extra space after the and

src/rotation.zig
```diff
@@ -321,6 +321,193 @@ pub const RotationModes = struct {
...
+
+				var currentData: BranchData = @bitCast(@as(u6, @truncate(currentBlock.data)));
+				// Branch block upon placement should extend towards a block it was placed
+				// on if the block is solid or also uses branch model.
```
IntegratedQuantum, line 416: I think the comment would be more helpful, if the `targetVal` declaration was also below it.

IntegratedQuantum (changes requested):
I think we shouldn't have the extra face on the outside, if this is used everywhere in the forest, we need to save every triangle we can. Sure it disables some potential future terrain options, but if we ever do want them I'd prefer adding them with an extra bit for it in the rotation mode, instead of always dragging around these extra faces.

Furthermore I still think that branches should not connect to viewThrough blocks, this is even more true with the extra face due to z-fighting:


Screenshot at 2025-03-01 21-44-28

Argmaster:
Extra faces removed, code cleaned.

If we don't connect to viewThrough blocks player can never connect a branch to leaves. You can generate them in worldgen since you don't cause block updates and you can insert any data you want, but it is kind of inconvenient. Maybe we could add exception for them at some point?

careeoki:
Yeah, I think branches connecting to leaves is important. I didn't find it too noticeable that there was open faces when connecting to leaves, since the leaf texture is in the way.

src/rotation.zig
```diff
@@ -328,6 +328,10 @@ pub const RotationModes = struct {
 		const BranchData = packed struct(u6) {
 			enabledConnections: u6,
 
+			pub fn init(blockData: u16) BranchData {
+				return .{.enabledConnections = @intCast(blockData)};
```
IntegratedQuantum, line 332: Should be `@truncate`, `@intCast` will panic or cause undefined behavior if the blockdata is corrupted (which may happen when changing rotation mode)

IntegratedQuantum:
> Yeah, I think branches connecting to leaves is important.

I would prefer branches having like one leaf block around it if they are inside the tree. This would also allow for branches without making the leaf culling any less efficient.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+			}
+		}
+
+		pub fn createBlockModel(modelId: []const u8) u16 {
+			if(branchModels.get(modelId)) |modelIndex| return modelIndex;
+
+			const baseModelIndex = main.models.getModelIndex(modelId);
+			const baseModel = main.models.models.items[baseModelIndex];
+
+			const modelIndex: u16 = baseModel.transformModel(branchTransform, .{@as(BranchData, @bitCast(@as(u6, 0)))});
```
IntegratedQuantum, line 383: Can be `BranchData.init` now.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+			for(1..64) |branchData| {
+
+				_ = baseModel.transformModel(branchTransform, .{@as(BranchData, @bitCast(@as(u6, @intCast(branchData))))});
```
IntegratedQuantum, line 386: Can be `BranchData.init` now.

Argmaster:
Ok, I think we should keep it as is for now (ie. no player created leaves connections) and possibly tweak it further down the road, so this PR can be finally merged. We will see how important that actually is after biomes and structures are adapted to presence of new block. If players even notice this we will surely get an issue for that anyway.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+			_: Vec3i,
+			neighbor: ?Neighbor,
+			currentBlock: *Block,
+			neighborBlock: Block,
+			blockPlacing: bool,
+		) bool {
+			const blockBaseModel = blocks.meshes.modelIndexStart(currentBlock.*);
+			const neighborBaseModel = blocks.meshes.modelIndexStart(neighborBlock);
+
+			if(blockPlacing or (blockBaseModel == neighborBaseModel) or neighborBlock.solid()) {
```
IntegratedQuantum, line 410: == always has higher precedence, there is not need to put it in parenthesis.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+
+				if((block.data & directionBitMask) != 0) {
+					const modelIndex = blocks.meshes.modelIndexStart(block) + directionBitMask;
+					if(RotationMode.DefaultFunctions.rayModelIntersection(modelIndex, relativePlayerPos, playerDir)) |intersection| {
+						// If we did have intersection with center part of the branch we must check
+						// if intersection we have right now is better than the one we had before.
+						// If we were to not do that, even accidental intersection with the connection
+						// would delete the connection even though player was aiming at the center.
+						if(resultIntersection == null or @abs(resultIntersection.?.distance) > @abs(intersection.distance)) {
+							return direction.bitMask();
```
IntegratedQuantum, line 484: Now that I think about, I think we actually cannot have an early return here, since two intersections may be behind each other on screen, and we may accidentally iterate through the further one first.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+		) ?u16 {
+			var resultIntersection: ?RayIntersectionResult = null;
+			var resultBitMask: ?u16 = null;
+			// Check intersection with central part of the branch block.
+			{
+				const modelIndex = blocks.meshes.modelIndexStart(block);
+				if(RotationMode.DefaultFunctions.rayModelIntersection(modelIndex, relativePlayerPos, playerDir)) |intersection| {
+					// We can't return here, as any intersection that accidentally passes
+					// through the central part would delete the whole block, even if player
+					// was aiming at the connection.
```
IntegratedQuantum, line 466: I don't like that this function now has more comments than actual code.
This one here for example should have been just a review comment on my request to use early returns.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+			block.data = result;
+			return true;
+		}
+
+		fn closestRay(
+			block: Block,
+			relativePlayerPos: Vec3f,
+			playerDir: Vec3f
+		) ?u16 {
+			var resultIntersection: ?RayIntersectionResult = null;
```
IntegratedQuantum, line 458: Should be called `closestIntersection` in my opinion, since that's what you are trying to find here.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+					if(RotationMode.DefaultFunctions.rayModelIntersection(modelIndex, relativePlayerPos, playerDir)) |intersection| {
+						// If we did have intersection with center part of the branch we must check
+						// if intersection we have right now is better than the one we had before.
+						// If we were to not do that, even accidental intersection with the connection
+						// would delete the connection even though player was aiming at the center.
+						if(resultIntersection == null or @abs(resultIntersection.?.distance) > @abs(intersection.distance)) {
+							return direction.bitMask();
+						}
+						// Since branch can have multiple connections, we must continue iteration
+						// in case connection with different intersection yields better results.
```
IntegratedQuantum, line 487: That should be obvious.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+			// In case we did not intersect with the central part, check all the side
+			// connections.
+			for(Neighbor.iterable) |direction| {
+				const directionBitMask = Neighbor.bitMask(direction);
+
+				if((block.data & directionBitMask) != 0) {
+					const modelIndex = blocks.meshes.modelIndexStart(block) + directionBitMask;
+					if(RotationMode.DefaultFunctions.rayModelIntersection(modelIndex, relativePlayerPos, playerDir)) |intersection| {
+						// If we did have intersection with center part of the branch we must check
+						// if intersection we have right now is better than the one we had before.
```
IntegratedQuantum, line 480: This is just describing what the if below is doing, which is just redundant since the code is right there.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+				if(RotationMode.DefaultFunctions.rayModelIntersection(modelIndex, relativePlayerPos, playerDir)) |intersection| {
+					// We can't return here, as any intersection that accidentally passes
+					// through the central part would delete the whole block, even if player
+					// was aiming at the connection.
+					resultIntersection = intersection;
+					resultBitMask = 0;
+				}
+			}
+			// In case we did not intersect with the central part, check all the side
+			// connections.
```
IntegratedQuantum, line 472: should be one line.

src/rotation.zig
```diff
@@ -321,6 +321,190 @@ pub const RotationModes = struct {
...
+					}
+				}
+			}
+			return resultBitMask;
+		}
+
+		pub fn onBlockBreaking(_: ?main.items.Item, relativePlayerPos: Vec3f, playerDir: Vec3f, currentData: *Block) void {
+			if(closestRay(currentData.*, relativePlayerPos, playerDir)) |directionBitMask| {
+				// If player destroys a central part of branch block, branch block is completely
+				// destroyed.
```
IntegratedQuantum, line 497: should be one line.

IntegratedQuantum (changes requested):
some more small things

src/rotation.zig
Argmaster: iuy9fxt300811

src/rotation.zig
Argmaster: What is the strict column limit enforced in this repository?

src/rotation.zig
Argmaster: Possible, but this is assuming you have memorized operator precedence. This is not true and not necessary, I prefer it explicit. Always. I can change it if you really want tho.

src/rotation.zig
Argmaster: Also, this comment is outdated and makes no sense with current implementation, i will remove it.

IntegratedQuantum: I think line limits are a relic from the past. Nowadays we can display 200 characters or more on the screen.
I think the focus should be one writing readable code, and breaking a line at an arbitrary width does not help with that in my opinion.

IntegratedQuantum: > Also, this comment is outdated and makes no sense with current implementation, i will remove it.

Yeah, that's one of the reasons, why I don't like seeing so many comments, they tend to get outdated soon.

Argmaster: The fact that you can do something doesn't mean you should do it. Realistically, it is very handy to open two files side by side and with eg. my screen (4K) it imposes ~150 column limit before I have to scroll to see the whole line of both files. I might be too sensitive most of the time since I'm usually cutting at 88 or 110. If not a strict rule, a guideline should be there, to stick with X columns max unless doing otherwise improves readability.

IntegratedQuantum: If you want a fixed number for your own then I'd say 200 maybe.

src/rotation.zig
IntegratedQuantum: There are some operators where precedence should obvious, and in all languages that I'm aware of (logical) `or` always has the lowest precedence.
Furthermore even if you did not know this, in this case `or` cannot act on numbers, so there is only one way to read this.

So please remove them.

IntegratedQuantum:

Argmaster:
That:

```
 		fn branchTransform(quad: *main.models.QuadInfo, data: BranchData) void {
 			for(&quad.corners) |*corner| {
-				if(
-					(!data.isConnected(Neighbor.dirNegX) and corner[0] == 0)
-					or (!data.isConnected(Neighbor.dirPosX) and corner[0] == 1)
-					or (!data.isConnected(Neighbor.dirNegY) and corner[1] == 0)
-					or (!data.isConnected(Neighbor.dirPosY) and corner[1] == 1)
-					or (!data.isConnected(Neighbor.dirDown) and corner[2] == 0)
-					or (!data.isConnected(Neighbor.dirUp) and corner[2] == 1)
-				) return degenerateQuad(quad);
+				if((!data.isConnected(Neighbor.dirNegX) and corner[0] == 0) or (!data.isConnected(Neighbor.dirPosX) and corner[0] == 1) or (!data.isConnected(Neighbor.dirNegY) and corner[1] == 0) or (!data.isConnected(Neighbor.dirPosY) and corner[1] == 1) or (!data.isConnected(Neighbor.dirDown) and corner[2] == 0) or (!data.isConnected(Neighbor.dirUp) and corner[2] == 1)) return degenerateQuad(quad);
 			}
 		}
```

Is not happening.

src/rotation.zig
```diff
@@ -321,6 +321,173 @@ pub const RotationModes = struct {
...
+				// zig fmt: off
+				if(
+					(!data.isConnected(Neighbor.dirNegX) and corner[0] == 0)
+					or (!data.isConnected(Neighbor.dirPosX) and corner[0] == 1)
+					or (!data.isConnected(Neighbor.dirNegY) and corner[1] == 0)
+					or (!data.isConnected(Neighbor.dirPosY) and corner[1] == 1)
+					or (!data.isConnected(Neighbor.dirDown) and corner[2] == 0)
+					or (!data.isConnected(Neighbor.dirUp) and corner[2] == 1)
+				) return degenerateQuad(quad);
+				// zig fmt: on
```
IntegratedQuantum, line 368: If you put the `or` at the end of the line, the formatter seems to keep the lines mostly separate. I think that's the way to go here instead of disabling it entirely.

src/rotation.zig
```diff
@@ -321,6 +321,173 @@ pub const RotationModes = struct {
...
+			}
+		}
+
+		fn degenerateQuad(quad: *main.models.QuadInfo) void {
+			for(&quad.corners) |*corner| {
+				corner[0] = 0.5;
+				corner[1] = 0.5;
+				corner[2] = 0.5;
```
IntegratedQuantum, line 376: Just in case you were not aware of it:
You could also use `corner.* = @splat(0.5)` to fill the entire thing in one statement.

src/rotation.zig
```diff
@@ -321,6 +321,173 @@ pub const RotationModes = struct {
...
+
+			const result: u16 = currentData.enabledConnections;
+			if(result == block.data) return false;
+
+			block.data = result;
+			return true;
+		}
+
+		fn closestRay(block: Block, relativePlayerPos: Vec3f, playerDir: Vec3f) ?u16 {
+			var closestIntersection: ?RayIntersectionResult = null;
```
IntegratedQuantum, line 453: We only need the distance, so how about storing just the distance starting at Infinity, so we can avoid the extra null checks.

IntegratedQuantum (changes requested):
Functionally it seems to work great now, just a few more small fixes and I am ready to merge this.

PR #1121, opened 2025-02-27
Move wood blocks and textures into nested wood folder

Argmaster:
This pull request moves wood like blocks `.zig.zon` and texture files into nested wood folders for blocks and textures.

---

List of blocks considered wood like:

```python
{
    "baobab_fence",
    "baobab_log",
    "baobab_planks",
    "baobab_top",
    "birch_fence",
    "birch_log",
    "birch_planks",
    "birch_top",
    "mahogany_fence",
    "mahogany_log",
    "mahogany_planks",
    "mahogany_top",
    "oak_fence",
    "oak_log",
    "oak_planks",
    "oak_top",
    "pine_fence",
    "pine_log",
    "pine_planks",
    "pine_top",
    "willow_fence",
    "willow_log",
    "willow_planks",
    "willow_top",
}
```

List is based on old ".class" property.


Argmaster:
Actually, I might need to redo that one, I did it with script and completely missed `src` folder due to mistake in script.

Argmaster:
Fixed! I indeed missed 2 files from `src` folder.

Argmaster:
In case we would ever want to do something similar, here is core part of script I have used:

```python
from itertools import chain
from pathlib import Path
import shutil
import pyparsing as pp
import itertools as itr

from typing import Any

struct = pp.Forward()
key = pp.Regex(r"\.[a-zA-Z_][a-zA-Z0-9_]*")
boolean = (pp.Literal("true") | pp.Literal("false")).set_parse_action(
    lambda t: t[0] == "true"
)
string = pp.Literal('"').suppress() + pp.Regex(r'[^"]*') + pp.Literal('"').suppress()
number = pp.Regex(r"[-+]?(\d+(\.\d+)?|(\d+)?\.\d+)").set_parse_action(lambda t: float(str(t[0])))
hex_integer = pp.Regex(r"0x[a-fA-F0-9]+").set_parse_action(lambda t: int(str(t[0]), 16))

comma_separated_list = pp.Forward()

value = key | string | hex_integer | number | boolean | struct | comma_separated_list


key_value_pair = key + pp.Literal("=").suppress() + value

comma_separated_list <<= pp.Group(
    pp.Literal(".{").suppress()
    + pp.ZeroOrMore(value + pp.Literal(",").suppress())
    + pp.Opt(value + pp.Opt(pp.Literal(",")).suppress())
    + pp.Literal("}").suppress()
).set_parse_action(pp.ParseResults.as_list)

struct <<= (
    pp.Literal(".{").suppress()
    + pp.ZeroOrMore(key_value_pair + pp.Literal(",").suppress())
    + pp.Opt(key_value_pair + pp.Opt(pp.Literal(",")).suppress())
    + pp.Literal("}").suppress()
).set_parse_action(lambda t: {k: v for k, v in itr.batched(t.as_list(), 2)})


def parse_zon(source: str) -> dict[str, Any]:
    return struct.parse_string(source)[0]  # type: ignore


assets_path = Path("assets")
source_path = Path("src")

blocks_folder_path = assets_path / "cubyz" / "blocks"
block_textures_folder_path = assets_path / "cubyz" / "blocks" / "textures"

wood_folder_path = blocks_folder_path / "wood"
wood_folder_path.mkdir(exist_ok=True)

textures_wood_folder_path = block_textures_folder_path / "wood"
textures_wood_folder_path.mkdir(exist_ok=True)

for file in blocks_folder_path.glob("*.zon"):
    item_name = file.with_suffix("").with_suffix("").name
    if item_name not in wood_block_list:
        continue

    original_namespace_name = f"cubyz:{item_name}"
    new_namespace_name = f"cubyz:wood/{item_name}"

    for replace_file in chain(
        assets_path.rglob("*.zon"),
        assets_path.rglob("*.zig"),
        source_path.rglob("*.zon"),
        source_path.rglob("*.zig"),
    ):
        try:
            replace_file.write_bytes(
                replace_file.read_bytes().replace(
                    original_namespace_name.encode("utf-8"), new_namespace_name.encode("utf-8")
                ),
            )
        except Exception as e:
            print(replace_file, e)

    for texture_file in block_textures_folder_path.glob("*.png"):
        if item_name in texture_file.name:
            shutil.move(
                texture_file.as_posix(), textures_wood_folder_path / texture_file.name
            )

    shutil.move(file.as_posix(), wood_folder_path / file.name)

```

Argmaster:
> I would prefer having one folder for each wood type, or maybe one for each block type (one folder for leaves, log and planks) might be better. I'm not sure on that yet. 

I think you have already made this decision by grouping chalk, cloth, glass etc. ie. by grouping block variants inside of folder of block type. Therefore to be consistent we should have leaves folder with variants of leaves, logs folder with variants of logs, etc. 

I don't think it matters in what order of grouping we choose, but since you already did grouping one way we can just stick with it.

Issue #1081, opened 2025-02-21
Damage Update


ikabod-kee:
Easy stuff [handled with player status effects?]
- [ ] #933 
- [ ] Cacti
- [ ] Magma
- [ ] Lava

Hard stuff:
- [ ] Weapons
- [ ] PVE
- [ ] PVP

SEGenjoysBlivits:
A damage update would go well with the addition of local immunity frames.

PR #1126, opened 2025-02-27
Add block touch detection

qFelixDev:
I have added a function in `blocks.zig` that runs every time a block is touched. This is one step towards making cacti prickly and lava hot as requested by ikabod in Issue #1081. Note that this system does not support incomplete blocks yet.


qFelixDev:
I just noticed that the way that `touchBlocksFromDirection()` gets block data causes a crash when connected to an external server, I'll fix this later this afternoon.

qFelixDev:
Closing for now due to the poor support for special block shapes, I will come back to this later.

PR #1142, opened 2025-03-02
Block Resistance!

ikabod-kee:
I have to make this one a draft, because I would like to compare it with the current version before it is merged. If I could get @argmaster to show the graphs, it would be greatly appreciated.


assets/cubyz/items/uranium_ingot.zig.zon
careeoki, line 1: I feel these modifiers would be better fit for a different material

Argmaster:
Sure, although, did the formula change since last graphs?
Right now I have:

```python
".breakTime": s[".swingTime"] / s[".damage"]
```

Argmaster:
Okay, so there is a slight problem with setting stuff to 0 when they are taken as inverse for some property:

For example:

```
.elasticity
{'.density': 2.5, '.strength': 1.5, '.elasticity': 0.0, '.grip': 0.0, '.hardness': 2.5, '.textureRoughness': 2.0, '.colors': [4285429119, 4287007118, 4288980906, 4290362559, 4291480782], '.modifiers': [{'.id': 'fragile', '.strength': 0.35}]}
glacite
```

glacite has elasticity of 0, but when used in tip which uses:

```

		.{
			.source = .elasticity,
			.destination = .swingTime,
			.functionType = .inverse,
			.factor = 0.005,
		},
```

We get float division by zero

Argmaster:
This also mean that setting it low would increase swing time by insane amount, as inverse is not "decrease instead of increase" but `1/x`

Argmaster:
I think `factor` is a float32 so it can be negative. Set functionType .linear and .factor = -0.005 to penalize player.

Argmaster:
What formula do you have in mind? Can you just use squareroot with negative factor?

Argmaster:
Apply this patch to your PR:

```
diff --git a/src/items.zig b/src/items.zig
index cf19597..a2cd036 100644
--- a/src/items.zig
+++ b/src/items.zig
@@ -373,6 +373,7 @@ const FunctionType = enum {
 			.inverseSquare => return 1.0/(val*val),
 			.squareRoot => return @sqrt(val),
 			.inverseSquareRoot => return 1.0/@sqrt(val),
+			.symmetricSquareRoot => return @sqrt(@abs(val)),
 			.exp2 => return @exp2(val),
 			.log2 => return @log2(val),
 		}

```

Issue #1060, opened 2025-02-17
Add chests


ikabod-kee:
Carrie Suggestion (the best one):

> Chests holding bigger stacks seems weird, especially when taking items out of it.
> 
> Here are my suggestions for storage blocks:
> 
> Chest: A standard chest, it has slots you can put items into. It's the cheapest storage option.
> Crate: Holds a total number of items rather than individual slots. This is great for storing a variety of items you only have a few of.
> Barrel: Holds a very large amount of one item. Perfect for bulk storage, but only holds one type of item and takes up a full block of space.

Ikabod Suggestion:

> Chests hold 9 slots, but each slot can hold 2x (240)
> Having more slots would mean that players wouldn't just dump all their items in. These chests encourage players to organize with the same item.

Lemon Suggestion:

> more slots (and less stacking) i feel like would be better because the slots relate more to the inventory, and you can fit what you have in the inventory into the chest. also you'd be able to fit a wider range of items, and i think that encourages organizing

careeoki:
Chests holding bigger stacks seems weird, especially when taking items out of it. 


Here are my suggestions for storage blocks:

- **Chest:** A standard chest, it has slots you can put items into. It's the cheapest storage option.
- **Crate:** Holds a total number of items rather than individual slots. This is great for storing a variety of items you only have a few of.
- **Barrel:** Holds a very large amount of one item. Perfect for bulk storage, but only holds one type of item and takes up a full block of space.

Argmaster:
> Chests holding bigger stacks seems weird, especially when taking items out of it.
> 
> Here are my suggestions for storage blocks:
> 
> * **Chest:** A standard chest, it has slots you can put items into. It's the cheapest storage option.
> * **Crate:** Holds a total number of items rather than individual slots. This is great for storing a variety of items you only have a few of.
> * **Barrel:** Holds a very large amount of one item. Perfect for bulk storage, but only holds one type of item and takes up a full block of space.

What would be the suggested capacity? How much space will they occupy?
[Create](https://www.curseforge.com/minecraft/mc-mods/create) has quite well balanced mechanics of vault block:
it can be connected with more vault blocks to create shared inventory of higher capacity at cost of occupying more space. Create vaults do not have inventory that can be directly opened by player, but I don't think its a necessary trait.

careeoki:
> Should a chest be a full block or eg. 1 pixel smaller like in minecraft?

Yeah, full block should be fine for now

PR #1151, opened 2025-03-05
Furnaces!

Argmaster:
This pull request aims to implement furnaces in compliance with ideas from #870, #1115, #197

I assume that furnace should have persistent storage per block (for fuel, input and temporarily output, until picked up), thus relation with #1060, please correct me if I am wrong.
My understanding is that currently we have not blocks with persistent inventories, hence storage mechanism has to be implemented (?)

Twist I am considering is giving furnace 3 fuel slots. The result could be some of:

- different temperature and thus ability to smelt different materials, while destroying others, eg. smelting copper with one fuel, smelting iron with two, smelting titanium with three. If we don't plan separate cooking station, higher power levels could destroy food.
- different smelting speed, increasing by x% with every additional slot
- requiring 0, 1 or 2 bellows depending on used fuel slots

Maybe we could also use multiple input slots, for example to have crafting options for steel, alloys and all the overcomplicated that does not fit Cubyz ❤️

I am considering including mandatory `ingot mold` item slot (and also item, made out of dirt and planks). Mold would not be required to be handled manually, but would have to be placed in specially designated slot in furnace for it to be able to smelt metals, thus making furnace a 4 component station to increase inconvenience of trying to carry it. 

---

Screenshot 2025-03-05 020511

Screenshot 2025-03-05 020643

---

Consider textures temporary, as I am no pixel artist. I hope that @careeoki will help with that matter at some point.

Related to: #197
Related to: #870
Related to: #1115
Related to: #1060


Argmaster:
- Should temperature decrease over time or drop after X time to lower level?
- Should fuel affect how fast temperature drops
- Should fuel type affect how high temperature can go
- After what time one fuel unit is used

I am thinking about minecraft style furnace, then temperature and ignition concepts make sense. 

About the bellows, I think it should be limited how much you have jump on it. I think it should be necessary to initialize, to raise power level and from time to time to keep it. Otherwise it will become annoying due to repetition.

My proposal is to express temperature as a vertical bar with different colors:

image

Colors represent furnace's energy levels and are intentionally of different sizes and they have different number of "energy units" within them.

You have to jump on bellow to start ignition, fuel has to be inside. Ignition will use 1 fuel to raise energy by N energy units depending on the fuel type. Proposed energy units for fuel types:

- leaves could add 1 unit
- planks could add 2 units
- wood could add 8 units (to keep 1:1 ratio)
- coal could add 16 units

After you ignite, without jumping on a bellow furnace will use 1 fuel item every X seconds to add corresponding amount of energy until 1st level is filled.

You can use bellow to raise energy units above first level, using items for N energy units.

Furnace will naturally keep first K levels filled,, where K is based on the type of the fuel:

- planks etc. could naturally reach only 1st level
- wood could reach 2nd level
- coal could reach 3rd level

K should depend on the **next** item in fuel queue.

You have to use bellow to jump from level 1 to level 2 etc, even if K is 3 or so.

When K level loses energy and is no longer full, furnace will automatically use fuel to fill it with N energy units using 1 fuel item. This means that furnace can temporary reach higher energy level, but it will not keep it automatically.

Every X seconds temperature should drop by 1, 2, 3, 4 or 5 energy units (depending on the energy level)

For fuel that provide low energy units the limit how high temperature can go should be implied by how fast energy level depletes in comparison to how much energy one fuel unit can provide.

Leaves, branches etc. should provide only 1 energy unit so to never naturally grow without use of bellow and to never be able to go outside of first level.

Below should be only usable every Y seconds to make it impossible to pump temperature with low energy fuel to max level.

In that model blasting should not use up energy units.

Argmaster:
Yeah, well, I am not so keen on investing a week of my work into a quite complicated implementation of feature just to hear that it is too complicated and to throw it away. I would much rather continue the discussion here until we agree on some compromise on core mechanics, let this PR be a nudge to continue the discussion on this, as I already have shown interest in implementing this.

Since bags were requested as slot for fuel this is blocked anyway, since I am not going to include bag code in this PR.
We should also probably figure out storage blocks (and thus storage of storage data) before, in case we want to go with minecraft style furnace.

IntegratedQuantum:
> > Should fuel type affect how high temperature can go?
>
> Yes. Right on the money with that.

I think this is problematic. The player can't see what item is next in the bag, and it kind of defeats the purpose of the bag as a trash can where you can put your leftover wood materials.

Furthermore it means that we now need to parametrize two values for every item that can be a fuel. And I think we already have 50 of them that would need to be parametrized and balanced.

To be honest I'm not even sold on the temperature idea as a whole. It does add quite a lot of complexity. I prefer Ikabod's idea of just making it faster if you hop the bellows. That's such a simple concept that it can be one sentence in the tooltip of the bellows: "Jumping on it increases the speed of neighboring furnaces". And if you don't notice it, then you'll just have to wait longer.

Argmaster:
Ok. What about multiple input slots for crafting colored glass and alloys (eg. steel)?

IntegratedQuantum:
Yes, I think two input slots would be nice.
Though we'd need some special handling to allow using one dye item for multiple glasses. Maybe we could do it like the Grinding balls in EnderIO's pulverizer, with a little bar after an item is consumed.

IntegratedQuantum:
> Or we can avoid introducing new mechanics to the engine and just convert one die into multiple die powder? This is "the shards problem" (https://github.com/PixelGuys/Cubyz/issues/1116) yet again, but here for good reason.

I don't like adding more clutter, we already have so many different blocks.

I guess we could just make the furnace consume 8 sand and one ingot in one go, with a longer smelting time of course.
But how would we tell that player that they need 8 glass per ingot?

careeoki:
I worry it would be unclear what ingots make what colours, and also having enough unique metals for all the colours. It would also be completely different from created coloured cloth and chalk, which might be weird.
I think it would be better to have a specific dyeing workstation. I had an idea awhile ago like this:
image
With something like this you don't have to add any new items and there can be multiple sources of colours

careeoki:
> We could just bite the bullet and just use chalk to dye the glass

~but how do you dye the chalk?~

Argmaster:
Idk guys, I think colored sand is useful for texture variance while building stuff. If you have a look on most of those crazy builds ppl do in minecraft they are never solid stone bricks, they interleave them with other similar color blocks with different textures. Colored sand is good for that purpose. Also, I don't understand why you want to avoid adding new block types, when textures are bascially all the variance you can provide without harming performance. 

Regardless, I agree with @careeoki about coloring station, but I the second I saw CMYK i went hell no I have no idea what I am doing. This is good for consistency, but do we actually want consistency? Those colorful blocks couldn't be more different from each other. Do we color glowing crystals there too?

Argmaster:
Without coloring station how would you obtain colors?

- `aqua` -
- `black` - 
- `blue` - 
- `brown` - iron? 
- `crimson` - 
- `cyan` -
- `dark_grey` -
- `green` -
- `grey` -
- `indigo` - copper
- `lime` -
- `magenta` -
- `orange` - silver?
- `pink` - 
- `purple` - 
- `red` - gold
- `violet` -
- `viridian` - 
- `white` -
- `yellow` -

Any suggestions?

careeoki:
> Hey wait, what if you could just dye the blocks like that? Input the colored item into the color slot (for example, a toadstool), then another input for the block that will get dyed. So if you ever have excess items, you can just use them for dye.

This is basically my idea, but now you need 20 different items for all of the colours.
For the dye station, you would input the toadstool and get X yellow and X magenta, then you can use those to make a variety of the different colours.

Argmaster:
This is good if we are keep restricting ourselves from adding items and blocks to fill those gaps in recipes.

Argmaster:
I don't think items that have actual recipes and well defined use are clutter. Sure, there is rotten meat, but I would never call dyes or colored concrete powder a clutter, they are things you make and use when you build cool stuff. Rotten meat is trash with literally no other purpose than to be a trash.

IntegratedQuantum:
Please rebase before marking this as ready for review so the formatter can run on the included .zon files

careeoki:
I would like it to be make of clay. It looks nice.

careeoki:
It needs to be added to a lot more biomes, of course. It would be pretty easy to get if rivers or lakes existed.
I would also add a "surface cave" biome which has clay patches and more flora than regular deep caves.

IntegratedQuantum:
> Yeah, you'd probably need to bake the clay first, or we could find another material to craft it out of?

We could just say that the clay bakes itself when you first use the furnace.

careeoki:
Yeah, that works.

careeoki:
I would still have the surface caves made of regular stone, to allow for some ores to appear. You kinda automatically get dirty caves in most biomes anyway.

Argmaster:
Clay sounds like one of those things that could stop player progression, even if it is fairly common, player with wooden/stone tools is really slow and running around for clay can get very annoying. Also we currently have no way to learn the recipe unless we have materials for it, this is probably first recipe for which it would be really important that player knows it has to look for it.

careeoki:
> even if it is fairly common

We would make it as common as soil or stone. The requirement could be just to make a shovel, if you can't break clay with your hands.

> Also we currently have no way to learn the recipe unless we have materials for it

Yeah, this is something we need to think about in general

Argmaster:
Well, where I leave clay is basically everywhere just below thin layer of soil, so that could also be done for many biomes. Not a specialist tho.

Argmaster:
Or do you want to have organicMatter tag and just use that?

Argmaster:
Well, my opinion about composter in minecraft is that its such a cool addition I have better things to do than afk right click with all the trash.

IntegratedQuantum:
> Do you mean specifically cubyz:stone, or any stone type on the surface?

Any of the surface stones, you should be able to make the basic furnace regardless of your starting biome.

IntegratedQuantum:
And yeah, making it faster with each tier would also be reasonable.

careeoki:
So for right now we would only need stone furnace, clay furnace, chimney, and bellows for smelting, correct?

careeoki:
Also I would give each tier a unique name make them more distinct
tier 1: Kiln
tier 2: Furnace
tier 3: Forge or blast furnace or something

codemob-dev:
> We could just bite the bullet and just use chalk to dye the glass

What if in cubyz glass is a crystal and spawns in caves?

Argmaster:
Closing until appropriate tools appear in engine.

PR #1163, opened 2025-03-06
Reduce Cacti spawn rates

Garrett96:
This is my first contribution to Cubyz. There were too many cacti spawning in the desert. I adjusted the rates so they would appear more infrequently and make the desert look more natural.

*Before*
Before

*After*
After


IntegratedQuantum:
> Real life has cacti more akin to the after picture

meanwhile in real life:

IntegratedQuantum:
Jokes aside, I think it is pretty bad how green the desert is, especially in LODs. So I'm not opposed to reducing the spawn.

@ikabod-kee I would like to also hear your opinion on this.

PR #1169, opened 2025-03-07
Added Birch Forest and a rare Chopped Forest

ikabod-kee:
image
image


Argmaster:
I feel like it could really use some foliage ❤️ 

image

In forest it is never just grass on the ground, there are bushes, flowers etc. 

We could create a issue for adding foliage to biomes with checklist of biomes which require foliage and what foliage they require. Right now I really feel like its really empty. 

Also we could update `SimpleTreeModel.zig` to use branches and mushrooms when generating trees, what do you think @IntegratedQuantum?

careeoki:
You should swap out daisies for dandelions in the birch forest for some more colour variety.

PR #1170, opened 2025-03-07
Birch, Thin Birch + Rare Chopped Forests

ikabod-kee:
https://discord.com/channels/443805812390100992/485498631156138005/1327726176193482752

@careeoki Could you review the changes I made to cubyz:forest/birch to see if it's yellow enough for you?


careeoki:
It looks good.
You should definitely mix some big birch trees into the thin birch forest and thin birch trees into the regular birch forest. It would add more variety.
Other biomes should do that as well, but i don't expect you to do that for every biome right away :P

careeoki:
Can I see a screenshot? I'm not at my computer anymore.

careeoki:
Looks good! Though, why aren't the birch trees using log tops for the branches?

Argmaster:
For tops use "cubyz:birch_branch:63", it should extend it in all directions.

I think that thin birches should be shorter  than thick ones, as I would assume the thin ones are younger.

IntegratedQuantum (changes requested):
Looks nice!

PR #1174, opened 2025-03-07
Glimmergills and Glimmergill caves

careeoki:
Introducing our obligatory glowing mushroom: the Glimmergill!
image
image
The caves are pretty basic, especially since I can't generate shelf mushrooms on the walls. I'd like to add big mushrooms, but I'm still deciding how to design them so they'll come later.


IntegratedQuantum (approved):
This looks great! I think we found the right color.

PR #1182, opened 2025-03-08
Darken grass plants for better contrast

careeoki:
Makes grass plants blend into grass blocks more.
grass_compare


careeoki:
I'll try that

careeoki:
Done @ikabod-kee 
grass_compare_2

PR #1183, opened 2025-03-09
Change branches so that the textures can be rotated

OneAvargeCoder193:
They are similar to branches, but they can have different textures depending on the connections. Will probably replace branch after all the textures are converted.
Screenshot 2025-03-08 225857

New vs. Old:
Screenshot 2025-03-09 012741

You could use this to have the thin logs in trees not have the bark be pointing up constantly.


OneAvargeCoder193:
An example would have the images
$RKWWVB8
$R3G19OB
$RCRLB4J
$RKLLNPB
$RCKGK7I
$RCK19ZB

And a block zig.zon of
```zig
.{
	.tags = .{.wood},
	.blockHealth = 4,
	.drops = .{
		.{.items = .{.auto}},
	},
	.absorbedLight = 0x202830,
	.model = "cubyz:cube",
	.rotation = .pipe,
	.texture_top = "cubyz:pipe_0",
	.texture_bottom = "cubyz:pipe_1",
	.texture_right = "cubyz:pipe_2",
	.texture_left = "cubyz:pipe_3",
	.texture_front = "cubyz:pipe_4",
	.texture_back = "cubyz:pipe_5",
}
```

src/chunk.zig
```diff
@@ -47,6 +47,24 @@ pub const Neighbor = enum(u3) { // MARK: Neighbor
 	pub fn relPos(self: Neighbor) Vec3i {
 		return .{self.relX(), self.relY(), self.relZ()};
 	}
+
+	pub fn fromRelPos(pos: Vec3i) ?Neighbor {
```
Argmaster, line 51: Can we get away with this (?):
```zig
	pub fn fromRelPos(pos: Vec3i) ?Neighbor {
		if(@reduce(.And, @abs(pos)) != 1) {
			return null;
		}
		return switch(pos[0]) {
			1 => return .dirPosX,
			-1 => return .dirNegX,
			else => switch(pos[1]) {
				1 => return .dirPosY,
				-1 => return .dirNegY,
				else => switch(pos[2]) {
					1 => return .dirUp,
					-1 => return .dirDown,
					else => return null,
				},
			},
		};
	}
```

OneAvargeCoder193: probably

src/rotation.zig
```diff
@@ -486,6 +486,383 @@ pub const RotationModes = struct {
...
+		const PatternType = enum(u3) {
+			dot = 0,
+			halfLine = 1,
+			line = 2,
+			bend = 3,
+			intersection = 4,
+			cross = 5,
+		};
+
+		const Pattern = union(PatternType) {
```
IntegratedQuantum, line 528: You can use `union(enum)`, we don't need the explicit backing type here.

IntegratedQuantum (changes requested):
Please merge this with the branch rotation, then we hopefully also lose another 100 lines of the diff.

And for the textures @careeoki might be willing to add them into your PR.

src/vec.zig
```diff
@@ -80,6 +80,18 @@ pub fn rotateZ(self: anytype, angle: @typeInfo(@TypeOf(self)).vector.child) @Typ
...
+
+	const sin = @sin(angle);
+	const cos = @cos(angle);
+
+	return @TypeOf(self){
+		cos*(self[0] - center[0]) - sin*(self[1] - center[1]),
+		sin*(self[0] - center[0]) + cos*(self[1] - center[1]),
+	} + center;
+}
+
```
Argmaster, line 94: Will optimizer extract `self[i] - center[i]` that subtraction to make it vector optimized?

To have it guaranteed I think we can do:

```suggestion
pub fn rotate2d(self: anytype, angle: @typeInfo(@TypeOf(self)).vector.child, center: @TypeOf(self)) @TypeOf(self) {
	if(@typeInfo(@TypeOf(self)).vector.len != 2) @compileError("Only available for vectors of length 2.");

	const sin = @sin(angle);
	const cos = @cos(angle);
	const pos = self - center;

	return @TypeOf(self){
		cos*(pos[0]) - sin*(pos[1]),
		sin*(pos[0]) + cos*(pos[1]),
	} + center;
}

```

src/rotation.zig
```diff
@@ -486,6 +486,383 @@ pub const RotationModes = struct {
...
+				@as(Vec3f, @floatFromInt(side.textureX()))*@as(Vec3f, @splat(corners[0][0] - offX)) + @as(Vec3f, @floatFromInt(side.textureY()))*@as(Vec3f, @splat(corners[0][1] - offY)),
+				@as(Vec3f, @floatFromInt(side.textureX()))*@as(Vec3f, @splat(corners[1][0] - offX)) + @as(Vec3f, @floatFromInt(side.textureY()))*@as(Vec3f, @splat(corners[1][1] - offY)),
+				@as(Vec3f, @floatFromInt(side.textureX()))*@as(Vec3f, @splat(corners[2][0] - offX)) + @as(Vec3f, @floatFromInt(side.textureY()))*@as(Vec3f, @splat(corners[2][1] - offY)),
+				@as(Vec3f, @floatFromInt(side.textureX()))*@as(Vec3f, @splat(corners[3][0] - offX)) + @as(Vec3f, @floatFromInt(side.textureY()))*@as(Vec3f, @splat(corners[3][1] - offY)),
+			};
+
+			const tex: u32 = @intCast(@intFromEnum(pattern));
+
+			var offset: Vec3f = .{0.0, 0.0, 0.0};
+			offset[@intFromEnum(side.vectorComponent())] = if(side.isPositive()) 12.0/16.0 else 4.0/16.0;
```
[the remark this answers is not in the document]
IntegratedQuantum: You make the model when a block requests it, and cache it in a hashmap, like it's done for all the other rotation modes.

src/rotation.zig
```diff
@@ -486,6 +486,383 @@ pub const RotationModes = struct {
...
+
+			const floatingPointFix: Vec3f = @splat(16.0);
```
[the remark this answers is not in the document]
OneAvargeCoder193: without it, there appears to be small gaps, so idk what would be causing that because this seems to fix it

src/rotation.zig
```diff
@@ -486,6 +486,383 @@ pub const RotationModes = struct {
...
+			};
+
+			return res;
+		}
+
+		fn addQuads(pattern: Pattern, side: Neighbor, out: *main.List(main.models.QuadInfo)) void {
+			switch(pattern) {
+				.dot => {
+					out.append(rotateQuad(.{
+						.{0.25, 0.25},
```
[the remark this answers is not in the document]
OneAvargeCoder193: yes, unless i find a way to make them variable width

src/rotation.zig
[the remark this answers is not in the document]
IntegratedQuantum: There are small gaps even with it. It's the old T-junction issue.

src/rotation.zig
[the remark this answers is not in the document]
Argmaster: That would be a really unfortunate change IMO

src/rotation.zig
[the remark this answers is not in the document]
OneAvargeCoder193: Where do you see the gaps? The images in the pr were from before i fixed this issue.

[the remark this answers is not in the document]
IntegratedQuantum: By the fact that you reverted this I assume you saw the same gaps with it?

src/rotation.zig
```diff
@@ -357,35 +357,242 @@ pub const RotationModes = struct {
...
+					return .cross;
+				},
+				else => undefined,
+			};
+		}
+
 		pub fn createBlockModel(zon: ZonElement) u16 {
-			const modelId = zon.as([]const u8, "cubyz:cube");
-			if(branchModels.get(modelId)) |modelIndex| return modelIndex;
+			const radius = zon.get(u32, "radius", 4);
```
IntegratedQuantum, line 572: Should be an `f32`, while we should not use partial pixels in the base game, there is no need to restrict it for addons.
Furthermore even without that `3.5` (→ 7 pixels) would also make sense.

src/rotation.zig
```diff
@@ -357,35 +357,242 @@ pub const RotationModes = struct {
...
+						.{1.0, 0.0},
+						.{1.0, 1.0},
+					}, pattern, minF, maxF, side));
+				},
 			}
 		}
 
+		fn getPattern(data: BranchData, side: Neighbor) ?Pattern {
+			const posX = Neighbor.fromRelPos(side.textureX()).?;
+			const posNegX = Neighbor.fromRelPos(side.textureX()).?.reverse();
```
IntegratedQuantum, line 490: positive negative x?

src/rotation.zig
```diff
@@ -357,35 +357,242 @@ pub const RotationModes = struct {
...
+						}
+
+						return .{.line = .{.dir = dir}};
+					}
+
+					var dir: Direction = .negXDir;
+
+					if(connectedNegY) {
+						dir = .negYDir;
+					}
```
IntegratedQuantum, line 534: This seems redundant, it's running the same code 2 lines down.

src/rotation.zig
```diff
@@ -357,35 +357,242 @@ pub const RotationModes = struct {
...
+				.normal = norm,
+				.opaqueInLod = 1,
+				.textureSlot = tex,
+			};
+
+			return res;
+		}
+
+		fn addQuads(pattern: Pattern, side: Neighbor, radius: u32, out: *main.List(main.models.QuadInfo)) void {
+			const minF: f32 = @as(f32, @floatFromInt(8 - radius))/16.0;
```
IntegratedQuantum, line 434: No hungarian notation please (unless the F means something else?)

src/rotation.zig
```diff
@@ -357,35 +357,242 @@ pub const RotationModes = struct {
...
 
-		fn degenerateQuad(quad: *main.models.QuadInfo) void {
-			for(&quad.corners) |*corner| {
-				corner.* = @splat(0.5);
+			const offX: f32 = @floatFromInt(@intFromBool(@reduce(.Add, side.textureX()) < 0));
+			const offY: f32 = @floatFromInt(@intFromBool(@reduce(.Add, side.textureY()) < 0));
+
+			const norm: Vec3f = @floatFromInt(side.relPos());
+
+			const corns = .{
```
IntegratedQuantum, line 405: I don't like that you use so many unusual abbreviations here, `corns` (🌽) `norm`, `tex`.
Here you could just inline `norm` and `tex` into there single use case. Same with `res`, why not just return directly?
As for 🌽s, I'd rename the above variable to corners2d and then you don't have the naming conflict.

IntegratedQuantum (changes requested):

src/rotation.zig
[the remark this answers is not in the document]
OneAvargeCoder193: it looked fine without

src/rotation.zig
```diff
@@ -400,101 +400,97 @@ pub const RotationModes = struct {
 			const offX: f32 = @floatFromInt(@intFromBool(@reduce(.Add, side.textureX()) < 0));
 			const offY: f32 = @floatFromInt(@intFromBool(@reduce(.Add, side.textureY()) < 0));
 
-			const norm: Vec3f = @floatFromInt(side.relPos());
-
-			const corns = .{
+			const corners2d = .{
```
IntegratedQuantum, line 403: But these are 3d!

src/rotation.zig
```diff
@@ -357,35 +357,234 @@ pub const RotationModes = struct {
...
+			const res: main.models.QuadInfo = .{
+				.corners = .{
+					corners2d[0] + offset,
+					corners2d[1] + offset,
+					corners2d[2] + offset,
+					corners2d[3] + offset,
+				},
+				.cornerUV = originalCorners,
+				.normal = @floatFromInt(side.relPos()),
+				.opaqueInLod = 1,
```
IntegratedQuantum, line 422: You should not set this manually, it's intended mainly for internal usage.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #1195, opened 2025-03-11
Add cactus flowers, tussock, and cactus top texture

careeoki:
image
Cactus flowers appear uncommonly on top of cacti. Tussock occasionally generates in small patches throughout the desert.
image
Cactus now has log rotation and a top texture. Cactus flowers have log rotation as well.


careeoki:
@ikabod-kee 
Let me know what you think. I'd like to keep the top primarily pink.
image

IntegratedQuantum:
Just an idea, but why not use a approximative cone for some flowers?
You could easily do this with three quads that intersect like a tetrahedron.
Then here you could make the outside pink and the inside white.

careeoki:
> Just an idea, but why not use a approximative cone for some flowers? 

I tried something similar, but I didn't like how it looked. It was also difficult to see from the ground on taller cacti.

careeoki:
I pushed the new cactus flower.

IntegratedQuantum:
How would the cactus flower work with the new branch cacti?

careeoki:
> How would the cactus flower work with the new branch cacti?

I guess it wouldn't, since they can't connect to eachother. This flower would be too big for a branch cactus anyway.
I imagine that only the large/old (full block) cacti would produce flowers.

PR #1197, opened 2025-03-12
Add `rotateZ` for all rotations

Argmaster:
This pull request adds functions for rotating block data around Z axis.

### DIFF
--- a/src/chunk.zig
+++ b/src/chunk.zig
@@ -108,6 +108,24 @@ pub const Neighbor = enum(u3) { // MARK: Neighbor
 			},
 		}
 	}
+
+	// Returns the neighbor that is rotated by 90 degrees counterclockwise around the x axis.
+	pub inline fn rotateX(self: Neighbor) Neighbor {
+		const arr = [_]Neighbor{.dirPosY, .dirNegY, .dirPosX, .dirNegX, .dirDown, .dirUp};
+		return arr[@intFromEnum(self)];
+	}
+
+	// Returns the neighbor that is rotated by 90 degrees counterclockwise around the y axis.
+	pub inline fn rotateY(self: Neighbor) Neighbor {
+		const arr = [_]Neighbor{.dirPosX, .dirNegX, .dirDown, .dirUp, .dirPosY, .dirNegY};
+		return arr[@intFromEnum(self)];
+	}
+
+	// Returns the neighbor that is rotated by 90 degrees counterclockwise around the z axis.
+	pub inline fn rotateZ(self: Neighbor) Neighbor {
+		const arr = [_]Neighbor{.dirUp, .dirDown, .dirPosY, .dirNegY, .dirNegX, .dirPosX};
+		return arr[@intFromEnum(self)];
+	}
 };
 
 /// Gets the index of a given position inside this chunk.
--- a/src/chunk.zig
+++ b/src/chunk.zig
@@ -109,18 +109,6 @@ pub const Neighbor = enum(u3) { // MARK: Neighbor
 		}
 	}
 
-	// Returns the neighbor that is rotated by 90 degrees counterclockwise around the x axis.
-	pub inline fn rotateX(self: Neighbor) Neighbor {
-		const arr = [_]Neighbor{.dirPosY, .dirNegY, .dirPosX, .dirNegX, .dirDown, .dirUp};
-		return arr[@intFromEnum(self)];
-	}
-
-	// Returns the neighbor that is rotated by 90 degrees counterclockwise around the y axis.
-	pub inline fn rotateY(self: Neighbor) Neighbor {
-		const arr = [_]Neighbor{.dirPosX, .dirNegX, .dirDown, .dirUp, .dirPosY, .dirNegY};
-		return arr[@intFromEnum(self)];
-	}
-
 	// Returns the neighbor that is rotated by 90 degrees counterclockwise around the z axis.
 	pub inline fn rotateZ(self: Neighbor) Neighbor {
 		const arr = [_]Neighbor{.dirUp, .dirDown, .dirPosY, .dirNegY, .dirNegX, .dirPosX};
--- a/src/rotation.zig
+++ b/src/rotation.zig
@@ -567,14 +567,12 @@ pub const RotationModes = struct {
 
 					const x: f32 = (@as(f32, @floatFromInt(i)) - 0.5)*2.0;
 					const y: f32 = (@as(f32, @floatFromInt(j)) - 0.5)*2.0;
-					const z: f32 = (@as(f32, @floatFromInt(k)) - 0.5)*2.0;
 
 					const rX = @intFromBool(x*cos - y*sin > 0);
 					const rY = @intFromBool(x*sin + y*cos > 0);
-					const rZ = @intFromBool(z > 0);
 
 					if(hasSubBlock(@intCast(old), @intCast(i), @intCast(j), @intCast(k))) {
-						new |= subBlockMask(rX, rY, rZ);
+						new |= subBlockMask(rX, rY, @intCast(k));
 					}
 				};
 				rotateTable[old] = new;
--- a/src/rotation.zig
+++ b/src/rotation.zig
@@ -618,7 +618,9 @@ pub const RotationModes = struct {
 				const y: u1 = @truncate(j);
 				const z: u1 = @truncate(k);
 				const xyz = rotateTable[i*4 + j*2 + k];
-				new |= subBlockMask(xyz[0], xyz[1], xyz[2]) * @intFromBool(hasSubBlock(@truncate(data), x, y, z));
+				if(hasSubBlock(@truncate(data), x, y, z)) {
+					new |= subBlockMask(xyz[0], xyz[1], xyz[2]);
+				}
 			};
 			return new;
 		}
--- a/src/rotation.zig
+++ b/src/rotation.zig
@@ -19,10 +19,10 @@ const RayIntersectionResult = struct {
 };
 
 pub const Degrees = enum(u2) {
-	@"90" = 0,
-	@"180" = 1,
-	@"270" = 2,
-	@"0" = 3,
+	@"0" = 0,
+	@"90" = 1,
+	@"180" = 2,
+	@"270" = 3,
 };
 
 // TODO: Why not just use a tagged union?
--- a/src/rotation.zig
+++ b/src/rotation.zig
@@ -27,6 +27,15 @@ pub const RotationMode = struct { // MARK: RotationMode
 		fn model(block: Block) u16 {
 			return blocks.meshes.modelIndexStart(block);
 		}
+		fn rotateX(data: u16) u16 {
+			return data;
+		}
+		fn rotateY(data: u16) u16 {
+			return data;
+		}
+		fn rotateZ(data: u16) u16 {
+			return data;
+		}
 		fn generateData(_: *main.game.World, _: Vec3i, _: Vec3f, _: Vec3f, _: Vec3i, _: ?Neighbor, _: *Block, _: Block, blockPlacing: bool) bool {
 			return blockPlacing;
 		}
@@ -112,6 +121,13 @@ pub const RotationMode = struct { // MARK: RotationMode
 
 	model: *const fn(block: Block) u16 = &DefaultFunctions.model,
 
+	// Rotates block data 90 degrees counterclockwise round X axis.
+	rotateX: *const fn(data: u16) u16 = DefaultFunctions.rotateX,
+	// Rotates block data 90 degrees counterclockwise round X axis.
+	rotateY: *const fn(data: u16) u16 = DefaultFunctions.rotateY,
+	// Rotates block data 90 degrees counterclockwise round X axis.
+	rotateZ: *const fn(data: u16) u16 = DefaultFunctions.rotateZ,
+
 	createBlockModel: *const fn(zon: ZonElement) u16 = &DefaultFunctions.createBlockModel,
 
 	/// Updates the block data of a block in the world or places a block in the world.
@@ -181,6 +197,21 @@ pub const RotationModes = struct {
 
 		pub fn model(block: Block) u16 {
 			return blocks.meshes.modelIndexStart(block) + @min(block.data, 5);
+		}
+
+		fn rotateX(data: u16) u16 {
+			var neighbor: Neighbor = @enumFromInt(data);
+			return @intFromEnum(neighbor.rotateX());
+		}
+
+		fn rotateY(data: u16) u16 {
+			var neighbor: Neighbor = @enumFromInt(data);
+			return @intFromEnum(neighbor.rotateY());
+		}
+
+		fn rotateZ(data: u16) u16 {
+			var neighbor: Neighbor = @enumFromInt(data);
+			return @intFromEnum(neighbor.rotateZ());
 		}
 
 		pub fn generateData(_: *main.game.World, _: Vec3i, _: Vec3f, _: Vec3f, _: Vec3i, neighbor: ?Neighbor, currentData: *Block, _: Block, blockPlacing: bool) bool {
@@ -261,6 +292,17 @@ pub const RotationModes = struct {
 
 		fn reset() void {
 			fenceModels.clearRetainingCapacity();
+		}
+
+		fn rotateZ(data: u16) u16 {
+			var currentData: FenceData = @bitCast(@as(u4, @truncate(data)));
+			currentData = .{
+				.isConnectedNegY = currentData.isConnectedNegX,
+				.isConnectedPosY = currentData.isConnectedPosX,
+				.isConnectedPosX = currentData.isConnectedNegY,
+				.isConnectedNegX = currentData.isConnectedPosY,
+			};
+			return @as(u4, @bitCast(currentData));
 		}
 
 		fn fenceTransform(quad: *main.models.QuadInfo, data: FenceData) void {
@@ -345,15 +387,15 @@ pub const RotationModes = struct {
 		const BranchData = packed struct(u6) {
 			enabledConnections: u6,
 
-			pub fn init(blockData: u16) BranchData {
+			pub inline fn init(blockData: u16) BranchData {
 				return .{.enabledConnections = @truncate(blockData)};
 			}
 
-			pub fn isConnected(self: @This(), neighbor: Neighbor) bool {
+			pub inline fn isConnected(self: @This(), neighbor: Neighbor) bool {
 				return (self.enabledConnections & Neighbor.bitMask(neighbor)) != 0;
 			}
 
-			pub fn setConnection(self: *@This(), neighbor: Neighbor, value: bool) void {
+			pub inline fn setConnection(self: *@This(), neighbor: Neighbor, value: bool) void {
 				if(value) {
 					self.enabledConnections |= Neighbor.bitMask(neighbor);
 				} else {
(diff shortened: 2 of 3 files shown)


src/chunk.zig
```diff
@@ -108,6 +108,24 @@ pub const Neighbor = enum(u3) { // MARK: Neighbor
 			},
 		}
 	}
+
+	// Returns the neighbor that is rotated by 90 degrees counterclockwise around the x axis.
+	pub inline fn rotateX(self: Neighbor) Neighbor {
```
Argmaster, line 113: Note about `inline` here:

I [checked it with godbolt](https://godbolt.org/z/sWbhrjo3W) and it seems like those functions as well as some of the other ones I marked `inline` are not automatically inlined by optimizer, even in `ReleaseFast`. Since this will be in critical path of world generation, I though it would be beneficial to avoid excessive calls in favor of unconditional jumps of inlined functions.

src/rotation.zig
```diff
@@ -507,13 +591,88 @@ pub const RotationModes = struct {
...
+		inline fn subBlockMask(x: u1, y: u1, z: u1) u8 {
 			return @as(u8, 1) << ((@as(u3, x)*2 + @as(u3, y))*2 + z);
 		}
-		fn hasSubBlock(stairData: u8, x: u1, y: u1, z: u1) bool {
+		inline fn hasSubBlock(stairData: u8, x: u1, y: u1, z: u1) bool {
 			return stairData & subBlockMask(x, y, z) == 0;
 		}
 
+		fn rotateX(data: u16) u16 {
+			comptime var rotateTable: [8][3]u1 = undefined;
```
Argmaster, line 602: My sincere hope here is that this table will be calculated once at compile time and used as lookup table at runtime, hence sparing us some math and sin cos lookups.

src/rotation.zig
```diff
@@ -507,13 +591,88 @@ pub const RotationModes = struct {
...
+					@intFromBool(y*sin + z*cos > 0),
+				};
+			};
+			var new: u8 = 0;
+			for(0..2) |i| for(0..2) |j| for(0..2) |k| {
+				const x: u1 = @truncate(i);
+				const y: u1 = @truncate(j);
+				const z: u1 = @truncate(k);
+				const xyz = rotateTable[i*4 + j*2 + k];
+				new |= subBlockMask(xyz[0], xyz[1], xyz[2]) * @intFromBool(hasSubBlock(@truncate(data), x, y, z));
```
Argmaster, line 621: This part here is to avoid conditional branch, it is possible that this is excessive, maybe optimizer does that anyway, I didn't check that.

IntegratedQuantum: Yes, most optimizers will just use a `cmov` instruction here. Using multiplication may actually make it harder for the optimizer to discover your intentions.

IntegratedQuantum: Furthermore a branch might even be better, if we assume that most of the time stair blocks are fully filled.

Argmaster:
```suggestion
				if(hasSubBlock(@truncate(data), x, y, z)) {
					new |= subBlockMask(xyz[0], xyz[1], xyz[2]);
				}
```

src/rotation.zig
Argmaster: Yes, this would probably be optimal.

src/chunk.zig
Argmaster: Depends if we cache rotated structures. 
If we do, with palettes only on IO we will perform:
- 1 translation (`O(xyz)` complexity, X, Y, Z are dimensions of structure), 
- N pastes (`O(xyz)` complexity),
- up to 1+1+1 rotations (90, 180, 270 degrees) (single 90 degree rotation has `O(xyz)` complexity since rotations are O(1) as we use lookup tables for most of them)

This is `O((N+4)*xyz)`.

If we don't cache once rotated structures them we perform:

- 1 translation `O(xyz)`
- N pastes `O(xyz)`
- up to A+2B+3C rotations where A+B+C=N `O(xyz)`

Which yields `O((N+1)*xyz + 2*N*xyz) = O((3*N+1)*xyz)` 

(I assume we get similar number of 0, 90, 180 an 270 degree rotations, hence second 2*N*xyz).

If we use palette with no cache we get:

- N translations `O(xyz)`
- N pastes `O(xyz)`
- A+2B+3C rotations where A+B+C=N `O(xyz)`

Which yields `O(2*N*xyz + 2*N*xyz) = O(4*N*xyz)`

If we get palette with cache we get:

- N translations `O(xyz)`
- N pastes `O(xyz)`
- up to `1+1+1` rotations

Which yields `O((2*N+3)*xyz)`

Best time complexity (if i didn't F anything up) is still no palette with cache for rotated structures.

Argmaster: Ok, I messed up a bit, cause I overlooked configuration where we rotate only palette, then it would be another variable, M effectively yielding `O(2*N*xyz + 6*M)` and `O(2*N*xyz + 3*M)`, IMO still worse, assuming we can afford few hundred MB for cache.

Argmaster: Yeah, and I overlooked that there are 3 axis of rotation, lets assume that we only count Z rotations as they should be most common ones.

Argmaster: As for the cache, if storage space turns out to be a problem, we could limit ourselves to storing one type array and multiple data arrays for required rotations. Then rotation of actual blueprint could be done by changing iteration order. This approach has a caveat  though, a very poor cache locality in certain cases, I am not sure if this is worth the effort, as it might be very punishing for performance.

IntegratedQuantum: If storage space becomes a problem, we can always use palette compression for blueprints. But I hope we don't end up with billions of blocks in blueprints anyways (keep in mind that someone has to place all of them in the first place!)

IntegratedQuantum:
I wonder if it would be best to just restrict the rotations to only the z axis, then we could focus on optimizing these cases (to just one function call, instead of calling rotateZ 3 times in the worst case), without having to worry about any other rotation axis.

@ikabod-kee maybe you have an opinion about this? Are there many cases where you may really want to rotate a structure in any other direction? If so, would be fine about manually rotation things in that case?

src/chunk.zig
Argmaster: > (keep in mind that someone has to place all of them in the first place!)

Well, actually no, eg. for a tree only <40% of blocks in blueprint will be actual blocks that somebody placed, the rest will be air since a non-cuboid build has to be wrapped by a cuboid selection. For example 8 blocks tall tree stem with branches in every direction requires 3x3x10 cuboid, 7.5x times more blocks than the actual build. This gets even worse for things like branches which are even further from a full cuboid.

Screenshot 2025-03-14 044155

IntegratedQuantum: Yeah, but even if we make the conservative assumption that only 10% need to be placed, you'd still need to place over 10 million blocks to get to 1 GB of data.

Argmaster:
I don't see how having all 3 axis available stops you from having bigger rotation tables, since every rotation is applied separately and simply there is no way to create rotation table for combination of rotations in different axis due to how API is designed.

I don't agree with removing rotations around X and Y, these have valid use cases of creating eg. XYZ junctions and in general anything that is symmetrical around X or Y axis. If you don't have  X and Z rotations you can't use commands to create builds that are symmetrical around X or Y since even if you build it symmetrical along Z axis there is no tool to rotate it to horizontal position. This applies for example to tunnels.

Argmaster:
What about spikes, stalagmites and stalactites, reusing branches as flipped top-down, having leaves spawn child nodes point up and down? We will manually create exactly same model twice because we can't rotate it around X and Y?

IntegratedQuantum:
> I don't see how having all 3 axis available stops you from having bigger rotation tables, since every rotation is applied separately and simply there is no way to create rotation table for combination of rotations in different axis due to how API is designed.

I mean that in the sense of focusing our efforts on making one thing faster, rather than having to work on 3 different things (6 including flips).

> What about spikes, stalagmites and stalactites, reusing branches as flipped top-down, having leaves spawn child nodes point up and down? We will manually create exactly same model twice because we can't rotate it around X and Y?

Well, that is an argument for implementing flipping, not for implementing X or Y rotations. Like Ikabod said, there rarely is a reason to turn something on its side.

Argmaster:
Well then you can keep implementation of XY rotations suboptimal but present instead of removing it at all.

IntegratedQuantum:
> Well then you can keep implementation of XY rotations suboptimal but present instead of removing it at all.

Well, it's not just about this code here, but also about the future rotation code. And you can't deny that it would make the structure rotation simpler too to restrict it on flipping and z rotations.
And of course it's just simply more code, and any code has to be maintained.

Argmaster:
So for ship wrecks you will also have separate model laying on side? No complicated fallen tree models either, unless you build separate set of assets for them. I am fairly certain that this will come back as a request, thankfully It won't be me who will be implementing that. Things that were build vertical will stay vertical.

IntegratedQuantum:
I'm sure that both shipwrecks and fallen trees are better to built them sideways to begin with.
For shipwrecks especially, since you'd probably want to build a mast out of fences, which is impossible when it lies on the side.
Generally I think there are also just too many rotation modes that cannot be equivalently displayed sidewards (e.g. fences, torches, mushrooms).

Finally if we limit outselves to just z rotations it could make your idea
> Then rotation of actual blueprint could be done by changing iteration order. 

more viable, since then you could achieve cache locality by iterating the columns (and for flips we could precompute just one separate model)

src/rotation.zig
```diff
@@ -112,6 +115,9 @@ pub const RotationMode = struct { // MARK: RotationMode
 
 	model: *const fn(block: Block) u16 = &DefaultFunctions.model,
 
+	// Rotates block data 90 degrees counterclockwise round X axis.
```
IntegratedQuantum, line 118:
```suggestion
	// Rotates block data 90 degrees counterclockwise round Z axis.
```

src/rotation.zig
```diff
@@ -410,6 +437,20 @@ pub const RotationModes = struct {
 			return blocks.meshes.modelIndexStart(block) + (block.data & 63);
 		}
 
+		fn rotateZ(data: u16) u16 {
+			var old: BranchData = .init(data);
```
IntegratedQuantum, line 441: Should be const

src/rotation.zig
```diff
@@ -507,13 +548,40 @@ pub const RotationModes = struct {
 		pub const id: []const u8 = "stairs";
 		var modelIndex: u16 = 0;
 
-		fn subBlockMask(x: u1, y: u1, z: u1) u8 {
+		inline fn subBlockMask(x: u1, y: u1, z: u1) u8 {
 			return @as(u8, 1) << ((@as(u3, x)*2 + @as(u3, y))*2 + z);
 		}
-		fn hasSubBlock(stairData: u8, x: u1, y: u1, z: u1) bool {
+		inline fn hasSubBlock(stairData: u8, x: u1, y: u1, z: u1) bool {
```
IntegratedQuantum, line 554: These aren't used at runtime in the rotateZ function anymore, so I don't see a reason anymore to mark them inline.

src/rotation.zig
```diff
@@ -507,13 +548,40 @@ pub const RotationModes = struct {
...
+			comptime for(0..256) |old| {
+				var new: u8 = 0;
+
+				for(0..2) |i| for(0..2) |j| for(0..2) |k| {
+					const sin: f32 = @sin(std.math.pi/2.0);
+					const cos: f32 = @cos(std.math.pi/2.0);
+
+					const x: f32 = (@as(f32, @floatFromInt(i)) - 0.5)*2.0;
+					const y: f32 = (@as(f32, @floatFromInt(j)) - 0.5)*2.0;
+					const z: f32 = (@as(f32, @floatFromInt(k)) - 0.5)*2.0;
```
IntegratedQuantum, line 570: There is no need to include the calculations for `z` here, since it doesn't change.

IntegratedQuantum (changes requested):

src/rotation.zig
Argmaster:
```suggestion
			const old: BranchData = .init(data);
```

src/rotation.zig
```diff
@@ -507,13 +548,40 @@ pub const RotationModes = struct {
 		pub const id: []const u8 = "stairs";
 		var modelIndex: u16 = 0;
 
-		fn subBlockMask(x: u1, y: u1, z: u1) u8 {
+		inline fn subBlockMask(x: u1, y: u1, z: u1) u8 {
```
Argmaster, line 551:
```suggestion
		fn subBlockMask(x: u1, y: u1, z: u1) u8 {
```

src/rotation.zig
Argmaster: This covers all possible subblock combinations, to cover only changes in XY plane I would have to separately handle top and bottom half of the block. One is represented by odd bits, other by even bits. How would you implement it then?

Argmaster: This covers all possible subblock combinations, to cover only changes in XY plane I would have to separately handle top and bottom half of the block. One is represented by odd bits, other by even bits. How would you implement it then?

IntegratedQuantum: I would propose just this, and removing the then unused variables:
```diff
- subBlockMask(rX, rY, rZ);
+ subBlockMask(rX, rY, @intCast(k));
```

Argmaster: Ok, I wonder if we maybe should restore angle parameter if we want to have table for 90 and 180, 270 rotations.

IntegratedQuantum: Yeah, it might make sense to do. (probably best to do as an enum with `@"0"` `@"90"`, ... )

Argmaster:
Seems to be working more or less

image

src/rotation.zig
```diff
@@ -18,6 +18,13 @@ const RayIntersectionResult = struct {
 	max: Vec3f,
 };
 
+pub const Degrees = enum(u2) {
+	@"90" = 0,
+	@"180" = 1,
+	@"270" = 2,
+	@"0" = 3,
```
IntegratedQuantum, line 25: shouldn't 0° be mapped to 0?
That's also how you are using it, if I interpret this correctly:
https://github.com/PixelGuys/Cubyz/blob/150ac3c71d6dc2e604cf8050a6d77e20610ed004/src/rotation.zig#L200-L203

src/rotation.zig
```diff
@@ -189,9 +196,18 @@ pub const RotationModes = struct {
...
+			comptime for(0..6) |i| {
+				rotationTable[i] = i;
+			};
+			comptime for(1..4) |a| {
+				for(0..6) |i| {
+					const neighbor: Neighbor = @enumFromInt(rotationTable[i + 6*(a - 1)]);
+					rotationTable[i + 6*a] = neighbor.rotateZ().toInt();
+				}
+			};
+			return rotationTable[data + @as(u16, @intCast(@intFromEnum(angle)))*6];
```
IntegratedQuantum, line 210: Should do bounds checking before plugging a random u16 into a table that only supports 6 values.

Otherwise worlds where the rotation mode of a block was different will crash.

In case of an incompatible value, you can just return 0 for the data.

src/rotation.zig
```diff
@@ -237,9 +253,18 @@ pub const RotationModes = struct {
...
+			comptime for(0..4) |i| {
+				rotationTable[i] = i + 2;
+			};
+			comptime for(1..4) |a| {
+				for(0..4) |i| {
+					const neighbor: Neighbor = @enumFromInt(rotationTable[i + 4*(a - 1)]);
+					rotationTable[i + 4*a] = neighbor.rotateZ().toInt();
+				}
+			};
+			return rotationTable[data + @as(u16, @intCast(@intFromEnum(angle)))*4] - 2;
```
IntegratedQuantum, line 267: same here

src/rotation.zig
```diff
@@ -279,15 +304,24 @@ pub const RotationModes = struct {
...
+						.isConnectedNegY = old.isConnectedNegX,
+						.isConnectedPosY = old.isConnectedPosX,
+						.isConnectedPosX = old.isConnectedNegY,
+						.isConnectedNegX = old.isConnectedPosY,
+					};
+					rotateTable[i + 16*a] = @as(u4, @bitCast(new));
+				}
 			};
-			return @as(u4, @bitCast(currentData));
+			return rotateTable[data + @as(u16, @intCast(@intFromEnum(angle)))*16];
```
IntegratedQuantum, line 324: same here

src/rotation.zig
```diff
@@ -555,29 +601,32 @@ pub const RotationModes = struct {
...
+
+						if(hasSubBlock(@intCast(old), @intCast(i), @intCast(j), @intCast(k))) {
+							new |= subBlockMask(rX, rY, @intCast(k));
+						}
+					};
+					rotateTable[old + a*256] = new;
+				}
 			};
-			return rotateTable[data];
+			return rotateTable[data + @as(u16, @intCast(@intFromEnum(angle)))*256];
```
IntegratedQuantum, line 629: same here

src/rotation.zig
```diff
@@ -956,16 +1005,26 @@ pub const RotationModes = struct {
...
+						.negY = old.negX,
+						.posY = old.posX,
+						.posX = old.negY,
+						.negX = old.posY,
+					};
+					rotateTable[i + 32*a] = @as(u5, @bitCast(new));
+				}
+			};
+
+			return rotateTable[data + @as(u16, @intCast(@intFromEnum(angle)))*32];
```
IntegratedQuantum, line 1027: same here

src/rotation.zig
```diff
@@ -1072,17 +1131,27 @@ pub const RotationModes = struct {
...
+						.posY = old.posX,
+						.negY = old.negX,
+						.negX = old.posY,
+						.posX = old.negY,
+					};
+					rotateTable[i + 64*a] = @as(u6, @bitCast(new));
+				}
+			};
+
+			return rotateTable[data + @as(u16, @intCast(@intFromEnum(angle)))*64];
```
IntegratedQuantum, line 1154: same here

src/rotation.zig
```diff
@@ -189,9 +196,18 @@ pub const RotationModes = struct {
 			return blocks.meshes.modelIndexStart(block) + @min(block.data, 5);
 		}
 
-		fn rotateZ(data: u16) u16 {
-			var neighbor: Neighbor = @enumFromInt(data);
-			return @intFromEnum(neighbor.rotateZ());
+		fn rotateZ(data: u16, angle: Degrees) u16 {
+			comptime var rotationTable: [24]u8 = undefined;
```
IntegratedQuantum, line 200: Why not use a `[4][6]u8`? This would reduce the access code and make it more safe (less multiplications by hardcoded array length 6):
```diff
- rotationTable[data + @as(u16, @intCast(@intFromEnum(angle)))*6]
+ rotationTable[@intFromEnum(angle)][data]
```

IntegratedQuantum (changes requested):

src/rotation.zig
Argmaster: Yes, definitely, I was experimenting with having zero covered by condition, but I ended up just making tables for everything.

src/rotation.zig
Argmaster: No particular reason, except avoiding second dereference, but I am all in for that change.

IntegratedQuantum: In Zig arrays are flat in memory, the assembly code of both approaches should be identical.

src/rotation.zig
```diff
@@ -112,6 +122,9 @@ pub const RotationMode = struct { // MARK: RotationMode
 
 	model: *const fn(block: Block) u16 = &DefaultFunctions.model,
 
+	// Rotates block data 90 degrees counterclockwise round Z axis.
```
IntegratedQuantum, line 125: The doc comment is no longer correct.
```suggestion
	// Rotates block data counterclockwise around the Z axis.
```

src/rotation.zig
```diff
@@ -226,6 +254,21 @@ pub const RotationModes = struct {
...
+				rotationTable[0][i] = i + 2;
+			};
+			comptime for(1..4) |a| {
+				for(0..4) |i| {
+					const neighbor: Neighbor = @enumFromInt(rotationTable[a - 1][i]);
+					rotationTable[a][i] = neighbor.rotateZ().toInt();
+				}
+			};
+			if(data >= 4) return 0;
+			return rotationTable[@intFromEnum(angle)][data] - 2;
```
IntegratedQuantum, line 269: Could you precompute the -2 into the table?

src/rotation.zig
```diff
@@ -994,6 +1135,29 @@ pub const RotationModes = struct {
...
+						.posY = old.posX,
+						.negY = old.negX,
+						.negX = old.posY,
+						.posX = old.negY,
+					};
+					rotationTable[a][i] = @as(u6, @bitCast(new));
+				}
+			};
+
+			return rotationTable[@intFromEnum(angle)][data];
```
IntegratedQuantum, line 1158: Here it's missing the bounds check.

IntegratedQuantum (changes requested):

src/rotation.zig
Argmaster: Yes, but you have to check if I didn't mess it up.

PR #1207, opened 2025-03-16
Blueprint asset loading

Argmaster:
This pull request implements loading blueprint and SBB configuration assets.

Depends on: #1226


Argmaster:
I am starting to get somewhere

image

There is a bug somewhere causing child structures to spawn incorrectly.

Argmaster:
image

IntegratedQuantum:
Well, for trees we currently use blue noise. We could definitely do the same thing at multiple different scales, and put larger structures into a bigger size class.

IntegratedQuantum:
(Also, I think that's outside the scope of this PR)

Argmaster:
Do we plan strongholds, mines etc? This algorithm doesn't really scale well for that purpose unfortunately, as noting prevents it from overwriting already existing blocks unless you basically create all segments of particular layout. If I recall correctly wave function collapse was deemed too expensive, but maybe we could have it from some of those more specialized structures?

Do we have a way to dynamically specify how many chunks structure affects? Is it possible to add such control? (ignoring exponential growth of generation cost)

Argmaster:
> (Also, I think that's outside the scope of this PR)

Indeed.

IntegratedQuantum:
> Do we plan strongholds, mines etc? This algorithm doesn't really scale well for that purpose unfortunately, as noting prevents it from overwriting already existing blocks unless you basically create all segments of particular layout.

Yeah, we definitely need a different approach there. I think it would be best to use code on the large scale (maze generators, or the room generators from roguelikes), and then use structures for individual rooms (or individual walls, depending on how finely you define the individual components).
We can also try wave function collapse in that context, though I do think that explicit generators still have an advantage over wave function collapse, which tends to generate isolated rooms if you are not careful with the input.

> Do we have a way to dynamically specify how many chunks structure affects? Is it possible to add such control? (ignoring exponential growth of generation cost)

Not yet, currently it's hardcoded to 32 blocks in the SImpleStructureGen.zig
The intention is that for larger structures you'd create a new, different structure map generator that produces small (chunk-sized) structures and puts them into the structure map.
The cost should be manageable, if the structures are rare enough.

IntegratedQuantum:
That's all the requirements done, right?
You just need to rebase and it's ready for review right?

Argmaster:
Umm, it kind of seg faults after rebase😝

Argmaster:
We should probably also have parameter for randomizing rotation of first structure, that is not included here yet

IntegratedQuantum:
> Ok it was caused by change of generate signature.

Yeah, it's kind of bad that I use ptrCast for these, maybe we can find a safer way to implement function pointer casts: #1223

> that is not included here yet

Yeah, this PR is probably big enough anyways.

IntegratedQuantum (changes requested):
I'm not too happy with the overall runtime storage here.
Structures are rather expensive to generate, so we should do as much as we can during loading and use a small and simple in-memory storage format:
- Structures should be rotated during loading instead of being rotated and cached during runtime
- Particularly there should be no mutexes and reference counting going on at runtime. These are not cheap (even without contention), and shouldn't be done for every structure (and we have to assume that people will use many small structures)
- There should be no hashmap lookups during generation:
  - Things like blueprints should be stored as pointers instead of storing ids
  - children should be slices (of less than 20 elements) instead of using the block id for indexing all the time. This also allow making them not nullable, saving another 8 byte for each entry there.
- There should be no string ids left after loading is done.
- Finally the things that are stored, should probably just use an arena, since they are only loaded once and never changed as long as the world is loaded. This would also help reduce the cleanup code (and it is awful if you have to wait 20 seconds for the game to close because it needs to free every tiny allocation manually)

Argmaster:
Ok, I did some thinking and realistically even with hundreds of models with all rotations precalculated we should fit in tens or hundreds of megabytes, which is not a big deal, definitely we should move in that direction.

src/rotation.zig
```diff
@@ -821,13 +821,16 @@ pub const RotationModes = struct {
...
 						}
 					};
 					rotationTable[a][old] = new;
 				}
 			};
-			if(data >= 256) return 0;
+			if(data > 0b11111111) {
+				std.log.err("Invalid stair data: {}", .{data});
+				return 0;
+			}
```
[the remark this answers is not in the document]
Argmaster: Indeed it is, I'll do that

assets/cubyz/blocks/sbb/origin.zig.zon
```diff
@@ -0,0 +1,8 @@
+.{
+	.tags = .{.sbbOrigin},
```
Argmaster, line 2: If I leave it blank I get `"Block {s} is missing 'tags' field"`

src/blueprint.zig
```diff
@@ -36,6 +41,44 @@ pub const Blueprint = struct {
...
+					std.debug.assert(newX < @as(i64, @intCast(new.blocks.width)));
+					std.debug.assert(newY >= 0);
+					std.debug.assert(newY < @as(i64, @intCast(new.blocks.depth)));
+
+					new.blocks.set(@intCast(newX), @intCast(newY), z, block.rotateZ(angle));
+				}
+			}
+		}
+		return new;
+	}
```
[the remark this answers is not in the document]
Argmaster: #1225

src/structure_building_blocks.zig
```diff
@@ -0,0 +1,309 @@
...
+	}
+	var list = arena_allocator.alloc(Child, zon.array.items.len);
+	for(zon.array.items, 0..) |entry, i| {
+		list[i] = Child.initFromZon(childName, stringId, i, entry);
+	}
+	return .init(arena_allocator, list);
+}
+
+const Child = struct {
+	structureId: []const u8,
```
Argmaster, line 185: Child structures must be resolved in second pass, as they may not be loaded when parent structure is, hence ID is required to fetch correct structure then.

IntegratedQuantum: Well then please store a separate data structure for the unresolved entries that's cleared after loading.
There is no need for this data to persist after loading. It's just clogging the cache lines.

IntegratedQuantum: You could literally just store a **StructureBuildingBlock to this location and it's string id in a list, then resolve them easily.

Argmaster: Do they have guaranteed pointer stability? There is return value optimization, but is that actually guaranteed?
And this will also mean that `structureCache` has to be `std.StringHashMapUnmanaged(*StructureBuildingBlock)`, not `std.StringHashMapUnmanaged(StructureBuildingBlock)`, unless entries have guaranteed pointer stability after inserting new elements?

IntegratedQuantum: Ah right, I guess you have to use id and child index instead then.

IntegratedQuantum (changes requested):
Please rebase to include changes from #1225

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,288 @@
...
+const parseBlock = main.blocks.parseBlock;
+const Degrees = main.rotation.Degrees;
+const hashInt = main.utils.hashInt;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+var arena = main.heap.NeverFailingArenaAllocator.init(main.globalAllocator);
+const arenaAllocator = arena.allocator();
+
+var structureCache: ?std.StringHashMapUnmanaged(StructureBuildingBlock) = null;
+var blueprintCache: ?std.StringHashMapUnmanaged([4]*BlueprintEntry) = null;
```
IntegratedQuantum, line 19: I don't think these should be optional.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,288 @@
...
+var childrenToResolve: List(struct {parentId: []const u8, colorName: []const u8, colorIndex: usize, childIndex: usize, structureId: []const u8}) = undefined;
+
+const BlueprintEntry = struct {
+	blueprint: Blueprint,
+	info: Info,
```
IntegratedQuantum, line 24: We have two fields here, `Info` has two fields, together the functions are less than 100 lines, `Info` is a pretty generic name.
I see no reason to have it as a seperate struct. I would just inline it here.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,288 @@
...
+			index: u32,
+			block: Block,
+
+			pub inline fn direction(self: StructureBlock) Neighbor {
+				return @enumFromInt(self.block.data);
+			}
+		};
+
+		originBlock: StructureBlock,
+		childrenBlocks: List(StructureBlock),
```
IntegratedQuantum, line 49: Should be ListUnmanaged in anticipation of #1181

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,288 @@
...
+
+		fn deinit(self: Info) void {
+			self.childrenBlocks.deinit();
+		}
```
IntegratedQuantum, line 53: We probably don't need this, it's using the arena right?

src/server/terrain/structure_building_blocks.zig
IntegratedQuantum: Actually this should be a slice, since after loading the number of elements is known.

IntegratedQuantum (changes requested):
Alright, this is still too big to review in my opinion, so I'd propose to move all the assets.zig changes to a separate PR, with some stubs for all the functions you need in structure_building_blocks.zig.

src/server/terrain/structure_building_blocks.zig
Argmaster: This is a list of blocks that appear in blueprint. Single color child block may appear multiple times. Only way to determine this value is to do one detection pass through whole blueprint and count how many child blocks there are. Will be as requested.

src/server/terrain/structure_building_blocks.zig
Argmaster: Ok, will be your way.

Argmaster:
Here, I cut out #1226 and #1227.
I would be also thankful if you could preserve pull request description as merge commit description instead of list of meaningless commit titles.

IntegratedQuantum:
> I would be also thankful if you could preserve pull request description as merge commit description instead of list of meaningless commit titles.

Sorry, I'm just working with the tools github gives me, and there is no option to put the PR description into the commit message. I'd have to manually copy it from the top, which seems rather annoying to do. Either way there is a link to the PR in the commit, so it shouldn't be a big deal.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+				continue;
+			};
+
+			std.log.debug("Resolving child structure '{s}'->'{s}'->'{d}' to '{s}'", .{entry.parentId, entry.colorName, entry.childIndex, entry.structureId});
+			parent.children[entry.colorIndex].items[entry.childIndex].structure = child;
+			continue;
+		}
+		childrenToResolve.deinit();
+	}
+	std.log.debug("Registered {} structure building blocks", .{structureCache.count()});
```
IntegratedQuantum, line 228: So it's printing the count 3 times? Twice here and once in assets.zig?
I think we should only print it once.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+	structureCache.ensureTotalCapacity(arenaAllocator.allocator, structures.count()) catch unreachable;
+	childrenToResolve = .init(main.stackAllocator);
+	{
+		var iterator = structures.iterator();
+		while(iterator.next()) |entry| {
+			const value = StructureBuildingBlock.initFromZon(entry.key_ptr.*, entry.value_ptr.*) catch continue;
+			const key = arenaAllocator.dupe(u8, entry.key_ptr.*);
+			const result = structureCache.getOrPut(arenaAllocator.allocator, key) catch unreachable;
+
+			if(result.found_existing) {
```
IntegratedQuantum, line 203: The key comes from another hashmap, so this cannot happen.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+				std.log.err("Ignoring duplicated structure building block: '{s}'", .{entry.key_ptr.*});
+				continue;
+			}
+			result.value_ptr.* = value;
+			std.log.debug("Registered structure building block: '{s}'", .{entry.key_ptr.*});
+		}
+	}
+	{
+		for(childrenToResolve.items) |entry| {
+			const parent = structureCache.getPtr(entry.parentId) orelse {
```
IntegratedQuantum, line 213: This should not happen either. Can use `.?` to assert it.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+				continue;
+			};
+			const child = structureCache.getPtr(entry.structureId) orelse {
+				std.log.err("Could not find child structure '{s}' for child resolution.", .{entry.structureId});
+				continue;
+			};
+
+			std.log.debug("Resolving child structure '{s}'->'{s}'->'{d}' to '{s}'", .{entry.parentId, entry.colorName, entry.childIndex, entry.structureId});
+			parent.children[entry.colorIndex].items[entry.childIndex].structure = child;
+			continue;
```
IntegratedQuantum, line 224:
```suggestion
```

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+			const parent = structureCache.getPtr(entry.parentId) orelse {
+				std.log.err("Could not find parent structure '{s}' for child resolution.", .{entry.structureId});
+				continue;
+			};
+			const child = structureCache.getPtr(entry.structureId) orelse {
+				std.log.err("Could not find child structure '{s}' for child resolution.", .{entry.structureId});
+				continue;
+			};
+
+			std.log.debug("Resolving child structure '{s}'->'{s}'->'{d}' to '{s}'", .{entry.parentId, entry.colorName, entry.childIndex, entry.structureId});
```
IntegratedQuantum, line 222:
```suggestion
			std.log.debug("Resolved child structure '{s}'->'{s}'->'{d}' to '{s}'", .{entry.parentId, entry.colorName, entry.childIndex, entry.structureId});
```

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+	}
+	std.log.debug("Registered {} structure building blocks", .{structureCache.count()});
+}
+
+pub fn registerChildBlock(numericId: u16, stringId: []const u8) void {
+	const index: u16 = @intCast(childBlockNumericIdMap.count());
+	childBlockNumericIdMap.put(arenaAllocator.allocator, numericId, index) catch unreachable;
+	// Take only color name from the ID.
+	var iterator = std.mem.splitBackwardsScalar(u8, stringId, '/');
+	const colorName = iterator.next().?;
```
IntegratedQuantum, line 236: Can use `.first()`

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+	childBlockStringId.append(arenaAllocator, arenaAllocator.dupe(u8, colorName));
+	std.log.debug("Child block '{s}' {} ('{s}' {}) ", .{colorName, index, stringId, numericId});
```
IntegratedQuantum, line 238:
```suggestion
	std.log.debug("Structure child block '{s}' {} ('{s}' {}) ", .{colorName, index, stringId, numericId});
```

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+}
+
+pub fn registerBlueprints(blueprints: *std.StringHashMap([]u8)) !void {
+	if(blueprintCache.capacity() != 0) {
```
IntegratedQuantum, line 242: Should be an assertion, see comment above.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+		std.log.err("Attempting to register new blueprints without resetting cache.", .{});
+		return error.AlreadyRegistered;
+	}
+
+	originBlockNumericId = parseBlock(originBlockStringId).typ;
```
IntegratedQuantum, line 247: I generally don't like seeing function aliases, because it makes me think that it is a local function.
Especially here, where it's only used once it's more confusing than just writing the path in my opinion.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+const ZonElement = main.ZonElement;
+const Blueprint = main.blueprint.Blueprint;
+const List = main.List;
+const ListUnmanaged = main.ListUnmanaged;
+const AliasTable = main.utils.AliasTable;
+const Neighbor = main.chunk.Neighbor;
+const Block = main.blocks.Block;
+const parseBlock = main.blocks.parseBlock;
+const Degrees = main.rotation.Degrees;
+const hashInt = main.utils.hashInt;
```
IntegratedQuantum, line 13: unused

IntegratedQuantum (changes requested):

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+var arena = main.heap.NeverFailingArenaAllocator.init(main.globalAllocator);
+const arenaAllocator = arena.allocator();
+
+var structureCache: std.StringHashMapUnmanaged(StructureBuildingBlock) = .{};
+var blueprintCache: std.StringHashMapUnmanaged([4]*BlueprintEntry) = .{};
+var childrenToResolve: List(struct {parentId: []const u8, colorName: []const u8, colorIndex: usize, childIndex: usize, structureId: []const u8}) = undefined;
```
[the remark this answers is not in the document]
Argmaster: Outcome of redoing same code over and over again, stuff just gets left behind if it compiles. They should all be here, since they are global variables/constants.

src/assets.zig
```diff
@@ -162,6 +171,68 @@ pub fn readAllZonFilesInAddons(
...
+	for(0..relativeFilePath.len - fileSuffixLen) |i| {
+		if(relativeFilePath[i] == '\\') {
+			assetId[addonName.len + 1 + i] = '/';
+		} else {
+			assetId[addonName.len + 1 + i] = relativeFilePath[i];
+		}
+	}
+
+	return assetId;
+}
```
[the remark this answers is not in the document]
Argmaster: #1229

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+}
+
+const Child = struct {
+	structure: *StructureBuildingBlock,
+	chance: f32,
+
+	fn initFromZon(parentId: []const u8, colorName: []const u8, colorIndex: usize, childIndex: usize, zon: ZonElement) Child {
+		const structureId = zon.get([]const u8, "structure", "");
+		if(structureId.len == 0) {
+			std.log.warn("['{s}'->'{s}'->'{d}'] Child node has empty structure field.", .{parentId, colorName, childIndex});
```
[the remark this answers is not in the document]
Argmaster: Yes, by misspelling field name. But it should be an error, there is no reason to leave it blank from asset designer perspective, it should always be treated as bug.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,290 @@
...
+		for(childBlockStringId.items, 0..) |colorName, colorIndex| {
+			self.children[colorIndex] = initChildTableFromZon(stringId, colorName, colorIndex, childrenZon.getChild(colorName));
+		}
+		return self;
+	}
+	pub fn getBlueprint(self: StructureBuildingBlock, rotation: Degrees) *BlueprintEntry {
+		return self.blueprint[@intFromEnum(rotation)];
+	}
+	pub fn pickChild(self: StructureBuildingBlock, block: BlueprintEntry.StructureBlock, seed: *u64) Child {
+		return self.children[block.index].sample(seed).*;
```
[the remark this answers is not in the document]
Argmaster: Because the function is called `pickChild` 😝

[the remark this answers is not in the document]
Argmaster: See this is a random change of contract I will have to account for in #1227, most likely spotted during rebuild, but since zig stops on first error found, I will have to rebuild the project for every single one of those small changes. The contract is not perfect, but it is not unacceptable either, is it?

src/server/terrain/structure_building_blocks.zig
Argmaster: This is true only after the value gets assigned, but the message is useful to know for which child structure assignment failed, so it has to be printed before the assignment. This is outcome of real debug done during development.

src/server/terrain/structure_building_blocks.zig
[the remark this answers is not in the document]
IntegratedQuantum: fair enough

src/assets.zig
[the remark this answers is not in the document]
IntegratedQuantum: Well, but if they do come up during implementing, then you could directly make a small PR, and it might even be merged before you finished fully implementing the feature.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -1,25 +1,31 @@
 const std = @import("std");
 
-const main = @import("root");
+// const main = @import("root");
+const main = @import("../../main.zig");
```
IntegratedQuantum, line 4: unintended change

src/server/terrain/structure_building_blocks.zig
```diff
@@ -1,25 +1,31 @@
...
 const Degrees = main.rotation.Degrees;
-const hashInt = main.utils.hashInt;
 const NeverFailingAllocator = main.heap.NeverFailingAllocator;
 
 var arena = main.heap.NeverFailingArenaAllocator.init(main.globalAllocator);
 const arenaAllocator = arena.allocator();
 
 var structureCache: std.StringHashMapUnmanaged(StructureBuildingBlock) = .{};
-var blueprintCache: std.StringHashMapUnmanaged([4]*BlueprintEntry) = .{};
+var blueprintCache: std.StringHashMapUnmanaged([4]BlueprintEntry) = .{};
```
IntegratedQuantum, line 19: Should this be `*[4]BlueprintEntry`? Otherwise we end up copying it in every place it's used, and BlueprintEntry is not a small struct.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -98,6 +89,8 @@ const BlueprintEntry = struct {
 			std.log.err("[{s}] No origin block found.", .{stringId});
 			return error.NoOriginBlock;
 		}
+		self.childBlocks = arenaAllocator.alloc(StructureBlock, childBlocks.items.len);
+		@memcpy(self.childBlocks, childBlocks.items);
```
IntegratedQuantum, line 93: Use `arenaAllocator.dupe`

src/server/terrain/structure_building_blocks.zig
```diff
@@ -111,59 +104,50 @@ pub fn isOriginBlock(block: Block) bool {
...
-var originBlockNumericId: u16 = 0;
-
-// Maps global child block numeric ID to index used to locally represent that child block.
-var childBlockNumericIdMap: std.AutoHashMapUnmanaged(u16, u16) = .{};
-var childBlockStringId: ListUnmanaged([]const u8) = .{};
-
 pub const StructureBuildingBlock = struct {
 	children: []AliasTable(Child),
-	blueprint: [4]*BlueprintEntry,
+	blueprints: [4]BlueprintEntry,
```
IntegratedQuantum, line 109: Should this be `*[4]BlueprintEntry`?

src/server/terrain/structure_building_blocks.zig
```diff
@@ -111,59 +104,50 @@ pub fn isOriginBlock(block: Block) bool {
...
 		const childrenZon = zon.getChild("children");
 		for(childBlockStringId.items, 0..) |colorName, colorIndex| {
-			self.children[colorIndex] = initChildTableFromZon(stringId, colorName, colorIndex, childrenZon.getChild(colorName));
+			self.children[colorIndex] = try initChildTableFromZon(stringId, colorName, colorIndex, childrenZon.getChild(colorName));
 		}
 		return self;
 	}
 	pub fn getBlueprint(self: StructureBuildingBlock, rotation: Degrees) *BlueprintEntry {
-		return self.blueprint[@intFromEnum(rotation)];
+		return self.blueprints[@intFromEnum(rotation)];
```
IntegratedQuantum, line 131: Should be `&self.blueprints...` otherwise this won't compile.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -111,59 +104,50 @@ pub fn isOriginBlock(block: Block) bool {
...
 	}
-	pub fn pickChild(self: StructureBuildingBlock, block: BlueprintEntry.StructureBlock, seed: *u64) Child {
-		return self.children[block.index].sample(seed).*;
+	pub fn pickChild(self: StructureBuildingBlock, block: BlueprintEntry.StructureBlock, seed: *u64) *StructureBuildingBlock {
```
IntegratedQuantum, line 133: Can return `*const StructureBuildingBlock`

IntegratedQuantum (changes requested):

src/server/terrain/structure_building_blocks.zig
Argmaster: Maybe, I can't compile it anyway cause there is no createAssetStringID implementation.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,149 @@
...
+
+var blueprintCache: std.StringHashMapUnmanaged(*[4]BlueprintEntry) = .{};
+var childrenToResolve: List(struct {parentId: []const u8, colorName: []const u8, colorIndex: usize, childIndex: usize, structureId: []const u8}) = undefined;
+
+const originBlockStringId = "cubyz:sbb/origin";
+var originBlockNumericId: u16 = 0;
+
+// Maps global child block numeric ID to index used to locally represent that child block.
+var childBlockNumericIdMap: std.AutoHashMapUnmanaged(u16, u16) = .{};
+var childBlockStringId: ListUnmanaged([]const u8) = .{};
```
IntegratedQuantum, line 25: Just for my own understanding: What is this needed for again? Was it for printing nicer log messages?

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,149 @@
...
+
+const BlueprintEntry = struct {
+	blueprint: Blueprint,
+	originBlock: StructureBlock,
+	childBlocks: []StructureBlock,
+
+	const StructureBlock = struct {
+		x: i32,
+		y: i32,
+		z: i32,
```
IntegratedQuantum, line 35: Could these be `i16`? Blueprints shouldn't be that large anyways.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,149 @@
...
+		index: u32,
```
IntegratedQuantum, line 36: This can also be `u16`, since we cannot have 65536 sbbChild blocks anyways.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -0,0 +1,149 @@
...
+		block: Block,
```
IntegratedQuantum, line 37: Do we need the `typ`, or could we just store the data?

src/server/terrain/structure_building_blocks.zig
Argmaster: In configuration files you specify what should spawn at given child block by specifying color name of that child block and since we load list of child blocks dynamically, string IDs are used to determine color names. [Example](https://github.com/Argmaster/Cubyz/blob/feature/SBBGen/assets/cubyz/sbb/tree/oak/stem0.zig.zon)

src/server/terrain/structure_building_blocks.zig
Argmaster: Its a bit late for limiting capacity of blueprints, we already chose u32 by using Array3D, although 60k blocks is reasonable limitation it makes interface inconsistent, some parts support bigger blueprints, other wont.

But there is a bug here, those should be `u32` to match Array3D capacity limitations, not `i32`.

If you insist I can change it to u16, I don't think there is a use for that sign here.

src/server/terrain/structure_building_blocks.zig
Argmaster: Depends if we restrict x, y, z to `u16`, if we don't it will be 4 bytes aligned anyway.

src/server/terrain/structure_building_blocks.zig
IntegratedQuantum: Shouldn't it be a hashmap then, so you can lookup the value without needing to iterate the list?

Argmaster: No, indexes of IDs match `StructureBlock.index` and can be obtained through mapping block typ to index with `childBlockNumericIdMap`

src/server/terrain/structure_building_blocks.zig
IntegratedQuantum: Oh, I thought you may have used the sign to store precomputed offsets from the origin.

Either way it should be 16 bits. The blueprint capturing just cannot reasonably capture such giant things anything, since the render distance (or actually I think it's limited by the entity animation distance even) doesn't get that high.

src/server/terrain/structure_building_blocks.zig
Argmaster: I suppose it will make more sense after #1240

IntegratedQuantum: Ah I see

src/server/terrain/structure_building_blocks.zig
Argmaster: Hmm I could be possible, but I am also not 100% sure I can do that, indexes don't like me, so I would prefer to tinker with it in #1227 to be able to see what are the results.

It will be u16 then.

Argmaster:
Should be all clear now.

IntegratedQuantum (approved):

PR #1210, opened 2025-03-17
Add hunger, hunger bar, and related functions on the Player struct.

catmeow72:
This PR implements what is required for hunger to work, including functions on the player struct for future accessories to use hunger, and for food to heal hunger.

Note that this PR does not implement any accessories or make food heal hunger just yet

This PR also includes three new icons for the hunger bar, from @careeoki


careeoki:
I've done a PR for new hunger textures that are more consistent with hearts.

src/game.zig
```diff
@@ -630,6 +631,28 @@ pub const Player = struct { // MARK: Player
...
+	pub fn useHunger(amount: f32) bool {
+		if(super.hunger >= amount) {
+			super.hunger -= amount;
+			return true;
+		} else {
+			return false;
+		}
+	}
+
+	pub fn healHunger(amount: f32) bool {
```
Argmaster, line 644: I don't think "heal" makes sense here, I would opt for addHunger and subtractHunger, or maybe even a single updateHunger function, since this is not something called often.

src/game.zig
```diff
@@ -630,6 +631,28 @@ pub const Player = struct { // MARK: Player
...
+		const added = super.hunger + amount;
+		if(super.hunger == super.maxHunger) {
+			return false;
+		}
```
Argmaster, line 648:
```suggestion
                 if(super.hunger == super.maxHunger) { 
                         return false; 
                 }
                const added = super.hunger + amount; 
```

src/gui/windows/hungerbar.zig
```diff
@@ -0,0 +1,70 @@
...
+	.isHud = true,
+	.showTitleBar = false,
+	.hasBackground = false,
+	.hideIfMouseIsGrabbed = false,
+	.closeable = false,
+};
+
+var heartTexture: Texture = undefined;
+var halfHeartTexture: Texture = undefined;
+var deadHeartTexture: Texture = undefined;
```
Argmaster, line 31: Update variable names to reflect that it is hunger not health.

Argmaster (changes requested):
As @IntegratedQuantum agreed to call secondary resource "energy" please rename it everywhere in PR.

src/game.zig
catmeow72: Had changed it when renaming to energy, and it should make more sense now.

src/game.zig
```diff
@@ -630,6 +631,28 @@ pub const Player = struct { // MARK: Player
 			}
 		}
 	}
+
+	pub fn useEnergy(amount: f32) bool {
```
Argmaster, line 635: That is a curious choice of words, I wonder what @IntegratedQuantum thinks about it.
I am used to less "special" names for functions that modify a value - `setX`, `updateX`, `addX`.
Aside from my concerns about words used, I see that for health there is single `addHealth` function which accepts positive and negative values. Can we be consistent and not introduce completely different logic for energy if it is not necessary? (`addEnergy`)

Argmaster: Although, that function is part of `Inventory.Sync` which you are not touching yet, I assume you will while implementing sync with server side?

catmeow72: Didn't notice that the Inventory.Sync was used for more than just inventory operations, but also things like health. I'm going to add the sync code there now that I know that needs to be done.

IntegratedQuantum: I think on function updateEnergy would be best, instead of two separate functions.

catmeow72: The intention was to make the energy functions like the health functions, so I moved the functionality to Inventory.Sync.

catmeow72:
Accidentally ran the wrong formatter, I'll amend the commit.

src/Inventory.zig
```diff
@@ -1674,6 +1715,50 @@ pub const Command = struct { // MARK: Command
 			};
 		}
 	};
+
+	const AddEnergy = struct { // MARK: AddEnergy
```
IntegratedQuantum, line 1719: In my opinion this should not be a separate thing.
Instead all the operations that change energy (e.g. eating food, using accessories) should update the energy themselves. These would need to be synchronized here anyways (to consume the food item, or to do other (healing) effects in the case of accessories).

IntegratedQuantum (changes requested):

src/Inventory.zig
```diff
@@ -510,46 +520,44 @@ pub const Command = struct { // MARK: Command
...
+	}, addHealth: struct {
+		target: ?*main.server.User,
+		health: f32,
+		cause: main.game.DamageType,
+		previous: f32,
+	}, addEnergy: struct {
+		target: ?*main.server.User,
+		energy: f32,
+		previous: f32,
+	}};
```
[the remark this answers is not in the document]
catmeow72: Should I add a trailing comma?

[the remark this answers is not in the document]
IntegratedQuantum: yes

src/Inventory.zig
catmeow72: Should I keep the addEnergy base operation, and remove the AddEnergy struct?

catmeow72: Did that for now, as it seems to make most sense for me.

IntegratedQuantum: Yeah, the base operation should be kept.

IntegratedQuantum (approved):
Looks good.
It would be nice if you could add eating next.

PR #1215, opened 2025-03-19
Bark block

Argmaster:
This pull request adds bark block for existing wood types. Basically a wood block but covered with bark on all sides. Additionally it used stair rotation mode to allow creation of greater detail on trees.

Bark blocks drop normal logs when broken.
There is a crafting recipe for converting logs into corresponding bark blocks and converting bark block into planks.

Screenshot 2025-03-19 231356


Argmaster:
Although they probably should not be used extensively in world generation due to possible performance penalties, players really 
 can make wonders with block like this.

OneAvargeCoder193:
aren't log tops used in the simple tree as branches? should those be replaced with log branches instead?

Argmaster:
@OneAvargeCoder193 current tree models are not here to stay, there is #1207 to replace current generation. Although I suppose I should check if the data values used for tops as branches should be updated in `SimpleTreeModel`.

careeoki:
I think we should wait for #357 before doing this. With log rotation, tops can be used to hide the log_top texture when making curvy trees or large branches.
Also, "oak top" wouldn't make sense anymore. They should be called "oak bark"

Argmaster:
Ok, maybe it should just be a separate thing, `cubyz:oak_bark` at all and we could allow tops to rest in peace.

IntegratedQuantum (changes requested):
Instead of adding a third block type for each tree, I would prefer having a rotation mode that combines them all into one block.
We could easily support 3 bits for log rotation, 2 bits for whether it has a bark top or bottom and 8 bits for all stair states.

Argmaster:
Ehh it's one of those.
I guess I will work on it.

Argmaster:
Taken by OneAvargeCoder193 in #1496

PR #1226, opened 2025-03-22
Add blocks for representing origin of structure building blocks and children spawn points 

Argmaster:
Extracted from #1207


IntegratedQuantum (approved):
Let's get them in already, then @ikabod-kee can maybe already start making some example structures.

PR #1227, opened 2025-03-22
Add Structure Building Blocks based Generator

Argmaster:
This pull request adds generator type which uses Structure Building Blocks and blueprints to generate structures.

[discord](https://discord.com/channels/443805812390100992/702167480654757991/1349008429087326249)

<p align="center">
  
</p>
<p align="center">
  
<p align="center">
  
</p>

Depends on: #1141
Depends on: #1197 
Depends on: #1207 
Depends on: #1240 
Depends on: #1226 
Depends on: #1241
Depends on: #1253 
Depends on: #1333

Related to: #67
Related to: #194
Related to: #1029
Related to: #1172


src/server/terrain/simple_structures/SBBGen.zig
```diff
@@ -0,0 +1,120 @@
...
+const ServerChunk = main.chunk.ServerChunk;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const parseBlock = main.blocks.parseBlock;
+const hashInt = main.utils.hashInt;
+const hashCombine = main.utils.hashCombine;
+
+pub const id = "cubyz:sbb";
+pub const generationMode = .floor;
+
+const SubstitutionMap = std.AutoHashMapUnmanaged(u16, u16);
```
Argmaster, line 21: This is duplicated, as importing it from blueprint causes stack overflow in `refAllDeclsRecursiveExceptCImports` due to reference cycle.

src/utils.zig
```diff
@@ -1497,6 +1497,20 @@ pub const BinaryWriter = struct {
 	}
 };
 
+// https://stackoverflow.com/questions/5889238/why-is-xor-the-default-way-to-combine-hashes
```
Argmaster, line 1500: #1241

IntegratedQuantum:
Would you mind adding your example structures to a biome, to show it works, so that @ikabod-kee could try and play around with it to see if there are any problems?

src/blueprint.zig
```diff
@@ -104,6 +106,41 @@ pub const Blueprint = struct {
...
+					if(sbb.isOriginBlock(block) or sbb.isChildBlock(block)) continue;
+
+					if(substitutions) |map| {
+						if(map.get(block.typ)) |entry| {
+							block.typ = entry;
+						}
+					}
+
+					sw: switch(mode) {
+						.all => chunk.updateBlockInGeneration(worldX, worldY, worldZ, block),
```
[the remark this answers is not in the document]
Argmaster: Well, `.all` is intended to break stuff, replace literally everything, that's why its called `.all` and it should not be used for structures that can overlap, we can have `.degradable` mode for that purpose, for now I have used `.replaceAir`. This needs polishing anyway when we introduce void blocks. For small houses it makes sense to have `.all` replace everything bcs void/air blocks will make sure to not destroy terrain while walls of house should replace terrain.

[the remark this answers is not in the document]
Argmaster: Btw paste modes should also be implemented for normal paste and here should be option to specify list of blocks to omit in either of them, or at least `noWater` option etc. for other fluids.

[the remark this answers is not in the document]
IntegratedQuantum: Checking a list of blocks (or worse a list of block tags, since if we'd want to be more general) during generation for millions of structures isn't going to work out well. Even the current pattern, having a runtime switch here might be too slow, but at least that can be easily migrated to `comptime` if needed.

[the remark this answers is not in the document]
Argmaster: I would prefer to leave it mostly as is for now and optimize this in the future. In particular, we could have function to create copy of blueprint with substitution table applied (we will need this for `/replace` anyway), then we could use it to substitute child and origin blocks with air during loading after we discover where they are located. `noWater` etc should likely be implemented as substitution table (or even better, a mask!). Substitution tables given to generator should likely pre-compute blueprints with those substitution tables applied. But this complicates the code a lot, and I don't think it is necessary from the beggining, it will take time before we use that particular generator everywhere, so there some margin for future developemnt I think.

src/blueprint.zig
```diff
@@ -270,3 +308,5 @@ pub const Blueprint = struct {
 		}
 	}
 };
+
+const sbb = main.server.terrain.structure_building_blocks;
```
IntegratedQuantum, line 312: Why is this at the bottom of the file?

src/server/terrain/simple_structures/SBBGen.zig
```diff
@@ -0,0 +1,122 @@
+const std = @import("std");
```
IntegratedQuantum, line 1: file name should be sbb_gen according to the name guidelines

src/blueprint.zig
```diff
@@ -104,6 +106,42 @@ pub const Blueprint = struct {
...
+
+				for(0..self.blocks.height) |z| {
+					const worldZ = startZ + @as(i32, @intCast(z));
+					if(!chunk.liesInChunk(worldX, worldY, worldZ)) continue;
+
+					var block = self.blocks.get(x, y, z);
+					if(sbb.isOriginBlock(block) or sbb.isChildBlock(block)) continue;
+
+					if(substitutions) |map| {
+						if(map.get(block.typ)) |entry| {
```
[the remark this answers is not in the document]
IntegratedQuantum: Then we could also avoid the hashing problems entirely.

src/server/terrain/simple_structures/SBBGen.zig
```diff
@@ -0,0 +1,122 @@
...
+const parseBlock = main.blocks.parseBlock;
+const hashInt = main.utils.hashInt;
+const hashCombine = main.utils.hashCombine;
+
+pub const id = "cubyz:sbb";
+pub const generationMode = .floor;
+
+const SubstitutionMap = std.AutoHashMapUnmanaged(u16, u16);
+
+const SBBGen = @This();
```
IntegratedQuantum, line 23: file name should be SbbGen according to the name guidelines

src/server/terrain/simple_structures/SBBGen.zig
```diff
@@ -0,0 +1,122 @@
...
+		placeSbb(self, child, pasteX + childBlock.x, pasteY + childBlock.y, pasteZ + childBlock.z, childBlock.direction(), chunk, seed);
+	}
+}
+
+fn alignDirections(input: Neighbor, desired: Neighbor) !usize {
+	var current = input;
+	for(0..4) |i| {
+		if(current == desired) return i;
+		current = current.rotateZ();
+	}
```
IntegratedQuantum, line 120: Please use a lookup table here, to avoid doing so many operations at runtime.

src/server/terrain/simple_structures/SBBGen.zig
```diff
@@ -0,0 +1,122 @@
...
+	placeSbb(self, self.structureRef, x, y, z - 1, Neighbor.dirUp, chunk, seed);
+}
+
+fn placeSbb(self: *SBBGen, structure: *const sbb.StructureBuildingBlock, x: i32, y: i32, z: i32, placementDirection: Neighbor, chunk: *ServerChunk, seed: *u64) void {
+	const origin = structure.blueprints[0].originBlock;
+	const rotationCount = alignDirections(origin.direction(), placementDirection) catch |err| {
+		std.log.err("Could not align directions {s} and {s} error: {s}", .{@tagName(origin.direction()), @tagName(placementDirection), @errorName(err)});
+		return;
+	};
+	const rotated = structure.blueprints[rotationCount];
```
IntegratedQuantum, line 100: to avoid the copy
```suggestion
	const rotated = &structure.blueprints[rotationCount];
```

src/server/terrain/simple_structures/SBBGen.zig
```diff
@@ -0,0 +1,122 @@
...
+	const rotatedOrigin = rotated.originBlock;
+
+	const pasteX: i32 = x - rotatedOrigin.x - placementDirection.relX();
+	const pasteY: i32 = y - rotatedOrigin.y - placementDirection.relY();
+	const pasteZ: i32 = z - rotatedOrigin.z - placementDirection.relZ();
```
IntegratedQuantum, line 105:
```suggestion
	const pastePosition = Vec3i{x, y, z} - rotatedOrigin - placementDirection.relPos();
```

IntegratedQuantum (changes requested):

src/blueprint.zig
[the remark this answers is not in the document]
Argmaster: We already discussed why palette compression and decompression can result in bad time complexity. Since replacements are known at instantiation time we can apply them once and create a modified copy of the blueprint(s) for that purpose (I hope so at least) This would also make it viable to do more complicated replacements like with masks from #1284 and patterns from the /set command. Blueprints could be stored globally if we append hash or sth of replacement map to the ID to avoid duplicating child blueprints. That's an optimization tho, maybe not necessary.

I was thinking about the relation between commands like /set and /undo and I came to conclusion that (since we have to capture and store blueprint of world before editing) it would be better to have all the WE functions as methods on Blueprint. This would make the pattern for each command the same (capture, store, modify, paste) instead of requiring dedicated undo implementation for each command. With the current state you also can't use WE commands programatically since they are tied to the command system. With that change Blueprint would have the replace function which would return a modified copy which we could use here to generate modified blueprints.

That said I agree that substitutions will be better off added separately.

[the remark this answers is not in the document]
IntegratedQuantum: > We already discussed why palette compression and decompression can result in bad time complexity.

I cannot remember that discussion, in fact palette compression is so good that we use it for chunks already.

> Since replacements are known at instantiation time we can apply them once and create a modified copy of the blueprint(s)

Well this has quite a big memory footprint, to the point where palette compression might be a requirement anyways.

> This would also make it viable to do more complicated replacements like with masks from https://github.com/PixelGuys/Cubyz/pull/1284 and patterns from the /set command.

Masks could be done by modifying only the palette as well, I think.

As for patterns, they don't seem particularly useful when caching the structure, so we should not need to worry about it in this discussion.

> I was thinking about the relation between commands like /set and /undo

I think that world edit commands have fundamentally different performance requirements, they are only executed once so they can get away with doing a lot more stuff.

Chunk generation on the other hand makes up 50% of the chunk generation+loading time which is one of the main bottlenecks for render distance currently. So we have to make it as fast as possible.
And I think it's safe to assume that this system will be used for a majority of structures in the game, starting with trees.

src/server/terrain/simple_structures/SBBGen.zig
Argmaster: Well, either sbb_gen or SbbGen, can't be both, I assume latter since this is a file struct.

src/blueprint.zig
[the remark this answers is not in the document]
Argmaster: Ok, I am not sure, we discussed something about using palettes for IO possibly in first blueprint PR but there are a lot of comments there so its hard to find.
Mind that unlike chunks blueprints don't have compile time known size limitations so this would involve implementation of dynamically sized PaletteCompressedRegion.

src/blueprint.zig
Argmaster: I think I was searching for source of stack overflow in refAllDeclsRecursiveExceptCImports, tried if order of definitions matters and didn't clean up afterwards.

assets/cubyz/biomes/forest/base.zig.zon
```diff
@@ -13,12 +13,18 @@
 		"2 to 3 cubyz:soil",
 	},
 	.structures = .{
+		.{
+			.id = "cubyz:sbb",
+			.structure = "cubyz:tree/oak/stem0",
+			.placeMode = .replaceAir,
```
Argmaster, line 19:
```suggestion
			.placeMode = .degradable,
```

assets/cubyz/biomes/grassland.zig.zon
```diff
@@ -13,6 +13,12 @@
 		"2 to 3 cubyz:soil",
 	},
 	.structures = .{
+		.{
+			.id = "cubyz:sbb",
+			.structure = "cubyz:tree/oak/stem0",
+			.placeMode = .replaceAir,
```
Argmaster, line 19:
```suggestion
			.placeMode = .degradable,
```

src/server/terrain/simple_structures/SBBGen.zig
IntegratedQuantum: Yeah, if it's a struct then SbbGen

src/blueprint.zig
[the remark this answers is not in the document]
IntegratedQuantum: yes indeed, shouldn't be too difficult though.

src/server/terrain/simple_structures/_list.zig
```diff
@@ -5,3 +5,4 @@ pub const GroundPatch = @import("GroundPatch.zig");
 pub const SimpleTreeModel = @import("SimpleTreeModel.zig");
 pub const SimpleVegetation = @import("SimpleVegetation.zig");
 pub const Stalagmite = @import("Stalagmite.zig");
+pub const SBBGen = @import("SBBGen.zig");
```
IntegratedQuantum, line 8:
```suggestion
pub const SbbGen = @import("SbbGen.zig");
```

Argmaster:
Ok, this should wait for #1324 and it should already incorporate void blocks into `pasteInGeneration` behavior.

Argmaster:
Alright, now it has void blocks handling implemented, unfortunately this breaks assets I have added as an example, as they used air. For purpose of fixing assets broken like this it would be nice to have `-c` flag for `/set` command that would apply that command to clipboard instead of selection.

src/blueprint.zig
```diff
@@ -20,6 +21,8 @@ const BlockStorageType = u32;
 const BinaryWriter = main.utils.BinaryWriter;
 const BinaryReader = main.utils.BinaryReader;
 
+const SubstitutionMap = std.AutoHashMapUnmanaged(u16, u16);
+
```
IntegratedQuantum, line 25: unused

src/blueprint.zig
```diff
@@ -105,9 +108,47 @@ pub const Blueprint = struct {
 		}
 		return .{.success = self};
 	}
+
+	pub const PasteMode = enum {all, degradable};
+
+	pub fn pasteInGeneration(self: Blueprint, pos: Vec3i, chunk: *ServerChunk, mode: PasteMode) void {
+		switch(mode) {
+			.all => _pasteInGeneration(self, pos, chunk, .all),
+			.degradable => _pasteInGeneration(self, pos, chunk, .degradable),
```
IntegratedQuantum, line 117:
```suggestion
			inline else => |comptimeMode| _pasteInGeneration(self, pos, chunk, comptimeMode),
```

src/blueprint.zig
```diff
@@ -105,9 +108,47 @@ pub const Blueprint = struct {
...
+
+		for(0..self.blocks.width) |x| {
+			const worldX = startX + @as(i32, @intCast(x));
+
+			for(0..self.blocks.depth) |y| {
+				const worldY = startY + @as(i32, @intCast(y));
+
+				for(0..self.blocks.height) |z| {
+					const worldZ = startZ + @as(i32, @intCast(z));
+					if(!chunk.liesInChunk(worldX, worldY, worldZ)) continue;
```
IntegratedQuantum, line 134: `liesInChunk` expects relative coordinates, but you give it world coordinates?
I think you got your names wrong.

src/blueprint.zig
```diff
@@ -105,9 +108,47 @@ pub const Blueprint = struct {
...
+
+					const block = self.blocks.get(x, y, z);
+					if(sbb.isOriginBlock(block) or sbb.isChildBlock(block) or block.typ == voidType) continue;
```
IntegratedQuantum, line 137: Could we replace all origin and child blocks by void blocks during loading?
Then we would only need one comparison for each placed block, instead of 2 + a hashmap lookup.

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -0,0 +1,95 @@
...
+const ZonElement = main.ZonElement;
+const Neighbor = main.chunk.Neighbor;
+const ServerChunk = main.chunk.ServerChunk;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const parseBlock = main.blocks.parseBlock;
+
+pub const id = "cubyz:sbb";
+pub const generationMode = .floor;
+
+const SubstitutionMap = std.AutoHashMapUnmanaged(u16, u16);
```
IntegratedQuantum, line 20: unused

src/server/terrain/simple_structures/_list.zig
```diff
@@ -5,3 +5,4 @@ pub const GroundPatch = @import("GroundPatch.zig");
 pub const SimpleTreeModel = @import("SimpleTreeModel.zig");
 pub const SimpleVegetation = @import("SimpleVegetation.zig");
 pub const Stalagmite = @import("Stalagmite.zig");
+pub const SbbGen = @import("SbbGen.zig");
```
IntegratedQuantum, line 8: Please keep it sorted alphabetically.

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -0,0 +1,95 @@
...
+}
+
+pub fn generate(self: *SbbGen, _: GenerationMode, x: i32, y: i32, z: i32, chunk: *ServerChunk, _: CaveMapView, _: CaveBiomeMapView, seed: *u64, _: bool) void {
+	placeSbb(self, self.structureRef, Vec3i{x, y, z}, Neighbor.dirUp, chunk, seed);
+}
+
+fn placeSbb(self: *SbbGen, structure: *const sbb.StructureBuildingBlock, placementPosition: Vec3i, placementDirection: Neighbor, chunk: *ServerChunk, seed: *u64) void {
+	const origin = structure.blueprints[0].originBlock;
+	const rotationCount = alignDirections(origin.direction(), placementDirection) catch |err| {
+		std.log.err("Could not align directions {s} and {s} error: {s}", .{@tagName(origin.direction()), @tagName(placementDirection), @errorName(err)});
```
IntegratedQuantum, line 54: Not a blocker for this PR, but in the future we definitely need a better error message for this, including the names of the structures.

IntegratedQuantum (changes requested):

src/blueprint.zig
Argmaster: Ideally I would like this alongside `/set` being a method instead of function, with all the replacements being implemented as load time.

Argmaster: I would also prefer to have EditMasks for that

Argmaster: Alright, i suppose back to being draft until #1284 #1352 #1337 are done

Argmaster: Generic way of replacing some blocks with different blocks

IntegratedQuantum: But should we use the generic way? We already know where all the child and origin blocks are, why not just update them directly instead of iterating through all blocks in the volume again, running some generic (=expensive) test function on all blocks?

Argmaster: Alright, maybe we should do it non generic way

Argmaster: Okay, so only drawback is that rotated copies need to be created before BlueprintEntry.init because it removes origin and child blocks in place.

IntegratedQuantum: Can't you replace them after you found all the child blocks?

Argmaster: I could, but why bother, if I need to make those copies anyway. Code would get much more complicated if I had to actually rotate those blocks instead of just rediscovering them for every rotation. I also do It only once, while opening world, so again, why bother.

Argmaster:
Alr cooked as requested

src/blueprint.zig
```diff
@@ -105,9 +105,46 @@ pub const Blueprint = struct {
...
+			inline else => |comptimeMode| _pasteInGeneration(self, pos, chunk, comptimeMode),
+		}
+	}
+
+	fn _pasteInGeneration(self: Blueprint, pos: Vec3i, chunk: *ServerChunk, comptime mode: PasteMode) void {
+		const startX = pos[0];
+		const startY = pos[1];
+		const startZ = pos[2];
+
+		for(0..self.blocks.width) |x| {
```
IntegratedQuantum, line 122: You should iterate in steps of `chunk.pos.voxelSize`, otherwise this is going to be quite wasteful when generating structures in LOD chunks.

src/blueprint.zig
```diff
@@ -280,3 +317,8 @@ pub fn registerVoidBlock(block: Block) void {
 	voidType = block.typ;
 	std.debug.assert(voidType != 0);
 }
+
+pub fn getVoidBlock() Block {
+	std.debug.assert(voidType != null);
```
IntegratedQuantum, line 322: `voidType.?` already is an assertion.
```suggestion
```
Also I think we should find a way to make this not optional, maybe even have a defined id for it (like air which is always 0). But not in this PR of course.

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -0,0 +1,93 @@
...
+const GenerationMode = terrain.biomes.SimpleStructureModel.GenerationMode;
+const CaveMapView = terrain.CaveMap.CaveMapView;
+const CaveBiomeMapView = terrain.CaveBiomeMap.CaveBiomeMapView;
+const sbb = terrain.structure_building_blocks;
+const Blueprint = main.blueprint.Blueprint;
+const ZonElement = main.ZonElement;
+const Neighbor = main.chunk.Neighbor;
+const ServerChunk = main.chunk.ServerChunk;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const parseBlock = main.blocks.parseBlock;
```
IntegratedQuantum, line 15: unused, also avoid aliasing functions as stated in the contributing guidelines.

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -0,0 +1,93 @@
...
+
+pub const id = "cubyz:sbb";
+pub const generationMode = .floor;
+
+const SbbGen = @This();
+
+structure: []const u8,
```
IntegratedQuantum, line 22: All uses of this can be replaced by `structureRef.id`.

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -0,0 +1,93 @@
...
+	const self = arenaAllocator.create(SbbGen);
+	self.* = .{
+		.structure = structureId,
+		.structureRef = structureRef,
+		.placeMode = std.meta.stringToEnum(Blueprint.PasteMode, parameters.get([]const u8, "placeMode", "degradable")) orelse Blueprint.PasteMode.degradable,
+	};
+	return self;
+}
+
+pub fn generate(self: *SbbGen, _: GenerationMode, x: i32, y: i32, z: i32, chunk: *ServerChunk, _: CaveMapView, _: CaveBiomeMapView, seed: *u64, _: bool) void {
```
IntegratedQuantum, line 45: For the future: It might be worth exploring to use the CaveMap here.
E.g. we could use it to only generate buildings in a reasonably flat area, or check if there is enough open space for a tree to grow..

src/blueprint.zig
```diff
@@ -105,9 +105,46 @@ pub const Blueprint = struct {
...
+
+		for(0..self.blocks.width) |x| {
+			const localX = startX + @as(i32, @intCast(x));
+
+			for(0..self.blocks.depth) |y| {
+				const localY = startY + @as(i32, @intCast(y));
+
+				for(0..self.blocks.height) |z| {
+					const localZ = startZ + @as(i32, @intCast(z));
+					if(!chunk.liesInChunk(localX, localY, localZ)) continue;
```
IntegratedQuantum, line 130: Also instead of checking this for every block, why not just clamp the inputs to the chunk, as is done in most other generators?
We don't need to iterate through a volume if it's not even a part of the current chunk.

IntegratedQuantum (changes requested):

src/server/terrain/simple_structures/SbbGen.zig
Argmaster: Ye, sorry, old change

src/server/terrain/simple_structures/SbbGen.zig
Argmaster: #1357

Not that trivial, you first have to resolve all children before you know how high the structure will be.

src/blueprint.zig
Argmaster: Alr I changed that code a bit but honestly I am not sure how to implement it, I had a look on other generators, but I just don't understand what exactly is happening :C

src/server/terrain/simple_structures/SbbGen.zig
IntegratedQuantum: Or, you just let the addon creator give you a bounding box. It doesn't really need to be precise anyways.

Argmaster: I mean, you can calculate upper bound during loading (when resolving references to be precise) without making structure creators update height of their structures every time they modify one segment of the structure. Just pick tallest blueprint for each child block. Technically you could also just resolve the whole structure while placing it and discard it if it is too tall, but that is quite some waste.

IntegratedQuantum: But what if the addon creator wants to allow some leaves cut off by the cave ceiling, but leave enough room for the stem?

Argmaster: Alr, I guess they will have to do maths then

src/blueprint.zig
Argmaster: Umm I did this:

```zig
	fn _pasteInGeneration(self: Blueprint, pos: Vec3i, chunk: *ServerChunk, comptime mode: PasteMode) void {
		const chunkWidth: i32 = @intCast(chunk.super.width);

		const blueprintWidth: i32 = @intCast(self.blocks.width);
		const blueprintDepth: i32 = @intCast(self.blocks.depth);
		const blueprintHeight: i32 = @intCast(self.blocks.height);

		var blueprintX: u32 = @max(0, -pos[0]);
		const blueprintEndX = @min(blueprintWidth, chunkWidth - pos[0]);

		var chunkX = @max(0, pos[0]);
		const chunkEndX = @min(chunkWidth, blueprintWidth + pos[0]);

		while(chunkX < chunkEndX and blueprintX < blueprintEndX) : ({
			chunkX += chunk.super.pos.voxelSize;
			blueprintX += chunk.super.pos.voxelSize;
		}) {
			var blueprintY: u32 = @max(0, -pos[1]);
			const blueprintEndY = @min(blueprintDepth, chunkWidth - pos[1]);

			var chunkY = @max(0, pos[1]);
			const chunkEndY = @min(chunkWidth, blueprintDepth + pos[1]);

			while(chunkY < chunkEndY and blueprintY < blueprintEndY) : ({
				chunkY += chunk.super.pos.voxelSize;
				blueprintY += chunk.super.pos.voxelSize;
			}) {
				var blueprintZ: u32 = @max(0, -pos[2]);
				const blueprintEndZ = @min(blueprintHeight, chunkWidth - pos[2]);

				var chunkZ = @max(0, pos[2]);
				const chunkEndZ = @min(chunkWidth, blueprintHeight + pos[2]);

				while(chunkZ < chunkEndZ and blueprintZ < blueprintEndZ) : ({
					chunkZ += chunk.super.pos.voxelSize;
					blueprintZ += chunk.super.pos.voxelSize;
				}) {
					const block = self.blocks.get(blueprintX, blueprintY, blueprintZ);
					if(block.typ == voidType) continue;

					switch(mode) {
						.all => chunk.updateBlockInGeneration(chunkX, chunkY, chunkZ, block),
						.degradable => chunk.updateBlockIfDegradable(chunkX, chunkY, chunkZ, block),
					}
				}
			}
		}
	}
```

and it kind of works?

IntegratedQuantum: kind of?

IntegratedQuantum: Also why not just calculate the `chunkX` from the `blueprintX`, like you did before?

IntegratedQuantum: > Well, I didn't check every single edge case, but those I saw in the world seemed okay.

Ideally we'd create mipmaps with the same algorithm that's used to generate LODs from player terrain, but that's out of scope for this PR.

Argmaster: I fixed the air part, now they use void as they should

IntegratedQuantum: I don't want to optimize it, just simplifying it.
Having multiple loop variables with different limits is just hard to read in my opinion.
Please figure something out.

src/blueprint.zig
```diff
@@ -105,9 +105,56 @@ pub const Blueprint = struct {
...
+
+	pub const PasteMode = enum {all, degradable};
+
+	pub fn pasteInGeneration(self: Blueprint, pos: Vec3i, chunk: *ServerChunk, mode: PasteMode) void {
+		switch(mode) {
+			inline else => |comptimeMode| _pasteInGeneration(self, pos, chunk, comptimeMode),
+		}
+	}
+
+	fn _pasteInGeneration(self: Blueprint, pos: Vec3i, chunk: *ServerChunk, comptime mode: PasteMode) void {
```
Argmaster, line 117: Is that particularily more readable? I am not even sure if it works rn :C

src/blueprint.zig
```diff
@@ -115,45 +115,33 @@ pub const Blueprint = struct {
...
-				const chunkEndZ = @min(chunkWidth, blueprintHeight + pos[2]);
-
-				while(chunkZ < chunkEndZ and blueprintZ < blueprintEndZ) : ({
-					chunkZ += chunk.super.pos.voxelSize;
-					blueprintZ += chunk.super.pos.voxelSize;
-				}) {
+		const blueprintOffset = @max(@as(Vec3i, @splat(0)), -pos);
+		const chunkOffset = @max(@as(Vec3i, @splat(0)), pos);
+
+		const indexEndX: i32 = @min(@as(i32, @intCast(chunk.super.width)) - chunkOffset[0], @as(i32, @intCast(self.blocks.width)) - blueprintOffset[0]);
```
IntegratedQuantum, line 121: Something like this could work.
```suggestion
		const blueprintEndX: i32 = @min(chunk.super.width - pos[0], @as(i32, @intCast(self.blocks.width)));
```
Then you could iterate the blueprint coordinates directly, and the chunk-relative coordinates are just `blueprintX - pos[0]`

Argmaster: pos is chunk local relative so it can be negative, cant be used directly as index for blueprint. Changed to iterate blueprint coordinates.

src/blueprint.zig
```diff
@@ -105,9 +105,49 @@ pub const Blueprint = struct {
...
+		while(indexX < indexEndX) : (indexX += chunk.super.pos.voxelSize) {
+			var indexY: i32 = blueprintOffset[1];
+			while(indexY < indexEndY) : (indexY += chunk.super.pos.voxelSize) {
+				var indexZ: i32 = blueprintOffset[2];
+				while(indexZ < indexEndZ) : (indexZ += chunk.super.pos.voxelSize) {
+					const block = self.blocks.get(@intCast(indexX), @intCast(indexY), @intCast(indexZ));
+
+					if(block.typ == voidType) continue;
+
+					const chunkX = indexX + chunkOffset[0] - blueprintOffset[0];
```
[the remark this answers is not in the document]
Argmaster: No, because it is clamped.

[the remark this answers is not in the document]
IntegratedQuantum: https://www.wolframalpha.com/input?i=simplify+max%280%2C+x%29+-+max%280%2C+-x%29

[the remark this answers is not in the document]
Argmaster: I guess for me math is not mathing

src/blueprint.zig
```diff
@@ -105,9 +105,49 @@ pub const Blueprint = struct {
...
+		switch(mode) {
+			inline else => |comptimeMode| _pasteInGeneration(self, pos, chunk, comptimeMode),
+		}
+	}
+
+	fn _pasteInGeneration(self: Blueprint, pos: Vec3i, chunk: *ServerChunk, comptime mode: PasteMode) void {
+		const blueprintOffset = @max(@as(Vec3i, @splat(0)), -pos);
+		const chunkOffset = @max(@as(Vec3i, @splat(0)), pos);
+
+		const indexEndX: i32 = @min(@as(i32, @intCast(chunk.super.width)) - chunkOffset[0], @as(i32, @intCast(self.blocks.width)));
```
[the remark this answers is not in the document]
Argmaster: This cast is necessary, otherwise we get integer overflow if we get into negative values:

```
thread 11260 panic: integer overflow
C:\Users\argma\dev\Cubyz\feature-sbbgen\src\blueprint.zig:122:49: 0x7ff7290683eb in _pasteInGeneration__anon_167578 (Cubyzig.exe.obj)
  const indexEndZ: i32 = @min(chunk.super.width - chunkOffset[2], @as(i32, @intCast(self.blocks.height)));
                                                ^
C:\Users\argma\dev\Cubyz\feature-sbbgen\src\blueprint.zig:113:52: 0x7ff728ff0fd0 in pasteInGeneration (Cubyzig.exe.obj)
   inline else => |comptimeMode| _pasteInGeneration(self, pos, chunk, comptimeMode),
                                                   ^
C:\Users\argma\dev\Cubyz\feature-sbbgen\src\server\terrain\simple_structures\SbbGen.zig:56:37: 0x7ff728f747d7 in placeSbb (Cubyzig.exe.obj)
 rotated.blueprint.pasteInGeneration(pastePosition, chunk, self.placeMode);
                                    ^
C:\Users\argma\dev\Cubyz\feature-sbbgen\src\server\terrain\simple_structures\SbbGen.zig:43:10: 0x7ff728efbb53 in generate (Cubyzig.exe.obj)
 placeSbb(self, self.structureRef, Vec3i{x, y, z}, Neighbor.dirUp, chunk, seed);
         ^
```

Argmaster:
Applied suggestions to Y and Z, please review.

src/blueprint.zig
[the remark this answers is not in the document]
IntegratedQuantum: Oh right, well either way a cast is not necessary, `@as(i32, ...)` is enough.

src/blueprint.zig
```diff
@@ -105,9 +105,48 @@ pub const Blueprint = struct {
...
+	pub fn pasteInGeneration(self: Blueprint, pos: Vec3i, chunk: *ServerChunk, mode: PasteMode) void {
+		switch(mode) {
+			inline else => |comptimeMode| _pasteInGeneration(self, pos, chunk, comptimeMode),
+		}
+	}
+
+	fn _pasteInGeneration(self: Blueprint, pos: Vec3i, chunk: *ServerChunk, comptime mode: PasteMode) void {
+		const chunkOffset = @max(@as(Vec3i, @splat(0)), pos);
+
+		const indexEndX: i32 = @min(@as(i32, chunk.super.width) - chunkOffset[0], @as(i32, @intCast(self.blocks.width)));
```
IntegratedQuantum, line 120: I think it should be `pos` here, instead of `chunkOffset`, otherwise it will be wrong if the structure is bigger than the chunk.

IntegratedQuantum:
Also since we are getting close to the merge, please test one last time if it works and then remove the test structures.

src/blueprint.zig
Argmaster: Sure.

assets/cubyz/biomes/forest/base.zig.zon
```diff
@@ -18,7 +18,7 @@
 			.leaves = "cubyz:oak_leaves",
 			.log = "cubyz:oak_log",
 			.top = "cubyz:oak_top",
-			.chance = 0.4,
+			.chance = 0.1,
```
Argmaster, line 21:
```suggestion
			.chance = 0.4,
```

assets/cubyz/biomes/forest/base.zig.zon
```diff
@@ -30,7 +30,7 @@
 			.leaves = "cubyz:birch_leaves",
 			.log = "cubyz:birch_log",
 			.top = "cubyz:birch_top",
-			.chance = 0.2,
+			.chance = 0.1,
```
Argmaster, line 33:
```suggestion
			.chance = 0.2,
```

Argmaster:
Looks like it still works, test structure removed.

IntegratedQuantum (approved):
Let's go!

PR #1236, opened 2025-03-24
Add `/set` command

Argmaster:
This pull request adds `/set` command which fills current selection with pattern from #1237

This implementation does **not** yet support undo/redo from #1228 

Depends on: #1237
Related to: #1214


Argmaster:
Screenshot 2025-03-24 043937

src/server/command/worldedit/pattern.zig
Argmaster, line 1: No idea where this file should be placed, so its here for now, I am open for suggestions.

src/server/command/worldedit/pattern.zig
```diff
@@ -0,0 +1,60 @@
+const std = @import("std");
+
+const main = @import("root");
+const AliasTable = main.utils.AliasTable;
+const Block = main.blocks.Block;
+const ListUnmanaged = main.ListUnmanaged;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+blocks: AliasTable(Entry),
```
Argmaster, line 9: First I have implemented it as a struct within a file, but it was only thing here, so I figured it doesn't make sense to have a single struct within a file of the same name.

Argmaster:
Actually, I suspect that there will be a request to split this into two PRs, so I will do this in advance. 

Extracted #1237

src/server/command/worldedit/set.zig
```diff
@@ -0,0 +1,53 @@
...
+		return;
+	}
+	const pos1 = source.worldEditData.selectionPosition1 orelse {
+		return source.sendMessage("#ff0000Position 1 isn't set", .{});
+	};
+	const pos2 = source.worldEditData.selectionPosition2 orelse {
+		return source.sendMessage("#ff0000Position 2 isn't set", .{});
+	};
+
+	var seed: u64 = @as(u64, @intCast(pos1[0])) ^ (@as(u64, @intCast(pos1[1])) << 15) ^ (@as(u64, @intCast(pos1[2])) << 31);
```
IntegratedQuantum, line 25: This code breaks for negative coordinates.
Also you could just work with `main.seed`, to get a different starting seed every time.
```
const seed = &main.seed;
```
(also there is main.random.initSeed3D, in case you actually want to have the same seed for each position)

src/server/command/worldedit/set.zig
```diff
@@ -0,0 +1,53 @@
...
+	const pattern = Pattern.initFromString(main.globalAllocator, args) catch |err| {
```
IntegratedQuantum, line 26: Always use the stackAllocator for local allocations

IntegratedQuantum (changes requested):

src/server/command/worldedit/set.zig
Argmaster: I think it makes more sense to have it always random, despite my initial intentions. Then you can re-do the thing to get different pattern if you don't like it first time.

Argmaster: I think it will also be nice to add optional --seed/-s flag to specify seed manually and add print for current seed, but that should be separate PR.

IntegratedQuantum: I don't think that makes sense. If you want the exact same structure, then you can just copy paste it.
And for other purposes the seed is kind of useless anyways.

Argmaster: You may want to change size of the selection but keep the pattern. Although, since this is just noise there shouldn't be anything that special about that pattern. I guess we scrap that idea.

IntegratedQuantum (approved):

PR #1238, opened 2025-03-24
Add `/redo` command

Argmaster:
This pull request adds `/redo` command that restores changes undone with `/undo` command.

Extracted from: #1228 
Depends on: #1228


IntegratedQuantum (approved):
Nice and simple PR, nothing to complain, that's how I like them.

PR #1251, opened 2025-03-28
Directional drops

Argmaster:
This pull requests causes drops of blocks which don't convert to air to drop to the side they were mined from.

Resolves: #1058


Argmaster:
https://github.com/user-attachments/assets/01b88fa0-4753-442a-9383-c67deff0bab1

IntegratedQuantum:
There are several edge cases that aren't handled correctly (and these are non-trivial to handle either):

https://github.com/user-attachments/assets/b216e0ec-1b54-41cd-895e-e4dd7e18be56

https://github.com/user-attachments/assets/5afc5ee1-7de7-46b2-96ce-616bbcd038fb

In my opinion, the drop location and direction should be dictacted by the rotation mode. This might also allow us to implement the behavior that @ikabod-kee proposed in #1058 (which would be really nice in my opinion)

Argmaster:
For first one i agree it should not behave like this, but for second one I don't see anything overly bad, I mean, yeah, it maybe should just drop in place, but current behavior is not terrible. Could you be a bit more precise about what behavior you mean? The video or something else?

IntegratedQuantum:
> for second one I don't see anything overly bad

It's not overly bad. But it's incosistent behavior. It is a problem that should be fixed in my opinion.

> Could you be a bit more precise about what behavior you mean? The video or something else?

Yes the video, where the ore nicely pops out of the side of the block that was mined.

> As for using rotation mode to dictate drop direction, well, that's a lot of code and rotation mode is not only API that shows interest in controlling drops - so does EntityDataClass to eg. drop chest contents, so rotation mode should not have exclusive control over this.

Same when the player drops something, but these are different functions entirely, each of which would call `world.drop` in a completely different context.

> Maybe we could keep this as is as a dreaded workaround and revisit it when we have requirements from EntityDataClass ready?

I don't like the idea of adding temporary workarounds like this. They tend to stick around for a while since the source problem is no longer an issue. Furthermore you'll likely discover that revisiting this won't be needed for chests anyways.
I'd prefer if this was done right the first time, instead of adding a workaround that breaks in other cases.

Argmaster:
I guess ok, I was not supposed to approach new features, but I guess feature approached me.

As for making items drop do arc in air, well, this requires them to have enough Z+ and <away-from-block> velocity to overcome gravity, but without friction this will make them run away far from player. 

For making rotations handle drops, what exactly you think they should do? Because that stair case is basically about checking if spawn can happen in that neighbor block, not about moving spawn action itself into rotation code. Do we do both, or do go with some `canSpawnDropHere` or sth that will check if drop can be spawned in that particular location?

IntegratedQuantum:
> Because that stair case is basically about checking if spawn can happen in that neighbor block

Just to write it down: The item should drop on the side that the player mined. With this it's safe to assume that there is enough empty space there for the item to drop, and if there isn't, regular item unstucking (which is done during the physics update and currently a bit broken) should take care of it.

Argmaster:
https://github.com/user-attachments/assets/00cbefd8-7e74-41c6-9d42-93c8f9ed2a68

src/Inventory.zig
```diff
@@ -1620,17 +1622,56 @@ pub const Command = struct { // MARK: Command
...
+					main.server.world.?.drop(itemStack.clone(), dropPos, @floatCast(dropDir), dropVelocity);
+				}
+			} else {
+				const sign = std.math.sign(_dir_hint);
+				const factors = Vec3d{
+					@sin(@abs(std.math.atan(_dir_hint[0]))),
+					@sin(@abs(std.math.atan(_dir_hint[1]))),
+					@sin(@abs(std.math.atan(_dir_hint[2]))),
+				};
+				const edgeOffset = _dir_hint + factors*@as(Vec3d, @splat(std.math.sqrt(2) - 1));
```
IntegratedQuantum, line 1657: This math seems needlessly complex.
Couldn't you just also send a position hint/box (which would be useful for e.g. torches anyways) to avoid having to retrieve it here from just the direction?

IntegratedQuantum (changes requested):
Could find a way to simplify the block dropping code?
Does it really need two branches? Couldn't it be solved by giving extra parameters to the function?
I would pass it the direction, a random offset weight for the direction and a bounding box for the position. This should make the logic on the receiving end simpler and allow more freedom for other effects (I'm thinking about dropping torches here)

src/Inventory.zig
```diff
@@ -1554,6 +1555,7 @@ pub const Command = struct { // MARK: Command
 	const UpdateBlock = struct { // MARK: UpdateBlock
 		source: InventoryAndSlot,
 		pos: Vec3i,
+		dropDirection: Vec3d = Vec3d{0, 0, 1},
```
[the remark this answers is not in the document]
Argmaster: I mean, we can agree that this will be normalized on receiver side and make it u8, then you have like two decimal places of precision to work with by manipulating proportion between xyz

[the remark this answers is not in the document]
IntegratedQuantum: It might make sense to also pass a velocity in it, so it would not really make sense to always normalize it.

Either way that would be premature optimization, since handling normalized coordinates requires adding a lot more code and it's likely that this will not be called very often.
Switching to f32 is just "non-pessimization" to put it in Casey Muratori's terms. (Maybe I should add a section about this to the contributing guidelines?)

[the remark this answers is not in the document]
Argmaster: `main.server.world.?.drop` expect direction to be normalized, so that velocity thing we would have to revisit later.

Yeah, anything you find enforceable in reviews should be welcome in contribution guidelines.

src/Inventory.zig
```diff
@@ -1620,17 +1623,33 @@ pub const Command = struct { // MARK: Command
...
 					if(drop.chance == 1 or main.random.nextFloat(&main.seed) < drop.chance) {
-						blockDrop(self.pos, drop);
+						blockDrop(self.dropPos, self.dropDir, drop);
 					}
 				}
 			}
 		}
 
-		fn blockDrop(pos: Vec3i, drop: main.blocks.BlockDrop) void {
+		fn blockDrop(dropPosHint: Vec3d, dirHint: Vec3f, drop: main.blocks.BlockDrop) void {
```
IntegratedQuantum, line 1632: These are no longer just hints, they are the real thing.

src/Inventory.zig
```diff
@@ -1554,6 +1555,8 @@ pub const Command = struct { // MARK: Command
 	const UpdateBlock = struct { // MARK: UpdateBlock
 		source: InventoryAndSlot,
 		pos: Vec3i,
+		dropPos: Vec3d,
```
IntegratedQuantum, line 1558: In my opinion it would make more sense to use a relative position (relative to `pos`), after all that's what you got on the client side anyways.

src/rotation.zig
```diff
@@ -65,10 +66,27 @@ pub const RotationMode = struct { // MARK: RotationMode
 			const boxTMin = @reduce(.Max, @min(t1, t2));
 			const boxTMax = @reduce(.Min, @max(t1, t2));
 			if(boxTMin <= boxTMax and boxTMax > 0) {
+				var face = Neighbor.dirUp;
```
IntegratedQuantum, line 69: Should be initialized to undefined since you overwrite it in every reachable path.

src/renderer.zig
```diff
@@ -915,7 +917,13 @@ pub const MeshSelection = struct { // MARK: MeshSelection
...
 
 	fn updateBlockAndSendUpdate(source: main.items.Inventory, slot: u32, x: i32, y: i32, z: i32, oldBlock: blocks.Block, newBlock: blocks.Block) void {
-		main.items.Inventory.Sync.ClientSide.executeCommand(.{.updateBlock = .{.source = .{.inv = source, .slot = slot}, .pos = .{x, y, z}, .oldBlock = oldBlock, .newBlock = newBlock}});
+		const pos = Vec3i{x, y, z};
+		const dropDir = if(newBlock.typ == 0) Vec3f{0, 0, 1} else @as(Vec3f, @floatFromInt(selectionFace.relPos()));
+		const dropPosOffset = if(newBlock.typ == 0) Vec3d{0.5, 0.5, 0.0} else @as(Vec3d, @splat(0.5)) + dropDir*@as(Vec3d, @splat(0.5));
+		const dropPos = @as(Vec3d, @floatFromInt(pos)) + dropPosOffset;
+		main.items.Inventory.Sync.ClientSide.executeCommand(.{
+			.updateBlock = .{.source = .{.inv = source, .slot = slot}, .pos = pos, .dropDir = dropDir, .dropPos = dropPos, .oldBlock = oldBlock, .newBlock = newBlock},
+		});
```
IntegratedQuantum, line 926: This has still too some edge cases, torches don't drop correctly when there are multiple in one block.

Also I don't like how everything now drops exactly from the center. The best solution would be to give a bounding box to the server and let it choose randomly within it.
Furthermore I think we should make a new interface function for this (or rename the rayIntersection function and give it more return types), to avoid all the edge cases, and allow torches to specify their drop location more exact in the future.

IntegratedQuantum (changes requested):

src/renderer.zig
Argmaster: > Also I don't like how everything now drops exactly from the center. The best solution would be to give a bounding box to the server and let it choose randomly within it. 

You would have to elaborate, however generally speaking I don't think we have to send bbox to server, we can calculate viable offset and pos here. IMO when dropping from side it makes most sense to have it drop from more or less center of the block. As for things that don't have to drop from the side, I made it randomized again.

> Furthermore I think we should make a new interface function for this (or rename the rayIntersection function and give it more return types), to avoid all the edge cases, and allow torches to specify their drop location more exact in the future.

We can already get nice results by just using selectionMin and selectionMax since they are equivalent to bbox of torch that is being broken. See current behavior. I had to also invert face value for carpets, as they were for some reason preferring collision with face that had normal in same direction as breaking direction. I must admit I don't fully understand math of those intersection, although I tried understanding it C:

Argmaster: Ok, adjusted as requested.

src/Inventory.zig
```diff
@@ -1555,9 +1556,80 @@ pub const Command = struct { // MARK: Command
 	const UpdateBlock = struct { // MARK: UpdateBlock
 		source: InventoryAndSlot,
 		pos: Vec3i,
+		drop: BlockDrop,
 		oldBlock: Block,
 		newBlock: Block,
 
+		const BlockDrop = struct {
```
IntegratedQuantum, line 1563: I would prefer calling this `BlockDropLocation`

src/Inventory.zig
```diff
@@ -1555,9 +1556,80 @@ pub const Command = struct { // MARK: Command
...
+				for(_drop.items) |itemStack| {
+					main.server.world.?.drop(itemStack.clone(), self.insidePos(pos), self.dropDir(), self.dropVelocity());
+				}
+			}
+			fn dropOutside(self: BlockDrop, pos: Vec3i, _drop: main.blocks.BlockDrop) void {
+				for(_drop.items) |itemStack| {
+					main.server.world.?.drop(itemStack.clone(), self.outsidePos(pos), self.dropDir(), self.dropVelocity());
+				}
+			}
+			fn insidePos(self: BlockDrop, _pos: Vec3i) Vec3d {
```
IntegratedQuantum, line 1585: I still don't understand how you order things.
If you want it to read like a book, then shouldn't `insidePos` be after `dropInside`?

src/Inventory.zig
```diff
@@ -1555,9 +1556,80 @@ pub const Command = struct { // MARK: Command
...
+			}
+			fn randomOffset(self: BlockDrop) Vec3d {
+				const itemMargin = itemHitBoxMargin();
+				const max: Vec3d = @min(@as(Vec3d, @splat(1.0)) - itemMargin, @max(itemMargin, self.max - itemMargin));
+				const min: Vec3d = @min(max, @max(itemMargin, self.min + itemMargin));
+				const center: Vec3d = (max + min)/@as(Vec3d, @splat(2));
+				const width: Vec3d = (max - min)/@as(Vec3d, @splat(2));
+				const factors = main.random.nextDoubleVectorSigned(3, &main.seed)*@as(Vec3d, @splat(0.5));
+				const offset = center + width*factors;
+				std.log.debug("min {d:.2} max {d:.2} offset {d:.2}", .{min, max, offset});
```
IntegratedQuantum, line 1597: This seems like it would needlessly spam the log, I assume you only used this for debugging?

src/Inventory.zig
```diff
@@ -1555,9 +1556,80 @@ pub const Command = struct { // MARK: Command
...
+				return @floatFromInt(self.dir.relPos());
+			}
+			inline fn major(self: BlockDrop) Vec3d {
+				return @floatFromInt(@abs(self.dir.relPos()));
+			}
+			inline fn minor(self: BlockDrop) Vec3d {
+				return @floatFromInt(self.dir.orthogonalComponents());
+			}
+			fn itemHitBoxMargin() Vec3d {
+				return @as(Vec3d, @splat(main.itemdrop.ItemDropManager.radius*2));
```
IntegratedQuantum, line 1618:
```suggestion
				return @splat(main.itemdrop.ItemDropManager.radius*2);
```

src/Inventory.zig
```diff
@@ -1555,9 +1556,80 @@ pub const Command = struct { // MARK: Command
...
+			}
+			fn dropDir(self: BlockDrop) Vec3f {
+				return @floatCast(vec.normalize(Vec3d{
+					self.direction()[0] + main.random.nextDoubleSigned(&main.seed)*0.25,
+					self.direction()[1] + main.random.nextDoubleSigned(&main.seed)*0.25,
+					if(self.direction()[2] < 0) self.direction()[2] else self.direction()[2] + 2.0,
```
IntegratedQuantum, line 1624: Shouldn't there be some randomness in the velocity too?

src/Inventory.zig
```diff
@@ -1555,9 +1556,80 @@ pub const Command = struct { // MARK: Command
...
+				const min: Vec3d = @min(max, @max(itemMargin, self.min + itemMargin));
+				const center: Vec3d = (max + min)/@as(Vec3d, @splat(2));
+				const width: Vec3d = (max - min)/@as(Vec3d, @splat(2));
+				const factors = main.random.nextDoubleVectorSigned(3, &main.seed)*@as(Vec3d, @splat(0.5));
+				const offset = center + width*factors;
+				std.log.debug("min {d:.2} max {d:.2} offset {d:.2}", .{min, max, offset});
+				return offset;
+			}
+			fn directionOffset(self: BlockDrop) Vec3d {
+				const half = @as(Vec3d, @splat(0.5));
```
IntegratedQuantum, line 1601: You use this quite often, sometimes also as `/@splat(2)`, maybe just make a struct-level constant for this?

src/Inventory.zig
```diff
@@ -1555,9 +1556,80 @@ pub const Command = struct { // MARK: Command
...
+			}
+			fn dropDir(self: BlockDrop) Vec3f {
+				return @floatCast(vec.normalize(Vec3d{
+					self.direction()[0] + main.random.nextDoubleSigned(&main.seed)*0.25,
+					self.direction()[1] + main.random.nextDoubleSigned(&main.seed)*0.25,
+					if(self.direction()[2] < 0) self.direction()[2] else self.direction()[2] + 2.0,
+				}));
+			}
+			fn dropVelocity(self: BlockDrop) f32 {
+				if(self.direction()[2] < -0.5) return 0.0;
```
IntegratedQuantum, line 1628: Shouldn't this only zero out the z component? Otherwise the thing just drops straight down.

IntegratedQuantum (changes requested):

src/Inventory.zig
Argmaster:
```suggestion
```

Yeah, it can be removed.

src/Inventory.zig
Argmaster: yes

src/Inventory.zig
Argmaster: I don't think it will be that noticeable with random drop direction, while it will have potential to screw up how nice the arc of drop flight  looks due to changed proportion between gravity and directional velocity.

src/Inventory.zig
Argmaster: Well, I will need to check again what the effect is when it is not zeroed, but I definitely did that on purpose.

Argmaster:
Ok, I tweaked the velocity so that its not completely zeroed, but items falling downwards still can't compete with Flash

IntegratedQuantum (changes requested):

src/rotation/carpet.zig
```diff
@@ -158,6 +158,7 @@ fn closestRay(comptime typ: enum {bit, intersection}, block: Block, _: ?main.ite
 			if(RotationMode.DefaultFunctions.rayModelIntersection(modelIndex, relativePlayerPos, playerDir)) |intersection| {
 				if(result == null or result.?.distance > intersection.distance) {
 					result = intersection;
+					result.?.face = intersection.face.reverse();
```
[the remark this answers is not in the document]
Argmaster: Firstly, this was determined empirically, so is definitely necessary. Secondly, I have no idea, but it works.

src/rotation/stairs.zig
```diff
@@ -296,15 +296,12 @@ pub fn rayIntersection(block: Block, item: ?main.items.Item, relativePlayerPos:
...
 					if(intersectionPos(block, relativePlayerPos, playerDir)) |intersection| {
 						const offset: Vec3f = @floatFromInt(intersection.minPos);
 						const half: Vec3f = @splat(0.5);
-						return .{
-							.distance = intersection.minT,
-							.min = half*offset,
-							.max = half + half*offset,
-						};
+						const fullIntersection = RotationMode.DefaultFunctions.rayIntersection(block, item, relativePlayerPos, playerDir) orelse unreachable;
+						return .{.distance = intersection.minT, .min = half*offset, .max = half + half*offset, .face = fullIntersection.face};
```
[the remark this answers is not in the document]
Argmaster: To avoid implementing direction detection for eighths.

src/Inventory.zig
```diff
@@ -1555,9 +1556,84 @@ pub const Command = struct { // MARK: Command
...
 
+		const half = @as(Vec3d, @splat(0.5));
+
+		const BlockDropLocation = struct {
+			dir: Neighbor,
+			min: Vec3f,
+			max: Vec3f,
+
+			pub fn drop(self: BlockDropLocation, pos: Vec3i, newBlock: Block, _drop: main.blocks.BlockDrop) void {
+				if(newBlock.solid() and newBlock.collide()) {
```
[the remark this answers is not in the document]
Argmaster: Yeah, this should probably be or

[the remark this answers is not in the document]
Argmaster:
```suggestion
				if(newBlock.solid() or newBlock.collide()) {
```

src/rotation/carpet.zig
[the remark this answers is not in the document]
IntegratedQuantum: But this value isn't used anywhere from how I understand it, since the direction is only used when the new block has collision, which is not the case for carpets anyways.
Either way, maybe your code should account for the actual collision box, instead of blindly assuming that it's always a full cube.

src/Inventory.zig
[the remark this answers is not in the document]
IntegratedQuantum: Why solid anyways?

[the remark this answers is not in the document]
Argmaster: What solid means? My understanding was that you either can't see inside it or walk inside it, hence drops shouldn't be inside it.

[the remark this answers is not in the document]
IntegratedQuantum: That was the original intention, yes, but now collision is exclusively handled by the collide parameter.
Solid only means that you cannot replace the current block by placing another one in the same location, I'm still not sure if we should even allow that at all, other than for fluids.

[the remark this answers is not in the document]
IntegratedQuantum: #1358

src/rotation/stairs.zig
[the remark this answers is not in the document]
Argmaster: I need to check that.

src/rotation/carpet.zig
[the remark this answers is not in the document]
Argmaster: I need to check that.

src/Inventory.zig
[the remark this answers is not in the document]
Argmaster: Ok, then I guess we can remove it from that condition.

src/rotation/carpet.zig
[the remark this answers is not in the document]
Argmaster: Yes, this is no longer relevant as we have two different drop modes, one inside block and one to the side.

src/rotation/stairs.zig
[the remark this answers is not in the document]
IntegratedQuantum: It should probably just do it like the torch code instead.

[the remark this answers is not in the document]
Argmaster: Possibly, but I am pretty sure that is out of scope of this PR, so I would advocate for compromise with having full block intersection until this is changed by some brave warrior.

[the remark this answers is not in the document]
IntegratedQuantum: Well, I'm pretty sure a brave knight is going to need an issue for it.

src/Inventory.zig
```diff
@@ -1555,9 +1556,82 @@ pub const Command = struct { // MARK: Command
 	const UpdateBlock = struct { // MARK: UpdateBlock
 		source: InventoryAndSlot,
 		pos: Vec3i,
+		dropLocation: BlockDropLocation,
 		oldBlock: Block,
 		newBlock: Block,
 
+		const half = @as(Vec3f, @splat(0.5));
+		const itemHitBoxMargin: f32 = @floatCast(main.itemdrop.ItemDropManager.radius);
+		const itemHitBoxMarginVec = @as(Vec3f, @splat(itemHitBoxMargin));
```
IntegratedQuantum, line 1565:
```suggestion
		const itemHitBoxMarginVec: Vec3f = @splat(itemHitBoxMargin);
```

src/rotation/stairs.zig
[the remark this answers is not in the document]
Argmaster: #1360

src/Inventory.zig
```diff
@@ -1555,9 +1556,82 @@ pub const Command = struct { // MARK: Command
...
+				const pos: Vec3d = @floatFromInt(_pos);
+				return pos + self.randomOffset();
+			}
+			fn randomOffset(self: BlockDropLocation) Vec3f {
+				const max: Vec3f = @min(@as(Vec3f, @splat(1.0)) - itemHitBoxMarginVec, @max(itemHitBoxMarginVec, self.max - itemHitBoxMarginVec));
+				const min: Vec3f = @min(max, @max(itemHitBoxMarginVec, self.min + itemHitBoxMarginVec));
+				const center: Vec3f = (max + min)*half;
+				const width: Vec3f = (max - min)*half;
+				const factors: Vec3f = main.random.nextFloatVectorSigned(3, &main.seed)*half;
+				const offset: Vec3f = center + width*factors;
```
IntegratedQuantum, line 1594: I'm sure most of these don't need the explicit type declaration.

src/Inventory.zig
```diff
@@ -1555,9 +1556,82 @@ pub const Command = struct { // MARK: Command
...
+				return offset;
```
IntegratedQuantum, line 1595:
```suggestion
				return center + width*factors;
```

IntegratedQuantum (changes requested):

src/Inventory.zig
Argmaster: But is that harmful, that they have names?

Argmaster: Anyway, I have removed factors variable as it didn't provide any value, but I would prefer to keep the rest.

IntegratedQuantum: Well, but adding the types everywhere doesn't make it clearer to read, the function is short, the inputs and outputs are Vec3f, why write this everywhere?

IntegratedQuantum: > to ensure all the math is done with the precision I want

The return type should be enough for this promise. Zig will never narrow from f64 to f32.

Argmaster: Ok, so I stripped the type annotations, is it enough or do you insist on removing the variables?

IntegratedQuantum (approved):

Issue #1031, opened 2025-02-10
Climbing/Climbable Blocks


careeoki:
The player should be able to climb up certain blocks, like ropes, vines, ladders, etc.

PR #1266, opened 2025-04-02
Add climbing

ChibChi:
This pull request adds the `climbable: bool` block trait, block detection for climbable blocks and player input handeling.

Related to: https://github.com/PixelGuys/Cubyz/issues/1031


IntegratedQuantum (changes requested):
Instead of doing 4 more collision checks per frame, could you maybe reuse the code from block touch detection added in #1143?

ChibChi:
Climbing on blocks with collision works pretty good. Blocks without collision on the other hand... they start i climb when passing through. Should the player be able to walk through without obstruction?

IntegratedQuantum (changes requested):
> [Use intersections instead of collisions](https://github.com/PixelGuys/Cubyz/pull/1266/commits/a2d08dffaf5072e3aa0b8aca605083809b70e329)

I think you misunderstood me. I want you to reuse the intersection results from there, by e.g. adding a touch callback for climbable blocks or something like that.
What you have done here is better than before, but it's still one extra pass that iterates through all blocks in the player's hitbox.

Also sorry for the late review, I had to focus on the new network system.

ChibChi:
> I think you misunderstood me. I want you to reuse the intersection results from there, by e.g. adding a touch callback for climbable blocks or something like that. What you have done here is better than before, but it's still one extra pass that iterates through all blocks in the player's hitbox.

I didn't do this because I felt like it would be too restrictive. I'm using a smaller hitbox to prevent the player from skipping gaps.

IntegratedQuantum:
> I didn't do this because I felt like it would be too restrictive.

If it's too restrictive, then please improve the interface of touch events.

We can't keep adding a new collision test call for every single addition to the physics system. These are not cheap at all!

> I'm using a smaller hitbox to prevent the player from skipping gaps.

According to the code you are using a larger hitbox, also what do you mean by skipping gaps?

src/game.zig
```diff
@@ -836,36 +872,12 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
-
-				// Check blocks below to get over the edge
-				const downBlock = main.renderer.mesh_storage.getBlock(x, y, z - 1);
-				const diagonalDownBlock = main.renderer.mesh_storage.getBlock(x + blockOffsetX, y + blockOffsetY, z - 1);
-
-				break :isClimbing Player.climbing and (downBlock.?.climbable() or diagonalDownBlock.?.climbable());
-			}
-		} else break :isClimbing false;
-	};
+	const footHitbox: collision.Box = .{.min = -Player.outerBoundingBoxExtent, .max = Player.outerBoundingBoxExtent*Vec3d{1, 1, -0.1}};
```
ChibChi: > According to the code you are using a larger hitbox

I did use a smaller hitbox unless I'm missing something.

> what do you mean by skipping gaps?

In Minecraft it used to be possible to climb ladders where every other block is a ladder block.

src/game.zig
```diff
@@ -414,8 +414,9 @@ pub const collision = struct {
 					if(block == null)
 						continue;
 
-					// TODO Find a way to prevent the entity to skip gaps
-					if(block.?.climbable() and !entity.climbing) {
+					// Prevent the entity from skipping gaps
+					const isBelowTop = posZ < maxZ;
```
ChibChi, line 418: Now i kind of ignore the top blocks which feels cheaty

src/game.zig
IntegratedQuantum: Oh, I see, I was looking at the wrong thing.

> In Minecraft it used to be possible to climb ladders where every other block is a ladder block.

In my mind this should be possible, but slower. Mabye we could have a climbing speed parameter that gets averaged of the entire area of the player hitbox in the movement direction. Then you could also assign different climbing speeds to different climbable blocks. E.g. ivy could have a slower climbing speed.

Take a look at the ice code for an example of this averaging on a different axis.

src/game.zig
```diff
@@ -904,6 +1002,9 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 						movementSpeed = @max(movementSpeed, 5.5);
 						movementDir[2] += 5.5;
 					}
+				} else if(Player.super.climbing) {
+					movementSpeed = @max(movementSpeed, Player.currentClimbSpeed*0.7);
```
IntegratedQuantum, line 1006: This feels wrong. When jumping next to a climbable block you should be able to keep your jumping speed, slowly slowing down to the ladder speed due to gravity.

Furthermore using `@min` and `@max` in collision code is always bad, since it has the ability to cut your velocity instantly as soon as you touch a certain block.
It's better to use friction to slow it down slowly instead.

src/game.zig
```diff
@@ -904,6 +1002,9 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
+					movementDir += up*@as(Vec3d, @splat(walkingSpeed));
```
IntegratedQuantum, line 1007: `movementDir` is normalized.
This has the side effect off slowing you down even if you just walk alongside a climbable block.
Up direction should be handled separately here.

src/game.zig
```diff
@@ -1010,6 +1111,10 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 			move[i] = a/frictionCoefficient*deltaTime - c_1/frictionCoefficient*@exp(-frictionCoefficient*deltaTime) + c_1/frictionCoefficient;
 		}
 
+		if(Player.super.climbing and Player.crouching and move[2] < 0) {
+			move[2] = 0;
```
IntegratedQuantum, line 1115: This is bad since it allows you to negate any fall damage by jumping onto a ladder. There should be a velocity limit beyond which you cannot hold onto the climbable block.

src/game.zig
```diff
@@ -407,13 +487,24 @@ pub const collision = struct {
 				while(posZ <= maxZ) : (posZ += 1) {
 					const block: ?Block =
 						if(side == .client) main.renderer.mesh_storage.getBlock(posX, posY, posZ) else main.server.world.?.getBlock(posX, posY, posZ);
-					if(block == null or block.?.touchFunction() == null)
+					if(block == null)
 						continue;
+
+					// climbing
+					if(block.?.climbable() and !entity.climbing) {
+						const touchX: bool = isBlockIntersecting(block.?, posX, posY, posZ, center, extendClimableX);
```
IntegratedQuantum, line 495: Is there any reason why you cannot use the `touchX` defined below?

IntegratedQuantum (changes requested):
Also could you make ivy climbable as an example to make testing easier?

src/game.zig
```diff
@@ -1193,7 +1298,13 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 		} else if(Player.eyeCoyote > 0) {
 			Player.eyePos[2] -= move[2];
 		}
-		collision.touchBlocks(Player.super, hitBox, .client);
+
+		var climbAmount: f32 = 0.01;
+		if(!Player.onGround or KeyBoard.key("jump").pressed) {
+			climbAmount = 0.4;
```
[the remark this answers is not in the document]
ChibChi: I made it this big as a QOL feature. With an amount of `0.01` its incredibly difficult to grab onto blocks. It could be smaller of course.

src/game.zig
ChibChi: I'm using a different extend (`climbAmount`) than the original touch extend of `0.01`. I could reuse this when I use the same extend. But this makes climbing down really hard.

src/game.zig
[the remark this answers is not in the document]
IntegratedQuantum: Well, you just have to walk right into it until you hit the hitbox, and then you can just climb.
I don't see how that's difficult.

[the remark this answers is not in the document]
ChibChi: Climbing up isn't as much of a problem, but trying to get down from higher elevations.

https://github.com/user-attachments/assets/ee80e71c-5b2b-4e62-8de6-e7765e1155aa

ChibChi:
> I was skeptical about this PR at first, but it looks to be better than Minecraft's janky climbing (plus, you added slip velocity! No ladder clutches XD)

Being by first PR I also expected this to go differently (my math is rusty)

ChibChi:
> Found a bug!

Fixed

ChibChi:
It is now possible to crouch off of blocks onto climbable blocks. Took me longer than I want to admit. You need to crouch backwards towards the edge and let go of crouch to go down. It's still a little death trap.

careeoki:
Climbing gets really slow when you reach the top of the block?
Also you're not able to crouch forwards against a ledge with ivy on it (it sends you plummeting to your doom)

ChibChi:
> Climbing gets really slow when you reach the top of the block? 

It gets slower because it is based on the area of the collision, so you get slowed down with air gaps. I could increase the default climb speed to make it a little faster. How do i differentiate between gaps and platforms without blasting collision checks

careeoki:
> It gets slower because it is based on the area of the collision, so you get slowed down with air gaps.

You also get slowed down if you're not centered on the climbable block, which logically makes sense but feels annoying while playing.

Also, it's weird not being able to move backwards while climbing (i.e. to get off the block.) 

Oh, and a probably unintended mechanic: you can climb on ceilings with ivy on them! This is fun, I hope you're able to keep it.
(not being able to move backwards is especially weird here)
image
However, you can also climb on the sides of blocks if they have ivy on the bottom.
image
and if you're crouching and climbing upwards, then start climbing on a ceiling, you get stuck crouching and can't fall down.

IntegratedQuantum:
> How do i differentiate between gaps and platforms without blasting collision checks

You could restrict the climbing collision box to the legs, however that would break the climbing on the ceiling thing.

The alternative would be to detect if there is a ledge, but that is indeed rather complicated (though a nice byproduct of this would be that you would be able to make your own ladders by stacking slabs or fences)

IntegratedQuantum:
> What if crouching on a climbable made you go down, and not having any inputs kept you still?

Yes please, that would be great. This has always bothered me in minecraft.
In my opinion not pressing any keys should keep you still, like what if you need to get away from your computer while climbing a big ladder? Or what if you just want to open chat or the inventory?
Now all your progress is gone because it sent you back to the bottom.

IntegratedQuantum:
> I could increase the default climb speed to make it a little faster.

Actually instead of increasing the climb speed, you could change the interpolation. e.g. a weighted pythagorean sum is always closer to the largest element than just a weighted sum.

IntegratedQuantum:
The area detection doesn't seem to work correctly. I can climb on the side of an ivy covered pillar:

https://github.com/user-attachments/assets/74f0b006-f703-4cc1-8f9e-6243ee0a0c1f

src/server/Entity.zig
```diff
@@ -15,6 +15,8 @@ health: f32 = 8,
 maxHealth: f32 = 8,
 energy: f32 = 8,
 maxEnergy: f32 = 8,
+climbing: bool = false,
+touchingClimbable: bool = false,
```
IntegratedQuantum, line 19: This Data should only be stored in the client-side player struct for now.

src/game.zig
```diff
@@ -509,6 +509,12 @@ pub const collision = struct {
 
 						entity.touchingClimbable = entity.touchingClimbable or isSideBlock or isBottomBlock;
 						entity.climbing = isSideBlock and (!isBottomBlock or isFullBlock);
+
+						const normal = if (model.internalQuads.len > 0) model.internalQuads[0].quadInfo().normal else break :missingNormal;
```
IntegratedQuantum: Wait, so you cannot make full blocks climbable? That doesn't make sense.
Also does this work on other blocks, like e.g. if you were to make a climbable pole?

IntegratedQuantum (changes requested):

codemob-dev:
This might have to be partially rewritten once we have #1726 and #1728

codemob-dev:
although #1728 is a long ways away

IntegratedQuantum:
This PR seems stale anyways, please create a new PR if you intend to continue working on this.
And if you do, it would be best to start directly with #1728 before continuing work on climbing.

PR #1268, opened 2025-04-02
Add copper ore and make iron generate deeper

careeoki:
related to #1115 

Copper has about half the strength of Iron, though it's lighter and has higher durability.
image
image

**New distribution of ores**:
z >20: Copper starts
z >-50: Iron starts

**Iron is not longer found on the surface**: Now you actually have to enter a cave to get iron tools. And to explore caves, you need torches, so it takes some preparation to get to iron.
Copper now appears on the surface where iron used to.

**You can still mine iron ore with a stone pickaxe**: My thought is that the player would smelt 1 copper ingot, make a copper-tipped pickaxe, then just skip right to iron tools. To me, that doesn't give progression any more depth, it just makes it longer.
Instead, copper could be used alongside iron to make tools more durable or cheaper, and it would have other niche uses in crafting, i.e. bellows.
Also, a copper pickaxe is way better than a stone pickaxe, so it still might be worth it to make one even if it's not mandatory.
Of course, let me know what you think.

**Copper stops generating at -100**: This way, if the player wants copper, they have to seek it out near the surface. And, when in deeper caves, they won't be pestered by copper ore they don't want.


Argmaster:
Let me start with complimenting your textures - they are awesome, we are so lucy to have you!
And now I am going to be annoying and complain about balance:
1. Copper is softer than iron (on Mohs scale copper is 3, iron is 4), it should not be more durable. Moreover, i think it should have a modifier that causes it to deal exponentially less damage the less durability it has, simulating the tool getting blunt. 
2. I can easily find a cave that spans from Z=0 to Z=-150 from the surface, it took me ~14 minutes from the place I spawned. IMO diamond should be deeper, while other minerals (which we don't yet have) should spawn closer to surface.

du82:
Would be cool to see the other phases of copper too, just like Minecraft

Argmaster:
Imo meh, that's bloat

du82:
> Imo meh, that's bloat, do we add rust too? Do we add every decay that can happen or is copper special?

As a creative mode builder, I prefer a wide variety of textures and designs that suit many use cases, rather than a small palette.

careeoki:
I want atleast a fully oxidized copper, because it's a unique colour seen in a lot of real life architecture.
Not right now though, that's something I would add in the future when there's a way to actually age the copper.

Argmaster:
> As a creative mode builder, I prefer a wide variety of textures and designs that suit many use cases, rather than a small palette.

IMO this is material for addons, especially since they can be per-world.

PR #1283, opened 2025-04-03
Meteorite assets

careeoki:
You can find meteorites scattered throughout the world. It has a unique ore, Meteorite. To mine it, you need a pick that deals at least 15 damage, so you will have to do some caving before being able to obtain it. 
If your pick isn't quite good enough, you can at least grab some of the iron ore on the outside and come back later.

Compared to iron, meteorite trades some damage and swing speed for significantly higher durability. This is ideal for tools you bring along on surface adventures, since your tool will last longer, and you don't really need high damage while exploring.

image
image

- [ ] meteorite bars

<details><summary>Old description</summary>
part of #1132 

I was reading about meteorites and got carried away (I actually did this before copper.) I'm leaving it as a draft until we can generate meteorites in case things change or break.
...So for now I'll demonstrate with my awesome build:
image
Meteorites are made of kamacite, which needs a diamond pick to be broken. If you don't have a diamond pick yet, you can still gather the iron ore on the surface of the meteorite. After #1268, this is the highest point iron ore can appear at.

The other ore is Meteorite, also requiring a diamond pick. The inside of the meteorite is densely filled with meteorite ore.
While not dealing too much more damage than iron, it has significantly higher durability. My idea is that, while meteorite is not very powerful, it would be good for exploring the surface since you wouldn't have to worry about tools breaking as often.
It could also have a unique modifier, maybe.
image
image
</details>


codemob-dev:
Is there a reason this is still a draft, or can it be added now?

IntegratedQuantum:
Well, it needs some system for placing biome-independent structures, like #1805

careeoki:
Biome independent structures would just use SBB, right? So I could build some meteorites now in preparation?

IntegratedQuantum:
I think here a combination of #1793 and #2002 might do better than pure SBBs.

BoySanic:
Depends on #1805, which would be fixed by #2131 and #2129

IntegratedQuantum:
this is now possible

careeoki:
> this is now possible

Except I can't build meteorites since ores are broken in structures

BoySanic:
> > this is now possible
> 
> Except I can't build meteorites since ores are broken in structures


Are they?
I made a structure with ores recently and it seemed to work.

Unless you mean #2477

IntegratedQuantum:
> I made a structure with ores recently and it seemed to work.

It will work locally, but as soon as the folder structure changes or if you try opening a world where the ids are mapped differently, it will fail.

careeoki:
I now have a basic meteorite structure. The crater isn't as large or as good as what we could do with some kind of heightmap manipulation, but I think it looks good enough for now. When the time comes, we can just remove the crater, since the meteorites themselves are separate sbbs.

image

Crepestrom:
you should still be able to cut away into the terrain with sbbs

careeoki:
> you should still be able to cut away into the terrain with sbbs

But then that causes other problems: floating plants, trees cut off halfway, weird terrain with bites taken out of it

assets/cubyz/blocks/meteorite_ore.zig.zon
```diff
@@ -0,0 +1,14 @@
+.{
+	.tags = .{.stone, .mineable},
+	.blockHealth = 30,
+	.blockResistance = 15,
```
IntegratedQuantum: I think we should lower this, in my opinion this should be the same resistance as iron.

IntegratedQuantum (changes requested):

assets/cubyz/blocks/meteorite_ore.zig.zon
careeoki: Done

IntegratedQuantum (approved):

PR #1284, opened 2025-04-04
Edit masks

Argmaster:
This pull request adds `Mask` struct which can parse and match mask expression against blocks. General idea is based on [WorldEdit masks](https://worldedit.enginehub.org/en/latest/usage/general/masks/).

Masks can consist of block IDs, block tags (prefixed with `$`) and property checks (prefixed with `@`) separated by `|` (logical or) or `&` (logical and). And operator has tighter binding.
Each of those expressions can be prefixed with `!` to invert it.

Examples:

- `$stone` - masks blocks that have `stone` tag
- `@solid` - masks all blocks that have `solid` property set to `true`
- `cubyz:stone|cubyz:grass` - masks stone or grass blocks.
- `@solid&@degradable` - masks blocks that are marked as solid **and** degradable at once.

Related to: #1214 
Related to: #1337


Argmaster:
Ready for review.

Argmaster:
I will convert it back to draft because implementation may change depending on #1352

Argmaster:
Updated `/set` command to accept mask as announced in #1352

src/blocks.zig
```diff
@@ -229,9 +229,25 @@ pub fn getTypeById(id: []const u8) u16 {
...
+			block.data = block.mode().naturalStandard;
+		}
+		return block;
+	} else {
+		std.log.err("Couldn't find block {s}. Replacing it with air...", .{id});
+		return .{.typ = 0, .data = 0};
+	}
+}
+
+pub fn parseBlock2(data: []const u8) ?struct {typ: u16, data: ?u16} {
```
Argmaster, line 247: I need more information about block expression I have parsed, since block with no data specifier and block with data specifier have different valid meanings - block with no data specifier means match that block with any data, while block with data specifier means match that block with exact data match. I decided that since I don't know what is the preferred way to implement this I will just go with simple-stupid way and see what comes up in review.

IntegratedQuantum:
I don't expect you to do it in this PR, but to be logically complete, it should also allow &-ing conditions.

IntegratedQuantum:
> @solid

I would propose `.solid` to match the syntax of the zon.

src/blocks.zig
IntegratedQuantum: I would prefer if there is just one global function to parse a block. 
So maybe for now just check if it contains two colons at the call site, and maybe in the future (since this change would be too big) we can maybe return an extra enum parameter that specifies whether it contained the data or not.

Argmaster:
`.` can be easily confused with `,` which is not desireable. I intentionally omitted anding, because most of the time this is too restrictive, but If you give me good use case example that can not be achieved in different way, sure. If we want both operators, then maybe `|` instead of `,` would be a better choice.

src/blocks.zig
Argmaster: `parseBlock` is used in ~20 places and current contract is not acceptable for many reasons. It completely supresses any parsing errors, but adding same error handling to every of those ~20 places I waste of code and time IMO, hence it was made a adapter function. This way parsing is done indeed with one function, but it is `parseBlock2`, while `parseBlock` is just an adapter with error handling. Maybe we should change name of `parseBlock2` to `tryParseBlock`? We discussed something similar before, if I recall correctly in network rewrite for `receive`.

src/blocks.zig
```diff
@@ -256,6 +271,9 @@ pub fn hasRegistered(id: []const u8) bool {
 pub const Block = packed struct { // MARK: Block
 	typ: u16,
 	data: u16,
+
+	pub const Air = Block{.typ = 0, .data = 0};
```
IntegratedQuantum, line 275: This is not a `type`, so it should be lowercase.
```suggestion
	pub const air = Block{.typ = 0, .data = 0};
```

src/blocks.zig
IntegratedQuantum: Alright, how about this:
1. `parseBlock`, which behaves like it does now.
2. `getBlockById` which returns only a `?u16` for the block type, and ignores the data part
3. `getBlockData` which returns only a `?u16` for the data, and ignores the block id

src/blueprint.zig
```diff
@@ -335,6 +337,116 @@ pub const Pattern = struct {
...
+	const Entry = struct {
+		inner: Inner,
+		isInverse: bool,
+
+		const Inner = union(enum) {
+			block: struct {typ: u16, data: ?u16},
+			blockTag: Tag,
+			property: Property,
+
+			const Property = enum {transparent, collide, solid, selectable, degradable, viewThrough, allowOres, isEntity};
```
IntegratedQuantum, line 357: I would suggest to extract this generically, otherwise we will need to maintain and update this list.
You can use `std.meta.declarations` to get a list of all the function, and then you just need to check if the result is a bool.

src/blueprint.zig
```diff
@@ -335,6 +337,116 @@ pub const Pattern = struct {
 	}
 };
 
+pub const Mask = struct {
+	entries: ListUnmanaged(Entry),
+
+	pub const separator = ',';
+	pub const inverse = '!';
+	pub const tag = '$';
+	pub const property = '@';
```
IntegratedQuantum, line 346: Should these be named? They are only used once, and it's clear from context what they are for.
(If you think they should be named, then they should definitely be private)

src/blueprint.zig
```diff
@@ -335,6 +337,116 @@ pub const Pattern = struct {
...
+
+	const Entry = struct {
+		inner: Inner,
+		isInverse: bool,
+
+		const Inner = union(enum) {
+			block: struct {typ: u16, data: ?u16},
```
IntegratedQuantum, line 353: I would suggest to split this into two type `blockType` and `block`, then we don't need the optional.

src/blueprint.zig
```diff
@@ -335,6 +337,116 @@ pub const Pattern = struct {
...
+			}
+
+			fn match(self: Inner, block: Block) bool {
+				return switch(self) {
+					.block => block.typ == self.block.typ and (self.block.data == null or block.data == self.block.data),
+					.blockTag => |desired| {
+						for(block.blockTags()) |current| {
+							if(desired == current) return true;
+						}
+						return false;
```
IntegratedQuantum, line 385: Should we add a `block.hasTag` function, like for items?

src/blueprint.zig
```diff
@@ -335,6 +337,116 @@ pub const Pattern = struct {
...
+					},
+					.property => |prop| return switch(prop) {
+						.transparent => block.transparent(),
+						.collide => block.collide(),
+						.solid => block.solid(),
+						.selectable => block.selectable(),
+						.degradable => block.degradable(),
+						.viewThrough => block.viewThrough(),
+						.allowOres => block.allowOres(),
+						.isEntity => block.entityDataClass() != null,
```
IntegratedQuantum, line 395: This is problematic, now we have a parameter name, that's different to the block parameter name, and can only be used for masks. If someone asks what parameters you can check in masks, do you want to tell them
> "all block parameters hat are bools"

or do you want to tell them
> "all block parameters that are bools, except maybe we forgot to add some, oh and also there is this is one special case `isEntity` which is not a block property, instead it checks if the entityDataClass exists"

If we need any non-bool parameters (and seriously, why would anyone want to replace all blocks with block entities?), then we should probably implement a more generic comparing mechanism.

src/blueprint.zig
```diff
@@ -335,6 +337,116 @@ pub const Pattern = struct {
...
+			switch(specifier[0]) {
+				inverse => {
+					const entry = try Inner.initFromString(specifier[1..]);
+					return .{.inner = entry, .isInverse = true};
+				},
+				else => {
+					const entry = try Inner.initFromString(specifier);
+					return .{.inner = entry, .isInverse = false};
+				},
+			}
```
IntegratedQuantum, line 411:
```suggestion
			const isInverse = specifier[0] == '!';
			const entry = try Inner.initFromString(specifier[if(isInverse) 1 else 0..]);
			return .{.inner = entry, .isInverse = isInverse};
```

IntegratedQuantum (changes requested):

src/blueprint.zig
Argmaster: Well, I think it is beneficial to have centralized place for declarations of reserved characters alongside their meaning encoded in variable names instead of comments.
I'll make them private.

src/blueprint.zig
Argmaster: Probably, especially since #1398 is going to check some tags too, right?

src/blueprint.zig
Argmaster: That optional implies whether the data match is mandatory or not how would I determine in suggested layout?

src/blocks.zig
Argmaster: Do I use `getBlockById` and `getBlockData` to implement `parseBlock` or do I leave it with current implementation?

Argmaster:
@IntegratedQuantum I didn't implement all of the suggestions yet, but I am really curious what do you thing about the way I enabled testing with basically mocking with use of generic types, if you could quickly scroll through the code and see `GenericMask`, `RuntimeContext` and `Test` struct. I basically replaced the part that relied on global assets being initialized with `Context` type provided as template parameter and then I created `pub const Mask = GenericMask(RuntimeContext);` to be used outside testing context. This way I can test at least some subset of Mask parsing and matching features without initializing all the assets. For now I did proof of concept with  single ID and two IDs ORed. With this approach its rather impossible to test anything except block ID matching without large amount of mocking, but it checks quite a lot of logic so maybe its worth?

src/blueprint.zig
Argmaster: I changed the implementation in away that outdates this suggestion, hence I will reject it for now.

src/blueprint.zig
IntegratedQuantum: yes

IntegratedQuantum:
> I am really curious what do you thing about the way I enabled testing with basically mocking with use of generic types

I don't like that it becomes generic, adding an extra layer of indentation, just for testing.

I'd prefer something like
```
fn parseBlockLike(..) {
   if(in test) return parseBlockLikeTest(...);
   ...
}

...
fn parseBlockLikeTest(...) {
    if(eql(..."test:1_1")) return ...
    if(eql(..."test:1_null")) return ...
    ...
}
```

Argmaster:
Almost done, except do you know about a way to get name of currently ran test? I don't seem to find any.

Argmaster:
Alr, I don't know how to get currently executed test name, maybe we could craft something with out custom test runner but atm `@src` is not doing the trick and I can't find any other way.

Argmaster:
@IntegratedQuantum Ready 4 review

src/blueprint.zig
IntegratedQuantum: This suggestion still applies.

> That optional implies whether the data match is mandatory or not, how would I determine that in suggested layout?

Well you just have one separate tag for each case, instead of using an optional data field:
```zig
block: Block, // Matches type + data
blockType: u16, // Matches only the type
```

src/blueprint.zig
```diff
@@ -335,6 +337,231 @@ pub const Pattern = struct {
...
+fn parseBlockLike(block: []const u8) error{DataParsingFailed, IdParsingFailed}!BlockLike {
+	if(@import("builtin").is_test) return try Test.parseBlockLikeTest(block);
+	const typ = main.blocks.getBlockById(block) catch return error.IdParsingFailed;
+	const data = try main.blocks.getBlockData(block) catch return error.DataParsingFailed;
+	return .{.typ = typ, .data = data};
+}
+
+const Test = struct {
+	var testingAllocator = main.heap.ErrorHandlingAllocator.init(std.testing.allocator);
+	var allocator = testingAllocator.allocator();
```
IntegratedQuantum, line 502: Should use `std.heap.testingAllocator` if #1414 gets merged first.

IntegratedQuantum (changes requested):

src/blueprint.zig
```diff
@@ -277,10 +278,11 @@ pub const Blueprint = struct {
 			},
 		}
 	}
-	pub fn set(self: *Blueprint, pattern: Pattern) void {
+	pub fn set(self: *Blueprint, pattern: Pattern, mask: ?Mask) void {
```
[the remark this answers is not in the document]
Argmaster: This is let's say, a destructive mask, i.e. set can modify blocks that don't match it. It is also an implementation detail, user facing /set command doesn't have mask parameter, instead one can set global mask that automatically limits all the commands. Finally, /replace is already planned with a constructive mask, i.e. modify all the blocks that match the mask. This is for convenience, so players don't have to explicitly invert every part of a mask. Also /replace mask is local, per command.

src/blueprint.zig
```diff
@@ -335,6 +337,231 @@ pub const Pattern = struct {
...
+			blockProperty: Property,
+
+			const Property = generatePropertyEnum();
+
+			fn initFromString(specifier: []const u8) !Inner {
+				switch(specifier[0]) {
+					tag => {
+						const blockTag = specifier[1..];
+						if(blockTag.len == 0) return error.MaskSyntaxError;
+						return .{.blockTag = Tag.find(blockTag)};
```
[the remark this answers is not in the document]
Argmaster: Yes, that is very unfortunate.

src/blueprint.zig
```diff
@@ -335,6 +337,231 @@ pub const Pattern = struct {
...
+	const inverse = '!';
+	const tag = '$';
+	const property = '@';
+
+	const Entry = struct {
+		inner: Inner,
+		isInverse: bool,
+
+		const Inner = union(enum) {
+			blockLike: BlockLike,
```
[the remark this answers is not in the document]
Argmaster: Yes, I will look into it, now I understand what you mean

src/blueprint.zig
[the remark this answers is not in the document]
Argmaster: Ok, so `replace` with two nullable parameters: old and mask. A lot more ifs but the name is better?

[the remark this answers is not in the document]
Argmaster:
```suggestion
	// Replace all blocks matching old (if present) unless matched by mask (if present) with blocks from new. 
	pub fn replace(self: *Blueprint, new: Pattern, old: ?Mask, mask: ?Mask) void {
```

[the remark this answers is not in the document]
Argmaster: Can I leave it as-is and we will revisit this as part of `/replace` command implementation? Just after `/mask` command.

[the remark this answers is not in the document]
IntegratedQuantum: Yeah, you can leave it for now

Argmaster:
Updated to latest master, all change requests applied, R4R

src/tag.zig
```diff
@@ -34,6 +34,10 @@ pub const Tag = enum(u32) {
 		loadDefaults();
 	}
 
+	pub fn get(tag: []const u8) ?Tag {
```
IntegratedQuantum, line 37: Maybe we should name this one `find`, and the other one `findOrInsert`, but I'll leave that for a future PR.

src/blueprint.zig
```diff
@@ -372,6 +374,246 @@ pub const Pattern = struct {
...
+					property => {
+						const propertyName = specifier[1..];
+						const propertyValue = std.meta.stringToEnum(Property, propertyName) orelse return error.PropertyNotFound;
+						return .{.blockProperty = propertyValue};
+					},
+					else => {
+						switch(try parseBlockLike(specifier)) {
+							.block => |block| return .{.block = block},
+							.blockType => |blockType| return .{.blockType = blockType},
+						}
```
IntegratedQuantum, line 446: Why not return `Inner` directly from `parseBlockLike`? That would some of the boilerplate here
```suggestion
						return try parseBlockLike(specifier));
```

src/blueprint.zig
```diff
@@ -372,6 +374,246 @@ pub const Pattern = struct {
...
+
+fn generatePropertyEnum() type {
+	var tempFields: [@typeInfo(Block).@"struct".decls.len]std.builtin.Type.EnumField = undefined;
+	var count = 0;
+
+	for(std.meta.declarations(Block)) |decl| {
+		const declInfo = @typeInfo(@TypeOf(@field(Block, decl.name)));
+		if(declInfo != .@"fn") continue;
+		if(declInfo.@"fn".return_type != bool) continue;
+		if(declInfo.@"fn".params.len > 1) continue;
```
IntegratedQuantum, line 390: More correct would be `!= 1` since you probably don't want to catch 0-argument functions either.

src/blocks.zig
```diff
@@ -248,13 +248,31 @@ pub fn parseBlock(data: []const u8) Block {
...
+	const blockIdEndIndex = std.mem.indexOfScalarPos(u8, idAndData, 1 + addonNameSeparatorIndex, ':') orelse idAndData.len;
+	const id = idAndData[0..blockIdEndIndex];
+	return reverseIndices.get(id) orelse return error.NotFound;
+}
+
+pub fn getBlockData(idLikeString: []const u8) !?u16 {
+	const addonNameSeparatorIndex = std.mem.indexOfScalar(u8, idLikeString, ':') orelse return error.MissingAddonNameSeparator;
+	const blockIdEndIndex = std.mem.indexOfScalarPos(u8, idLikeString, 1 + addonNameSeparatorIndex, ':') orelse return null;
+	const dataString = idLikeString[blockIdEndIndex + 1 ..];
+	if(dataString.len == 0) return null;
```
IntegratedQuantum, line 262: Wouldn't that also be invalid? Having a colon at the end with no data afterwards?

src/blueprint.zig
```diff
@@ -372,6 +374,246 @@ pub const Pattern = struct {
...
+			blockTag: Tag,
+			blockProperty: Property,
+
+			const Property = generatePropertyEnum();
+
+			fn initFromString(specifier: []const u8) !Inner {
+				switch(specifier[0]) {
+					tag => {
+						const blockTag = specifier[1..];
+						if(blockTag.len == 0) return error.TagNotFound;
```
IntegratedQuantum, line 434: Already covered by the case below,
also technically it is a valid tag, so we should probably allow it if it exists.

src/blueprint.zig
```diff
@@ -372,6 +374,246 @@ pub const Pattern = struct {
...
+		}
+	};
+
+	pub fn initFromString(allocator: NeverFailingAllocator, source: []const u8) !@This() {
+		var orStorage: OrList = .{};
+		errdefer orStorage.deinit(allocator);
+
+		var oredExpressions = std.mem.splitScalar(u8, source, or_);
+		while(oredExpressions.next()) |subExpression| {
+			if(subExpression.len == 0) continue;
```
IntegratedQuantum, line 484: Should this be an error? Otherwise `cubyz:air||cubyz:void` is a valid mask. Or is this intentional to allow `||` as used in many programming languages?

src/blueprint.zig
```diff
@@ -372,6 +374,246 @@ pub const Pattern = struct {
...
+
+			var andStorage: AndList = .{};
+			errdefer andStorage.deinit(allocator);
+
+			var andedExpressions = std.mem.splitScalar(u8, subExpression, and_);
+			while(andedExpressions.next()) |specifier| {
+				if(specifier.len == 0) continue;
```
IntegratedQuantum, line 491: same here

src/blueprint.zig
```diff
@@ -372,6 +374,246 @@ pub const Pattern = struct {
...
+
+	pub fn deinit(self: @This(), allocator: NeverFailingAllocator) void {
+		for(self.entries.items) |andStorage| {
+			andStorage.deinit(allocator);
+		}
+		self.entries.deinit(allocator);
+	}
+
+	pub fn match(self: @This(), block: Block) bool {
+		for(self.entries.items) |andEdExpressions| {
```
IntegratedQuantum, line 518:
```suggestion
		for(self.entries.items) |andedExpressions| {
```

IntegratedQuantum (changes requested):

src/blocks.zig
Argmaster: That's your choice, there is no formal syntax declaration for ID, so I was guessing and figured such case doesn't hurt anyone, we can just ignore it.

src/blueprint.zig
Argmaster: This is not intentional, but such deduplication of operators should not have any performance or syntactic implications.

src/blocks.zig
IntegratedQuantum: ok, then please make this return an error.

src/tag.zig
Argmaster: There is already similar construct in HashMap - `get` and `getOrPut`

src/blueprint.zig
Argmaster: Changed to error

Argmaster:
Applied change requests, R4R

src/blueprint.zig
```diff
@@ -372,6 +374,228 @@ pub const Pattern = struct {
 	}
 };
 
+fn generatePropertyEnum() type {
```
IntegratedQuantum, line 377: Functions returning a type should start with a capital letter.
Also why even make this a function when you alias the result to `Property` anyways?
```zig
const Property = blk: {
	...
	break :blk @Type(...);
};
```

src/blueprint.zig
```diff
@@ -372,6 +374,228 @@ pub const Pattern = struct {
...
+			if(self.isInverse) {
+				return !isMatch;
+			}
+			return isMatch;
+		}
+	};
+
+	pub fn initFromString(allocator: NeverFailingAllocator, source: []const u8) !@This() {
+		var orStorage: OrList = .{};
+		errdefer orStorage.deinit(allocator);
```
IntegratedQuantum, line 462: This is leaking memory if the error happens not on the first iteration. Instead you'd need to have a loop here that also frees all the `and` lists. At this point, I'd suggest to just use
```
var result: @This() = .{.entries = .{}};
errdefer result.deinit();
```

IntegratedQuantum (changes requested):

Argmaster:
Alr, I hope I fixed all the memory leaks in the world, I added some tests and I suppose I will fix rest when we finally have use case for Masks

IntegratedQuantum (approved):
everything looks good now

PR #1287, opened 2025-04-04
Sulfur ore, block, and torches

careeoki:
part of #1272
Neither of the proposed modifiers actually exist, so it just has basic material values.

The "item drop" and ore block are the same block item, so it can be placed back into stone or used to craft or make tools. We should do this for older ores (#1061)

Sulfur ore starts generating at -200. In the future, its Z level could be balanced so that it generates deep enough to have the player manually mine for some time before unlocking explosives.
Veins are small and somewhat uncommon, but are very dense of ore.
image
image


IntegratedQuantum (approved):
Looks great, I can't wait to put some small yellow rocks everywhere :​P

PR #1299, opened 2025-04-12
Big mushroom blocks

careeoki:
Cap, stem, and branch blocks for all mushrooms! We can't generate (proper) mushrooms yet, so this is just blocks.
Mushroom branches make good tool handles. Bolete and toadstool branches are a bit better than planks, glimmergill branches are better than the other mushroom branches.
image
image
image
image


IntegratedQuantum (approved):
This would be really nice with #1227

PR #1306, opened 2025-04-15
Add silver ore

careeoki:
image
image
Has material values based on its real life characteristics. In the future, it should have some sort of anti-magic modifier. I imagine it as the counterpart to gold, which boosts magic power instead.


IntegratedQuantum (approved):
Oh, more shiny rocks!

Also in my mind silver would be a (the only?) alternative to gold for encasing gems.

----

I would however suggest to stop adding more metals for now, we need to leave room for other materials as well, like gems, mob drops, special wood types, dungeon rewards.

Of course we'll probably need more metals later in progression when it comes to the root or the sky islands or other future layers of the world, but I think for the regular cave layer we got enough different kinds of metals for now.

careeoki:
Yeah, I'm done with metals for now.

PR #1309, opened 2025-04-16
Dynamite!

Argmaster:
Guys hear me out

image


Argmaster:
image

IntegratedQuantum:
I cannot accept this until #610 (to avoid lag spikes) and #81 (to avoid griefing) are resolved.
Alternatively, a gamerule to disable explosives could be a short-term workaround for the survival world until we have those features.

careeoki:
The dynamite block model is clever, but it stops you from making a giant pile of dynamite (without there being a bunch of gaps). I think a standard cube would be fine:
image

Argmaster:
Both, preferably as emergent behavior. I would prefer to avoid dynamite blocks flying around like in minecraft tho. With #1261 we can make dynamite explosion damage blocks it can not break, so consecutive explosions would break blocks damaged by previous ones. As for range, I would like to count how many TNT blocks are touching and merge their explosion strength and range as an optimization.

That said current tag based system (`explosionSmall` and friends) for explosion control is not acceptable long term. IMO we should allow for arbitrary attributes (that are not performance critical, not for rendering control) for blocks to be stored in hash map of `ZonElement`s. This problem will re-appear every time we add some non-binary configurable trait of a block (or item!), 
to name few: 
- food value, 
- explosion strength and radius, 
- fluid spread radius, viscosity
- inventory capacity for storage blocks
 3-rd party modifications will be interested in this kind of parameters too.

assets/cubyz/recipes/special_recipes.zig.zon
```diff
@@ -131,4 +131,12 @@
 		.inputs = .{"cubyz:gravel"},
 		.output = "4 cubyz:pebbles",
 	},
+	.{
+		.inputs = .{"2 cubyz:sulfur_ore", "2 cubyz:coal"},
+		.output = "cubyz:dynamite/stick",
+	},
+	.{
+		.inputs = .{"16 cubyz:dynamite/stick"},
```
Argmaster, line 139:
```suggestion
		.inputs = .{"8 cubyz:dynamite/stick"},
```

With new texture it is not that obvious how much dynamite there is in a block, so I think we can make this cheaper to make.

Argmaster:
Resolved conflicts with network refactor.

assets/cubyz/blocks/dynamite/block.zig.zon
```diff
@@ -0,0 +1,12 @@
+.{
+	.tags = .{.canBeIgnited, .explosionMedium},
```
IntegratedQuantum, line 2: Shouldn't the explosion radius be specifiable as a separate property or something? I'm sure we can spare a `u8` per block type for that.

assets/cubyz/recipes/special_recipes.zig.zon
```diff
@@ -147,4 +147,12 @@
...
 		.output = "4 cubyz:pebbles",
 	},
+	.{
+		.inputs = .{"2 cubyz:sulfur_ore", "2 cubyz:coal"},
+		.output = "cubyz:dynamite/stick",
+	},
+	.{
+		.inputs = .{"8 cubyz:dynamite/stick"},
+		.output = "cubyz:dynamite/block",
+	},
```
IntegratedQuantum, line 157: Please also add a recipe to uncraft the block again.

src/blocks.zig
```diff
@@ -317,6 +317,10 @@ pub const Block = packed struct { // MARK: Block
 		return _blockTags[self.typ];
 	}
 
+	pub inline fn hasTag(self: Block, tag: []const u8) bool {
+		return std.mem.indexOfScalar(Tag, _blockTags[self.typ], Tag.find(tag)) != null;
```
IntegratedQuantum, line 321: `std.mem.containsAtLeast` is a the intended function to use in cases like this.

src/gui/windows/save_creation.zig
```diff
@@ -82,6 +82,7 @@ fn flawedCreateWorld() !void {
 
 		gamerules.put("default_gamemode", @tagName(gamemode));
 		gamerules.put("cheats", allowCheats);
+		gamerules.put("allow_explosives", true);
```
IntegratedQuantum, line 85: Please add a checkbox for it.

IntegratedQuantum (changes requested):
preliminary review, I haven't looked at the actual explosion logic yet.

assets/cubyz/blocks/dynamite/block.zig.zon
Argmaster: That is wasteful and not a sustainable model. This property will be used by 2 blocks in base game (maybe 5 if you let us be creative). And it is completely irrelevant to performance. Will you add a field for every property that makes sense for 3 blocks? And what if a modification wants to add one of those? Instead we could have a sparse set of hash maps string to zon and just solve that problem in generic way.

src/gui/windows/save_creation.zig
Argmaster: I widened the window just a bit (to 128+32) so label doesn't wrap

image

assets/cubyz/blocks/dynamite/block.zig.zon
IntegratedQuantum: Well, I think it's only a matter of time until addon creators want to make nuclear explosions with this.
Either way this can be done in the future if you want.

src/network.zig
```diff
@@ -1277,6 +1277,95 @@ pub const Protocols = struct {
...
+					const yOffset: i32 = @intFromFloat(y*distance);
+					const zOffset: i32 = @intFromFloat(z*distance);
+
+					const p = pos + Vec3i{xOffset, yOffset, zOffset};
+
+					var oldBlock = main.server.world.?.getBlock(p[0], p[1], p[2]) orelse continue;
+					if(oldBlock.typ == 0) continue;
+
+					if(damage < oldBlock.blockResistance()) break;
+					damage -= oldBlock.blockHealth();
```
IntegratedQuantum, line 1333: This would be more in line with how block breaking works.
```suggestion
					damage -= oldBlock.blockHealth() + oldBlock.blockResistance();
```

src/network.zig
```diff
@@ -1277,6 +1277,95 @@ pub const Protocols = struct {
...
+					const xOffset: i32 = @intFromFloat(x*distance);
+					const yOffset: i32 = @intFromFloat(y*distance);
+					const zOffset: i32 = @intFromFloat(z*distance);
+
+					const p = pos + Vec3i{xOffset, yOffset, zOffset};
+
+					var oldBlock = main.server.world.?.getBlock(p[0], p[1], p[2]) orelse continue;
+					if(oldBlock.typ == 0) continue;
+
+					if(damage < oldBlock.blockResistance()) break;
```
IntegratedQuantum, line 1332:
```suggestion
					if(damage < oldBlock.blockHealth() + oldBlock.blockResistance()) break;
```

IntegratedQuantum (changes requested):

src/network.zig
```diff
@@ -1277,6 +1277,95 @@ pub const Protocols = struct {
...
+
+				for(0..radiusInt) |j| {
+					const distanceFactor = @as(f32, @floatFromInt(j))/@as(f32, @floatFromInt(radiusInt));
+					const distance = radius*distanceFactor;
+
+					const xOffset: i32 = @intFromFloat(x*distance);
+					const yOffset: i32 = @intFromFloat(y*distance);
+					const zOffset: i32 = @intFromFloat(z*distance);
+
+					const p = pos + Vec3i{xOffset, yOffset, zOffset};
```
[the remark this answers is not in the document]
OneAvargeCoder193: Wouldn't it make more sense to be +|?

src/network.zig
```diff
@@ -1277,6 +1277,95 @@ pub const Protocols = struct {
...
+					const zCentered: usize = @intCast(zOffset + @as(i32, @intCast(diameter/2)));
+
+					exploded.set(xCentered, yCentered, zCentered, true);
+
+					damage -= strength/@as(f32, @floatFromInt(radiusInt));
+				}
+			}
+
+			for(0..diameter) |_x| {
+				const x = @as(i32, @intCast(_x)) + pos[0] - @as(i32, @intCast(diameter/2));
```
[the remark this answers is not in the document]
OneAvargeCoder193: Wouldn't it make more sense for it to be +| here too?

src/network.zig
```diff
@@ -1277,6 +1277,95 @@ pub const Protocols = struct {
...
+
+				for(0..diameter) |_y| {
+					const y = @as(i32, @intCast(_y)) + pos[1] - @as(i32, @intCast(diameter/2));
```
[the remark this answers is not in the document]
OneAvargeCoder193: And here

src/network.zig
```diff
@@ -1277,6 +1277,95 @@ pub const Protocols = struct {
...
+
+					for(0..diameter) |_z| {
+						const z = @as(i32, @intCast(_z)) + pos[2] - @as(i32, @intCast(diameter/2));
```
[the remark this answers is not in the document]
OneAvargeCoder193: And here as well

src/network.zig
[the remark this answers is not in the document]
IntegratedQuantum: Then the explosion would stop at the world border, instead of wrapping around #310

Argmaster:
Damage for stone and mud

image

image

Argmaster:
Future improvements:
- [ ] Sticks of dynamite should be throwable
- [ ] Sticks of dynamite should be placeable on bottom of a block
- [ ] Delayed explosions with dedicated EntityDataClass
- [ ] Chained explosions
- [ ] Ignition animation
- [ ] Explosion particles
- [ ] Damage to blocks which were not broken but were in explosion radius (#1261)
- [ ] Entity damage
- [ ] Scale damage with number of dynamite sticks placed in one block
- [ ] Block drops
- [ ] Batched block updates
- [ ] Placeable string / gunpowder to be used as fuse for igniting dynamite from afar
- [ ] Torches replace dynamite when placed on same block and vice versa
- [ ] VTable based item interactions?
- [ ] Maybe we should have system for arbitrary properties? Explosives could use `explosionStrength` and `explosionRadius` but adding them for all blocks doesn't make sense.

src/network.zig
```diff
@@ -1277,6 +1277,95 @@ pub const Protocols = struct {
...
+		pub const id: u8 = 14;
+		pub const asynchronous = false;
+		fn receive(conn: *Connection, reader: *utils.BinaryReader) !void {
+			const pos = try reader.readVec(Vec3i);
+			const strength = try reader.readFloat(f32);
+			const radius = try reader.readFloat(f32);
+
+			if(conn.user != null) {
+				if(!main.server.world.?.allowExplosives) return;
+				doExplode(pos, strength, radius);
```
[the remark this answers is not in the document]
Argmaster: Made it so client can only nudge server to go and explode a block that exists and server determines the damage based on block.

src/gui/windows/save_creation.zig
```diff
@@ -148,15 +154,18 @@ pub fn onOpen() void {
 	}
 	const name = std.fmt.allocPrint(main.stackAllocator.allocator, "Save{}", .{num}) catch unreachable;
 	defer main.stackAllocator.free(name);
-	textInput = TextInput.init(.{0, 0}, 128, 22, name, .{.callback = &createWorld}, .{});
+
+	const widgetWidth = 160;
```
IntegratedQuantum, line 158: We call them components, not widgets

src/network.zig
```diff
@@ -1289,6 +1289,106 @@ pub const Protocols = struct {
...
+		pub const id: u8 = 14;
+		pub const asynchronous = false;
+		fn receive(conn: *Connection, reader: *utils.BinaryReader) !void {
+			const pos = try reader.readVec(Vec3i);
+
+			if(conn.user == null) return error.InvalidPacket;
+			if(!main.server.world.?.allowExplosives) return;
+			doExplode(pos);
+		}
+		fn doExplode(pos: Vec3i) void {
```
IntegratedQuantum, line 1302: I do not believe this belongs in the network code. Maybe the server world would be a more appropiate place?

IntegratedQuantum (changes requested):

src/network.zig
Argmaster: Honestly I think they should be block entities and this code should be inside / next to onInteract, tho it is a big change and big redesign. This also would mean we are back to believing whatever client says, unless we introduce server side onInteract.
IMO we should consider it though because it would be an example of block entity with such interaction and some attached metadata (explosion radius).

IntegratedQuantum: Well, you could just propagate it as a block entity data change, then you can check on the server side, if the gamerule is set.
Implementing it as a state change could also allow adding a short delay and some fuse particles before detonating.

Argmaster:
Closing until it's a priority to integrate similar stuff into engine.

PR #1313, opened 2025-04-17
Batch block updates

Argmaster:
I completely removed neighbor updates on client side, now server has to send block updates for changed blocks and any neighbors that got affected. This is also a step towards fixing #1217, It would require adding `updateNeighbors` parameter to `cmpxchgBlock` or some other way to signalize that neighbors should remain unchanged. Depending on significance of network overhead maybe it would be good to allow `updateBlock` protocol to carry multiple blocks at once.

Resolves: #610
Related to: #1309 
Related to: #1214


Argmaster:
w/o this change filling area selected on image below with stone takes >1min (didn't wait for it to finish, it may even crash)

Screenshot 2025-04-17 161616

with change, it takes few seconds.

https://github.com/user-attachments/assets/947afa3e-8b96-44b3-a923-e2d6c33a0e81

src/chunk.zig
```diff
@@ -173,6 +173,20 @@ pub const ChunkPosition = struct { // MARK: ChunkPosition
...
+	pub const HashContext = struct {
+		pub fn hash(_: HashContext, a: ChunkPosition) u64 {
+			var h = std.hash.Wyhash.init(0);
+			h.update(std.mem.asBytes(&a.wx));
+			h.update(std.mem.asBytes(&a.wy));
+			h.update(std.mem.asBytes(&a.wz));
+			h.update(std.mem.asBytes(&a.voxelSize));
+			return h.final();
+		}
+		pub fn eql(_: HashContext, a: ChunkPosition, b: ChunkPosition) bool {
```
IntegratedQuantum, line 185: If you don't use a the HashContext, why add a new namespace for it?
Also both of these functions exist already, so either add a context to those, or call them internally so we don't have two different implementations of everything.

src/renderer/mesh_storage.zig
```diff
@@ -859,6 +853,61 @@ pub fn updateMeshes(targetTime: i64) void { // MARK: updateMeshes()
 	}
 }
 
+fn batchUpdateBlocks() void {
+	var lightRefreshList = main.List(*ChunkMesh).init(main.stackAllocator);
+	defer lightRefreshList.deinit();
+
+	var regenerateMesh = std.HashMapUnmanaged(chunk.ChunkPosition, *ChunkMesh, chunk.ChunkPosition.HashContext, 80){};
```
IntegratedQuantum, line 860: Any reason why we can't just use an AutoHashMap?

src/renderer/mesh_storage.zig
```diff
@@ -859,6 +853,61 @@ pub fn updateMeshes(targetTime: i64) void { // MARK: updateMeshes()
...
+		while(regenerateMeshIterator.next()) |entry| {
+			const mesh = entry.value_ptr.*;
+			mesh.generateMesh(&lightRefreshList);
+		}
+	}
+	{
+		// TODO: Don't reschedule light refresh for duplicate meshes.
+		for(lightRefreshList.items) |mesh| {
+			if(mesh.needsLightRefresh.load(.unordered)) {
+				mesh.scheduleLightRefreshAndDecreaseRefCount1();
```
IntegratedQuantum, line 895: Why did I put a `1` there? Could you investigate this?

src/renderer/mesh_storage.zig
```diff
@@ -859,6 +853,61 @@ pub fn updateMeshes(targetTime: i64) void { // MARK: updateMeshes()
...
+	}
+	{
+		var regenerateMeshIterator = regenerateMesh.iterator();
+		while(regenerateMeshIterator.next()) |entry| {
+			const mesh = entry.value_ptr.*;
+			mesh.generateMesh(&lightRefreshList);
+		}
+	}
+	{
+		// TODO: Don't reschedule light refresh for duplicate meshes.
```
IntegratedQuantum, line 892: I think it's already deduplicated elsewhere, so this probably wouldn't make a difference.

src/renderer/mesh_storage.zig
IntegratedQuantum: Also why not just make this a list anyways? Using a hashmap does seem premature since it only has a few entries.

src/server/world.zig
```diff
@@ -1061,12 +1065,21 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+
 			var neighborBlock = ch.getBlock(nx & chunk.chunkMask, ny & chunk.chunkMask, nz & chunk.chunkMask);
 			if(neighborBlock.mode().dependsOnNeighbors) {
 				if(neighborBlock.mode().updateData(&neighborBlock, neighbor.reverse(), newBlock)) {
 					ch.updateBlockAndSetChanged(nx & chunk.chunkMask, ny & chunk.chunkMask, nz & chunk.chunkMask, neighborBlock);
+					const wnx = wx + neighbor.relX();
+					const wny = wy + neighbor.relY();
+					const wnz = wz + neighbor.relZ();
+					for(userList) |user| {
+						std.log.debug("Neighbor update ({d:.2} {d:.2} {d:.2}) {s}:{d}", .{wnx, wny, wnz, neighborBlock.id(), neighborBlock.data});
```
IntegratedQuantum, line 1080: Why are you printing this in a loop?

src/server/world.zig
```diff
@@ -1061,12 +1065,21 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+						main.network.Protocols.blockUpdate.send(user.conn, wnx, wny, wnz, neighborBlock);
```
IntegratedQuantum, line 1081: Could you instead batch it and only send it once in bulk in the end? We shouldn't do 7 network operations (each locking the network mutex) per user here, and the protocol overhead is also a thing (even if it's just 2 bytes here).

IntegratedQuantum (changes requested):

src/server/world.zig
Argmaster: Oh, debug artifact

src/chunk.zig
Argmaster: Right, lets use list anyway. 
Different topic: what do you think about that Wyhash thingy? I found it in AutoHashMap implementation and it looks really universal.

src/renderer/mesh_storage.zig
Argmaster: I don't see any reason right now, maybe during refactoring you have duplicated the function and removed original afterwards? [Original commit](https://github.com/PixelGuys/Cubyz/commit/04f9bae91c1757bd523d68911ab046d6e97c4ef2) shows just that you have added that 1 along other changes.

src/chunk.zig
IntegratedQuantum: I don't know, I'd need to take a look inside to figure out if it's actually any useful.
It might be doing too much slow stuff, I know that's true for Zig's random number generators.

src/renderer/mesh_storage.zig
IntegratedQuantum: Alright, could you remove the 1 then in a future PR, please?

src/network.zig
```diff
@@ -884,24 +884,49 @@ pub const Protocols = struct {
...
 		fn receive(conn: *Connection, reader: *utils.BinaryReader) !void {
-			const x = try reader.readInt(i32);
-			const y = try reader.readInt(i32);
-			const z = try reader.readInt(i32);
-			const newBlock = Block.fromInt(try reader.readInt(u32));
-			if(conn.user != null) {
-				return error.InvalidPacket;
-			} else {
-				renderer.mesh_storage.updateBlock(x, y, z, newBlock);
+			const count = try reader.readInt(u32);
```
IntegratedQuantum, line 905: Can be implicit, just read until there is nothing left.

src/network.zig
```diff
@@ -884,24 +884,49 @@ pub const Protocols = struct {
 	pub const blockUpdate = struct {
 		pub const id: u8 = 7;
 		pub const asynchronous = false;
+
+		pub const BlockUpdate = struct {
```
IntegratedQuantum, line 888: How about reusing `mesh_storage.BlockUpdate` for this? Then you could also adjust the interface of `renderer.mesh_storage.updateBlock` to take a `BlockUpdate`

src/renderer/mesh_storage.zig
Argmaster: Sure

src/renderer/chunk_meshing.zig
```diff
@@ -1184,56 +1182,14 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
...
-				}
-				self.mutex.unlock();
-				neighborBlocks[neighbor.toInt()] = neighborBlock;
-			}
-		}
-		if(newBlock.mode().dependsOnNeighbors) {
-			for(chunk.Neighbor.iterable) |neighbor| {
-				_ = newBlock.mode().updateData(&newBlock, neighbor, neighborBlocks[neighbor.toInt()]);
-			}
-		}
```
IntegratedQuantum, line 1236: Removing the neighbor block checking on the client side means that it has to do a round-trip on the network to e.g. connect fences.
I don't think this is a good experience, especially considering that round trip time can be 500 ms or more.

src/renderer/mesh_storage.zig
```diff
@@ -859,6 +853,55 @@ pub fn updateMeshes(targetTime: i64) void { // MARK: updateMeshes()
...
+				.regenerateMesh => {
+					var found = false;
+					for(regenerateMesh.items) |other| {
+						if(other.pos.equals(mesh)) {
+							mesh.decreaseRefCount();
+							found = true;
+							break;
+						}
+					}
+					if(!found) regenerateMesh.append(main.stackAllocator, mesh);
```
IntegratedQuantum, line 882: Zig supports breaking from arbitrary blocks, which in my opinion is more readable than bools.
```suggestion
				.regenerateMesh => blk: {
					for(regenerateMesh.items) |other| {
						if(other.pos.equals(mesh)) {
							mesh.decreaseRefCount();
							break :blk;
						}
					}
					regenerateMesh.append(main.stackAllocator, mesh);
```

IntegratedQuantum (changes requested):

src/renderer/chunk_meshing.zig
Argmaster: Alright, the delay is annoying even locally, that's unfortunate. I guess getting around #1217 will require sending some flag / using separate protocol after all.

src/network.zig
```diff
@@ -885,23 +886,30 @@ pub const Protocols = struct {
...
-				renderer.mesh_storage.updateBlock(x, y, z, newBlock);
+			while(true) {
+				const update: BlockUpdate = .{
+					.x = reader.readInt(i32) catch break,
+					.y = try reader.readInt(i32),
+					.z = try reader.readInt(i32),
+					.newBlock = Block.fromInt(try reader.readInt(u32)),
+				};
+				if(conn.user != null) {
+					return error.InvalidPacket;
```
IntegratedQuantum, line 897: Also this should probably be outside the loop.

src/renderer/chunk_meshing.zig
Argmaster: Okay, I restored neighbor updates, unfortunately in the process I came across a memory leak and touched few more likes of code than I should have while looking for it. I promise I didn't break anything 🤞

Basically I have extracted `addSelfToLightRefreshList` and merged `neighborBlock.mode().dependsOnNeighbors` and `neighborBlock.mode().updateData(&neighborBlock, neighbor.reverse(), newBlock)` conditions into single if statement to have single else with decref

IntegratedQuantum (changes requested):

src/renderer/chunk_meshing.zig
```diff
@@ -1268,16 +1273,18 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
...
-		for(lightRefreshList.items) |other| {
-			if(other.needsLightRefresh.load(.unordered)) {
-				other.scheduleLightRefreshAndDecreaseRefCount1();
-			} else {
-				other.decreaseRefCount();
+
+		appendIfNotContainsOrDecreaseRefCount(regenerateMeshList, self);
+	}
+
+	fn appendIfNotContainsOrDecreaseRefCount(list: *main.List(*ChunkMesh), mesh: *ChunkMesh) void {
```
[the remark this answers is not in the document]
Argmaster: Oh no my biggest enemy

src/renderer/chunk_meshing.zig
```diff
@@ -1183,47 +1181,49 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
 		const oldBlock = self.chunk.data.getValue(chunk.getIndex(x, y, z));
 
 		if(oldBlock == newBlock) {
+			self.decreaseRefCount();
```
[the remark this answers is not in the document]
Argmaster: Alright, what if I reverse the behavior? Always decrease ref count on call side, but increase ref count if retaining copy in regenerateMeshList, rename `appendIfNotContainedOrDecreaseRefCount` to `appendIfNotContainedAndIncreaseRefCount`

Argmaster:
Alright, depending on your opinion on reversing mesh ownership logic for `regenerateMeshList` it should be ready for re-review.

Argmaster:
Fixed conflicts with latest master.

src/network.zig
```diff
@@ -943,23 +944,29 @@ pub const Protocols = struct {
...
-			const y = try reader.readInt(i32);
-			const z = try reader.readInt(i32);
-			const newBlock = Block.fromInt(try reader.readInt(u32));
 			if(conn.isServerSide()) {
 				return error.InvalidPacket;
-			} else {
-				renderer.mesh_storage.updateBlock(x, y, z, newBlock);
+			}
+			while(reader.remaining.len != 0) {
+				const update: BlockUpdate = .{
```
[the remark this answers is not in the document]
Argmaster: Noo. Btw do zig guarantee order of evaluation of members in initialization like this? I know that C++ does not, but it is used quite often across Cubyz codebase, so I guess zig does?

src/renderer/chunk_meshing.zig
```diff
@@ -1188,42 +1186,44 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
...
-					}
+
+				if(neighborBlock.mode().dependsOnNeighbors and neighborBlock.mode().updateData(&neighborBlock, neighbor.reverse(), newBlock)) {
+					neighborChunkMesh.chunk.data.setValue(index, neighborBlock);
+					neighborChunkMesh.mutex.unlock();
+
+					neighborChunkMesh.updateBlockLight(nnx, nny, nnz, neighborBlock, lightRefreshList);
+					appendIfNotContainedAndIncreaseRefCount(regenerateMeshList, neighborChunkMesh);
+
+					neighborChunkMesh.mutex.lock();
```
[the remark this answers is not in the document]
Argmaster: Aaaand that depends on how you look at this stuff. For me it is setValue that is wrapped with lock/unlock and that other lock at the end of the scope is to match unlock below, hence the formatting. Although since you pointed that out I can change it to your liking.

Argmaster:
Alright, changed mostly as requested, ofc I had to talk back 😝

src/network.zig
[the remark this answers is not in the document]
IntegratedQuantum: I think it guarantees that all function calls happen in the order they are declared in.

src/renderer/chunk_meshing.zig
```diff
@@ -1234,6 +1233,11 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
 				_ = newBlock.mode().updateData(&newBlock, neighbor, neighborBlocks[neighbor.toInt()]);
 			}
 		}
+
+		if(oldBlock.entityDataClass()) |class| {
+			class.onBreakClient(.{_x, _y, _z}, self.chunk);
```
IntegratedQuantum, line 1238: What about the new block?

Maybe better do the missing entity data class methods in a future PR.

src/renderer/chunk_meshing.zig
```diff
@@ -1268,16 +1272,18 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
...
 		self.mutex.unlock();
-		self.generateMesh(&lightRefreshList); // TODO: Batch mesh updates instead of applying them for each block changes.
-		self.mutex.lock();
-		for(lightRefreshList.items) |other| {
-			if(other.needsLightRefresh.load(.unordered)) {
-				other.scheduleLightRefreshAndDecreaseRefCount1();
-			} else {
-				other.decreaseRefCount();
+
+		appendIfNotContainedAndIncreaseRefCount(regenerateMeshList, self);
```
IntegratedQuantum, line 1276: The naming is confusing. Function that contain increaseRefCount are a sign to the caller to decrease it again, which does not happen here.

src/renderer/chunk_meshing.zig
```diff
@@ -1188,42 +1186,43 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
...
 		var neighborBlocks: [6]Block = undefined;
 		@memset(&neighborBlocks, .{.typ = 0, .data = 0});
+
 		for(chunk.Neighbor.iterable) |neighbor| {
 			const nx = x + neighbor.relX();
 			const ny = y + neighbor.relY();
 			const nz = z + neighbor.relZ();
-			if(nx & chunk.chunkMask != nx or ny & chunk.chunkMask != ny or nz & chunk.chunkMask != nz) {
+
+			if(!self.chunk.liesInChunk(nx, ny, nz)) {
```
IntegratedQuantum, line 1197: The function actually works on a different coordinate space (world space relative coordinates), while you give it chunk space relative coordinates.
This is not relevant now, since the function only operates on voxelSIze=1 chunks, but it might be relevant in the future.

IntegratedQuantum (changes requested):

src/renderer/chunk_meshing.zig
Argmaster: So remove that part completely?

IntegratedQuantum: yes

src/renderer/chunk_meshing.zig
Argmaster: That was moved by mistake while restoring neighbor updates.

src/renderer/chunk_meshing.zig
```diff
@@ -1234,6 +1237,7 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
 				_ = newBlock.mode().updateData(&newBlock, neighbor, neighborBlocks[neighbor.toInt()]);
 			}
 		}
+
```
Argmaster, line 1240:
```suggestion
```

src/renderer/chunk_meshing.zig
Argmaster: Well, what do I do about it?

src/renderer/chunk_meshing.zig
Argmaster: Moved back, as for new block, it gets `onPlaceClient` instead in line 1244

src/renderer/chunk_meshing.zig
IntegratedQuantum: You revert to the old logic, or add a function with the old logic with a name that clearly distinguishes it from `liesInChunk`.

Argmaster: > clearly distinguishes it from `liesInChunk`.

Yeah, unlikely to happen, lets just get this over with.

```suggestion
			if(nx & chunk.chunkMask != nx or ny & chunk.chunkMask != ny or nz & chunk.chunkMask != nz) {
```

src/items.zig
```diff
@@ -463,7 +463,7 @@ pub const Tool = struct { // MARK: Tool
 	pub fn deinit(self: *const Tool) void {
 		// TODO: This is leaking textures!
 		//if(self.texture) |texture| {
-			//texture.deinit();
```
IntegratedQuantum, line 466: please rebase

Argmaster:
Alr getting rid of that formatting change was harder than I though xD
But its gone now.

IntegratedQuantum (approved):

PR #1336, opened 2025-04-22
Use `readVec` and `writeVec` where applicable

Argmaster:
Resolves: #1264


IntegratedQuantum (approved):
Thanks for taking this.

Issue #1303, opened 2025-04-13
First placement of `texture_pile` does not consume the block


Argmaster:
i.e. you can play infinitely many first level leaf piles!

![Image](https://github.com/user-attachments/assets/754b3740-99c4-4a9b-a8c3-05bffb24b242)

PR #1340, opened 2025-04-24
Fix `texture_pile`

Argmaster:
Resolves: #1303

### DIFF
--- a/src/rotation/texture_pile.zig
+++ b/src/rotation/texture_pile.zig
@@ -86,7 +86,7 @@ pub fn canBeChangedInto(oldBlock: Block, newBlock: Block, item: main.items.ItemS
 		.no, .yes_costsDurability, .yes_dropsItems => return .no,
 		.yes_costsItems => |r| return .{.yes_costsItems = r},
 		.yes => {
-			const oldData = @min(oldBlock.data, oldBlock.modeData() - 1);
+			const oldAmount = if(oldBlock.typ == newBlock.typ) @min(oldBlock.data, oldBlock.modeData() - 1) else 0;
 			if(oldData == newBlock.data) return .no;
 			if(oldData < newBlock.data) {
 				if(!isThisItem(newBlock, item)) return .no;
--- a/src/rotation/texture_pile.zig
+++ b/src/rotation/texture_pile.zig
@@ -73,7 +73,7 @@ pub fn onBlockBreaking(_: ?main.items.Item, _: Vec3f, _: Vec3f, currentData: *Bl
 	if(currentData.data == 0) {
 		currentData.* = .{.typ = 0, .data = 0};
 	} else {
-		currentData.data = currentData.data - 1;
+		currentData.data = @min(currentData.data, currentData.modeData() - 1) - 1;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Argmaster:
(and hopefully doesn't introduce more bugs)

src/rotation/texture_pile.zig
```diff
@@ -77,16 +77,26 @@ pub fn onBlockBreaking(_: ?main.items.Item, _: Vec3f, _: Vec3f, currentData: *Bl
...
+					return .{.yes_costsItems = @intCast(newBlock.data - oldBlock.data)};
+				} else {
+					return .{.yes_dropsItems = @intCast(oldBlock.data - newBlock.data)};
+				}
+			} else if(oldBlock.typ == 0) {
+				return .{.yes_costsItems = 1};
 			} else {
-				if(item.item == null or item.item.? != .baseItem or item.item.?.baseItem.block != newBlock.typ) return .no;
-				return .{.yes_costsItems = @intCast(amountChange)};
+				return .no;
```
IntegratedQuantum, line 99: This case is already handled in more generic form in `DefaultFunctions.canBeChangedInto`, you can remove it together with the `if(oldBlock.typ == 0)` check above.

Argmaster:
Ready for re-review!

I did `const result =` because `.yes` and `.yes_coststems` have different backing type and this disallows using `=> |result|` syntax.

src/rotation/texture_pile.zig
```diff
@@ -77,16 +77,27 @@ pub fn onBlockBreaking(_: ?main.items.Item, _: Vec3f, _: Vec3f, currentData: *Bl
...
+					return .{.yes_costsItems = newBlock.data - oldBlock.data};
+				} else {
+					return .{.yes_dropsItems = oldBlock.data - newBlock.data};
+				}
+			} else if(oldBlock.typ == 0) {
+				return .{.yes_costsItems = newBlock.data};
 			} else {
-				if(item.item == null or item.item.? != .baseItem or item.item.?.baseItem.block != newBlock.typ) return .no;
-				return .{.yes_costsItems = @intCast(amountChange)};
+				return result;
```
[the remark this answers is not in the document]
Argmaster: `.yes_costsItems` needs to be special as it is invoked when placing `texture_pile` and first state of texture pile has data=0 same as air, so cost would be 0.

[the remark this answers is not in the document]
Argmaster: I also added value clamping to fix old leaves which were generated with carpet rotation mode

src/rotation/texture_pile.zig
```diff
@@ -73,20 +73,26 @@ pub fn onBlockBreaking(_: ?main.items.Item, _: Vec3f, _: Vec3f, currentData: *Bl
...
 pub fn canBeChangedInto(oldBlock: Block, newBlock: Block, item: main.items.ItemStack, shouldDropSourceBlockOnSuccess: *bool) RotationMode.CanBeChangedInto {
 	switch(RotationMode.DefaultFunctions.canBeChangedInto(oldBlock, newBlock, item, shouldDropSourceBlockOnSuccess)) {
 		.no, .yes_costsDurability, .yes_dropsItems => return .no,
-		.yes, .yes_costsItems => {
-			const amountChange = @as(i32, newBlock.data) - if(oldBlock.typ == newBlock.typ) @as(i32, oldBlock.data) else 0;
-			if(amountChange <= 0) {
-				return .{.yes_dropsItems = @intCast(-amountChange)};
+		.yes_costsItems => |r| return .{.yes_costsItems = r},
+		.yes => {
+			const oldData = @min(oldBlock.data, oldBlock.modeData() - 1);
```
IntegratedQuantum, line 89: This doesn't handle the case of replacing non-solid blocks correctly:
```suggestion
			const oldAmount = if(oldBlock.typ == newBlock.typ) @min(oldBlock.data, oldBlock.modeData() - 1) else 0;
```

src/rotation/texture_pile.zig
```diff
@@ -73,20 +73,26 @@ pub fn onBlockBreaking(_: ?main.items.Item, _: Vec3f, _: Vec3f, currentData: *Bl
 	if(currentData.data == 0) {
 		currentData.* = .{.typ = 0, .data = 0};
 	} else {
-		currentData.data = currentData.data - 1;
+		currentData.data = @min(currentData.data, currentData.modeData() - 1) - 1;
 	}
 }
 
+fn isThisItem(block: Block, item: main.items.ItemStack) bool {
```
IntegratedQuantum, line 80: This function name is a bit misleading, my first instinct is to read it as "Is this an item?"
I would suggest `isItemBlock`

IntegratedQuantum (changes requested):

Argmaster:
Changed as requested.

IntegratedQuantum (approved):

PR #1356, opened 2025-04-26
Blueprint scopes

Argmaster:
This pull request adds `--scope`/`-s` flag to simplify specifying where blueprint should be placed. Currently placing blueprint in `assets/cubyz/blueprints/` requires either manual action or tricks with relative paths. After that change to save blueprint in `assets/cubyz/blueprints` all you have to do is include `-s game` and blueprint will be placed in mentioned directory **on server side**. To download blueprint you can use `-s local`, then blueprint will be placed in local `./blueprints/` directory.
On top of that this PR changes how locations of blueprints are specified. Now instead of raw path user should specify ID of blueprint.

## Example:
`/blueprint save -s game cubyz:stem` will save clipboard to `./assets/cubyz/blueprints/stem.blp` i.e. exactly in place where sbb system will look for it.
`/blueprint list -s game` will list all blueprints in `assets` (in form of IDs)
`/blueprint save argmaster:stone -s world` will save clipboard in `./saves/{current}/assets/argmaster/blueprints/stone.blp`

### DIFF
--- a/src/blueprint.zig
+++ b/src/blueprint.zig
@@ -129,7 +129,7 @@ pub const Blueprint = struct {
 			}
 		}
 	}
-	pub fn load(allocator: NeverFailingAllocator, inputBuffer: []u8) !Blueprint {
+	pub fn load(allocator: NeverFailingAllocator, inputBuffer: []const u8) !Blueprint {
 		var compressedReader = BinaryReader.init(inputBuffer);
 		const version = try compressedReader.readInt(u16);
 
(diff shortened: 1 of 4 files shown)


Argmaster:
This will be handled by #1581, we could consider adding separate `/blueprint download` command to allow downloading server side blueprints to client. Blueprints should still be changed to use IDs instead of paths, but this can be done after #1425 so It doesn't require manually crafted parser.

PR #1363, opened 2025-04-27
Add serialization and deserialization functions to entity_data.zig

OneAvargeCoder193:


Argmaster:
I am pretty sure that is what onLoad and onUnload functions are for, exact details should be handled by the struct that is stored.

PR #1367, opened 2025-04-29
Particles

RanPix:
https://github.com/user-attachments/assets/de82e298-acfb-431d-973a-2d37b9a707ae


RanPix:
i did something wrong

assets/cubyz/shaders/particles/particles.vs
```diff
@@ -0,0 +1,90 @@
...
+struct ParticleData {
+	vec3 pos;
+	float align1;
+	vec3 vel; // unused
+	float align2;
+	float lifeTime;
+	float lifeLeft;
+	uint typ;
+	uint light;
+	float size;
```
Argmaster, line 23: Can size be stored in ParticleTypeData?

Argmaster: Can we maybe scale particle size with texture size (in pixels)? Imo size of single pixel should remain more or less constant across all the textures, this kind of applies to other stuff too.

RanPix: the thing is that we could have the same particle type used with different sizes applied to them, also particles could have "dynamic" size

RanPix: > Can we maybe scale particle size with texture size (in pixels)? Imo size of single pixel should remain more or less constant across all the textures, this kind of applies to other stuff too.

then i think it would be harder to make nice small scale particles, like torch smoke in minecraft

src/particles.zig
```diff
@@ -0,0 +1,386 @@
...
+		textures.deinit();
+		emissionTextures.deinit();
+		arenaForWorld.deinit();
+		textureArray.deinit();
+		emissionTextureArray.deinit();
+		particleTypeHashmap.deinit();
+		ParticleSystem.deinit();
+	}
+
+	pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
```
Argmaster, line 58: Don't return ID if you don't need it on call side.

src/particles.zig
```diff
@@ -0,0 +1,386 @@
...
+		particleType.isBlockTexture = zon.get(bool, "isBlockTexture", false);
+		particleType.animationFrames = zon.get(u32, "animationFrames", 1);
+
+		const textureId = zon.get([]const u8, "texture", "cubyz:spark");
+		var splitter = std.mem.splitScalar(u8, textureId, ':');
+		const mod = splitter.first();
+		const _id = splitter.rest();
+
+		// this is so confusing i just hardcoded that thing
+		const path = std.fmt.allocPrint(main.stackAllocator.allocator, "assets/{s}/particles/textures/{s}.png", .{mod, _id}) catch unreachable;
```
Argmaster, line 71: What if texture comes from world specific assets?

RanPix: i dont really know how this asset system works so i did that, when i was using the asset folder for the path it just straight up didnt load the images because the path was wrong

RanPix: i will look into this again

assets/cubyz/shaders/particles/particles.vs
```diff
@@ -0,0 +1,90 @@
...
+uniform mat4 projectionMatrix;
+uniform mat4 viewMatrix;
+uniform mat4 billboardMatrix;
+uniform ivec3 playerPositionInteger;
+uniform vec3 playerPositionFraction;
+
+struct ParticleData {
+	vec3 pos;
+	float align1;
+	vec3 vel; // unused
```
IntegratedQuantum, line 17: You shouldn't send unused data to the GPU.

assets/cubyz/shaders/particles/particles.vs
```diff
@@ -0,0 +1,90 @@
...
+
+const vec3 facePositions[4] = vec3[4]
+(
+	vec3(-0.5f, 0.5f, 0.0f),
+	vec3(0.5f, -0.5f, 0.0f),
+	vec3(0.5f, 0.5f, 0.0f),
+	vec3(-0.5f, -0.5f, 0.0f)
+);
+
+const int indices[6] = int[6](0, 1, 2, 1, 0, 3);
```
IntegratedQuantum, line 57: With manual indexing the GPU has no chance to cache vertex shader invocations of the same index, which leads to 50% more work (assuming that particles are vertex bound).
Please use an index buffer, you can reuse the one from chunk rendering. Then you could also use the faster `gl_VertexID/4` and `%4`.

src/main.zig
```diff
@@ -703,6 +707,7 @@ pub fn main() void { // MARK: main()
 
 		if(game.world != null) { // Update the game
 			game.update(deltaTime);
+			particles.ParticleManager.update(deltaTime);
```
IntegratedQuantum, line 710: Should be part of `game.update`

src/particles.zig
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	pub fn render(playerPosition: Vec3d, ambientLight: Vec3f) void {
+		ParticleSystem.render(playerPosition, ambientLight);
+	}
+};
+
+pub const ParticleSystem = struct {
+	pub const maxCapacity: u32 = 65536;
+	var particles: []Particle = undefined;
```
IntegratedQuantum, line 137: We don't need to heap-allocate this
```suggestion
	var particles: [maxCapacity]Particle = undefined;
```

assets/cubyz/shaders/particles/particles.vs
```diff
@@ -0,0 +1,90 @@
...
+	textureIndex = particleType.animationFrames != 1 ?
+		floor(float(particleType.startFrame) + (particle.lifeLeft/particle.lifeTime)*float(particleType.animationFrames)) :
+		particleType.startFrame;
+
+	vec3 position = (billboardMatrix*vec4(particle.size*facePositions[indices[vertexID]], 1.0)).xyz;
+	position += vec3(particle.pos - playerPositionInteger);
+	position -= playerPositionFraction;
+
+	vec4 mvPos = viewMatrix*vec4(position, 1);
+	gl_Position = projectionMatrix*mvPos;
```
IntegratedQuantum, line 87: Please premultiply projection and view matrix on the CPU.

src/particles.zig
```diff
@@ -0,0 +1,385 @@
...
+
+	pub fn deinit() void {
+		shader.deinit();
+		particlesSSBO.deinit();
+		particles.len = maxCapacity;
+		main.globalAllocator.free(particles);
+	}
+
+	pub fn update(deltaTime: f32) void {
+		const vdt: Vec3f = @as(Vec3f, @splat(deltaTime));
```
IntegratedQuantum, line 184: Please use a more descriptive name `vdt` is just gibberish

src/particles.zig
```diff
@@ -0,0 +1,385 @@
...
+				particles.len -= 1;
+				continue;
+			}
+
+			particle.vel += properties.gravity*vdt;
+			particle.vel *= @splat(@max(0, 1 - properties.drag*deltaTime));
+			const vel = particle.vel*vdt;
+
+			// TODO: OPTIMIZE THE HELL OUT OF THIS
+			if(particle.collides) {
```
IntegratedQuantum, line 201: We can start optimizing right here. What particles wouldn't have collision?

src/particles.zig
```diff
@@ -0,0 +1,385 @@
...
+				@as(u32, light[3] >> 3) << 10 |
+				@as(u32, light[4] >> 3) << 5 |
+				@as(u32, light[5] >> 3));
+
+			particles[i] = particle;
+			i += 1;
+		}
+	}
+
+	pub fn spawn(id: []const u8, count: u32, pos: Vec3f, size: f32, collides: bool, shape: EmmiterShape) void {
```
IntegratedQuantum, line 246: Could you optimize all Emitter related parameters into a separate emitter struct?
Also please move the bulk of this function there, implementation details for all the shapes don't belong into the ParticleSystem struct in my opinion.

src/particles.zig
```diff
@@ -0,0 +1,385 @@
...
+};
+
+pub const ParticleType = struct {
+	texture: u32,
+	animationFrames: u32,
+	startFrame: u32,
+	isBlockTexture: bool,
+};
+
+// TODO?: separate the data which is sent to the gpu into another struct
```
IntegratedQuantum, line 374: Please do it now.

src/renderer.zig
```diff
@@ -188,6 +189,10 @@ pub fn renderWorld(world: *World, ambientLight: Vec3f, skyColor: Vec3f, playerPo
 	Skybox.render();
 	gpu_performance_measuring.stopQuery();
 
+	gpu_performance_measuring.startQuery(.particle_rendering);
+	particles.ParticleManager.render(playerPos, ambientLight);
```
IntegratedQuantum, line 193: Particles should be rendered after opaque chunks to prevent lag spikes from the occlusion culling system.

IntegratedQuantum (changes requested):
Preliminary review

IntegratedQuantum:
Also I'm getting a shader compilatio error:
```
OpenGL Shader Compiler error: 0:43(8): error: syntax error, unexpected NEW_IDENTIFIER, expecting ')' or ','
Error compiling shader assets/cubyz/shaders/particles/particles.vs:
0:43(8): error: syntax error, unexpected NEW_IDENTIFIER, expecting ')' or ','
```

src/particles.zig
RanPix: some types of ambient, or really small particles that should not be bothered with collisions. Like "void" particles or tiny smoke particles from torches

IntegratedQuantum: Well, in that case these should be stored in a separate array, otherwise we get expensive branch mispredicts here all the time.

RanPix: got it

src/particles.zig
```diff
@@ -368,18 +390,19 @@ pub const ParticleType = struct {
...
 	startFrame: u32,
+	size: f32,
 	isBlockTexture: bool,
 };
 
-// TODO?: separate the data which is sent to the gpu into another struct
 pub const Particle = struct {
-	pos: Vec3f,
-	vel: Vec3f,
+	pos: Vec4f, // 4th element is rotation
```
IntegratedQuantum, line 395:
```suggestion
	posAndRotation: Vec4f,
```

src/particles.zig
```diff
@@ -368,18 +390,19 @@ pub const ParticleType = struct {
...
 	lifeTime: f32,
 	lifeLeft: f32,
-	typ: u32,
 	light: u32 = 0,
-	size: f32,
+	typ: u32,
+};
+
+pub const ParticleLocal = struct {
+	vel: Vec4f, // 4th element is rotation velocity
```
IntegratedQuantum, line 403: Comments age poorly
```suggestion
	velAndRoationVel: Vec4f,
```

src/particles.zig
RanPix: i think we are going to have different particle logic functions anyways do handle different movement types, like windy rain, snow, dust storms, leaves or fireflies?

RanPix: but then for that its probably better to allocate different particle buffers for that

assets/cubyz/shaders/particles/particles.vs
IntegratedQuantum: `c.glBindVertexArray(chunk_meshing.vao);`
`c.glDrawElements` instead of `c.glDrawArrays` with *4 instead of *6.
And you also should make sure it draws them in multiple groups since the index buffer has a limited size.

src/particles.zig
IntegratedQuantum: yes

IntegratedQuantum (changes requested):

IntegratedQuantum:
Also, maybe it's just me, but collision does not seem to work for the test particle.

src/particles.zig
RanPix: and this should be done in a later pr?

RanPix:
i disabled particle collisions for now (you can reenable them in the spawn call at the bottom of game.zig)

assets/cubyz/shaders/particles/particles.vs
RanPix: thanks!

src/particles.zig
```diff
@@ -0,0 +1,408 @@
...
+		const emission = readTextureFile(path, "_emission.png", Image.emptyImage);
+		typ.startFrame = @intCast(textures.items.len);
+		typ.size = @as(f32, @floatFromInt(base.width))/16;
+		for(0..animationFrames) |i| {
+			textures.append(extractAnimationSlice(base, i, animationFrames));
+			emissionTextures.append(extractAnimationSlice(emission, i, animationFrames));
+		}
+	}
+
+	fn extendedPath(_allocator: main.heap.NeverFailingAllocator, path: []const u8, ending: []const u8) []const u8 {
```
IntegratedQuantum, line 95: Just use `std.mem.concat`

src/particles.zig
IntegratedQuantum: Yes, please, also this should be a generic method in assets.zig, instead of repeating this logic everywhere.

src/particles.zig
```diff
@@ -0,0 +1,408 @@
...
+		defer main.stackAllocator.free(path);
+		return Image.readFromFile(arenaForWorld.allocator(), path) catch default;
+	}
+
+	fn extractAnimationSlice(image: Image, frame: usize, frames: usize) Image {
+		if(image.height < frames) return image;
+		var startHeight = image.height/frames*frame;
+		if(image.height%frames > frame) startHeight += frame else startHeight += image.height%frames;
+		var endHeight = image.height/frames*(frame + 1);
+		if(image.height%frames > frame + 1) endHeight += frame + 1 else endHeight += image.height%frames;
```
IntegratedQuantum, line 110: Instead of worrying about partial textures so much, you should just print an error if the image size is not divisbile by the number of frames, and then just use the `image.height/frames*frame` directly.

src/particles.zig
```diff
@@ -0,0 +1,408 @@
...
+		var result = image;
+		result.height = @intCast(endHeight - startHeight);
```
IntegratedQuantum, line 112: Can be avoided if you just pass `frameHeight: u32 = image.height/frames*frame` into the function

src/particles.zig
```diff
@@ -0,0 +1,408 @@
...
+		result.imageData = result.imageData[startHeight*image.width .. endHeight*image.width];
+		return result;
+	}
+
+	pub fn generateTextureArray() void {
+		std.log.debug("Particle texture sizes: {d} {d}", .{textures.items.len, emissionTextures.items.len});
```
IntegratedQuantum, line 118: No need to print two copies of the same number.
If you want to check if they are the same, use an assertion.

src/particles.zig
```diff
@@ -0,0 +1,408 @@
...
+const Vec3d = vec.Vec3d;
+const Vec3f = vec.Vec3f;
+const Vec4f = vec.Vec4f;
+const Vec3i = vec.Vec3i;
+
+var arena = main.heap.NeverFailingArenaAllocator.init(main.globalAllocator);
+const allocator = arena.allocator();
+
+pub const ParticleManager = struct {
+	pub var particleTypesSSBO: ?SSBO = null;
```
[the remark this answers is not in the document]
IntegratedQuantum: Also please deinit this somewhere.

src/particles.zig
```diff
@@ -0,0 +1,408 @@
...
+		ParticleSystem.update(dt);
+	}
+
+	pub fn render(projectionMatrix: Mat4f, viewMatrix: Mat4f, playerPosition: Vec3d, ambientLight: Vec3f) void {
+		ParticleSystem.render(projectionMatrix, viewMatrix, playerPosition, ambientLight);
+	}
+};
+
+pub const ParticleSystem = struct {
+	pub const maxCapacity: u32 = 1165536;
```
IntegratedQuantum, line 139: Please choose a power of 2

src/particles.zig
```diff
@@ -0,0 +1,408 @@
...
+	var particleCount: u32 = 0;
+	var particles: [maxCapacity]Particle = undefined;
+	var particlesLocal: [maxCapacity]ParticleLocal = undefined;
+	var properties: EmmiterProperties = undefined;
```
IntegratedQuantum, line 143: Why are these here? I don't think they should be hardcoded in the particle system.

IntegratedQuantum (changes requested):

src/particles.zig
```diff
@@ -0,0 +1,408 @@
...
+const Vec3f = vec.Vec3f;
+const Vec4f = vec.Vec4f;
+const Vec3i = vec.Vec3i;
+
+var arena = main.heap.NeverFailingArenaAllocator.init(main.globalAllocator);
+const allocator = arena.allocator();
+
+pub const ParticleManager = struct {
+	pub var particleTypesSSBO: ?SSBO = null;
+	pub var types: main.List(ParticleType) = undefined;
```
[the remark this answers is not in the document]
RanPix: the types? they are used to get the particle size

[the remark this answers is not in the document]
IntegratedQuantum: But they aren't accessed outside this files.

src/particles.zig
```diff
@@ -0,0 +1,405 @@
...
+	var particleCount: u32 = 0;
+	var particles: [maxCapacity]Particle = undefined;
+	var particlesLocal: [maxCapacity]ParticleLocal = undefined;
+	var properties: EmmiterProperties = undefined;
+	var seed: u64 = undefined;
+	var previousPlayerPos: Vec3f = undefined;
+
+	var particlesSSBO: SSBO = undefined;
+
+	var shader: Shader = undefined;
```
IntegratedQuantum, line 148: Please also switch from a Shader to the new graphics.Pipeline now.

IntegratedQuantum: And also fix the shader compiler errors from glslang

assets/cubyz/shaders/particles/particles.vs
```diff
@@ -84,5 +80,6 @@ void main() {
 
 	gl_Position = projectionAndViewMatrix*vec4(position, 1);
 
-	uv = uvPositions[vertexID];
+	float textureIndex = floor(float(particleType.startFrame) + particle.lifeRatio*float(particleType.animationFrames));
```
IntegratedQuantum, line 80: Why not store the frame data as a float to begin with? Casts are not free.

assets/cubyz/shaders/particles/particles.vs
```diff
@@ -67,18 +67,18 @@ void main() {
 	light = max(sunLight*ambientLight, blockLight)/31;
 
 	float rot = particle.rot;
-	vec3 pos = facePositions[vertexID];
+	vec3 facePos = facePositions[vertexID];
```
IntegratedQuantum, line 68: That name is kind of confusing, since that is not the position of the face.

src/particles.zig
```diff
@@ -141,15 +141,14 @@ pub const ParticleSystem = struct {
 	var particlesLocal: [maxCapacity]ParticleLocal = undefined;
 	var properties: EmmiterProperties = undefined;
 	var seed: u64 = undefined;
+	var previousPlayerPos: Vec3f = undefined;
```
IntegratedQuantum, line 144: Should probably be Vec3d, otherwise you get precision issues when calculating the distance

assets/cubyz/shaders/particles/particles.vs
```diff
@@ -77,6 +77,6 @@ void main() {
 	const vec3 vertexPos = (billboardMatrix*vec4(particleType.size*vertexRotationPos, 1)).xyz + particle.pos;
 	gl_Position = projectionAndViewMatrix*vec4(vertexPos, 1);
 
-	float textureIndex = floor(float(particleType.startFrame) + particle.lifeRatio*float(particleType.animationFrames));
```
IntegratedQuantum: Well, now it's an implicit cast, but it's still a cast.

src/particles.zig
```diff
@@ -0,0 +1,407 @@
...
+		billboardMatrix: c_int,
+		ambientLight: c_int,
+		textureSampler: c_int,
+		emissionTextureSampler: c_int,
+	};
+	var uniforms: UniformStruct = undefined;
+
+	pub fn init() void {
+		std.log.debug("Particle alignment: {d} size: {d}\n", .{@alignOf(Particle), @sizeOf(Particle)});
+		shader = Shader.initAndGetUniforms("assets/cubyz/shaders/particles/particles.vs", "assets/cubyz/shaders/particles/particles.fs", "", &uniforms);
```
IntegratedQuantum, line 161: Also please rebase and rename the shaders to .frag/.vert as in #1427

src/particles.zig
```diff
@@ -0,0 +1,411 @@
...
+				particlesLocal[i] = particlesLocal[particleCount];
+				continue;
+			}
+
+			var rot = particle.posAndRotation[3];
+			const rotVel = particleLocal.velAndRotationVel[3];
+			rot += rotVel*deltaTime;
+
+			particleLocal.velAndRotationVel += vec.combine(properties.gravity, 0)*vecDeltaTime;
+			particleLocal.velAndRotationVel *= @splat(@max(0, 1 - properties.drag*deltaTime));
```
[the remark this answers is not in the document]
RanPix: i dont think that really matters much? i tested it on different framerates and it looks good enough, also exponentiations are probably too expensive for particles

RanPix:
i think its almost done, all we need to figure out probably in this or next pr is how to allow for more spawning flexibility to allow for more diverse behaviors

src/particles.zig
RanPix: separate particle pools should be probably done in other prs

src/particles.zig
```diff
@@ -0,0 +1,413 @@
+const std = @import("std");
```
Argmaster, line 1:
```suggestion
const std = @import("std");

```

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+const main = @import("main");
+const chunk_meshing = @import("renderer/chunk_meshing.zig");
+const graphics = @import("graphics.zig");
+const SSBO = graphics.SSBO;
```
Argmaster, line 5: Aliases should be sorted below the import the come from.

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+	var particleTypesSSBO: SSBO = undefined;
+	var types: main.List(ParticleType) = undefined;
+	var textures: main.List(Image) = undefined;
+	var emissionTextures: main.List(Image) = undefined;
+
+	var textureArray: TextureArray = undefined;
+	var emissionTextureArray: TextureArray = undefined;
+
+	const ParticleIndex = u16;
+	var particleTypeHashmap: std.StringHashMapUnmanaged(ParticleIndex) = undefined;
```
Argmaster, line 36:
```suggestion
	var IdToIndex: std.StringHashMapUnmanaged(ParticleIndex) = undefined;
```

Names of those hash maps are insanely inconsistent, can we make a final decision here and apply it everywhere @IntegratedQuantum? And reverseIndeces is not the name.

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+		textureArray.deinit();
+		emissionTextureArray.deinit();
+		particleTypeHashmap.deinit(arenaAllocator.allocator);
+		ParticleSystem.deinit();
+		arena.deinit();
+	}
+
+	pub fn register(assetsFolder: []const u8, id: []const u8, zon: ZonElement) void {
+		const animationFrames = zon.get(u32, "animationFrames", 1);
+		const textureId = zon.get([]const u8, "texture", "cubyz:spark");
```
Argmaster, line 60: I don't like this default, I think it should just fail and report that ID must be specified. I hate silent failures. Use null as default and use orelse to log and early return.

Argmaster (changes requested):

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+
+		return graphics.Image.readFromFile(arenaAllocator, worldAssetsPath) catch graphics.Image.readFromFile(arenaAllocator, gameAssetsPath) catch {
+			if(status == .isMandatory) std.log.err("Particle texture not found in {s} and {s}.", .{worldAssetsPath, gameAssetsPath});
+			return default;
+		};
+	}
+
+	fn extractAnimationSlice(image: Image, frame: usize, frames: usize, imageName: []const u8, textureType: enum {base, emmision}) Image {
+		if(image.height%frames != 0) {
+			std.log.err("Particle texture size is not divisible by its frame count for {s} in {s} texture", .{imageName, @tagName(textureType)});
```
Argmaster, line 110: I would suggest suggest including those numbers in the error message
```suggestion
			std.log.err("Particle texture size ({}) is not divisible by its frame count ({}) for {s} in {s} texture", .{image.height, frames, imageName, @tagName(textureType)});
```

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+		emissionTextureArray.generate(emissionTextures.items, true, false);
+		c.glTexParameterf(c.GL_TEXTURE_2D_ARRAY, c.GL_TEXTURE_MAX_ANISOTROPY, @floatFromInt(main.settings.anisotropicFiltering));
+
+		particleTypesSSBO = SSBO.initStatic(ParticleType, ParticleManager.types.items);
+		particleTypesSSBO.bind(14);
+	}
+};
+
+pub const ParticleSystem = struct {
+	pub const maxCapacity: u32 = 524288;
```
Argmaster, line 134: Where did the number come from?

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+	var particleCount: u32 = 0;
+	var particles: [maxCapacity]Particle = undefined;
+	var particlesLocal: [maxCapacity]ParticleLocal = undefined;
+	// TODO: add different emitters for different types of movements like windy, normal, no collisions and etc.
```
Argmaster, line 138: Better make it an issue, people tend to forgot about TODOs and we have dedicated system for tracking tasks.

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+	var pipeline: graphics.Pipeline = undefined;
+	const UniformStruct = struct {
+		projectionAndViewMatrix: c_int,
+		billboardMatrix: c_int,
+		ambientLight: c_int,
+	};
+	var uniforms: UniformStruct = undefined;
+
+	pub fn init() void {
+		std.log.debug("Particle alignment: {d} size: {d}\n", .{@alignOf(Particle), @sizeOf(Particle)});
```
Argmaster, line 153: A debugging relict?

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+		particleTypesSSBO = SSBO.initStatic(ParticleType, ParticleManager.types.items);
+		particleTypesSSBO.bind(14);
+	}
+};
+
+pub const ParticleSystem = struct {
+	pub const maxCapacity: u32 = 524288;
+	var particleCount: u32 = 0;
+	var particles: [maxCapacity]Particle = undefined;
+	var particlesLocal: [maxCapacity]ParticleLocal = undefined;
```
Argmaster, line 137: What is particleLocal?

src/particles.zig
careeoki: Yeah, textures should always be square.

src/particles.zig
```diff
@@ -0,0 +1,413 @@
...
+
+			var rot = particle.posAndRotation[3];
+			const rotVel = particleLocal.velAndRotationVel[3];
+			rot += rotVel*deltaTime;
+
+			particleLocal.velAndRotationVel += vec.combine(properties.gravity, 0)*vecDeltaTime;
+			particleLocal.velAndRotationVel *= @splat(@max(0, 1 - properties.drag*deltaTime));
+			const posDelta = particleLocal.velAndRotationVel*vecDeltaTime;
+
+			// TODO: OPTIMIZE THE HELL OUT OF THIS
```
Argmaster, line 211: Unprofessional wording, code should not contain jokes so it is easy to read. Also this should be an issue.

src/particles.zig
IntegratedQuantum: For now it's fine in my opinion. Ultimate goal would be to make a separate datastructure for tracking this anyways.

I also think idToIndex is a bit too generic in my opinion.

src/particles.zig
IntegratedQuantum: Hmm, what about rain?

src/particles.zig
IntegratedQuantum: Yeah, please make an issue for it. As for the exact wording, I think a little fun in comments is nice sometimes.

src/particles.zig
RanPix: i guess the sides are going to be 0 alpha, tho i could try making particles have separate width and height sizes if needed

RanPix: but extra discarded pixels are probably not good

src/particles.zig
Argmaster: > I also think idToIndex is a bit too generic in my opinion.

It is within context of ParticleManager **and** part of its internal implementation. It has solid, unambiguous context. It should always be the case. It is as generic as `textures`. I would argue opposite, `particleTypeHashmap` is overspecified, there is no other type hash map, not even any other hash map.

src/particles.zig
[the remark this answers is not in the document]
IntegratedQuantum: It matters if we ever want media with larger friction (e.g. water, lava).
And we can always make it less precise if we can prove it to be a bottleneck, and given that right now collision and lighting are the main bottlenecks, I'd assume that we can afford one exponential function here.

src/particles.zig
Argmaster: > Hmm, what about rain?

What about it?

> but extra discarded pixels are probably not good

You will be discarding pixels for a lot of particles anyway, rarely any particle fills the whole square.

IntegratedQuantum: > You will be discarding pixels for a lot of particles anyway, rarely any particle fills the whole square.

There is a big difference between a 25% discard rate for a circular texture and a >90% discard rate for rain.

src/particles.zig
IntegratedQuantum: > I would argue opposite, particleTypeHashmap is overspecified, there is no other type hash map, not even any other hash map.

There is no other type hash map, but there very well could be a hashmap for emitter types or emitter shapes. If you don't know the details then you wouldn't know what this hashmap is for.

src/particles.zig
IntegratedQuantum: also next time don't forget to link it here: #1471

src/particles.zig
Argmaster: This is preparing for the future, future is unknown we can't prepare for. Also this can be solved with proper namespacing.

IntegratedQuantum: Well, mainly this is just personal preference, I prefer the current way of naming things.

And like I said, we should turn this into a generic data structure and avoid the problem altogether. So let's stop discussing about it.

src/particles.zig
RanPix: moved it to #1491

RanPix:
i think this is ready for next reviews, please also check if there are any problems with particle rendering, if they disappear randomly just in case

Argmaster:
I don't see any spawning, I assume this is intentional because we don't want players just spawning particles... unless? Maybe temporarily we could make players just leave trace of few, short living particles behind them? Just for a week or so as an easter egg?

RanPix:
forgot to mention that they spawn at 0, 0, 0 coordinates, also the example of spawn code is at the bottom of game.zig

IntegratedQuantum:
Now that #1289 is merged, please rebase and adapt to these asset loading changes.

Argmaster:
I'll likely hang around in development channel if you need help with that @RanPix

src/particles.zig
```diff
@@ -0,0 +1,441 @@
...
+const Vec4f = vec.Vec4f;
+const Vec3i = vec.Vec3i;
+
+var seed: u64 = undefined;
+
+var arena = main.heap.NeverFailingArenaAllocator.init(main.globalAllocator);
+const arenaAllocator = arena.allocator();
+
+pub const ParticleManager = struct {
+	var particleTypesSSBO: SSBO = undefined;
```
IntegratedQuantum, line 28: Please also deinit the SSBO.

src/particles.zig
```diff
@@ -0,0 +1,441 @@
...
+		return result;
+	}
+
+	pub fn generateTextureArray() void {
+		textureArray.generate(textures.items, true, true);
+		c.glTexParameterf(c.GL_TEXTURE_2D_ARRAY, c.GL_TEXTURE_MAX_ANISOTROPY, @floatFromInt(main.settings.anisotropicFiltering));
+		emissionTextureArray.generate(emissionTextures.items, true, false);
+		c.glTexParameterf(c.GL_TEXTURE_2D_ARRAY, c.GL_TEXTURE_MAX_ANISOTROPY, @floatFromInt(main.settings.anisotropicFiltering));
+
+		particleTypesSSBO = SSBO.initStatic(ParticleType, ParticleManager.types.items);
```
IntegratedQuantum, line 161: Please init it in the init function to avoid creating it twice when the texture array is regenerated.
Since you don't know the initial size, you can just use `particleTypesSSBO.bufferData` here.

src/particles.zig
```diff
@@ -0,0 +1,441 @@
...
+		result.height = @intCast(frameHeight);
+		result.imageData = result.imageData[startHeight*image.width .. endHeight*image.width];
+		return result;
+	}
+
+	pub fn generateTextureArray() void {
+		textureArray.generate(textures.items, true, true);
+		c.glTexParameterf(c.GL_TEXTURE_2D_ARRAY, c.GL_TEXTURE_MAX_ANISOTROPY, @floatFromInt(main.settings.anisotropicFiltering));
+		emissionTextureArray.generate(emissionTextures.items, true, false);
+		c.glTexParameterf(c.GL_TEXTURE_2D_ARRAY, c.GL_TEXTURE_MAX_ANISOTROPY, @floatFromInt(main.settings.anisotropicFiltering));
```
IntegratedQuantum, line 159: This is a dynamic setting, please check if it is actually updated (and this function called) when moving the slider in the settings menu.

src/particles.zig
```diff
@@ -0,0 +1,441 @@
...
+pub const Emitter = struct {
+	shapeType: EmitterShapeEnum,
+	directionMode: DirectionModeEnum = .spread,
+	count: u32,
+	size: f32 = 0,
+	dir: Vec3f = @splat(0),
+	id: []const u8,
+	collides: bool,
+
+	pub fn spawn(self: Emitter, pos: Vec3d) void {
```
IntegratedQuantum, line 371: I would prefer a more precise name 
```suggestion
	pub fn spawnParticles(self: Emitter, pos: Vec3d) void {
```

src/particles.zig
```diff
@@ -0,0 +1,441 @@
...
+		const typ = ParticleManager.particleTypeHashmap.get(self.id) orelse 0;
```
IntegratedQuantum, line 372: StringHashmap lookups should always be precomputed during load time, instead of redoing them on every access.

src/particles.zig
```diff
@@ -0,0 +1,441 @@
...
+		const properties = ParticleSystem.properties;
+
+		const to: u32 = @intCast(@min(ParticleSystem.particleCount + self.count, ParticleSystem.maxCapacity));
+		for(ParticleSystem.particleCount..to) |_| {
```
IntegratedQuantum, line 376: I would prefer computing the count directly, also the intCast seems pointless
```suggestion
		const count = @min(self.count, ParticleSystem.maxCapacity - ParticleSystem.particleCount);
		for(0..count) |_| {
```

IntegratedQuantum (changes requested):

src/particles.zig
IntegratedQuantum: Did you check this?

RanPix: i made the anisotropic filtering slider callback call this function on change, but im not sure if it actually changes anything

RanPix: i dont know how to test

IntegratedQuantum: You should be able to see it at a lower resolution scale with static (non-rotating, non-animated) particles.

src/particles.zig
```diff
@@ -0,0 +1,461 @@
...
+			i += 1;
+		}
+		previousPlayerPos = playerPos;
+	}
+
+	fn addParticle(typ: u32, pos: Vec3d, vel: Vec3f, collides: bool) void {
+		const lifeTime = properties.lifeTimeMin + random.nextFloat(&seed)*properties.lifeTimeMax;
+
+		particles[particleCount] = Particle{
+			.posAndRotation = vec.combine(@as(Vec3f, @floatCast(pos - previousPlayerPos)), 0),
```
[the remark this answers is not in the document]
RanPix: Do we want that to be the same as min and max rotation velocity settings, or just a boolean which will make them rotate from 0 to 360 degrees on spawn? I think the second option is better

src/particles.zig
```diff
@@ -0,0 +1,461 @@
...
+		c.glBindVertexArray(chunk_meshing.vao);
+
+		for(0..std.math.divCeil(u32, particleCount, chunk_meshing.maxQuadsInIndexBuffer) catch unreachable) |_| {
+			c.glDrawElements(c.GL_TRIANGLES, @intCast(particleCount*6), c.GL_UNSIGNED_INT, null);
+		}
+	}
+
+	pub fn getParticleCount() u32 {
+		return particleCount;
+	} // sdfjasidjgfasdpij
```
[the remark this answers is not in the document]
RanPix: i was testing the formatter there (forgot to clean up ❤️‍🔥❤️‍🔥)

src/particles.zig
RanPix: i guess that works 

https://github.com/user-attachments/assets/aafbc8a6-1448-452b-ab23-ae1729005286

src/particles.zig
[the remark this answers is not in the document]
IntegratedQuantum: Yeah, a bool would be good for now. We'll probably need more options in the future for stuff like rain, but that's a different issue and we shouldn't worry about it yet.

src/particles.zig
IntegratedQuantum: Now that I think about it: We shouldn't waste resources on anisotropic filtering for particles at all. They are always oriented towards the camera, anisotropic filtering is intended for looking at things from a shallow angle.
So just remove it I guess.

src/particles.zig
```diff
@@ -0,0 +1,450 @@
...
+			particle.light = compressedLight;
+
+			i += 1;
+		}
+		previousPlayerPos = playerPos;
+	}
+
+	fn addParticle(typ: u32, pos: Vec3d, vel: Vec3f, collides: bool) void {
+		const lifeTime = properties.lifeTimeMin + random.nextFloat(&seed)*properties.lifeTimeMax;
+		const rot = if(properties.randomizeRotationOnSpawn) random.nextFloat(&seed)*std.math.tau else 0;
```
IntegratedQuantum, line 281: Please just use 2*pi instead of tau.

src/particles.zig
```diff
@@ -0,0 +1,450 @@
...
+		radius: f32,
+		mode: DirectionMode,
+		position: Vec3d,
+
+		pub fn spawn(self: SpawnSphere) struct {Vec3d, Vec3f} {
+			const spawnPos: Vec3f = @splat(self.radius);
+			var offsetPos: Vec3f = undefined;
+			while(true) {
+				offsetPos = random.nextFloatVectorSigned(3, &seed);
+				if(vec.length(offsetPos) <= 1) break;
```
IntegratedQuantum, line 377: Use lengthSquare, to avoid doing a square root every attempt.

IntegratedQuantum (changes requested):
Two more things, then we can merge.

Also before merging please remove some of the dummy assets, they should be readded when they are actually used.

RanPix:
Done.

Argmaster (approved):

src/particles.zig
```diff
@@ -39,7 +39,9 @@ pub const ParticleManager = struct {
 	pub fn init() void {
 		types = .init(arenaAllocator);
 		textures = .init(arenaAllocator);
+		textures.append(Image.defaultImage);
```
IntegratedQuantum, line 42: This is an awful workaround. Please just fix the root issue.

RanPix: Ok, i just thought you wouldn't want me to change more things at this point.

IntegratedQuantum: Should be a small fix

RanPix: Setting max width and height at 1 on initialization in `TextureArray,generate()` seems to work but im not sure if thats a good fix...

IntegratedQuantum (approved):

PR #1380, opened 2025-05-02
Command name autocompletion

Argmaster:
## Description

This pull request adds the more trivial part of autocompletion - command name completion.

Behavior is as follows: use `tab` to find all partial matches, use `tab` again to cycle through them. Insert any character to confirm (eg. space). Completion for subcommands / command parameters is not supported yet. 

Depends on: #1244 due to optional callbacks, actual diff is smaller than presented by Github right now,

Implementation details:
- Moved command initialization to `fn main()` to guarantee command list is available. Right now we could WA this and just read command names, but this will be crucial for client side subcommand / parameter completion where callbacks for each command will be required.
- Completion prefers longest match, I decided that this is most likely to be most beneficial, although completion list is not sorted, only longest match is selected to be first in O(n) time while forming completion list.

## Links

Depends on: #1425

### DIFF
--- a/src/gui/components/TextInput.zig
+++ b/src/gui/components/TextInput.zig
@@ -33,6 +33,7 @@ maxHeight: f32,
 textSize: Vec2f = undefined,
 scrollBar: *ScrollBar,
 onNewline: gui.Callback,
+optional: OptionalCallbacks,
 
 pub fn __init() void {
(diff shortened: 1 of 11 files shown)


Argmaster:
I will try again after #1425, lets not clutter PR list with that PR, it needs a total overhaul anyway

PR #1377, opened 2025-05-02
Add coniferous trees the way I like them

Argmaster:
I still need more variants of lushy branches, also the average branch variants seem a bit bare, too bare IMO, tho maybe these are things we could tweak later?

Screenshot 2025-06-02 000603
Screenshot 2025-06-02 000620
Screenshot 2025-06-02 000636

Issue #1250, opened 2025-03-28
[Structure Building Block] Allow specifying empty / sentinel children


Argmaster:
Currently empty child structure list in structure building block configuration is treated as an error. However, it has a valid use case of leaving some of child spawn blocks empty on purpose.

See discussion in https://github.com/PixelGuys/Cubyz/pull/1240#discussion_r2017268765

PR #1382, opened 2025-05-02
Oak and Birch Tree Generators

ikabod-kee:
image
Now, I know what it looks like with all those changes, but in reality you don't have to check half of those, as they're just blueprints.
I also included "deco" and "null," which are for small decorations and nothing respectively.


assets/cubyz/blueprints/birch_1/base.blp
Argmaster, line 1: I don't like this exact structure you shown here, I seriously doubt that none of the birch models can be reused as oak models - block substitution is a planned feature - so I would suggest to prepare blueprints for that - blueprints don't exactly care if they are made of birch or oak wood, but they care if they are a stem or something else + they care about child block layout. Neither of those information is encoded clearly in ID of that asset. Take a look onto my PR blueprint layout in #1377 . I first group asset by tree type, then by what part of tree they belong too, then I have variants with just number or with encoded child block layout (applies to stems) I don't expect you to copy that 1:1, but I suggest you consider eg. separating that `birch_1` into `birch/1`. I will give a better review and possiby more concrete suggestions tomorrow.

Argmaster (changes requested):

assets/cubyz/blueprints/birch_1/base.blp
Argmaster: As for child block placement, It is really annoying that there is no simple way to view those blueprints or spawn SBBs on demand with fixed seed. Both of those features should get high priority alongside QOL features like masked set command.

Argmaster:
Oh, so you don't feel like I'm only criticizing - these trees are really cool trees, good job! C:

assets/cubyz/sbb/oak/1/trunk_6.zig.zon
```diff
@@ -0,0 +1,21 @@
+.{
+	.blueprint = "cubyz:oak/1/trunk_6",
+	.children = .{
+		.blue = .{
+			.{.structure = "cubyz:oak/1/branch_1", .chance = 0.333},
+			.{.structure = "cubyz:oak/1/branch_2", .chance = 0.333},
+			.{.structure = "cubyz:oak/1/branch_3", .chance = 0.334},
```
IntegratedQuantum, line 7: Note that the chances don't need to add up to 1, if you want all paths have the same chance, you can set it to `1` (which is the default, so you can just omit it)
```suggestion
			.{.structure = "cubyz:oak/1/branch_1"},
			.{.structure = "cubyz:oak/1/branch_2"},
			.{.structure = "cubyz:oak/1/branch_3"},
```

assets/cubyz/sbb/oak/1/trunk_6.zig.zon
```diff
@@ -0,0 +1,21 @@
...
+		.crimson = .{
+			.{.structure = "cubyz:oak/1/tree_blob_1", .chance = 0.5},
+			.{.structure = "cubyz:oak/1/tree_blob_3", .chance = 0.5},
+		},
+		.yellow = .{
+			.{.structure = "cubyz:oak/1/stub_1", .chance = 0.1},
+			.{.structure = "cubyz:oak/1/stub_2", .chance = 0.1},
+			.{.structure = "cubyz:oak/1/stub_3", .chance = 0.1},
+			.{.structure = "cubyz:oak/1/stub_4", .chance = 0.1},
+			.{.structure = "cubyz:null/null_side", .chance = 0.6},
```
IntegratedQuantum, line 18: @Argmaster could you please work on #1250 before we have too many of these?

Argmaster: Ye I will have it done today.

IntegratedQuantum:
And I also have to say: They look amazing in-game!

Argmaster:
Birches like these:

image

Really remind me of Acacia more than Birch trees, but I guess it might be just the style choice, noone said Birches have to look like birches all the time.

Although I would suggest adding few hanging leaves, so they don't have that obviously flat bottoms, I have same problem in #1377, In my case I will probably try chaning some of the leafed branches into normal ones so they stand out more, but here I would suggest this:

image

Don't get me wrong - I don't want the leaf balls back, I just want few hanging leaves, a bit like here:

image

Argmaster:
Hmm that is bad :C
What did we do wrong -_-

image

Argmaster:
In plains like these:

image

I would rather see a 2 block thick oaks, taller ones than those in forests and also placed more sparsely. Our placement method does not preserve any minimal distance between spawns which is a bit of a shame, but we have to cope for now.

assets/cubyz/sbb/oak/1/trunk_6.zig.zon
Argmaster: #1388

IntegratedQuantum:
> Hmm that is bad :C
> What did we do wrong -_-

Yeah, the degradable concept is not really working for everything. Ideally it would probably give each block a priority instead of being just a bool.

assets/cubyz/sbb/oak/1/trunk_6.zig.zon
IntegratedQuantum: merged, so please remove all the null files

assets/cubyz/sbb/birch/birch_forest_generator.zig.zon
Argmaster, line 1: What does that one do?

Argmaster:
- you are going to type most of the paths anyway, whether you chose to go with `blob_1` or `blob/1` doesnt even change char count
- autocompletion will work better with directories, implementing directory level completion is easier than implementing partial file name completion
- its our internal choice, it imposes no requirements on 3rd party extensions
- SBB system is tricky to get into for different reasons - for example you have no idea how those blueprints look like - I could find few others, but what matters is that number of directories on the way to asset is waaay less important.
- you should be using IDE to work with assets, and then it doesn't matter how many directories are on the way, since they are displayed as tree. If you are not using IDE, there are plenty of reasons to do so.

Argmaster:
Thats up to you. But don't duplicate identical componets if only difference is block type used.

Argmaster:
Or rather, be ready to remove duplicates, because I suppose until we have substitution we should keep different variants...?

For substitutions I need masks, then I can work on generating necessary assets during loading, since I already have `set` method,

IntegratedQuantum:
> Oh no... I forgot which of these are which.

@Argmaster another argument for #1387

Argmaster:
I guess, but for now I have few other things queued up and limited time.

assets/cubyz/sbb/tree/oak/1/leaf_1.zig.zon
```diff
@@ -0,0 +1,4 @@
+.{
+	.blueprint = "cubyz:tree/oak/1/leaf_1",
+	.children = .{},
+}
```
IntegratedQuantum, line 4: @Argmaster could we maybe allow adding end nodes without having to specify an empty file like this?

assets/cubyz/sbb/tree/oak/1/base_3.zig.zon
```diff
@@ -0,0 +1,22 @@
...
+			.{.structure = "cubyz:tree/oak/1/root_6", .chance = 0.125},
+			.{.structure = null, .chance = 0.25},
+		},
+		.brown = .{
+			.{.structure = "cubyz:tree/oak/1/trunk_1"},
+			.{.structure = "cubyz:tree/oak/1/trunk_2"},
+			.{.structure = "cubyz:tree/oak/1/trunk_3"},
+			.{.structure = "cubyz:tree/oak/1/trunk_4"},
+			.{.structure = "cubyz:tree/oak/1/trunk_5"},
+			.{.structure = "cubyz:tree/oak/1/trunk_6"},
```
IntegratedQuantum, line 19: @Argmaster should we add some kind of wildcard system?
Then here it would just be a single entry:
```zig.zon
.{.structure = "cubyz:tree/oak/1/trunk_*"},
```

IntegratedQuantum (changes requested):
one thing to change, the rest are just suggestions for future changes to the underlying system.

If @Argmaster doesn't have any other complaints then we can merge this.

assets/cubyz/sbb/tree/oak/1/base_3.zig.zon
Argmaster: Alright, but what are the rules?
The obvious matches:

```
cubyz:tree/oak/1/trunk_0
cubyz:tree/oak/1/trunk_1
cubyz:tree/oak/1/trunk_453
```

But, does that match to?

```
cubyz:tree/oak/1/trunk_/0
cubyz:tree/oak/1/trunk_/1
cubyz:tree/oak/1/trunk_/234
```

Do we go full glob with `?` and `**`?
What do we do with `cubyz:tree/oak/1/trunk_*/foobar_1`, do we even allow this?
Where is the wiki explaining what is and what is not allowed C:
Maybe instead of wildcards we would allow specifying directory-like syntax and we would take *everything starting with `cubyz:tree/oak/1/trunk/`*

Argmaster: The last one creates grouping that is visible both in code **and in directory structure** with no tricks to be done.

Argmaster: Generally speaking I am not happy with the `generator` structure types, they are clever trick but not indented use, it would be best to solve both those problems at once.

Argmaster:
I **really** dislike redundancy present in file names, stuff like `branch_1`, `branch_2`, `branch_3` should be instead a directory `branch` containing `1`, `2`, `3`. This is what directories are for, to avoid duplicating same information in every file name. And if you are refusing to follow that guideline, at least do this consistently with `birch/1` being instead `birch_1`.

assets/cubyz/sbb/tree/oak/1/leaf_1.zig.zon
Argmaster: Or do you mean like without having SBB file at all?
Maybe, it would mean I would have to allow inlined SBB definitions I will have to dive deeper to see what is the cost.

Argmaster: https://github.com/PixelGuys/Cubyz/issues/1403

assets/cubyz/sbb/tree/oak/1/base_3.zig.zon
IntegratedQuantum: > Alright, but what are the rules?

In my opinion it should only match one directory at a time, just like in bash.

> Do we go full glob with ? and **?

Let's stick to the simple things, and extend when we need it.

> What do we do with cubyz:tree/oak/1/trunk_*/foobar_1, do we even allow this?

Sure, that doesn't seem too difficult.

> Maybe instead of wildcards we would allow specifying directory-like syntax and we would take everything starting with cubyz:tree/oak/1/trunk/

That would be another option, I guess. It does however limit your freedom for organizing things.

IntegratedQuantum:
> There, I did the changes you wanted

Please also then make a PR with the corresponding changes to the contributing guidelines.

IntegratedQuantum:
If you can't find it, then tell me where you searched for it, so we can put a link there.

Also <https://github.com/PixelGuys/Cubyz/blob/master/CONTRIBUTING.md>

assets/cubyz/sbb/tree/oak/1/base_3.zig.zon
Argmaster: You know what I think about freedom.

Argmaster (approved):
There are few small things, but I consider them negligible for that PR:
- `.children = .{},` should be omitted, but #1403 is more general fix for that, so lets not waste more time on that
- contribution guidelines should be updated to inform what is the preferred layout of SBBs and blueprints. Techincally also could be affected by #1403, not sure yet, lets have this as separate PR too

IntegratedQuantum (approved):
Alright, let's do it then! Time for some new ~~lag biomes~~ trees!

PR #1425, opened 2025-05-09
Better argument parser

Argmaster:
## Description

This pull request introduces prototype of declarative, fully comptime argument parser based on introspection. The goal is to use to declaratively describe command syntax and automatically provide autocompletion for each of declared segments.

`/tp` command has been migrated to Parser based design as proof of concept.

## Example

Declaration of parser for `/tp` command:

```zig
const Args = union(enum) {
	@"/tp <x> <y> <z>": struct {x: f64, y: f64, z: ?f64},
	@"/tp <biome>": struct {biome: main.argparse.BiomeId(true)},
};

const ArgParser = main.argparse.Parser(Args, .{.commandName = "/tp"});
```

## Links

Related to: #1380


Argmaster:
TODO: "unpack" structs

src/server/command/tp.zig
```diff
@@ -6,93 +6,94 @@ const User = main.server.User;
...
-		z = source.player.pos[2];
+		},
+		.@"tp to xyz" => |params| {
+			var x = params.x;
+			var y = params.y;
+			var z = params.z;
+
+			if(z == null) {
+				z = source.player.pos[2];
+			}
```
IntegratedQuantum, line 92:
```suggestion
			var z = params.z orelse source.player.pos[2];
```

IntegratedQuantum:
> I need @IntegratedQuantum to decide if we keep callbacks as they are, do we try passing arbitrary arguments to them through parse call, or do we remove them completely.

Is this is still relevant? It seems like you removed the callbacks already.

IntegratedQuantum:
Oh right now I see it, I was only looking at the reference implementation so far.
Yeah, we can just remove it.

src/argparse.zig
```diff
@@ -0,0 +1,297 @@
+const std = @import("std");
+
+const main = @import("main");
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const ListUnmanaged = main.ListUnmanaged;
+
+pub fn Parser(comptime T: type, comptime callback: ?fn(self: T) anyerror!void) type {
+	return struct {
+		const Self = @This();
+		pub const Args = T;
```
IntegratedQuantum, line 10: Why do we need this?

src/argparse.zig
```diff
@@ -0,0 +1,297 @@
...
+
+		pub fn parse(allocator: NeverFailingAllocator, args: []const u8) !ParseResult(T) {
```
IntegratedQuantum, line 12: Doesn't need the error, if you remove the callback.

src/argparse.zig
```diff
@@ -0,0 +1,297 @@
...
+						inline .@"enum" => return std.meta.stringToEnum(Field, arg) orelse return error.InvalidEnum,
+						inline .float => |floatInfo| return try std.fmt.parseFloat(std.meta.Float(floatInfo.bits), arg),
+						inline .int => |intInfo| return try std.fmt.parseInt(std.meta.Int(intInfo.signedness, intInfo.bits), arg, 0),
+						inline else => |other| @compileError("Unsupported type " ++ @tagName(other)),
+					}
+				},
+			}
+		}
+		fn parseUnion(comptime u: std.builtin.Type.Union, allocator: NeverFailingAllocator, args: []const u8) ParseResult(T) {
+			var failureMessages: ListUnmanaged([]const u8) = .{};
```
IntegratedQuantum, line 82: Since you join them in the end anyways, just use a list of u8 and add a `\n` at the end of all messages.

Argmaster:
I also returned this PR to draft state because I figured I should also try how well autocompletion will fit into that design.

src/argparse.zig
```diff
@@ -0,0 +1,297 @@
...
+		pub const Args = T;
+
+		pub fn parse(allocator: NeverFailingAllocator, args: []const u8) !ParseResult(T) {
+			const result = _parse(allocator, args);
+			if(callback != null and result == .success) {
+				try callback.?(result.success);
+			}
+			return result;
+		}
+		pub fn _parse(allocator: NeverFailingAllocator, args: []const u8) ParseResult(T) {
```
IntegratedQuantum, line 19: maybe you could find a better name for this?
Also this should probably be private.

src/argparse.zig
```diff
@@ -0,0 +1,297 @@
...
+		}
+		fn parseStruct(comptime s: std.builtin.Type.Struct, allocator: NeverFailingAllocator, args: []const u8) ParseResult(T) {
+			var result: T = undefined;
+
+			var split = std.mem.splitScalar(u8, args, ' ');
+			var offset: usize = 0;
+			var count: usize = 0;
+
+			inline for(s.fields) |field| {
+				if(!std.mem.eql(u8, field.name, "flags")) {
```
IntegratedQuantum, line 39: How do flags work exactly?
Also this check should be moved to the PR that actually adds flag support.

src/argparse.zig
```diff
@@ -0,0 +1,297 @@
...
+}
+
+pub fn ParseResult(comptime SuccessT: type) type {
+	return union(enum) {
+		failure: struct {
+			message: []const u8,
+		},
+		success: SuccessT,
+
+		pub fn deinit(self: @This(), allocator: NeverFailingAllocator) void {
```
IntegratedQuantum, line 136: Hmm, this function does seem quite easy to miss. It seems you missed it in the reference implementation too.

Maybe a better pattern would be to pass in the result message List. This would allow using regular errors, with their syntax sugar, as well as moving this up one layer into the command code, so we could do the error message handling there. Either way, that would be a bigger rewrite, and doesn't need to be done here.

IntegratedQuantum (changes requested):

src/argparse.zig
Argmaster: They don't, it was an idea I never got to implement, I'll remove it.

src/argparse.zig
Argmaster: This most likely can be gone after callback is gone, it was used before because I don't like break to label syntax, I prefer early return, but now it is not really relevant I think

Argmaster:
Oh I see I don't push the removal of the callback parameter, sorry for my confusion.

src/argparse.zig
Argmaster: I think I will just have failure contain ListUnmanaged([]const u8), why bother joining, reallocating etc. If you can just print them separately. I will confirm readability of that solution tho

src/argparse.zig
Argmaster: Noted, but I don't fully understand the request unfortunately, I guess we will return to this topic. What I would like to preserve is the fact that this parser can be used both for commands and for CLI parameters for Cubyz executable if useful at some point.

src/argparse.zig
Argmaster: Replaced with `resolve` that is used for parts of parse and autocomplete that are really similar.

Argmaster:
image

Idk the message could be better tbh

Argmaster:
IMO much better now:

image

IntegratedQuantum:
> Idk the message could be better tbh

I'd suggest
```
Couldn't match argument list
/tp <x> <y> <z>
 Expected a number at argument 1 found "f" ← Note that normal people start counting at 1
/tp <biome>
 Invalid biome id "f"
```

Also if there is only one argument list, I'd suggest to only show the error message without all the extra information.

Argmaster:
Ewww being normal

Argmaster:
This kind of message with all the variants is only shown for unions, if there is only a struct with one set of arguments it will stop on first error and will report it.

src/argparse.zig
IntegratedQuantum: I think it would always be more readable to have this as a separate, named, struct instead of passing it anonymously.
Note that anymous structs also tend to make names in debug information more verbose.

Argmaster:
image

image

image

I would leave those, otherwise it is less readable, mind that adding a leading space will not work as good as in Github UI because chat is mach narrower and message is almost guaranteed to wrap.
Mind that system has to be generic, eg. BiomeId can be on different position in different use case, so it should print its position even if it is only position in the command.

Argmaster:
image

Argmaster:
I also "fixed" some of the funny combinations of optionals one may want to create, with optionals in the middle of argument list.

archbirdplus:
There are two naming schemes going on. First you give each of the arguments names, and then you proceed to not use them and have the player dereference argument indices instead.

If we continue to write `/tp x y z`, why not report using the argument name, `expected a number for x, found "foo"`?
Alternatively, I suggest writing the type signature like `/tp number number number` (with appropriate syntactic flair of course). Then it would make sense to keep the argument indices like `argument 1 expected to be a number, found "foo"` or whatever.

Having both type signature and argument name is probably silly.

Argmaster:
Good point, this is doable, but programmer must make sure that field names in struct overlap with what's described by the union field name, otherwise the message will make no sense.

Argmaster:
image

Argmaster:
@archbirdplus What do you think?

archbirdplus:
Just looking at this last screenshot:
- it seems you are missing a comma before `found`
- the quotes are inconsistent (double or single?)

Also, the grammar is different between those two lines. It might be worth standardizing it, if possible.

Argmaster:
image

IntegratedQuantum:
> Good point, this is doable, but programmer must make sure that field names in struct overlap with what's described by the union field name, otherwise the message will make no sense.

Instead of using the union field name, maybe just generate it from the command argument list?

Argmaster:
Isn't that convenience change? You can add that at any time since it relaxes requirements, doesn't impose new ones. Requires another recursive walk to determine all the possible combinations of commands, easily 100 lines of code, guaranteed discussion about best data structure, few more iterations of review... For me sounds like good separate PR.

Args are not aware about command name, but I guess we can have that as a dedicated constant or parser parameter.

Argmaster:
Different counterargument could be that you have to call those union and struct fields something readable and unambiguous, so full command signature is perfect for that.

IntegratedQuantum:
> Different counterargument could be that you have to call those union and struct fields something readable and unambiguous, so full command signature and names of arguments are perfect for that.

Well if it's visible to the user then it would still be good to enforce it to follow that form, right?

Argmaster:
Do you distinguish between optional and mandatory arguments with `<>` and `[]`? Mind that now there are only two branches in parsing, one with `x, y, optional z` and one with `biome`, so no `x, y` variant, unless you explicitly make one.

Argmaster:
Two things, IMO `<>` and `/` should be parts of parameter names and commandName, so parser remains generic enough so it can be used for command line parameters.

Argmaster:
TODO: use toOwnedSlice

IntegratedQuantum:
> Two things, IMO <> and / should be parts of parameter names and commandName, so parser remains generic enough so it can be used for command line parameters, eg. for headless server.

I don't quite understand this, the parser doesn't parse <> and / anywhere anyways, right?

Argmaster:
It doesn't, so it shouldn't contain code that automatically includes those in error messages, since they will not be a part of shell command syntax (at least `/` won't be). They should be explicitly specified by the user to allow us to adjust it for shell syntax.

IntegratedQuantum:
> since they will not be a part of shell command syntax (at least / won't be)

Why not? / is used to distinguish between chat messages and commands, shouldn't you be able to send messages to chat from the server console?

archbirdplus:
For "shell command syntax", do you mean the argv parser when launching the server?

Argmaster:
Yes, exactly

archbirdplus:
Surely you would want to be able to pass multiple arguments. Are you going to update the parser to accept something like `Cubyzig --new-world "My World" --spawn x y --seed ...`?

Argmaster:
TBD, if there will be use case we will work on it. For now all that's planned in near future are flags.

src/utils.zig
```diff
@@ -1935,3 +1935,7 @@ fn CastFunctionReturnToAnyopaqueType(Fn: type) type {
 pub fn castFunctionReturnToAnyopaque(function: anytype) *const CastFunctionReturnToAnyopaqueType(@TypeOf(function)) {
 	return @ptrCast(&function);
 }
+
+pub fn format(allocator: NeverFailingAllocator, comptime fmt: []const u8, args: anytype) []u8 {
```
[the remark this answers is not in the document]
Argmaster: Comments currently are not required, rejecting the suggestion.

src/server/terrain/biomes.zig
```diff
@@ -600,7 +600,7 @@ pub const TreeNode = union(enum) { // MARK: TreeNode
 var finishedLoading: bool = false;
 var biomes: main.List(Biome) = undefined;
 var caveBiomes: main.List(Biome) = undefined;
-var biomesById: std.StringHashMap(*Biome) = undefined;
+pub var biomesById: std.StringHashMap(*Biome) = undefined;
```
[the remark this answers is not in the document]
Argmaster: N/A

src/server/command/tp.zig
```diff
@@ -6,93 +6,92 @@ const User = main.server.User;
...
-					return;
-				}
+		return;
+	}
+	switch(result.success) {
+		.@"/tp <biome>" => |params| {
+			const biome = main.server.terrain.biomes.getById(params.@"<biome>".id);
+			if(!std.mem.eql(u8, biome.id, params.@"<biome>".id)) {
+				source.sendMessage("#ff0000Couldn't find biome with id \"{s}\"", .{params.@"<biome>".id});
+				return;
```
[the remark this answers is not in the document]
Argmaster: Who cares.

src/utils.zig
```diff
@@ -2099,6 +2099,10 @@ pub fn castFunctionReturnToAnyopaque(function: anytype) *const CastFunctionRetur
 	return @ptrCast(&function);
 }
 
+pub fn format(allocator: NeverFailingAllocator, comptime fmt: []const u8, args: anytype) []u8 {
```
IntegratedQuantum: As mentioned on another PR, if you want this, please make an issue for it first.
This is a bigger refactor/decision that doesn't belong here.

Argmaster: No, I'm good, I will just copy-paste `allocPrint`. I guess it is just 20 extra chars so who cares. I just didn't remember that it was there so I didn't change after our last discussion.

IntegratedQuantum: Well, I care, and so do you apparently, otherwise you wouldn't have made this.

Argmaster: Boy scout rule. We already have issue for that #920

IntegratedQuantum: That's specifically about file paths, which should be specialized.

Argmaster: Here #1595

src/server/command/tp.zig
```diff
@@ -6,93 +6,89 @@ const User = main.server.User;
 pub const description = "Teleport to location.";
 pub const usage = "/tp <x> <y>\n/tp <x> <y> <z>\n/tp <biome>";
 
+const Args = union(enum) {
+	@"/tp <x> <y> <z>": struct {@"<x>": f64, @"<y>": f64, @"<z>": ?f64},
```
IntegratedQuantum: I do not like how much noise the surrounding `@"<...>"` adds to the parameters, could we maybe apply the angle brackets at the location it is used?

src/argparse.zig
```diff
@@ -0,0 +1,424 @@
...
+
+				for(std.meta.fieldNames(field.type)) |name| {
+					messageSegments.append(allocator, " ");
+					messageSegments.append(allocator, name);
+					maxSize += name.len;
+				}
+
+				const message = std.mem.concat(allocator.allocator, u8, messageSegments.items) catch unreachable;
+				result.failure.messages.append(allocator, message);
+				result.failure.takeMessages(allocator, &fieldResult.failure);
```
IntegratedQuantum: You have one member function here, and one function which accesses the messages field directly.
So why not just do the same thing here as well and remove the then unused member function?
```suggestion
				result.failure.messages.appendSlice(allocator, fieldResult.failure.messages.items);
```

src/argparse.zig
```diff
@@ -0,0 +1,424 @@
...
+			inline for(u.fields) |field| {
+				var fieldResult = Parser(field.type, options).resolve(false, allocator, args);
+				defer fieldResult.deinit(allocator);
+
+				if(fieldResult == .success) {
+					result.deinit(allocator);
+					return .{.success = @unionInit(T, field.name, fieldResult.success)};
+				}
+
+				var messageSegments: ListUnmanaged([]const u8) = .{};
```
IntegratedQuantum: Just use a `ListUnmanaged(u8)`, then you don't need the extra concat at the end, instead you could use toOwnedSlice.

src/argparse.zig
```diff
@@ -0,0 +1,424 @@
...
+
+				result.takeSuggestions(allocator, &completion);
+			}
+
+			return result;
+		}
+	};
+}
+
+pub fn ParseResult(comptime SuccessT: type) type {
```
IntegratedQuantum: Honestly, I do not really like this Result pattern. Since there is no language support for types like this, you have these big switches everywhere.

So what would you think about using the out-parameter pattern instead?
```zig
fn resolveArgument(..., errorMessage: *List(u8)) !Field {
...
			return .{.success = std.fmt.parseFloat(std.meta.Float(floatInfo.bits), arg) catch |err| {
				errorMessages.writer().print("Expected a number for {s}, found \"{s}\"\n", .{name, arg});
				return err;
			}};
...
}
```

And at the callsite:
```zig
 // In the future we could even push this and the sending code outside the individual command and into the command system
const errorMessage: List(u8) = .init(main.stackAllocator);
defer errorMessage.deinit();

const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
	send...
	return;
};
```

IntegratedQuantum (changes requested):

src/server/command/tp.zig
Argmaster: > could we maybe apply the angle brackets at the location it is used?

I don't fully understand what you mean by "location it is used", could you post an example?

src/argparse.zig
Argmaster: Out parameters are the worst thing that happened to humanity, relict from Fortran and C, in the first one it was necessity, in the second one it was the laziness that made people do it. I discourage this kind of design regardless of the context, parameters should be the input of the function, not the mix of input and output, out parameters violate basic assumption you can have about parameters. As for support, zig does have support for that, the switch statements and enum unions are much better at dispatching result objects than tools many other languages have.
That said, if you insist, then the name of the parameter must be `outErrorMessage` to clearly indicate that it is out parameter.

src/server/command/tp.zig
IntegratedQuantum: It is used in the error messages, so I would suggest to change them to
```diff
- return .initWithFailure(allocator, utils.format(allocator, "Expected a number for {s}, found \"{s}\"", .{name, arg}));
+ return .initWithFailure(allocator, utils.format(allocator, "Expected a number for <{s}>, found \"{s}\"", .{name, arg}));
```
So that we can name the parameters `x` instead of `<x>`.

src/argparse.zig
IntegratedQuantum: I generally agree when in comes to simple things.
However I think that in the case, writing to a data structure, this is fine. And this is exactly what you are doing: Accumulating error messages in a data structure.

Furthermore, you'll often find this pattern for writing strings in the form of `writer: anytype`, and I'm sure this pattern is used in other languages as well, not just C and Fortran.

Argmaster:
Okay, I applied the requests. I tried returning optional to indicate that it was not possible to parse the string, but that failed miserably, so now I am returning ParseError instead and it seems to work.

src/server/command/tp.zig
Argmaster: I think this will be hard to keep it consistent, but this would be a problem anyway.

codemob-dev:
Is this still active?

IntegratedQuantum:
It is waiting for review

Argmaster:
Closing until integrating such feature becomes a priority.

thecreare:
>  Expected a number at argument 1 found "f" ← Note that normal people start counting at 1

To appease both parties, programmers and ""normal people"", we should follow the [GulfOfMexico](https://github.com/TodePond/GulfOfMexico?tab=readme-ov-file#arrays) standard and start counting at -1

src/argparse.zig
```diff
@@ -0,0 +1,427 @@
...
+pub const Options = struct {
+	commandName: []const u8,
+};
+
+pub fn Parser(comptime T: type, comptime options: Options) type {
+	return struct {
+		const Self = @This();
+
+		pub fn parse(allocator: NeverFailingAllocator, args: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!T {
+			return resolve(false, allocator, args, errorMessage);
```
IntegratedQuantum: Please avoid bool parameters, instead use an enum which is more readable.

src/argparse.zig
```diff
@@ -0,0 +1,427 @@
...
+				else => @compileError("Only structs and unions are supported"),
+			}
+		}
+
+		fn resolveStruct(comptime doAutocomplete: bool, comptime s: std.builtin.Type.Struct, allocator: NeverFailingAllocator, args: []const u8, errorMessage: *ListUnmanaged(u8)) if(doAutocomplete) AutocompleteResult else error{ParseError}!T {
+			var result: T = undefined;
+			var split = std.mem.splitScalar(u8, args, ' ');
+
+			var tempErrorMessage: ListUnmanaged(u8) = .{};
+			defer tempErrorMessage.deinit(allocator);
```
IntegratedQuantum: All local allocations should use the stackAllocator

@codemob-dev we need the fix for #1286

src/argparse.zig
```diff
@@ -0,0 +1,427 @@
...
+				result.suggestions.append(allocator, allocator.dupe(u8, id[arg.len..]));
+			}
+			return result;
+		}
+	};
+}
+
+const Test = struct {
+	var testingAllocator = main.heap.ErrorHandlingAllocator.init(std.testing.allocator);
+	var allocator = testingAllocator.allocator();
```
IntegratedQuantum: also not needed with #1286

src/argparse.zig
```diff
@@ -0,0 +1,427 @@
...
+
+			tempErrorMessage.appendSlice(allocator, "---");
+
+			inline for(u.fields) |field| {
+				tempErrorMessage.append(allocator, '\n');
+				tempErrorMessage.appendSlice(allocator, field.name);
+				tempErrorMessage.append(allocator, '\n');
+
+				const result = Parser(field.type, options).resolve(false, allocator, args, &tempErrorMessage);
+				if(result != error.ParseError) {
```
IntegratedQuantum: Better use catch of the corresponding if else construct for errors.

IntegratedQuantum (changes requested):
Alright, one thing is for sure I don't have the time and patience to review a 500 line PR.
First up instead of using `/tp`, I'd suggest to use a simpler command such as `/time` for demonstration. `/time` is a lot simpler but still has enough cases to be interesting.
With that the biome code is no longer needed, and I think it should be added in a future PR instead.

src/argparse.zig
codemob-dev: > @codemob-dev we need the fix for https://github.com/PixelGuys/Cubyz/issues/1286

Not again...

Argmaster:
Aight, there is nothing I can do about the allocator, so that is still pending, but I have removed tp and biome changes and implemented an example based on time command. 
That exposed some issues in field name concatenation, so I tried to improve how this is handled. 
I also noticed an issue with handling an empty argument list and/or trailing whitespace, so I also handled those in a more human friendly way.
I agnowledge certain weirdness in how messages for enum values are formulated - descriptions don't explicitly mention names of the arguments, while the error messages do.  However, there has to be a time to say enough and I think this is it.

image

Argmaster:
Honestly I am no longer sure this approach will suffice for our needs; It might be better to go with something like a PEG parser instead of a rigid struct with whitespace split based logic; current implementation will fail whenever we try to input eg. zon text with strings containing whitespace and that will be necessary for eg. particles. Any thoughts?

IntegratedQuantum:
Couldn't we just use parenthesis when any argument requires spaces, just like you do it in the terminal?

Argmaster:
Well, we could, but `std.mem.splitScalar` will still ignore those, right?
So we still need to switch to scanning+feedback to acomplish that.

IntegratedQuantum:
We can make our own iterator that is quote-aware

Argmaster:
Ok, maybe its not that bad. But it will have to be quote, parenthesis, brace and curly brace aware (and I wonder what else we decide to have as arguments). Realistically not something I would want to add now.

Apart from that, what else is blocking this PR? Don't tell me its testing allocator changes.

IntegratedQuantum:
yes indeed, it is the testing allocator

Argmaster:
I have updated the PR with latest test allocator changes.

src/argparse.zig
```diff
@@ -0,0 +1,386 @@
...
+const ListUnmanaged = main.ListUnmanaged;
+const utils = main.utils;
+
+pub const Options = struct {
+	commandName: []const u8,
+};
+
+const ResolveMode = enum {
+	Parse,
+	Autocomplete,
```
IntegratedQuantum: enum literals should be lowercase

src/server/command/time.zig
```diff
@@ -4,35 +4,54 @@ const main = @import("main");
...
+					main.server.world.?.doGameTimeCycle = false;
+					source.sendMessage("#ffff00Time stopped.", .{});
+					return;
+				},
+			}
+		},
+	};
+
+	main.server.world.?.gameTime = gameTime;
 	source.sendMessage("#ffff00{}", .{main.server.world.?.gameTime});
```
IntegratedQuantum: I think only the pure `/time` command should echo the time to you.

src/server/command/time.zig
```diff
@@ -4,35 +4,54 @@ const main = @import("main");
...
-pub const usage = "/time\n/time <day/night>\n/time <time>\n/time <start/stop>";
+pub const usage =
+	\\/time
+	\\/time <time>
+	\\/time <day/night>
+	\\/time <start/stop>"
+;
+
+const Args = union(enum) {
+	@"/time <phase>": struct { phase: enum { day, night } },
```
IntegratedQuantum, line 15: Shouldn't the names match the usage above?

src/argparse.zig
```diff
@@ -0,0 +1,386 @@
...
+	return struct {
+		const Self = @This();
+
+		/// Parse the string `args` according to the schema defined in `T` type parameter of the Parser.
+		/// Result is returned from this function as a value of type `T`.
+		///
+		/// Arguments:
+		/// - `allocator` - will be used for dynamic allocations of the parsing result returned.
+		/// - `args` - unprocessed string containing command arguments without command name.
+		/// - `errorMessage` - out parameter used to store and return errors, if any occur. Has to be allocated with stackAllocator.
```
IntegratedQuantum, line 27: Wouldn't it make sense to use `allocator` as well for `errorMessage`?

src/argparse.zig
```diff
@@ -0,0 +1,386 @@
...
+			comptime s: std.builtin.Type.Struct,
+			allocator: NeverFailingAllocator,
+			args: []const u8,
+			errorMessage: *ListUnmanaged(u8),
+		) switch (mode) {
+			.Autocomplete => AutocompleteResult,
+			.Parse => error{ParseError}!T,
+		} {
+			var result: T = undefined;
+			var split = std.mem.splitScalar(u8, args, ' ');
```
IntegratedQuantum, line 67: I think we should add a custom split iterator that handles all whitespaces, and automatically trims excess whitespaces between arguments. This would also avoid all the extra logic below to handle leftover whitespaces at the end.

Since this would require a custom iterator, I'd suggest to wait for it until the next PR, so we can get this through at some point.

src/argparse.zig
```diff
@@ -0,0 +1,386 @@
...
+					}
+				} else {
+					@field(result, field.name) = value catch unreachable;
+					tempErrorMessage.clearRetainingCapacity();
+					nextArgument = split.next();
+				}
+			}
+
+			const isLeftoverArgumentWhitespace = nextArgument == null or main.utils.all(std.ascii.isWhitespace, nextArgument.?);
+			const areUnusedArgumentsWhitespace = main.utils.all(std.ascii.isWhitespace, split.rest());
```
IntegratedQuantum: We don't even handle whitespaces properly when splitting. For now I'd prefer to just remove all this extra logic (and the corresponding `all` function) and handle it properly in the next PR (see comment above)

src/argparse.zig
```diff
@@ -0,0 +1,386 @@
...
+
+			if (!isLeftoverArgumentWhitespace or !areUnusedArgumentsWhitespace) {
```
IntegratedQuantum:
```suggestion
			if (nextArgument != null) {
```

src/argparse.zig
```diff
@@ -0,0 +1,386 @@
...
+					if (argument == null) return error.ParseError;
+					return resolveArgument(optionalInfo.child, allocator, name, argument, errorMessage) catch |err| {
+						return err;
+					};
+				},
+				inline .@"struct" => {
+					const arg = argument orelse {
+						failWithMessage(errorMessage, missingArgumentMessage, .{name});
+						return error.ParseError;
+					};
```
IntegratedQuantum: I'd prefer to treat optional as a separate case, then do this check, and then handle the other options in a switch to avoid repeating this everywhere. Then you also can inline the constant for the message

src/argparse.zig
```diff
@@ -0,0 +1,386 @@
...
+
+				result.takeSuggestions(allocator, &completion);
+			}
+
+			return result;
+		}
+	};
+}
+
+fn failWithMessage(errorMessage: *ListUnmanaged(u8), comptime fmt: []const u8, args: anytype) void {
```
IntegratedQuantum: Instead I'd suggest to introduce an equivalent to `List(u8).print` to the unmanaged variants.

src/argparse.zig
```diff
@@ -0,0 +1,386 @@
...
+}
+
+test "float negative" {
+	var errors: ListUnmanaged(u8) = .{};
+	defer errors.deinit(main.stackAllocator);
+
+	const result = Test.OnlyX.parse(main.stackAllocator, "foo", &errors);
+
+	try std.testing.expectError(error.ParseError, result);
+	try std.testing.expect(errors.items.len != 0);
```
IntegratedQuantum: Use expectEqual for better messages on failure

IntegratedQuantum (changes requested):

src/argparse.zig
Argmaster: What do you mean?

Argmaster: I think I know what you mean.

Argmaster: And the answer is no.
If you ever want to use stack allocator for error messages anywhere down the call stack, you have to use it everywhere, because parsing is recursive and you cannot tell if you got a `tempErrorMessage` (line 73) or one that came from `parse` call.
We can make `errorMessage` not be an out argument and use result struct pattern and use `allocator` then, but at that point it would be a full cycle from what originally was implemented and what you rejected.

Argmaster: [And you want to use stack allocator for tempErrorMessage](https://github.com/PixelGuys/Cubyz/pull/1425#discussion_r2596270022)

IntegratedQuantum: Hmm, yeah I guess the other options are worse.

src/server/command/time.zig
Argmaster: Well, its a little awkward and impractical to list everything (3 times) like that...
What usually happens is the arguments and subcommands offer attaching help messages and then the full help/error message is composed out of those when necessary.
We could try something like this using struct level constants maybe?
Whatever we decide, out of scope, I will change it for now as requested.

src/argparse.zig
Argmaster: > I'd suggest to wait for it until the next PR, so we can get this through at some point.

Sorry, but this sentence confuses me. It this PR (#1425) blocked until a PR that introduces mentioned better iterator appears, or not?

Some more thoughts on iterator; just whitespace is not enough, we need matching for quotes `"`, `'`, parenthesis `(`, `)`, brackets `[`, `]`, curly brackets `{`, `}`, and support for escape sequences `\"`, `\'` that break simple counting, that's from the top of my head, to support zon and string messages.

IntegratedQuantum: This PR is not blocked, I was saying you should do the iterator in the next PR after this one is merged.

src/argparse.zig
Argmaster: I think this is not enough; `splitScalar`, when provided with empty string, will return an empty string first time `next()` is called, so we have to check if what we got from `next()` is empty string, or we won't be able to call `/time` with no arguments.

src/server/command/time.zig
IntegratedQuantum: > Well, its a little awkward and impractical to list everything (3 times) like that...

That is a good point. We already have argument names (and possible values) in the struct. Now that I think about it, it makes no sense to manually specify `@"/time <phase>"` at all, we could just call it `phase` directly and save the surrounding noise, and the empty case could be called `get`.

src/utils.zig
```diff
@@ -2242,6 +2242,11 @@ pub fn obfuscateString(allocator: NeverFailingAllocator, string: []const u8) []c
 	return obfuscated;
 }
 
+pub fn all(comptime predicate: anytype, sequence: anytype) bool {
```
IntegratedQuantum: unused

src/argparse.zig
```diff
@@ -0,0 +1,420 @@
...
+	}
+
+	pub fn deinit(self: AutocompleteResult, allocator: NeverFailingAllocator) void {
+		for (self.suggestions.items) |item| {
+			allocator.free(item);
+		}
+		self.suggestions.deinit(allocator);
+	}
+};
+
```
IntegratedQuantum: Since this is a big files, please add a `MARK` to mark the beginning of the tests.

src/argparse.zig
```diff
@@ -0,0 +1,420 @@
...
+					return std.fmt.parseInt(std.meta.Int(intInfo.signedness, intInfo.bits), arg, 0) catch {
+						errorMessage.print(main.stackAllocator, "Expected an integer for <{s}>, found \"{s}\"", .{name, arg});
+						return error.ParseError;
+					};
+				},
+				inline else => |other| @compileError("Unsupported type " ++ @tagName(other)),
+			}
+		}
+
+		const missingArgumentMessage = "Missing argument at position <{s}>";
```
IntegratedQuantum: Only used once, can be inlined

src/argparse.zig
```diff
@@ -0,0 +1,420 @@
...
+
+		fn autocompleteArgument(comptime Field: type, allocator: NeverFailingAllocator, _arg: ?[]const u8) AutocompleteResult {
```
IntegratedQuantum: This function isn't called anywhere, maybe we should postpone the autocomplete functions for now until they are actually used/tested? You can keep the enum and everything else, since Zig won't complete about the missing functions.

IntegratedQuantum (changes requested):
Some small things, then we can merge this.

src/argparse.zig
Argmaster: I will remove most of the implementation, but I would prefer not remove the function signatures, `resolve()` integration and type declaration to leave the framework there, so the dual use of `resolve()` is not lost in changes.

src/server/command/time.zig
Argmaster: No can do, they are used to form error messages, so we would basically have to change how we define help messages to eliminate the duplication / allow some other way to specify what to put in error messages, out of scope

image

src/argparse.zig
IntegratedQuantum: Yes, please keep the rest, just remove two functions autocompleteArgument and autocompleteArgument function, all the rest should work fine without it.

src/server/command/time.zig
Argmaster: I think `const help="/time <day/night>"` placed inside struct would be a good way to specify how the command should be displayed in help:

```zig

const Args = union(enum) {
	@"phase": struct { const help = "/time <day/night>"; phase: enum { day, night } },
	@"/time <start/stop>": struct { subcommand: enum { start, stop } },
	@"/time <number>": struct { number: i64 },
	@"/time": struct {},
};
```

Argmaster: Actually, now I remember why it was `phase`, `subcommand` etc - look at the message. I don't think it makes sense to change it. help message deduplication still a problem.

IntegratedQuantum: > so we would basically have to change how we define help messages to eliminate the duplication / allow some other way to specify what to put in error messages, out of scope

I think we should do that, but yeah out of scope

IntegratedQuantum (approved):

Wunka:
Question for this. Is there a reason `source: *User` is not given to the parser? For example when I want to add `Axis` parsing ( ~ or value ) I need the `source` position. I could of course go by a two step thingy (first get if it a ~ or/and a position by the parser and then make the actual position out of that) but that was not the goal of the parser right?

Argmaster:
Argument parser is for parsing strings, not a do-everything-machine.
For ~ I would suggest to use a special enum value that you turn into coordinate in the command code.

Argmaster:
Actually a dedicated struct would make more sense, with offset field. There should be a custom struct parsing interface if we didn't remove it to shrink the PR

Wunka:
Yes there is a parser function. I already have done it with the two step solution, just wanted to ask if there is a reason.

Argmaster:
We could consider what would be the implications of passing Context object into there... Maybe in the future it will be strictly required for auto complete to access some engine runtime data?

Wunka:
~~I think it would be nice for e.g `playerIndex` parsing (the actual reason I asked, Axis was just the better example) auto complete could show you all the online players as options~~ scratch that, you don't need the context for that

IntegratedQuantum:
If we find more use-cases it might make sense to reconsider, but if it's only for position it's fine to handle this on the outside with an extra function call.

Issue #1030, opened 2025-02-10
Vine rotation mode


careeoki:
- Requires support
- Can only be placed on the underside of blocks
- Can only be placed downwards

Ideally:
- Trying to place vines horizontally will place vines downwards, like dropping down a spool of rope.
- There can be seperate textures/models for the last block in the chain of a vine

codemob-dev:
I think a more generic name, `hangingPlant` could be better.

codemob-dev:
<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/e17187ad-8ac9-473c-8bf2-3e1fa9cfb68d" />

codemob-dev:
I don't think it is possible to make the whole vine break if the support block is destroyed, but it works well otherwise.

Argmaster:
I would suggest even more generic name: `hanging`
Yes, it is not possible without chained block updates.

PR #1440, opened 2025-05-13
Revamped Swamp Trees [WIP]

ikabod-kee:
image

There are currently a few issues that need resolved before this is merged
- [ ] For better variation, structures need random rotation support
- [x] #1030 
- [ ] Roots need to be created for this tree type
- [ ] Fixed format check


Argmaster:
Reswamped wamp?

Argmaster:
Can you update this and just merge this as is? Random rotation will be the default after #1529 so you won't need to do any changes for that and do the rest in separate pull request? You already did the assets and you will have to change them for #1496, but this will take some more time, so we can just take what we have. Small incremental changes are generally good.

ikabod-kee:
I have to close this due to format changes in SBB and the new log rotation.

PR #1442, opened 2025-05-13
Add Surface Caves

ikabod-kee:
image

Temporary surface caves with clay patches, in preparation for #1151


careeoki:
I think it should literally just be stone, gravel, and clay (with primarily stone). The soil/dirt make it look too much like the surface.
The clay patches could be bigger on average, like 3 width + 4 variation or something.
copyImage
copyImage

PR #1447, opened 2025-05-15
Add sign assets

careeoki:
Sign models + textures since I had to set up the UVs anyway
image
image
Block zons not included! I only made the oak sign for testing.


careeoki:
Well... I didn't actually mean to do a PR on the branch, I guess I just called mine the same thing. 
I don't think it matters either way since master doesn't need these yet anyway?

IntegratedQuantum (approved):
> I don't think it matters either way since master doesn't need these yet anyway?

Yeah it's fine, it even saves me a rebase. Either way thanks for adding these so quickly!

PR #1456, opened 2025-05-17
Lower coal ore health

Argmaster:
As discussed during playtest, coal ore should be lowered, both because irl its soft and because it is annoying to mine in game.

### DIFF
--- a/assets/cubyz/blocks/coal_ore.zig.zon
+++ b/assets/cubyz/blocks/coal_ore.zig.zon
@@ -1,6 +1,6 @@
 .{
 	.tags = .{.stone},
-	.blockHealth = 60,
+	.blockHealth = 16,
 	.blockResistance = 1,
 	.ore = .{
 		.veins = 5,


IntegratedQuantum (approved):

Argmaster:
The problem is that even top-top-top pickaxe doesn't one-shot coal ore (actually, its not even close!), wooden pickaxe is ~1.5 dmg, stone is ~5 dmg, best ones you can make are around ~20 dmg if they are supposed to have reasonable swing times. Lowering the resistance would reduce the effective HP for wooden pickaxe from 100HP to 60HP, for stone from 72HP to 60HP, but doesn't fix late game annoyance with coal being hard to mine, in early game no one really minds, stuff is supposed to be mined slowly with weak tools. Also resistance does have the effect of cutting off tiers of tools that have too low damage, sure, we can lower it, but it doesn't change anything, either way wooden tools will be able to mine it, but now they have to mine through 24HP, with resistance 0 they would have to mine through 16HP. Let's adjust parts that solve our issue, not everything.

PR #1498, opened 2025-05-24
Merge `sbb` and `blueprint` asset directories

Argmaster:
## Description

This pull request merges `sbb` and `blueprint` directory trees into single directory tree as discussed in #1403 searched twice, once to load blueprint files, once to load sbb zon files.

## Links

Related to: #1403
Related to: #1408


Argmaster:
Quick PR so I can start working on inline SBBs :3

IntegratedQuantum (approved):

PR #1500, opened 2025-05-24
Allow direct use of blueprints as SBBs

Argmaster:
Resolves: #1403

### DIFF
--- a/src/server/terrain/structure_building_blocks.zig
+++ b/src/server/terrain/structure_building_blocks.zig
@@ -135,6 +135,24 @@ pub const StructureBuildingBlock = struct {
 		}
 		return self;
 	}
+	pub fn initInline(sbbId: []const u8) !StructureBuildingBlock {
+		const blueprints = blueprintCache.get(sbbId) orelse {
+			std.log.err("['{s}'] Could not find blueprint '{s}'.", .{sbbId, sbbId});
+			return error.MissingBlueprint;
+		};
+		if(blueprints[0].childBlocks.len != 0) {
+			std.debug.assert(blueprints[1].childBlocks.len == 0);
+			std.debug.assert(blueprints[2].childBlocks.len == 0);
+			std.debug.assert(blueprints[3].childBlocks.len == 0);
+			std.log.err("['{s}'] Inline structures cannot contain child blocks.", .{sbbId});
+			return error.InlineStructureCannotContainChildBlocks;
+		}
+		return .{
+			.id = sbbId,
+			.children = &.{},
+			.blueprints = blueprints,
+		};
+	}
 	pub fn getBlueprint(self: StructureBuildingBlock, rotation: Degrees) *BlueprintEntry {
 		return &self.blueprints[@intFromEnum(rotation)];
 	}
@@ -194,10 +212,28 @@ pub fn registerSBB(structures: *Assets.ZonHashMap) !void {
 		}
 	}
 	{
+		var iterator = blueprintCache.iterator();
+		while(iterator.next()) |entry| {
+			const blueprintId = entry.key_ptr.*;
+			const blueprints = entry.value_ptr.*;
+
+			if(structureCache.contains(blueprintId)) continue;
+			if(blueprints[0].childBlocks.len != 0) continue;
+
+			const value = StructureBuildingBlock.initInline(blueprintId) catch |err| {
+				std.log.err("Could not register inline structure building block '{s}' ({s})", .{blueprintId, @errorName(err)});
+				continue;
+			};
+			const key = arenaAllocator.dupe(u8, blueprintId);
+			structureCache.put(arenaAllocator.allocator, key, value) catch unreachable;
+			std.log.debug("Registered inline structure building block: '{s}'", .{blueprintId});
+		}
+	}
+	{
 		for(childrenToResolve.items) |entry| {
 			const parent = structureCache.getPtr(entry.parentId).?;
 			const child = structureCache.getPtr(entry.structureId) orelse {
-				std.log.err("Could not find child structure '{s}' for child resolution.", .{entry.structureId});
+				std.log.err("Could not find child structure nor blueprint '{s}' for child resolution.", .{entry.structureId});
 				continue;
 			};
 
--- a/assets/cubyz/sbb/tree/birch/1/root/1.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/root/1.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/root/1",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/root/2.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/root/2.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/root/1",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/root/3.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/root/3.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/root/1",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/small_branch/1.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/small_branch/1.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/small_branch/1",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/small_branch/2.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/small_branch/2.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/small_branch/2",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/small_branch/3.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/small_branch/3.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/small_branch/3",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/stub/1.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/stub/1.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/stub/1",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/stub/2.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/stub/2.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/stub/2",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/stub/3.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/stub/3.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/stub/3",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/stub/4.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/stub/4.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/stub/4",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/top/1.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/top/1.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/top/1",
-	.children = .{},
-}
--- a/assets/cubyz/sbb/tree/birch/1/top/2.zig.zon
+++ b/assets/cubyz/sbb/tree/birch/1/top/2.zig.zon
@@ -1,4 +1,0 @@
-.{
-	.blueprint = "cubyz:tree/birch/1/top/2",
-	.children = .{},
-}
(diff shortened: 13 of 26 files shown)


src/server/terrain/structure_building_blocks.zig
```diff
@@ -135,6 +135,24 @@ pub const StructureBuildingBlock = struct {
 		}
 		return self;
 	}
+	pub fn initInline(sbbId: []const u8) !StructureBuildingBlock {
+		const blueprints = blueprintCache.get(sbbId) orelse {
+			std.log.err("['{s}'] Could not find blueprint '{s}'.", .{sbbId, sbbId});
+			return error.MissingBlueprint;
+		};
+		if(blueprints[0].childBlocks.len != 0) {
+			std.debug.assert(blueprints[1].childBlocks.len == 0);
```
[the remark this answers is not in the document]
Argmaster: Hmm right I messed up the conditions

src/server/terrain/structure_building_blocks.zig
```diff
@@ -135,6 +135,24 @@ pub const StructureBuildingBlock = struct {
 		}
 		return self;
 	}
+	pub fn initInline(sbbId: []const u8) !StructureBuildingBlock {
+		const blueprints = blueprintCache.get(sbbId) orelse {
+			std.log.err("['{s}'] Could not find blueprint '{s}'.", .{sbbId, sbbId});
+			return error.MissingBlueprint;
+		};
+		if(blueprints[0].childBlocks.len != 0) {
+			std.log.err("['{s}'] Inline structures cannot contain child blocks.", .{sbbId});
```
IntegratedQuantum, line 144: An addon creator likely has no clue what "inline structure" is supposed to mean. It would be better to tell them what to do instead, something like "Please create a {s}.zig.zon file to be able to use child blocks."

src/server/terrain/structure_building_blocks.zig
```diff
@@ -135,6 +135,24 @@ pub const StructureBuildingBlock = struct {
...
+			return error.InlineStructureCannotContainChildBlocks;
```
IntegratedQuantum, line 145: Should this even be fatal? I think it might be useful to have these structures still spawn in the world in case the creator just wants to take a quick look how it fits in the world before actually defining the children.

IntegratedQuantum (changes requested):

src/server/terrain/structure_building_blocks.zig
Argmaster: If you insist. If I allow that I need to generate a dummy SBB for every blueprint that doesn't have an SBB with an identical ID. This makes on-demand generation of inline SBBs actually important optimization.

IntegratedQuantum: Couldn't you just not do the return error here and continue as you would normally?

Argmaster: Also I would have to check if pasting code doesn't crash as a result, since it is blindly indexing into child Alias tables which there will be none of.

IntegratedQuantum: Isn't that something that needs checking anyways? What does it do currently if a child is not bound and the file does exist?

src/server/terrain/structure_building_blocks.zig
```diff
@@ -135,23 +139,42 @@ pub const StructureBuildingBlock = struct {
...
+		}
+		std.log.warn("[{s}] Attempting to sample child structure from SBB with empty list of possible structures for that child color ({s}).", .{self.id, (block.id())});
+		return null;
 	}
 };
 
-fn initChildTableFromZon(parentId: []const u8, colorName: []const u8, colorIndex: usize, zon: ZonElement) !AliasTable(Child) {
-	if(zon == .null) return .init(arenaAllocator, &.{});
+fn initChildTableFromZon(parentId: []const u8, colorName: []const u8, colorIndex: usize, zon: ZonElement) !?AliasTable(Child) {
+	if(zon == .null) return null;
```
Argmaster, line 170: AliasTable with empty list panics.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -135,11 +139,27 @@ pub const StructureBuildingBlock = struct {
...
+			.blueprints = blueprints,
+		};
+	}
 	pub fn getBlueprint(self: StructureBuildingBlock, rotation: Degrees) *BlueprintEntry {
 		return &self.blueprints[@intFromEnum(rotation)];
 	}
 	pub fn pickChild(self: StructureBuildingBlock, block: BlueprintEntry.StructureBlock, seed: *u64) ?*const StructureBuildingBlock {
-		return self.children[block.index].sample(seed).structure;
+		if(self.children.len == 0) {
+			std.log.warn("[{s}] Attempting to sample child structure from SBB with no children defined.", .{self.id});
```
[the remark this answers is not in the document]
Argmaster: That is how we represent inline SBBs. I guess I could allocate empty AliasTables for them too if current solution is problematic.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -150,8 +170,8 @@ fn initChildTableFromZon(parentId: []const u8, colorName: []const u8, colorIndex
 		return .init(arenaAllocator, &.{});
 	}
 	if(zon.array.items.len == 0) {
-		std.log.err("['{s}'->'{s}'] Empty children list.", .{parentId, colorName});
-		return error.EmptyChildrenList;
+		std.log.warn("['{s}'->'{s}'] Empty children list.", .{parentId, colorName});
```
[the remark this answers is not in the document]
Argmaster: IMO we should allow multiple ways of specifying that structure has no children, as it is not intuitive why any of them would be incorrect.
We would have:

```zig
.{
	.blueprint = "cubyz:tree/birch/1/root/1",
	.children = .{},
}
```

```zig
.{
	.blueprint = "cubyz:tree/birch/1/root/1",
}
```

And inline form where there is no zon file at all.

Technically there is also no reason to disallow this:

```zig
.{
	.blueprint = "cubyz:tree/birch/1/root/1",
	.children = .{
        .blue = .{},
    },
}
```

Of course best one is the shortest one, but we should explain that in a wiki, not in one-line error message. In worst case scenario user will have to get through 4 error messages before they find out that they don't need zon at all. Quite annoying, especially if you are tinkering with something and you are changing the same configuration often.

src/server/terrain/structure_building_blocks.zig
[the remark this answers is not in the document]
Argmaster: > Also this should probably be an error, unless you have a valid use case where it could be useful?

You created the use case via https://github.com/PixelGuys/Cubyz/pull/1500#discussion_r2106165601
If we want to properly handle case where blueprint contains child blocks and the sbb doesn't define use for then we need this if statement or something that would handle such case.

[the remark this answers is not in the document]
Argmaster: Children list is fixed size because we are using indexes corresponding to different colors of children to index it, I can't just remove this alias table, I must either make it nullable, or use `sampleSafe` or something else.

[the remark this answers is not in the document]
Argmaster: Inline SBBs (the whole point of this PR) have empty list of children.

src/server/terrain/structure_building_blocks.zig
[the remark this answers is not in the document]
IntegratedQuantum: > IMO we should allow multiple ways of specifying that structure has no children, as it is not intuitive why any of them would be incorrect.

In my opinion we should only have one option to say discard child bocks, and that's explicitly stating for each color that the list of entries should be empty, so:
```
	.children = .{
		.blue = .{},
		.red = .{},
	},
```
Furthermore it should error if you mention a child block color that is not present in the actual structure.

And of course if there are no ssb child blocks, then we don't care, and we should allow the simple option to just have no children list.

> Quite annoying, especially if you are tinkering with something and you are changing the same configuration often.

That's why none of these errors should be breaking, during tinkering you should be able to safely ignore the little popups.

src/server/terrain/structure_building_blocks.zig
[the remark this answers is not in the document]
Argmaster: Ok, should I make StructureBuildingBlock a union then?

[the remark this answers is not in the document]
IntegratedQuantum: No, just don't add their SBBs to the child block list in the `BlueprintEntry`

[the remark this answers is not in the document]
Argmaster: But this is not how it works, BlueprintEntry detects the list of child blocks on its own based on the blueprint contents and it will be shared between multiple SBBs if multiple reference it, so I can't just throw away some of it's child blocks unless I copy all 4 rotated blueprint entries first. Eg. stems are used this way where you have one stem but multiple uses with different children.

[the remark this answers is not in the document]
IntegratedQuantum: Well then maybe we should have one copied BlueprintEntry per SBB. It wouldn't need to be a deep copy anyways.

Also I still think we shouldn't allow reusing a blueprint like that in the first place, but we've already discussed this.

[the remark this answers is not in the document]
Argmaster: Ok, I did what you asked for. I also had to move around some more assets to clean up errors in assets that already exist in cubyz and were reusing blueprints (eg. `cubyz:generator`)

assets/cubyz/biomes/grassland.zig.zon
```diff
@@ -48,7 +48,7 @@
 		},
 		.{
 			.id = "cubyz:sbb",
-			.structure = "cubyz:tree/oak/grassland_generator",
+			.structure = "cubyz:tree/oak/generator",
```
IntegratedQuantum, line 51: Why did you merge the grassland and forest generators? I think they were supposed to behave differently.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -120,10 +124,14 @@ pub const StructureBuildingBlock = struct {
 			std.log.err("['{s}'] Missing blueprint field.", .{stringId});
 			return error.MissingBlueprintIdField;
 		};
-		const blueprints = blueprintCache.get(blueprintId) orelse {
+		const blueprintsPtr = blueprintCache.get(blueprintId) orelse {
```
IntegratedQuantum, line 127:
```suggestion
		const blueprintsTemplate = blueprintCache.get(blueprintId) orelse {
```

src/server/terrain/structure_building_blocks.zig
```diff
@@ -133,8 +141,57 @@ pub const StructureBuildingBlock = struct {
...
+		for(self.children, 0..) |child, index| {
+			if(child.items.len == 0) continue;
+
+			var found = false;
+			for(self.blueprints[0].childBlocks) |blueprintChild| {
+				if(blueprintChild.index != index) continue;
+				found = true;
+				break;
+			}
+			if(found) break;
```
IntegratedQuantum, line 157: Wait, so it just ignores errors for all children that are past the first found one?

Also please use labelled break/continue instead of bools

src/server/terrain/structure_building_blocks.zig
```diff
@@ -133,8 +141,57 @@ pub const StructureBuildingBlock = struct {
...
+					if(index == 0) std.log.err("['{s}'] Missing child structure configuration for child '{s}'", .{self.id, child.id()});
+					continue;
+				}
+				childBlocks.append(main.stackAllocator, child);
+			}
+			blueprint.childBlocks = arenaAllocator.dupe(BlueprintEntry.StructureBlock, childBlocks.items);
+		}
+	}
+	pub fn initInline(sbbId: []const u8) !StructureBuildingBlock {
+		const blueprintsPtr = blueprintCache.get(sbbId) orelse {
```
IntegratedQuantum, line 175:
```suggestion
		const blueprintsTemplate = blueprintCache.get(sbbId) orelse {
```

src/utils.zig
```diff
@@ -2149,3 +2149,10 @@ test "NamedCallbacks registers functions" {
 	try std.testing.expectEqual(null, testFunctions.getFunctionPointer("functionTest"));
 	try std.testing.expectEqual(null, testFunctions.getFunctionPointer("wrongSignatureFunction"));
 }
+
+var panicMessageBuffer: [1 << 17]u8 = undefined;
+
+pub fn panicWithMessage(comptime fmt: []const u8, args: anytype) void {
+	const message = std.fmt.bufPrint(&panicMessageBuffer, fmt, args) catch unreachable; // Please don't overflow the buffer :3
```
IntegratedQuantum, line 2156: Can we not use allocPrint to avoid any overflow?

src/server/terrain/structure_building_blocks.zig
```diff
@@ -253,7 +332,18 @@ pub fn registerBlueprints(blueprints: *Assets.BytesHashMap) !void {
...
+	const structure = structureCache.getPtr(stringId) orelse return null;
+	const blueprint = structure.blueprints[0];
+	for(blueprint.childBlocks) |child| {
+		if(structure.children.len == 0) {
+			std.log.err("[{s}] Blueprint contains child blocks but there are no child structures defined for it.", .{structure.id});
+			break;
+		}
+		if(structure.children[child.index].items.len == 0) {
+			std.log.err("[{s}] Blueprint contains child block '{s}' but there are no child structures to fill it.", .{structure.id, child.id()});
+		}
```
IntegratedQuantum, line 344: Aren't these cases already handled above in `updateBlueprintChildLists`?

IntegratedQuantum (changes requested):

assets/cubyz/biomes/grassland.zig.zon
Argmaster: Only difference was in ratios between the children, they were slightly.

src/server/terrain/structure_building_blocks.zig
```diff
@@ -193,16 +247,38 @@ pub fn registerSBB(structures: *Assets.ZonHashMap) !void {
 			std.log.debug("Registered structure building block: '{s}'", .{entry.key_ptr.*});
 		}
 	}
+	{
+		var iterator = blueprintCache.iterator();
+		while(iterator.next()) |entry| {
+			const blueprintId = entry.key_ptr.*;
```
IntegratedQuantum, line 253: If you only want the key, you should us the keyIterator

IntegratedQuantum (changes requested):
One more thing then it is good to be merged

src/server/terrain/structure_building_blocks.zig
Argmaster: Fixed.

IntegratedQuantum (approved):

PR #1509, opened 2025-05-25
Use comptime modding for rotation loading

### LINKED ISSUE #1533: [MODS] Load rotations from mods directory
Rotation modes should be loaded from `rotations` directory inside `mods/<mod-name>/`. Ideally we would have same approach as with assets where built-in rotations would also be listed in a mod called `cubyz`.

OneAvargeCoder193:
Related to: #1507
closes #1533 

I'm pretty sure that the rest of the moving into a comptime mod interface can be done in a future pr


OneAvargeCoder193:
I marked this as ready for review since we have the basic implementation set up already, we just need to translate other parts of the code into the new modding interface in seperate prs, since that would probably be too much for 1 pr.

src/rotation.zig
```diff
@@ -196,8 +196,15 @@ pub fn deinit() void {
 
 pub fn getByID(id: []const u8) *RotationMode {
 	if(rotationModes.getPtr(id)) |mode| return mode;
-	std.log.err("Could not find rotation mode {s}. Using no_rotation instead.", .{id});
-	return rotationModes.getPtr("no_rotation").?;
+
+	if(!std.mem.containsAtLeastScalar(u8, id, 1, ':')) blk: {
+		const namespacedId = std.fmt.allocPrint(main.stackAllocator.allocator, "cubyz:{s}", .{id}) catch unreachable;
```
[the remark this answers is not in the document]
IntegratedQuantum: > Actually, maybe we could allow assets from the same namespace as the rotation to skip the namespace part of ID? Since they share the namespace anyway.

This is bad. This means you cannot just copy a file to a different addons, something that any new addon creator will likely do to get started.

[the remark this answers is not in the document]
Argmaster: Then we should be consistent and use full `namespace:object_name` and accept the consequences this implies for eg. commands.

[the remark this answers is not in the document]
Argmaster: There is also an option to have some files that allow users to alias IDs to some other values dynamically, but again - consistency - applicable to every ID space or to none.

build.zig
```diff
@@ -101,6 +144,7 @@ pub fn build(b: *std.Build) !void {
 		//.use_llvm = false,
 	});
 	exe.root_module.addImport("main", exe.root_module);
+	try addModFeatures(b, exe);
```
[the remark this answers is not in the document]
IntegratedQuantum: Create a custom build step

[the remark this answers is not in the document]
OneAvargeCoder193: done

OneAvargeCoder193:
I finished converting the last of the old rotation mode ids i forgot about.

build.zig
```diff
@@ -79,6 +79,78 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+	defer file.close();
+
+	try file.writeAll(out.items);
+	out.deinit(allocator);
+}
+
+pub fn addModFeatureStep(b: *std.Build, name: []const u8) !*std.Build.Step {
+	const step = try b.allocator.create(std.Build.Step);
+	step.* = std.Build.Step.init(.{
+		.id = .custom,
```
Argmaster: Can the ID of the step repeat? It will if we load multiple things?

build.zig
```diff
@@ -79,6 +79,78 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
 	}
 }
 
+pub fn makeModFeature(step: *std.Build.Step, _: std.Build.Step.MakeOptions) anyerror!void {
+	const b = step.owner;
+	const allocator = b.allocator;
```
Argmaster: aaaaa aliasing allocator in function how dare you!

build.zig
```diff
@@ -79,6 +79,78 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
 	}
 }
 
+pub fn makeModFeature(step: *std.Build.Step, _: std.Build.Step.MakeOptions) anyerror!void {
+	const b = step.owner;
```
Argmaster: One letter variable, very bad idea.

Argmaster (approved):
Cool, I got what I deserved for insisting that we use `cubyz:` everywhere - I scrolled through those 150 files and they seem to be all replaced correctly. Aside from that, I found few small things which may violate contribution guidelines, but otherwise I think the PR is fine.

build.zig
OneAvargeCoder193: .id isn't a string identifier, it seems to be what kind of step it is. for example, addWriteFIles has a write_file id

build.zig
OneAvargeCoder193: It's called b everywhere else in build.zig, i just recreated it here to make it more consistent with everything else

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
 	}
 }
 
+pub fn makeModFeature(step: *std.Build.Step, _: std.Build.Step.MakeOptions) anyerror!void {
+	var out: std.ArrayListUnmanaged(u8) = .{};
```
```suggestion
	var featureList: std.ArrayListUnmanaged(u8) = .{};
```

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+					addonEntry.name,
+					step.name,
+					rotationEntry.name,
+				},
+			));
+		}
+	}
+
+	const file_path = try std.fs.path.join(step.owner.allocator, &.{"mods", step.owner.fmt("{s}.zig", .{step.name})});
+	var file = try std.fs.cwd().createFile(file_path, .{.truncate = true});
```
IntegratedQuantum: Keep it simple and use writeFile directly, also you don't need to set the truncate flag.

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+	defer file.close();
+
+	try file.writeAll(out.items);
+	out.deinit(step.owner.allocator);
```
IntegratedQuantum: Should be in a defer next to the creation.
Down here it's leaking in 10 places.

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+	var out: std.ArrayListUnmanaged(u8) = .{};
+
+	var assetsDir = try std.fs.cwd().openDir("mods", .{.iterate = true});
+	defer assetsDir.close();
+
+	var iterator = assetsDir.iterate();
+	while(try iterator.next()) |addonEntry| {
+		if(addonEntry.kind != .directory) continue;
+
+		var addon = try assetsDir.openDir(addonEntry.name, .{});
```
```suggestion
		var mod = try assetsDir.openDir(addonEntry.name, .{});
```

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
 }
 
+pub fn makeModFeature(step: *std.Build.Step, _: std.Build.Step.MakeOptions) anyerror!void {
+	var out: std.ArrayListUnmanaged(u8) = .{};
+
+	var assetsDir = try std.fs.cwd().openDir("mods", .{.iterate = true});
+	defer assetsDir.close();
+
+	var iterator = assetsDir.iterate();
+	while(try iterator.next()) |addonEntry| {
```
```suggestion
	while(try iterator.next()) |modEntry| {
```

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
 	}
 }
 
+pub fn makeModFeature(step: *std.Build.Step, _: std.Build.Step.MakeOptions) anyerror!void {
+	var out: std.ArrayListUnmanaged(u8) = .{};
+
+	var assetsDir = try std.fs.cwd().openDir("mods", .{.iterate = true});
```
```suggestion
	var modDir = try std.fs.cwd().openDir("mods", .{.iterate = true});
```

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+	defer assetsDir.close();
+
+	var iterator = assetsDir.iterate();
+	while(try iterator.next()) |addonEntry| {
+		if(addonEntry.kind != .directory) continue;
+
+		var addon = try assetsDir.openDir(addonEntry.name, .{});
+		defer addon.close();
+
+		var rotationDir = addon.openDir(step.name, .{.iterate = true}) catch continue;
```
```suggestion
		var featureDir = addon.openDir(step.name, .{.iterate = true}) catch continue;
```

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+		defer rotationDir.close();
+
+		var rotationIterator = rotationDir.iterate();
```
```suggestion
		var featureIterator = rotationDir.iterate();
```

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+		while(try rotationIterator.next()) |rotationEntry| {
```
```suggestion
		while(try rotationIterator.next()) |featureEntry| {
```

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+			if(rotationEntry.kind != .file) continue;
+			if(!std.mem.endsWith(u8, rotationEntry.name, ".zig")) continue;
+
+			try out.appendSlice(step.owner.allocator, step.owner.fmt(
+				"pub const @\"{s}:{s}\" = @import(\"{s}/{s}/{s}\");\n",
```
IntegratedQuantum: Use multiline string to get rid of all the escape characters

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+		.id = .custom,
+		.name = name,
+		.owner = b,
+		.makeFn = makeModFeature,
+	});
+	return step;
+}
+
+pub fn addModFeature(b: *std.Build, exe: *std.Build.Step.Compile, name: []const u8) !void {
+	exe.step.dependOn(try addModFeatureStep(b, name));
```
IntegratedQuantum: Other steps, like testing should also depend on this.
To structure this I'd suggest to make a separate mod initialization step that depends on all the individual feature steps.

build.zig
```diff
@@ -79,6 +79,75 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+
+	const module = b.createModule(.{
+		.root_source_file = b.path(try std.fs.path.join(b.allocator, &.{"mods", b.fmt("{s}.zig", .{name})})),
+		.target = exe.root_module.resolved_target,
+		.optimize = exe.root_module.optimize,
+	});
+	module.addImport("main", exe.root_module);
+	exe.root_module.addImport(name, module);
```
IntegratedQuantum, line 130: other steps, like testing, also need the import.

IntegratedQuantum (changes requested):
Please, whenever you copy code, check if all the variable names still make sense.

build.zig
OneAvargeCoder193: So instead of creating a seperate step for each feature, I just make one step for all the features?

build.zig
OneAvargeCoder193: The target is passed into the function, I already call it on the testing step.

build.zig
IntegratedQuantum: You could also do both

build.zig
IntegratedQuantum: Why is it called `exe` then?

OneAvargeCoder193: the main program is called exe and the tests are called exe_tests, so i though it would make sense to call it exe

OneAvargeCoder193: More accurately, I call the function on exe_tests the same way I call it on exe

build.zig
```diff
@@ -79,6 +79,76 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+					featureEntry.name[0 .. featureEntry.name.len - 4],
+					modEntry.name,
+					name,
+					featureEntry.name,
+				},
+			));
+		}
+	}
+
+	const file_path = try std.fs.path.join(step.owner.allocator, &.{"mods", step.owner.fmt("{s}.zig", .{name})});
```
IntegratedQuantum: Why not use
```suggestion
	const file_path = try step.owner.fmt("mods/{s}.zig", .{name});
```

build.zig
```diff
@@ -79,6 +79,76 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+	try std.fs.cwd().writeFile(.{.data = featureList.items, .sub_path = file_path});
+}
+
+pub fn makeModFeaturesStep(step: *std.Build.Step, _: std.Build.Step.MakeOptions) anyerror!void {
+	try makeModFeature(step, "rotation");
+}
+
+pub fn addModFeatureModule(b: *std.Build, exe: *std.Build.Step.Compile, name: []const u8) !void {
+	const module = b.createModule(.{
+		.root_source_file = b.path(try std.fs.path.join(b.allocator, &.{"mods", b.fmt("{s}.zig", .{name})})),
```
IntegratedQuantum: same here

build.zig
```diff
@@ -79,6 +79,76 @@ fn linkLibraries(b: *std.Build, exe: *std.Build.Step.Compile, useLocalDeps: bool
...
+		}
+	}
+
+	const file_path = try std.fs.path.join(step.owner.allocator, &.{"mods", step.owner.fmt("{s}.zig", .{name})});
+	defer step.owner.allocator.free(file_path);
+
+	try std.fs.cwd().writeFile(.{.data = featureList.items, .sub_path = file_path});
+}
+
+pub fn makeModFeaturesStep(step: *std.Build.Step, _: std.Build.Step.MakeOptions) anyerror!void {
```
IntegratedQuantum: Could you move this below the `addModFeatures` function, I think it would helpful to have both usage of `"rotation"` and all the future strings closer to each other to make it easier to see that new features need to be added to both lists.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):
Alright, let's do this.
First mod support 🎉

PR #1582, opened 2025-06-02
Experimental Tool Rebalance

ikabod-kee:
This PR changes how pickaxe stats are calculated by combining flat sums and averages. Elasticity and Mass is no longer a factor for Swing Speed; pickaxes now have a flat mining speed that can be changed by modifiers. All ingots have been given the heavy modifier at different rates.


Argmaster:
Okay, so how did it go? Did you try it? Why did you add penalty for using better materials? What sps did you get with these changes? I am pretty sure that whatever changes we do, we should be able to make pickaxe that has 7 block breaks / second when breaking stone or other common filler stone types. That should include amber and good at modifier, but still.

ikabod-kee:
A full gold pickaxe gives me ~1.8 swings per second. Default swing speed is 3 swings per second.

ikabod-kee:
There is still much tweaking to do to get the desired results. Do keep in mind that accessories will make our swing speeds faster.

ikabod-kee:
This PR is more about future balance rather than Pre-Alpha balance

Argmaster:
I am annoyed with everyone staying that "oh no accessories will fix everything". Then hardcode values accessories should do and check if that is true. Either make a thought experiment or create table of what effects would accessories have to give to actually reach 7 bps (blocks per second). This is 233% increase of swing time + at least 20 (?) damage. What combination of materials will give you that? It should not be one magical combination of materials, rather something simple and fairly common in mid-late game, upper caps should be still higher if somebody wants to go for some min-maxing of design. Mind that replacing non-linear (averaging) with linear relation (addition) removes a lot if not all clever tricks we previously to get those min-maxed pickaxes. 
This is a excel spreadsheet task, not a programming task.

assets/cubyz/tools/pickaxe.zig.zon
```diff
@@ -17,81 +17,68 @@
 	.parameters = .{
 		.{
 			.source = .density,
-			.destination = .swingTime,
+			.destination = .damage,
```
Argmaster, line 20: You are converting density into damage twice, first time here, second time below. Choose one logical / good conversion instead. Mixing one that is bad with one that is good doesn't make sense, this introduces additional complexity without improving the outcome.

ikabod-kee:
I would appreciate if you could experiment with me by changing around the files!

assets/cubyz/tools/pickaxe.zig.zon
```diff
@@ -17,81 +17,68 @@
...
-				1.0, 1.5, 1.5, 0x0, 0x0,
+				1.0, 1.0, 1.0, 0x0, 0x0,
+				1.0, 1.0, 1.0, 0x0, 0x0,
 				1.0, 0x0, 0x0, 1.0, 0x0,
 				0x0, 0x0, 0x0, 0x0, 0.5,
 			},
-			.factor = 120,
-			.method = .average,
+			.factor = 15,
+			.method = .sum,
```
Argmaster, line 55: This is not logical. Mixing low durability material with low durability material doesn't magically create even more durable material, unless this is a very specific mix of materials.

Argmaster:
I don't see a point in changing the files. We can just do the math, it will have same effect without extra effort put into programming.

ikabod-kee:
Did you try it in-game yet?

Argmaster:
Ofc no I am here to complain 😜

ikabod-kee:
Try before you buy!

Argmaster:
What I noticed:
- Now there is no reason to do stone tipped pickaxes, because there is no tradeoff, you can only gain by using heavier material.
- Swing time of 0.3 for sickle it quite high, tool itself should be light, maybe default SPS should be set per tool type?
- Heavy modifier should be additive with no diminishing returns
- Difference between stone and copper tools is marginal - 1.32 vs 1.17 to break stone, 111 vs 119 blocks durability for minimal pickaxe, for maximal it is .99 vs .82 and 239 vs 281
- Best iron pickaxe you can make is just a bit off to one-short stone, very annoying.
- Specialized pickaxe with good at stone is as good at breaking stone as universal pickaxe with silver encased diamond for breaking stone, but the universal one is well, more universal. There is no reason to make specialized pickaxes, unless you want to min-max your resource use.
- Both universal pickaxe and specialized for stone pickaxe have BPS roughly equal to 2.3 which would need to be increased 300% to reach comfortable 7 BPS and 478% to reach even more comfortable 11 BPS which is reachable in m-word game.
How exactly do you plan to tackle that? What will have to be the numbers on accessories / materials to reach those values?

ikabod-kee:
Thank you for your feedback. I can see this is probably a step in the right direction, but I just need to get my values right. Calculating any of this makes my brain melt into goop.

Argmaster:
As I said, this is a spreadsheet task, just changing stuff randomly will not get us anywhere, there are too many possibilities, too many dependencies. Let me find my spreadsheet.

Argmaster:
Here, I did [spreadsheet](https://docs.google.com/spreadsheets/d/18RpA5kY1B9zlNR3WBnQ8TOUGhLAhScg9nOCq_5UX-H8/edit?usp=sharing) for current state for chosen pickaxe designs, I will do one for that PR soon.

ikabod-kee:
Yeah, I have no idea how to read that. I'll at least try to explain how my balancing is meant to be.

Argmaster:
Oh its dead simple really, you have few different designs of pickaxes and few different block types you check those designs against. For those designs I have exported metrics we already used before in discussions ie. 

- block damage (how much damage pickaxe x does to block y) 
- swings per block (how many swings of pickaxe X you need to break one block of type Y) 
- swings per second (literally how many swings pickaxe X can do per second)
- blocks per second (how many blocks Y can pickaxe X break in a second)
- total blocks breakable (how many blocks Y can pickaxe X break with its durability if it was only breaking block X)

Pickaxe designs checked are described in spreadsheet.
Order of columns is unlucky, but this is what Data Wrangler exported so it is what it is.

Argmaster:
Please use numbers for your balance goals, otherwise it is too easy to say things which are not technically possible due to math.

ikabod-kee:
I'm going to close this. I'm not in the right headspace today and I don't think I can comprehend the math behind it. I'll just stick to other things. I will make an issue though in hopes that someone else can handle it.

Issue #1326, opened 2025-04-21
[Gem] Amber


Argmaster:
This is a proposal to add **amber** to the game.

## Physical properties:
- hardness in Mohs scale: 2-2.5
- density (at 20° C): 1.1 g/cm3

## In game properties:
- behavior similar to other gemstones - provide special effect instead of stats

## Modifiers
- `light` #172 

## Spawn rules
- On beaches as an ore on the sand.

Related to: #1269

careeoki:
You should also be able to find amber in sandstone caves, or below deserts (hopefully, with #556). Amber on beaches should be pretty rare.

PR #1585, opened 2025-06-03
Amber ore & blocks

careeoki:
part of #1326
image
Amber starts generating at -1250. I would still want it to appear in sandstone caves, beaches, etc. but we don't support that right now.

image
Amber has the `light` modifier at 50%

image
Something unique compared to other gem blocks is that amber is translucent.


ikabod-kee:
Heck yes. In the future I can imagine amber only being available in certain cave biomes, namely ones made out of sedimentary rock

PR #1586, opened 2025-06-03
Tool Rebalance Proposal Mk.II

ikabod-kee:
This draft covers some issues I had with tool balancing in an attempt to make it more clear to the player what different materials pros and cons are. Here is what I got:

Wooden Pickaxe: Its role is to be the starting pickaxe, thus it breaks easily and is quickly replaced.
image

Copper Pickaxe: The next tier after stone (which I will be modifying later on in this draft.) Its role is to act as the metal of choice between the stone and iron age, and also works as a nice filler metal for damage and durability, as it's the lightest metal.
image

Iron Pickaxe: Its stats are extremely middling, providing decent damage and durability at the cost of a swing time penalty. 
image

Silver Pickaxe: Silver acts as a middle-ground between iron and gold. Its durability is slightly lower, but does provide a bit more damage than can cross the threshold on mining certain ores instantly. It also allows for gem and stone encasing!
image

Gold Pickaxe: Gold is the next tier. It is heavy, making it the perfect material for some extra damage; though it has the drawbacks of having slow swing times and low hardness. It's better to use gold as a filler material rather than it be THE material of choice for the pickaxe.
image

Uranium Pickaxe: The cream of the crop metal, uranium, is lighter than gold and harder than iron. It is the best metal as its benefits outweigh its drawbacks. This material is a reward for the player for managing to make it so deep within the world and living to tell the tale.
image

Tipping:
Tipping in theory should still work. Here's an iron pickaxe, then another with a uranium tip.
image
image

So, what's going on internally?
- All materials give a flat increase to the tool, period. There's no multipliers or averaging. It's all summing. I can re-add multipliers in select spots if need be.
- Elasticity was removed from the equation entirely.
- Swing speed is reliant on modifiers rather than density.

What still needs to be done?
- Balancing the other materials like stone, gems, and etc.
- Tools besides pickaxes.
- Slight adjustments to block hardness to account for balance changes.
- Durability balancing
- Grip. The best feature in any video game ever.
- Maybe remove diminishing returns on certain modifiers?? Big maybe.

Why?
- Depth comes from simplicity, not complexion, so having a unified system for this helps keep it from becoming over-tuned and beginner unfriendly.

No, but why make the swing times like, a modifier instead of a stat?
- It's better for controlling the balance of swing-times, which is currently very confusing. If you can find a better way of doing it, please do tell.

I would really like to hear your opinions on these, or perhaps even help **build onto the idea.** With a **pull request** maybe. I know I'm not a programmer, or a mathematician; I'm not really qualified to even make this pull request, but I can't help but feel that the whole system is heading in a bad direction, and I just want things to be simple for the sake of the players. After all, it's not just a game for developers and mathematicians, it's also for artists and gamers.

Also @Argmaster, try to let some other people chime in first.


ikabod-kee:
I'm going to increase swing speeds very shortly, as they are quite slow.

Argmaster:
> Also @Argmaster, try to let some other people chime in first.

Me expressing my opinions does not stop anyone else from expressing their opinion, so I'll just ignore this comment.

You have said it yourself that you are not a programmer, nor a mathematician, so why are you instantly jumping into the code? Take a damn piece of paper and scratch what pickaxes should be possible, what stats they roughly should have and reality check them later. Then you can incorporate features which are not yet implemented. We will be validating the idea through stats and metrics, not through manually checking all possible pickaxes anyway.

vertigofilip:
I think the multiplier would be a good idea. Handle area might have higher multiplayer for swing time, and lower to durability and damage, working edge (depending on tool) higher multiplier on damage, and rest higher multiplayer on durability.

ikabod-kee:
> @vertigofilip: I think the multiplier would be a good idea. Handle area might have higher multiplayer for swing time, and lower to durability and damage, working edge (depending on tool) higher multiplier on damage, and rest higher multiplayer on durability.

The tool's tip does get an exclusive stat bonus, which takes the material's hardness value and applies it to the pickaxe.
Also, I was thinking the same thing for handle vs tool head; instead of it being sort of weirdly linear, it could just be a hard separation! Thank you for your input.

> @Argmaster: You have said it yourself that you are not a programmer, nor a mathematician, so why are you instantly jumping into the code? Take a damn piece of paper and scratch what pickaxes should be possible, what stats they roughly should have and reality check them later. Then you can incorporate features which are not yet implemented. We will be validating the idea through stats and metrics, not through manually checking all possible pickaxes anyway.

Sorry for the call-out to you specifically. I feel like it's hard to get my thoughts down on paper, and easier for me to just modify files we already have and see instant results. As for not-yet-implemented features, such as the ever-amazing grip, it's more of a gut-feeling than anything.

ikabod-kee:
Also, explaining my most recent commit: I changed the default swing times back to 0.25s as to not completely overshadow the light modifier.

ChibChi:
IMO the tool creation is already pretty confusing. I didn't even know there was a 2.5 at the top of the pickaxe tool head. Personaly I would split the weights into handle, tool head and maybe tip, with even weights. Multipliers should be reserved for gems.

ikabod-kee:
New commit! Now the pickaxe has 3 parts: tool head, tool tip, and tool handle. Handles get a 2x multiplier for durability, but a 0.5x multiplier for damage.

ikabod-kee:
This new change should account for damage lost from Grip™. If not, I'll try a 3x. So far, so good! Pickaxes feel very nice to use now.

ikabod-kee:
New commit!
1. I've added the brittle modifier, which is calculated separately from the fragile modifier, but is virtually the same. This is so stone's brittleness and diamond's fragileness can stack.
2. Tool tips are now worth x0.33 durability when it comes to their hardness, to avoid durable, hard materials such as uranium from giving loads of durability.
3. Stone is now brittle, again. I know this was a controversial feature before it was removed, but this time it has more purpose: to balance out the new encased stone modifier.
4. Encased stone will not be brittle if surrounded by 4 precious materials (which I may change to metals in general.) It will remain brittle if not. This was put into place to prevent stacking with amber, and to make the pickaxe more situational.

Here's some pickaxe examples:
image
In this one, I tried making the best pickaxe for breaking stone without the use of uranium. Pretty worth it, I'd say! Note how heavy and light interact, essentially creating a sort of threshold. Gold could be used, but it would make it slower. There's strategy to this.

image
Also notice here that I'm sacrificing durability in the name of being able to insta-break stone. 

image
Switching around stone and amber will remove the brittleness, but also remove the lightness, creating a pickaxe that does indeed insta-break stone, but at a slower rate than it would've, all at the cost of increased durability.

Now, for non-specialized pickaxes!
image
The standard stone pickaxe will net you 52 stone blocks (yes, I counted!) So it's not useless! In fact, if you want to save copper and iron up, you will always be able to find stone around to use for stone pickaxes. This is why I made stone fragile, because it's so overwhelmingly abundant. Don't worry, I've learned from last time.

image
Here's a standard copper pickaxe, capable of breaking ~180 stone blocks at a faster rate, even with the slower swing speed, as it takes only 4 swings to break instead of 5. So it takes 1 second to break a block with a stone pickaxe, while it takes only 0.92 seconds for the copper pickaxe. Because of this, durability drain rate is also decreased.

image
This is an iron pickaxe. It breaks one stone block in 3 swings, so it breaks the block in a whopping 0.72 seconds! Also less swings means less durability drain, allowing this particular iron pickaxe to break up to 329 stone. Impressive, considering Minecraft's iron pickaxe only has 250 durability.

image
Golden pickaxe. One stone in 2 swings, meaning 1 block in 0.62 seconds! It can break up to 359 blocks, instantly break coal, and two-shot iron. Sure, it may have a similar durability to iron, but it's at least better, considering how much is saved from ores.

image
Now, I'm sad to say that the uranium pickaxe does not instantly break blocks. It's just below the damage threshold.

But wait, this is Cubyz!
image
image
Bam, just like that, we pass the 32 damage threshold, allowing us to instantly mine stone with a single swing. 0.29 seconds for one block. That's double the speed of the golden pickaxe, and double the durability too. The only drawback to this? Uranium is extremely rare.. or, well, it's planned to be at least, and I assume dangerous on its own as well. It's easily the best material of all.

-----------------------
But what's this? We're still going? You bet we are.
image
For this one, I just put in random materials. Seems balanced to me. 0.58 seconds per stone block, and you get a whopping 655 stone blocks to wallop with it.

image
Remove all the fluff, now you get 0.56 seconds per stone block, with 486 stone breakable with it. Seems better, though with less durability.

Lets take that optional fluff we used in the first pickaxe and put it in a new one.
image
0.75 seconds per stone block, 320 blocks breakable. If we combine the last pickaxe's effective durability with this one, we get 806 durability.
So this means that making these two pickaxes separately is more effective, right? It's more durability.
But you'd be wrong, because making those two pickaxes separately has made you use 3 additional copper ingots. Plus, if we take the mean effective durabilities of the first and last pickaxes, you get 815 effective durability, so technically it would've been more efficient to keep all that fluff. All this to say that optional slots are extremely balanced, and the player won't feel like their resources are wasted on using those slots.

------------------
That all being said, I feel pretty confident in this PR so far. I still have much to do, such as adjusting the other stones. Diminishing returns on modifiers have seemed to work in my favor after all!

Argmaster:
- If everything requires 4 items to encase then there is much less possibility of mixing and matching stuff, instead of small changes  by mixing multiple things we get huge leaps from few materials. 
  Yes, you balanced the system, by removing most of its complexity.
  Basically best you can do is use as much gold as possible, cool, "simplicity" is there.
- Making uranium the best material is completely absurd, uranium is just barely less dense than gold, I find it as intuitive as obsidian in m word game.
- You overshot the damage by far, I can easily get 75 dmg pickaxe, only block that is close to that damage requirement is uranium ore.
- Amber is atm most OP material, it exponentially increases performance of the tool.
- Behold new most powerful pickaxe in Cubyz at 8.33bps, insta-breaking every block in game, not even that heavy:
image
- Wait, why would I invest so many resources If I can just make a god wooden pickaxe:
image
- Wait! I found even more simplicity! That is 11 bps! cool. Who needs diamonds anyway.
image
- Oh I can make the last one out of uranium instead of wood, this is only .01 difference in swing time and +1000 durability, that is a fair tradeoff!
image
- With that tooling I don't need caves, I can just make them myself!

Anyway, how does that scale onto other tools exactly? Will we have 70 damage axes to break 8HP logs?

ikabod-kee:
I can agree that I've overshot damage. However, I didn't know that you could encase amber like that. I was under the impression that it had to be perpendicular to the amber. XD That's fine though, I'll get that fixed for ya.

ikabod-kee:
> Oh I can make the last one out of uranium instead of wood, this is only .01 difference in swing time and +1000 durability, that is a fair tradeoff!

To be fair, this is before grip is implemented. You probably won't want to be holding uranium by the handle, so the grip on it would definitely be very very negative.

Anyways, as for amber, I'll see what I can do about it.

ikabod-kee:
>You overshot the damage by far, I can easily get 75 dmg pickaxe, only block that is close to that damage requirement is uranium ore.

So, this is actually a funny thing that I forgot I did. I made the damage high to account for over-mining, if that's ever added.

Argmaster:
Sorry but at that point you are wasting my time, you are balancing stuff for not approved features and without any transparency, re-introducing bugs we are already either aware about or ones we removed. I will not be reviewing that until we come to an agreement on the tool balance with transparent and written spec for that purpose, math included. Maybe Quantum will review that. Good luck.

ikabod-kee:
You chose to waste your own time. I appreciate the feedback though, it honestly helped.

ikabod-kee:
Alright, I did some nerfing.
- Damage and durability overall got a nerf, as well as tip and handle durability multipliers. The highest damage achievable with modifiers is now 55-60.
- Handles no longer contribute to damage.
- Diamonds now need 5 precious metals surrounding it to rid it of its fragile modifier. The powerful modifier still requires 4 precious metals.
- Stone can now be surrounded by 4 of any metal to rid it of its brittle nature, but still needs 2 precious metals.

ikabod-kee:
@Argmaster revealed a few flaws. Some of them are fixable.
> It is unclear how modifiers stack, what is the output damage from mixing them, without taking a calculator,

I do agree with this, though with #1591 it could be remedied somewhat, or we could make modifiers non-diminishing.
> They produce enormous tooltips.

This can be remedied by combining tooltips that change the same stat, but also by making certain stats like the "heavy modifier" I applied to metals be part of the parameters. I wanted to do this all along, but due to a lack of programming skills, I'm left with having to use modifiers. Also, modifier descriptions can be shortened.
```
[Light] +50% speed
[Heavy] -50% speed
[Specialized] +50% damage to Stone
[Inept] -50% damage to Stone
[Durable] +50 durability
[Brittle] -50% durability
[Fragile] -50% durability
[Powerful] +50% damage
[Weak] -50% damage
[Disposable] Durability set to 100
```
> It is still the same problem of what values different pickaxes should have, just using different means to edit the values.

Yes. Another limitation on my part, but at the same time it does at least make things more visual for the player.

My main goal for this is to present a different way of balancing, where the tool grid is like a puzzle the player has to figure out without having confusing averaged values. Ideally, the experience of making an average tool should be easy and consistent, while making specialized tools should be more like a puzzle, letting the player experiment with different combinations of materials in order to fit in all the modifiers they want.

I think I would like to discuss this live with @Argmaster and @IntegratedQuantum at some point, so we can combine our brains together to create a definitive system.

ikabod-kee:
image
image

Doing stuff like this, trying to figure out how to make the best pickaxe for mining stone is seriously fun.

IntegratedQuantum:
> Doing stuff like this, trying to figure out how to make the best pickaxe for mining stone is seriously fun.

Well, let me just slap some gold on it, and bam, the pickaxe has +50% damage against stone and +66% more damage against anything else, than yours. And it's faster and more durable too.
Screenshot at 2025-06-04 15-40-36

In my opinion it's not fun to optimize if the solution is so obvious. If density is the only thing that really affects damage, besides the tip, then you just put gold into every slot add some gems and call it a days.

And there isn't even a real downside, you can even get 0.1 s time to swing with just a little amber:
Screenshot at 2025-06-04 15-32-00


Another thing I really don't like, which was much better with averaging, there is no reason not to use all the optional slots. Take this random iron pickaxe for example
Screenshot at 2025-06-04 15-37-35

As long as you got any leftover wood, or even garbage sulfur, it's just plain better to fill the optional slots with it:
Screenshot at 2025-06-04 15-49-35

ikabod-kee:
You actually make a good point there with the optional slots and density acting as a damager. I could make the optional slots contribute no stats, or I could make the optional slots not optional at all.

ikabod-kee:
Or somehow make each optional slot filled have diminishing returns??

ikabod-kee:
image
I made all the slots required now.
Now you *have* to use filler! Enjoy~

ikabod-kee:
I may have just accidentally done this to the wrong file. Oops

ikabod-kee:
Should be fixed now!

ikabod-kee:
Honestly if we can't get procedural tools down, we might have to just bite the bullet and do a tinkers' construct system, or even *shutters* crafting recipes.

IntegratedQuantum:
> Now you have to use filler!

Well, that just goes against the spirit of the entire system.

ikabod-kee:
Sigh

IntegratedQuantum:
> Honestly if we can't get procedural tools down, we might have to just bite the bullet and do a tinkers' construct system, or even shutters crafting recipes.

It's not that we can't balance it, in my opinion the current system is quite balanced, it's just that we can't agree how to balance it. And that won't change with any other system.

ikabod-kee:
I guess time will tell.

BoySanic:
Hi,

Is this still wanted? It has been some months since this opened and we'll have to close the PR in a month from now.

ikabod-kee:
Yeah I'll close it. I mostly used this as a discussion

PR #1616, opened 2025-06-07
Stop erasing the old logs from inventories that already have them

OneAvargeCoder193:


OneAvargeCoder193:
I re-added oak_top

IntegratedQuantum (approved):
thanks for fixing this

PR #1629, opened 2025-06-17
Calculate font unit per pixel dynamically

jingkaimori:


Argmaster:
How do UI and Signs react to this? Please post screenshots.
How do you handle dynamic char limits for signs?

jingkaimori:
> How do UI and Signs react to this? Please post screenshots.

This pr should not change outlook of UI or Signs.
图片


> How do you handle dynamic char limits for signs?

IMO this pr is not related to this feature.

Argmaster:
Well, if you change char width dynamically it implicitly changes how many chars can be rendered in single line, right?

jingkaimori:
> if you change char width dynamically it implicitly changes how many chars can be rendered in single line

No, this pr affects font file only, no visual change is made, and width of glyph(character) is not changed.

Argmaster:
Okay, so this basically dynamically squishes/stretches the chars? In such case what I said earlier does not apply. It was my misunderstanding of the matter.

jingkaimori:
> this basically dynamically squishes/stretches the chars?

No, only outer bounding box (`x_advance` returned by `fg_shape`) is squished/stretched, displayed part of glyph is not affected

src/graphics.zig
```diff
@@ -1149,6 +1149,7 @@ const TextRendering = struct { // MARK: TextRendering
 	var textureWidth: i32 = 1024;
 	const textureHeight: i32 = 16;
 	var textureOffset: i32 = 0;
+	var fontUnitPerPixel: f32 = undefined;
```
```suggestion
	var fontUnitsPerPixel: f32 = undefined;
```

src/graphics.zig
```diff
@@ -1176,6 +1177,7 @@ const TextRendering = struct { // MARK: TextRendering
 		try ftError(hbft.FT_Set_Pixel_Sizes(freetypeFace, 0, textureHeight));
 		harfbuzzFace = hbft.hb_ft_face_create_referenced(freetypeFace);
 		harfbuzzFont = hbft.hb_font_create(harfbuzzFace);
+		fontUnitPerPixel = @as(f32, @floatFromInt(freetypeFace.*.units_per_EM))/@as(f32, @floatFromInt(textureHeight));
```
IntegratedQuantum: I don't think you need to dereference here.
```suggestion
		fontUnitPerPixel = @as(f32, @floatFromInt(freetypeFace.units_per_EM))/@as(f32, @floatFromInt(textureHeight));
```

IntegratedQuantum (changes requested):
Thanks for looking into it, I wondered where that value came from, I think I only found it through trial and error.

src/graphics.zig
jingkaimori: error reported if access without dereferrence:
```
type '[*c]cimport.struct_FT_FaceRec_' does not support field access
```

IntegratedQuantum: Oh, right I guess it doesn't work for C pointers.

IntegratedQuantum (approved):
thanks

PR #1635, opened 2025-06-20
QoL changes for creative inventory

Argmaster:
Although major UI changes are planned, they are expected in rather far future. Unfortunately current state of creative inventory makes it really frustrating to use when building and when testing newly added blocks. Therefore I have implemented a dead simple search bar which filters out all items with IDs that don't contain the searched phrase. I also made a sin of (subjectively) improving the look of item list by padding last row of item list with empty immutable slots if it was not filled with items found.

Screenshot 2025-06-20 025239

Screenshot 2025-06-20 025312

Screenshot 2025-06-20 025245

### DIFF
--- a/src/gui/windows/creative_inventory.zig
+++ b/src/gui/windows/creative_inventory.zig
@@ -30,8 +30,6 @@ var items: main.List(Item) = undefined;
 var inventory: Inventory = undefined;
 var searchInput: *TextInput = undefined;
 var searchString: []const u8 = undefined;
-var searchSelectionStart: ?u32 = null;
-var searchCursor: ?u32 = null;
 
 fn lessThan(_: void, lhs: Item, rhs: Item) bool {
 	if(lhs == .baseItem and rhs == .baseItem) {
@@ -48,8 +46,6 @@ fn lessThan(_: void, lhs: Item, rhs: Item) bool {
 
 pub fn onOpen() void {
 	searchString = "";
-	searchSelectionStart = null;
-	searchCursor = null;
 	initContent();
 }
 
@@ -66,8 +62,6 @@ fn initContent() void {
 		const label = Label.init(.{0, 3}, 56, "Search:", .right);
 
 		searchInput = TextInput.init(.{0, 0}, 288, 22, searchString, .{.callback = &filter}, .{});
-		searchInput.cursor = searchCursor;
-		searchInput.selectionStart = searchSelectionStart;
 
 		row.add(label);
 		row.add(searchInput);
@@ -113,9 +107,6 @@ fn initContent() void {
 }
 
 fn deinitContent() void {
-	searchSelectionStart = searchInput.selectionStart;
-	searchCursor = searchInput.cursor;
-
 	if(window.rootComponent) |*comp| {
 		comp.deinit();
 	}
@@ -129,9 +120,13 @@ pub fn update() void {
 }
 
 fn filter(_: usize) void {
+	const cursor = searchInput.cursor;
+
 	main.globalAllocator.free(searchString);
 	searchString = main.globalAllocator.dupe(u8, searchInput.currentString.items);
 	deinitContent();
 	initContent();
+
+	searchInput.cursor = cursor;
 	searchInput.select();
 }
--- a/src/gui/windows/creative_inventory.zig
+++ b/src/gui/windows/creative_inventory.zig
@@ -50,45 +50,52 @@ pub fn onOpen() void {
 }
 
 fn initContent() void {
-	const list = VerticalList.init(.{padding, padding + 16}, 140, 0);
+	const root = VerticalList.init(.{padding, padding}, 300, 0);
 	{
+		const list = VerticalList.init(.{0, padding + padding}, 48, 0);
 		const row = HorizontalList.init();
-		const label = Label.init(.{0, 2}, 74, "Search: ", .right);
-		searchInput = TextInput.init(.{0, 0}, 284, 22, searchString, .{.callback = &filter}, .{});
+		const label = Label.init(.{0, 3}, 56, "Search:", .right);
+		searchInput = TextInput.init(.{0, 0}, 288, 22, searchString, .{.callback = &filter}, .{});
 		row.add(label);
 		row.add(searchInput);
 		list.add(row);
+		list.finish(.center);
+		root.add(list);
 	}
-
-	items = .init(main.globalAllocator);
-	var itemIterator = main.items.iterator();
-	while(itemIterator.next()) |item| {
-		if(searchString.len != 0 and std.mem.indexOf(u8, item.id(), searchString) == null) continue;
-		items.append(Item{.baseItem = item.*});
-	}
-
-	std.mem.sort(Item, items.items, {}, lessThan);
-	const slotCount = items.items.len + (slotsPerRow - items.items.len%slotsPerRow);
-	inventory = Inventory.init(main.globalAllocator, slotCount, .creative, .other);
-	for(0..items.items.len) |i| {
-		inventory.fillAmountFromCreative(@intCast(i), items.items[i], 1);
-	}
-	var i: u32 = 0;
-	while(i < items.items.len) {
-		const row = HorizontalList.init();
-		for(0..slotsPerRow) |_| {
-			if(i > slotCount) break;
-			if(i >= items.items.len) {
-				row.add(ItemSlot.init(.{0, 0}, inventory, i, .immutable, .immutable));
-			} else {
-				row.add(ItemSlot.init(.{0, 0}, inventory, i, .default, .takeOnly));
-			}
-			i += 1;
+	{
+		const list = VerticalList.init(.{0, padding}, 144, 0);
+		items = .init(main.globalAllocator);
+		var itemIterator = main.items.iterator();
+		while(itemIterator.next()) |item| {
+			if(searchString.len != 0 and std.mem.indexOf(u8, item.id(), searchString) == null) continue;
+			items.append(Item{.baseItem = item.*});
 		}
-		list.add(row);
+
+		std.mem.sort(Item, items.items, {}, lessThan);
+		const slotCount = items.items.len + (slotsPerRow - items.items.len%slotsPerRow);
+		inventory = Inventory.init(main.globalAllocator, slotCount, .creative, .other);
+		for(0..items.items.len) |i| {
+			inventory.fillAmountFromCreative(@intCast(i), items.items[i], 1);
+		}
+		var i: u32 = 0;
+		while(i < items.items.len) {
+			const row = HorizontalList.init();
+			for(0..slotsPerRow) |_| {
+				if(i > slotCount) break;
+				if(i >= items.items.len) {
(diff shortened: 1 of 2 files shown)


Argmaster:
image

IntegratedQuantum (changes requested):
I think we should improve the QoL here a bit:
The inventory should refresh every time the text changes.

Argmaster:
> I think we should improve the QoL here a bit: The inventory should refresh every time the text changes.

Sure, I added extra optional callbacks to TextInput for that.

IntegratedQuantum:
> Sure, I added extra optional callbacks to TextInput for that.

Well, in order to reliably do this with events you'd need to handle all events that can modify the string e.g. pasting is missing here.

Why not just go the direct route and check if the string has changed every frame?

Argmaster:
> Well, in order to reliably do this with events you'd need to handle all events that can modify the string e.g. pasting is missing here.
> 
> Why not just go the direct route and check if the string has changed every frame?

UI usually uses event driven design, not polling and even if we can get away with that here, there will be soner or later a request to provide `onChange` cb, this is just so common idiom. But sure, I can go the easy route here.

IntegratedQuantum:
If you think that we'll need this in the future then go for it. But please make sure it actually handles all cases.
And there should be one `onChange` callback, and not multiple for delete/inputCharacter/.. as you've started to implement here.

Argmaster:
> And there should be one `onChange` callback, and not multiple for delete/inputCharacter/.. as you've started to implement here.

Yeah, my mistake, I was lacking full understanding of the problem.

Argmaster:
Ready4Review

IntegratedQuantum (changes requested):
At a glance I see 3 more functions that aren't covered by the onChange event but can change the text. There might be more, so please check all the functions again.

Argmaster:
> At a glance I see 3 more functions that aren't covered by the onChange event but can change the text. There might be more, so please check all the functions again.

I intentionally omitted setString because it can only triggered by external factors, so TextInput API user already has control over that and its prone to causing infinite recursion if we dont inform callback what caused the event. I also intentionally omitted init and deinit, as they would cause infinite recursion. deleteSelection is handled through deleteLeft and deleteRight, I found clear, but I don't see other problems.

Unfortunately current implementation destroys and creates again all the components of the window, which means that if event is not last thing occurring, some actions could be performed post-deinit, but handling this differently quickly makes the change neither simple nor quick. Is it allowed to go through the component structure at runtime and just swap some of them with different ones?

IntegratedQuantum:
> deleteSelection is handled through deleteLeft and deleteRight

In my opinion the event should be triggered in all functions, and exclusively in functions, that modify the actual text buffer. Current approach seems to make it more difficult to track the individual changes, since it's unclear when and why the event is triggered.

> Unfortunately current implementation destroys and creates again all the components of the window, which means that if event is not last thing occurring, some actions could be performed post-deinit

Yeah, that's one of the reasons why I have avoided using events so far.
I think the simplest solution would be to just defer the callback execution to after the render/update loop is finished using a List in the gui system.

> Is it allowed to go through the component structure at runtime and just swap some of them with different ones?

Technically yes, but I do think that is a more complex solution.

Argmaster:
Idk, I think It will be better If I just do polling for now, its simply too problematic to do it with events with current state of UI framework. I do think however that long term we should consider creating more robust and more typical UI model (ie. event driven one). This would not be a requirement if we had just a bunch of fixed parts of UI and no modding, but here we have nearly OS level window system + modding and this requires more flexibility to avoid writing a lot of code and having a lot of bugs.

Argmaster:
> Why not just go the direct route and check if the string has changed every frame?

I assume you mean to check it in `render()`?
Which render? Of creative inventory or of TextInput?

IntegratedQuantum:
> I assume you mean to check it in render()?

You should use the window update function, you may get some glitches when you try to update it in the render function.

Argmaster:
Alright, implemented. Honestly I am really confused about function order here and at the same time I don't want to just shuffle stuff around for funzies (even if it means following contribution guidelines), so I will just wait for your requests if you think they should be in different order.

src/gui/windows/creative_inventory.zig
```diff
@@ -40,38 +45,83 @@ fn lessThan(_: void, lhs: Item, rhs: Item) bool {
...
 	if(window.rootComponent) |*comp| {
 		comp.deinit();
 	}
 	items.deinit();
 	inventory.deinit(main.globalAllocator);
 }
+
+pub fn update() void {
+	if(std.mem.eql(u8, searchInput.currentString.items, searchString)) return;
+	filter(0);
```
IntegratedQuantum: Use undefined if a value isn't used.

src/gui/windows/creative_inventory.zig
```diff
@@ -40,38 +45,83 @@ fn lessThan(_: void, lhs: Item, rhs: Item) bool {
...
+}
+
+fn filter(_: usize) void {
```
IntegratedQuantum, line 122: Also I don't think this needs the extra parameter now. It should work without being a callback for the text field.

src/gui/windows/creative_inventory.zig
```diff
@@ -40,38 +45,83 @@ fn lessThan(_: void, lhs: Item, rhs: Item) bool {
...
+	{
+		const list = VerticalList.init(.{0, padding + padding}, 48, 0);
 		const row = HorizontalList.init();
-		for(0..10) |_| {
-			if(i >= items.items.len) break;
-			row.add(ItemSlot.init(.{0, 0}, inventory, i, .default, .takeOnly));
-			i += 1;
-		}
+		const label = Label.init(.{0, 3}, 56, "Search:", .right);
+		searchInput = TextInput.init(.{0, 0}, 288, 22, searchString, .{.callback = &filter}, .{});
```
IntegratedQuantum, line 64: I think you shouldn't recreate the text field, since it will shuffle around the cursor.

IntegratedQuantum: Or at least copy and reapply the cursor/selection state on update.

src/gui/windows/creative_inventory.zig
```diff
@@ -40,38 +45,83 @@ fn lessThan(_: void, lhs: Item, rhs: Item) bool {
...
+	initContent();
+}
 
-	const list = VerticalList.init(.{padding, padding + 16}, 140, 0);
-	var i: u32 = 0;
-	while(i < items.items.len) {
+fn initContent() void {
+	const root = VerticalList.init(.{padding, padding}, 300, 0);
+	{
+		const list = VerticalList.init(.{0, padding + padding}, 48, 0);
```
IntegratedQuantum, line 60: Why is this second vertical list needed? Can't you just add the HorizontalList to the outer one directly?

src/gui/windows/creative_inventory.zig
```diff
@@ -40,38 +45,83 @@ fn lessThan(_: void, lhs: Item, rhs: Item) bool {
...
+
+fn filter(_: usize) void {
+	main.globalAllocator.free(searchString);
+	searchString = main.globalAllocator.dupe(u8, searchInput.currentString.items);
+	deinitContent();
+	initContent();
+	searchInput.select();
+}
+
+pub fn onClose() void {
```
IntegratedQuantum: I would put this next to onOpen

src/gui/windows/creative_inventory.zig
```diff
@@ -40,38 +45,83 @@ fn lessThan(_: void, lhs: Item, rhs: Item) bool {
...
+		const slotCount = items.items.len + (slotsPerRow - items.items.len%slotsPerRow);
+		inventory = Inventory.init(main.globalAllocator, slotCount, .creative, .other);
+		for(0..items.items.len) |i| {
+			inventory.fillAmountFromCreative(@intCast(i), items.items[i], 1);
+		}
+		var i: u32 = 0;
+		while(i < items.items.len) {
+			const row = HorizontalList.init();
+			for(0..slotsPerRow) |_| {
+				if(i > slotCount) break;
```
IntegratedQuantum: Is this ever true? I don't think so.

src/gui/windows/creative_inventory.zig
Argmaster: If I don't override onNewline cb then it is possible to input LF and that looks confusing in text window tweaked to fit exactly one line.

image

There are other ways to implement that, but what about not spending any more time than necessary at thinking about a simple callback?

src/gui/windows/creative_inventory.zig
Argmaster: Yes, but no, that screws up the paddings for some reason.

src/gui/windows/creative_inventory.zig
Argmaster: Well, deinit is recursive isn't it? So how exactly should i do it?

Argmaster: Anyway, I will just save cursor and selection. Although I can't think of a scenario where you can edit the entry to trigger the cb and keep the selection at once.

src/gui/windows/creative_inventory.zig
```diff
@@ -18,13 +20,18 @@ pub var window = GuiWindow{
...
 };
 
 const padding: f32 = 8;
+const slotsPerRow: u32 = 10;
 var items: main.List(Item) = undefined;
 var inventory: Inventory = undefined;
+var searchInput: *TextInput = undefined;
+var searchString: []const u8 = undefined;
+var searchSelectionStart: ?u32 = null;
+var searchCursor: ?u32 = null;
```
IntegratedQuantum: Instead of doing this with global variables, accessed from three different functions, better just do it all locally in the filter function.

IntegratedQuantum (changes requested):

Argmaster:
I have made `searchCursor` local and I have removed `searchSelectionStart` because only way to trigger `filter()` is by changing content of input and that inevitably removes selection.

IntegratedQuantum:
> and that inevitably removes selection.

That is not true. By the time we handle the change it is possibly already a few keypresses further, so if you do the relevant key presses during a lag spike, then you can select something before we update it here.

Argmaster:
Alr, if you say so.

src/gui/windows/creative_inventory.zig
```diff
@@ -40,38 +45,91 @@ fn lessThan(_: void, lhs: Item, rhs: Item) bool {
...
 		list.add(row);
+		list.finish(.center);
+		root.add(list);
+	}
+	{
+		const list = VerticalList.init(.{0, padding}, 144, 0);
+		items = .init(main.globalAllocator);
+		var itemIterator = main.items.iterator();
+		while(itemIterator.next()) |item| {
+			if(searchString.len != 0 and std.mem.indexOf(u8, item.id(), searchString) == null) continue;
```
IntegratedQuantum: Since you are interested in whether it contains the value and not the index, you should use `std.mem.containsAtLeast`

IntegratedQuantum (changes requested):
One more thing from my part.

@ikabod-kee please confirm that this is in line with the new UI design.

PR #1639, opened 2025-06-24
Pickaxe durability adjustments

Argmaster:
This is my take on pickaxe durability adjustment. Honestly, its neither complete nor precise, but if we never do changes we will never move closer to optimal state.
I have also increased impact of density of handle on swing time to increase impact of not using wood as handle.


Argmaster:
Lets just not do that now.

PR #1662, opened 2025-07-09
Binary storage for inventories

Argmaster:
Related to: #1642
Resolves: #1290


src/Inventory.zig
```diff
@@ -319,7 +321,12 @@ pub const Sync = struct { // MARK: Sync
 			defer mutex.unlock();
 			const inventory = ServerInventory.init(len, typ, source, .externallyManaged);
 			inventories.items[inventory.inv.id] = inventory;
-			inventory.inv.loadFromZon(zon);
+			switch(zon) {
+				.object => inventory.inv.loadFromZon(zon),
+				.string, .stringOwned => |str| inventory.inv.fromBase64(str),
```
IntegratedQuantum: Conversion from base64 should be handled on the outside, in the player data parsing/storing code.
Otherwise we will end up having to add a copy of this function once chests are stored.

src/Inventory.zig
```diff
@@ -364,7 +371,12 @@ pub const Sync = struct { // MARK: Sync
 
 					const inventoryZon = playerData.getChild(@tagName(source));
```
IntegratedQuantum: I think I may have forgot to remove this entire case here, when I introduced the externally managed method above. Let me investigate this.

src/Inventory.zig
```diff
@@ -1996,15 +2008,7 @@ pub fn getAmount(self: Inventory, slot: usize) u16 {
 }
 
 pub fn save(self: Inventory, allocator: NeverFailingAllocator) ZonElement {
```
IntegratedQuantum: Again, in my opinion the base64 stuff should be handled externally, since it's only supposed to be used in one place, for player data.

src/Inventory.zig
```diff
@@ -2024,3 +2028,49 @@ pub fn loadFromZon(self: Inventory, zon: ZonElement) void {
...
+		}
+		defer count -= 1;
+		stack.* = ItemStack.fromBytes(reader) catch |err| {
+			std.log.err("Failed to read item stack from bytes: {s}", .{@errorName(err)});
+			stack.clear();
+			continue;
+		};
+	}
+	for(0..count) |_| {
+		var stack = ItemStack.fromBytes(reader) catch continue;
```
IntegratedQuantum, line 2036: Should print an error. Just because the item is outside the range of the inventory, doesn't mean it can't be an indicator of a bug.

src/Inventory.zig
```diff
@@ -2024,3 +2028,49 @@ pub fn loadFromZon(self: Inventory, zon: ZonElement) void {
...
+		stack.deinit();
```
IntegratedQuantum, line 2040: Should print an error first, to let the user know that an item was lost and which one was lost.

src/Inventory.zig
```diff
@@ -2024,3 +2028,49 @@ pub fn loadFromZon(self: Inventory, zon: ZonElement) void {
...
+}
+
+fn fromBytes(self: Inventory, reader: *BinaryReader) void {
+	var count = reader.readVarInt(u32) catch 0;
+	for(self._items) |*stack| {
+		if(count == 0) {
+			stack.clear();
+			continue;
+		}
+		defer count -= 1;
```
IntegratedQuantum: Since it doesn't make a difference, why not just do
```suggestion
		count -= 1;
```

src/Inventory.zig
```diff
@@ -2024,3 +2028,49 @@ pub fn loadFromZon(self: Inventory, zon: ZonElement) void {
...
+	const destination: []u8 = main.stackAllocator.alloc(u8, std.base64.url_safe.Decoder.calcSizeForSlice(base64) catch unreachable);
+	defer main.stackAllocator.free(destination);
+
+	std.base64.url_safe.Decoder.decode(destination, base64) catch unreachable;
+	var reader = BinaryReader.init(destination);
+	fromBytes(self, &reader);
+}
+
+fn fromBytes(self: Inventory, reader: *BinaryReader) void {
+	var count = reader.readVarInt(u32) catch 0;
```
IntegratedQuantum: Should be called `remainingCount`, since it is changed and doesn't reflect the actual `count` for most of the function's runtime.

IntegratedQuantum (changes requested):
One thing I'm really missing here is the deprecation strategy.
Your approach keeps the zon alive and we will need to support it forever, as long as we want to support people joining worlds that were created long ago.

So instead, please bump the world format version and add a conversion path that automatically converts all player inventories to the new format on joining a world from the old version. This conversion path can then be removed after the ~~next~~first release, alongside all the zon inventory parsing code.

src/Inventory.zig
Argmaster: This will cause hundreds of errors to appear if we shrink any Inventory, maybe a debug log or sth?

IntegratedQuantum: That is exactly the point. We shouldn't mindlessly shrink inventories that the player put stuff in.
There should be no silent data loss.

Argmaster: Its pre-alpha, we should have freedom to adjust the inventory size to fit gameplay.
Although we could consider shuffling stuff into empty slots in such case.

IntegratedQuantum: Yes, but we shouldn't mindlessly break it. It will just frustrate players if their hard-earned stack of amber is gone.
We should shuffle things if possible, or drop it if not.

src/server/world.zig
```diff
@@ -489,6 +489,73 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 				}
 			} else |_| {}
 		}
+		convert_player_data_to_binary: { // TODO: Remove after #480
+			const playerDataPath = try std.fmt.allocPrint(main.stackAllocator.allocator, "saves/{s}/players", .{path});
+			defer main.stackAllocator.free(playerDataPath);
+
+			var playerDataDirectory = std.fs.cwd().openDir(playerDataPath, .{.iterate = true}) catch |err| {
+				std.log.err("Could not open player data directory to migrate file format: {s}. Conversion aborted.", .{@errorName(err)});
```
OneAvargeCoder193: This error seems to be printed when you first create a world

src/server/world.zig
```diff
@@ -864,8 +931,25 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
 
-		user.inventory = main.items.Inventory.Sync.ServerSide.createExternallyManagedInventory(main.game.Player.inventorySize, .normal, .{.playerInventory = user.id}, playerData.getChild("playerInventory"));
-		user.handInventory = main.items.Inventory.Sync.ServerSide.createExternallyManagedInventory(1, .normal, .{.hand = user.id}, playerData.getChild("hand"));
+		const base64EncodedEmptyInventory = "AA==";
+		{
+			const base64 = playerData.get([]const u8, "playerInventory", base64EncodedEmptyInventory);
+			const bytes: []u8 = main.stackAllocator.alloc(u8, std.base64.url_safe.Decoder.calcSizeForSlice(base64) catch unreachable);
+			defer main.stackAllocator.free(bytes);
+
+			std.base64.url_safe.Decoder.decode(bytes, base64) catch unreachable;
```
IntegratedQuantum: This data comes from the evil outside world. We cannot assert that it is correctly base64 encoded.

src/server/world.zig
```diff
@@ -864,8 +931,25 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 			main.items.Inventory.Sync.setGamemode(user, std.meta.stringToEnum(main.game.Gamemode, playerData.get([]const u8, "gamemode", @tagName(self.defaultGamemode))) orelse self.defaultGamemode);
 		}
 
-		user.inventory = main.items.Inventory.Sync.ServerSide.createExternallyManagedInventory(main.game.Player.inventorySize, .normal, .{.playerInventory = user.id}, playerData.getChild("playerInventory"));
-		user.handInventory = main.items.Inventory.Sync.ServerSide.createExternallyManagedInventory(1, .normal, .{.hand = user.id}, playerData.getChild("hand"));
+		const base64EncodedEmptyInventory = "AA==";
```
IntegratedQuantum: Maybe make the parser handle empty strings instead?
Right here this is just gibberish, and would require further explanation.

src/Inventory.zig
```diff
@@ -2046,7 +2038,7 @@ fn toBytes(self: Inventory, writer: *BinaryWriter) void {
 	}
 }
 
-fn fromBase64(self: Inventory, base64: []const u8) void {
+pub fn fromBase64(self: Inventory, base64: []const u8) void {
```
IntegratedQuantum: This shouldn't be needed anymore?

src/Inventory.zig
```diff
@@ -2029,7 +2021,7 @@ pub fn loadFromZon(self: Inventory, zon: ZonElement) void {
 	}
 }
 
-fn toBase64(self: Inventory, allocator: NeverFailingAllocator) []const u8 {
+pub fn toBase64(self: Inventory, allocator: NeverFailingAllocator) []const u8 {
```
IntegratedQuantum: For symmetry I think this should also be done on the outside.

src/server/world.zig
```diff
@@ -864,8 +931,25 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		{
+			const base64 = playerData.get([]const u8, "playerInventory", base64EncodedEmptyInventory);
+			const bytes: []u8 = main.stackAllocator.alloc(u8, std.base64.url_safe.Decoder.calcSizeForSlice(base64) catch unreachable);
```
IntegratedQuantum: When does calcSizeForSlice return an error? Why can you assert that it doesn't happen?

src/server/world.zig
```diff
@@ -864,8 +931,25 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		}
+		{
+			const base64 = playerData.get([]const u8, "hand", base64EncodedEmptyInventory);
+			const bytes: []u8 = main.stackAllocator.alloc(u8, std.base64.url_safe.Decoder.calcSizeForSlice(base64) catch unreachable);
+			defer main.stackAllocator.free(bytes);
+
+			std.base64.url_safe.Decoder.decode(bytes, base64) catch unreachable;
+			var reader: main.utils.BinaryReader = .init(bytes);
+			user.handInventory = main.items.Inventory.Sync.ServerSide.createExternallyManagedInventory(1, .normal, .{.hand = user.id}, &reader);
+		}
```
IntegratedQuantum: Ideally this should be extracted into a helper function `loadPlayerInventory(size, playerData.get(...), ...)` or something

src/server/world.zig
```diff
@@ -489,6 +489,73 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+									main.stackAllocator.free(temp._items);
+								}
+
+								temp.loadFromZon(zon);
+								const base64Data = temp.toBase64(main.stackAllocator);
+								const old = playerData.object.fetchPut(key, .{.stringOwned = base64Data}) catch unreachable orelse unreachable;
+								old.value.deinit(main.stackAllocator);
+							},
+							.string, .stringOwned, .null => continue,
+							else => unreachable,
```
IntegratedQuantum: This data comes from the evil outside world.

src/server/world.zig
```diff
@@ -489,6 +489,73 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 				}
 			} else |_| {}
 		}
+		convert_player_data_to_binary: { // TODO: Remove after #480
+			const playerDataPath = try std.fmt.allocPrint(main.stackAllocator.allocator, "saves/{s}/players", .{path});
```
IntegratedQuantum: Should use catch unreachable

src/server/world.zig
```diff
@@ -489,6 +489,73 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+								var temp: main.items.Inventory = undefined;
+								temp._items = main.stackAllocator.alloc(ItemStack, zon.get(u32, "count", 0));
+								for(temp._items) |*stack| {
+									stack.* = ItemStack{};
+								}
+								defer {
+									for(temp._items) |*stack| {
+										stack.deinit();
+									}
+									main.stackAllocator.free(temp._items);
```
IntegratedQuantum: The defer for this should be right below the corresponding alloc.

src/Inventory.zig
```diff
@@ -2024,3 +2020,55 @@ pub fn loadFromZon(self: Inventory, zon: ZonElement) void {
...
+			continue;
+		};
+	}
+	if(remainingCount > 0) {
+		std.log.err("Detected {} extras items while loading inventory.", .{remainingCount});
+	}
+	for(0..remainingCount) |_| {
+		var stack = ItemStack.fromBytes(reader) catch continue;
+		if(stack.item) |item| {
+			std.log.err("Lost {} of {s}", .{stack.amount, item.getName()});
```
IntegratedQuantum: The logs should always contain the ids. Printing a name, which might be chinese or whatever if we have language support, is not going to make things easy to debug.

src/Inventory.zig
IntegratedQuantum: I have removed this code in 362f5408254a67b29dd90843934cecf756001ec2, please rebase.

src/server/world.zig
Argmaster: I am asserting this because I don't think we can do a safe recovery. It will error if the string is malformed eg. not padded correctly, but that could be caused by many factors (broken file, manual modification) and if we load this anyway we can either get garbage or convert small mistake into completely wiping someones inventory. I suppose we could log something, but again, aborting is the safe thing to do if the string wasn't encoded correctly.

Argmaster: Well, technically we could do some major refactor and eg. refuse player to connect to the server instead of crashing the server, but its huge amount of work for very little benefit - save is on server side anyway, so you should have control over correctness of sever save files.

IntegratedQuantum: Well, not doing a recovery means you cannot open this world at all. I'd say that is worse than wiping your inventory.
Especially when all the player sees is `panic: reach unreachable code` and a stack trace that they can't understand.
Furthermore this will crash on the server side, so even worse.

And we can just log the malformed data, so it wouldn't really be lost.

src/Inventory.zig
Argmaster: What about tools? I can't do explicit dispatch here because of custom item support, should I add dedicated function for logging in failure case we have here? Something like `.getItemLostMessage()`?

IntegratedQuantum: Why not just `getId()` that returns the item or tool id?

Argmaster: This is not enough to restore the tool.

IntegratedQuantum: Just craft a new one. No one will cry about a lost tool. And if they do, then they'll likely know what materials it contained anyways.

src/Inventory.zig
Argmaster: I would argue against copying this code 3 times just for symmetry.

src/Inventory.zig
Argmaster: Adding `getId` to `Item` results in duplication we have to resolve in future - `BaseItemIndex` already has `id()` method which does exactly the same thing as `getId` (result of direct field name -> getter name mapping while implementing `*Index` enums) but to avoid manual dispatch in `Item` we are forced to have same method names for all enum entries in `Item` (condition for supporting custom item types for comptime modding). Converging towards `id()` is not acceptable IMO, we already have all the methods called `getX()` there, so we should converge towards `getX()` on `*Index` enums for consistency.

src/server/world.zig
```diff
@@ -863,9 +926,27 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		const bytes: []u8 = main.stackAllocator.alloc(u8, decodedSize);
+		defer main.stackAllocator.free(bytes);
+
+		var readerInput: []const u8 = bytes;
+
+		std.base64.url_safe.Decoder.decode(bytes, base64EncodedData) catch |err| {
+			std.log.err("Encountered incorrectly encoded inventory data ({s}) while loading data from file '{s}': '{s}'", .{@errorName(err), playerDataFilePath, base64EncodedData});
+			readerInput = "";
+		};
+		var reader: main.utils.BinaryReader = .init(bytes);
```
OneAvargeCoder193: Did you mean to pass readerInput into the BinaryReader instead of bytes?

Argmaster:
There is stack allocator memory leak somewhere :C

OneAvargeCoder193:
> There is stack allocator memory leak somewhere :C

I think the memory leak might be coming from when you place the base64 strings into the player zon. Since you are using `put`, it will place that address directly into the zon, as a `.string`, and in deinit `.string`s dont ever get deinited. I think you need to use putOwnedString, but that dupes the string so you also need to deinit the encoded string outside of it.

Argmaster:
@OneAvargeCoder193 thank you kind sir, that was the problem indeed.

OneAvargeCoder193:
I forgot you could put ZonElements directly thats way simpler than what I said

src/Inventory.zig
IntegratedQuantum: In my mind `getX()` means this is something we need to possibly calculate (e.g. getTexture lazily generates the texture), whereas just `x()` means it's just a simple parameter fetch, which I'd say is the case for indices.

IntegratedQuantum: Consequently the function here should also be called just `id()`.

src/Inventory.zig
IntegratedQuantum: Where is the third one?

And you can just make a helper function like you did for the storing.

src/Inventory.zig
```diff
@@ -2019,3 +2010,46 @@ pub fn loadFromZon(self: Inventory, zon: ZonElement) void {
...
+		}
+		remainingCount -= 1;
+		stack.* = ItemStack.fromBytes(reader) catch |err| {
+			std.log.err("Failed to read item stack from bytes: {s}", .{@errorName(err)});
+			stack.clear();
+			continue;
+		};
+	}
+	if(remainingCount > 0) {
+		std.log.err("Detected {} extras items while loading inventory.", .{remainingCount});
```
IntegratedQuantum: I don't think we should print anything here, in the best case all item slots are empty, then this would be pointlessly worrying the player.

IntegratedQuantum (changes requested):

Argmaster:
There is a bug in migration code or I am doing sth wrong, I need to investigate it before merging.

src/Inventory.zig
Argmaster: > Where is the third one?

image

migration, hand, playerInventory

> And you can just make a helper function like you did for the storing.

Well, this is that helper function.

Argmaster:
Alright, I had to move migration code below asset loading since we need to convert string IDs to indexes with palettes and these need to be loaded first.

src/server/world.zig
```diff
@@ -489,6 +489,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 				}
 			} else |_| {}
 		}
+
```
Argmaster:
```suggestion
```

src/Inventory.zig
IntegratedQuantum: Yeah, but it belongs into world.zig, next to the corresponding load function.

src/server/world.zig
```diff
@@ -863,9 +942,37 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		std.base64.url_safe.Decoder.decode(bytes, base64EncodedData) catch |err| {
+			std.log.err("Encountered incorrectly encoded inventory data ({s}) while loading data from file '{s}': '{s}'", .{@errorName(err), playerDataFilePath, base64EncodedData});
+			readerInput = "";
+		};
+		var reader: main.utils.BinaryReader = .init(readerInput);
+		return main.items.Inventory.Sync.ServerSide.createExternallyManagedInventory(size, .normal, source, &reader);
+	}
+
+	fn savePlayerInventory(inv: main.items.Inventory, allocator: NeverFailingAllocator) []const u8 {
```
IntegratedQuantum: allocator should always be the first argument

Argmaster:
```suggestion
	fn savePlayerInventory(allocator: NeverFailingAllocator, inv: main.items.Inventory) []const u8 {
```

src/server/world.zig
```diff
@@ -540,6 +540,85 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+								for(temp._items) |*stack| stack.* = ItemStack{};
+								defer for(temp._items) |*stack| stack.deinit();
+
+								temp.loadFromZon(zon);
+
+								for(temp._items, 0..) |*stack, i| {
+									std.log.debug("Item #{}: {} x {s}", .{i, stack.amount, if(stack.item) |item| item.id() else "null"});
+								}
+
+								const base64Data = savePlayerInventory(temp, main.stackAllocator);
```
Argmaster:
```suggestion
								const base64Data = savePlayerInventory(main.stackAllocator, temp);
```

src/server/world.zig
```diff
@@ -893,11 +1000,11 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 			main.items.Inventory.Sync.ServerSide.mutex.lock();
 			defer main.items.Inventory.Sync.ServerSide.mutex.unlock();
 			if(main.items.Inventory.Sync.ServerSide.getInventoryFromSource(.{.playerInventory = user.id})) |inv| {
-				playerZon.put("playerInventory", inv.save(main.stackAllocator));
+				playerZon.put("playerInventory", ZonElement{.stringOwned = savePlayerInventory(inv, main.stackAllocator)});
```
Argmaster:
```suggestion
				playerZon.put("playerInventory", ZonElement{.stringOwned = savePlayerInventory(main.stackAllocator, inv)});
```

src/server/world.zig
```diff
@@ -893,11 +1000,11 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
 			} else @panic("The player inventory wasn't found. Cannot save player data.");
 
 			if(main.items.Inventory.Sync.ServerSide.getInventoryFromSource(.{.hand = user.id})) |inv| {
-				playerZon.put("hand", inv.save(main.stackAllocator));
+				playerZon.put("hand", ZonElement{.stringOwned = savePlayerInventory(inv, main.stackAllocator)});
```
Argmaster:
```suggestion
				playerZon.put("hand", ZonElement{.stringOwned = savePlayerInventory(main.stackAllocator, inv)});
```

PR #1702, opened 2025-07-21
Add mushroom caves

codemob-dev:
Adds mushroom caves:
image
image
image


codemob-dev:
New boletes
image

codemob-dev:
image

codemob-dev:
This is done for now, bigger mushrooms can be in a separate pr.

Argmaster:
Haven't reviewed yet, but two things from me:
- I have **suggested** in Discord that you should split this PR into 3 smaller ones, each for different shroom type so we can discuss designs separately 
- as I previously communicated I am not satisfied with basic vertical stem + circle/half-sphere shroom designs. Although it's only my opinion and I am not an artist so you should take it with grain of salt. That said I had an impression that frequent contributors did agree with me.

ikabod-kee:
Yeah, I would rather have the mushrooms like ARG's

IntegratedQuantum:
> frequent contributors did agree with me.

I also agree with you, and I really liked the prototype you made.

ikabod-kee:
Gameplay reason for why I prefer ARG's mushrooms: they differentiate themselves from trees by being thicker, and you can even build a house out of one if you'd like!

codemob-dev:
> I have suggested in Discord that you should split this PR into 3 smaller ones, each for different shroom type so we can discuss designs separately

Done.

Issue #1580, opened 2025-06-01
[SBB] Right click to open children configuration menu


Argmaster:
Right clicking on origin block in creative mode should open a menu for configuring what structures should be spawned from children.
For simplicity of implementation for now player would specify child blocks. For each child block there should be an UI solution allowing to specify structure and change with completion / list of available structures.
Most likely this should be implemented as block entity.

PR #1714, opened 2025-07-23
Toadstool mushroom caves

codemob-dev:
image


codemob-dev:
New mushrooms:

image

I could also add bigger ones, but I think that would be best for another PR.

codemob-dev:
For the most part, bigger ones wouldn't fit in some of the caves and get cut off.

Argmaster:
Yes, this is why you have to be creative.
I can prep something you can split and modify if you want.

In the future we must explore if we can do some more informed structure resolution with cave maps so we can choose smaller structures when there is less room.

Argmaster:
This is my reference image:

image

Here are the models:

image

image

image


As you can see they don't have to be big to be unique. 

[blueprints.zip](https://github.com/user-attachments/files/21394355/blueprints.zip)

What do you think @ikabod-kee @careeoki  @IntegratedQuantum

codemob-dev:
ooh I like that!

Argmaster:
These blueprints need post processing to be a good structure material, but general idea is there.

ikabod-kee:
> what do you think @ikabod-kee

I'm a big, big fan.

codemob-dev:
> These blueprints need post processing to be a good structure material, but general idea is there.

I'm probably going to build mostly based off of the images to fit the stems I currently have

Argmaster:
Some options for alt caps for those shrooms:

image

image

image

As for the stems, you can angle them in X or Y or diagonally to get very big collection of different designs.

image

At the end to make almost every mushroom unique leave a lot of possible branch slots for small shrooms branching from the stem. just make sure that pool of caps you use can't collide with branches. Each of the children blocks should have a small chance to spawn a branch so its 0 to 2 branches on shrooms. If you do that well you are likely to get very nice parkour atomagically.

image

codemob-dev:
Should I keep the split?

image

ikabod-kee:
Hey, can't you guys just build this all in multiplayer? XD

codemob-dev:
> Hey, can't you guys just build this all in multiplayer? XD

Does `/blueprint save` save to the host or the player?

Argmaster:
Yeah we could, thats a good option. Server side i think, same way WE does.

codemob-dev:
I could try hosting to see how well that works on low-end machines

codemob-dev:
image

Much improved mushrooms!

codemob-dev:
Some better angles:
image
image
image
image

Argmaster:
Can you reduce frequency of branches? they appear too often IMO, more shrooms should have none of them. Also I wonder if there is a way to do wall shrooms... Probably no, we could use those too (right?) we may need a feature request for that.

Also some of the branches conflict with parts of stems / touch caps, these should be eliminated or caps should be raised by 1 block.

codemob-dev:
> Also I wonder if there is a way to do wall shrooms...

I was wondering that myself, but it seems that structures are limited to the floor and ceiling :disappointed:

codemob-dev:
> Can you reduce frequency of branches? they appear too often IMO, more shrooms should have none of them. 

Got that done.

Argmaster:
image

image

IMO they look great. We could use some wall caps and some huge shrooms too, but these need engine changes so they dont look awkward, therefore lets leave that for now.

This + #1704 #1707 can be a game changer when it comes to getting lower in huge caves (assuming we have a way to deliver light to the bottom of the cave w/o delivering our asses there)

Argmaster:
You could try to increase the frequency of shrooms themselves because its quite hard to jump from one to another.
Also we could have few very small ones, like these:

image

codemob-dev:
> This + https://github.com/PixelGuys/Cubyz/issues/1704 https://github.com/PixelGuys/Cubyz/pull/1707 can be a game changer when it comes to getting lower in huge caves (assuming we have a way to deliver light to the bottom of the cave w/o delivering our asses there)

YES it would be so fun to jump down a cave and start bouncing around mushrooms!

careeoki:
Looks great so far!!

codemob-dev:
> Looks great so far!!

Thanks! I'm switching gears to work on #1707 for now, but I'll add mini-mushrooms next.

codemob-dev:
Some small shroom designs
image

codemob-dev:
https://github.com/user-attachments/assets/c6ad3834-4f83-422f-894a-2b908c3322ec

assets/cubyz/sbb/mushroom/big/toadstool/branch/blue.zig.zon
Argmaster, line 1: Since when do we count ` `, `0`, `1`, ...?

assets/cubyz/sbb/mushroom/big/toadstool/branch/blue.zig.zon
```diff
@@ -0,0 +1,9 @@
+.{
+	.blueprint = "cubyz:mushroom/big/toadstool/empty_horizontal",
```
Argmaster: Whaaat is that?

codemob-dev: Its an empty blueprint with an origin and a white sbb block stacked horizontally. I've found it really handy to organize blueprints.

Argmaster: Ehh, yeah, I kind of expected that. I can't really blame you tbh, but I don't think we wan't blueprints like this lying around.

codemob-dev: The best option would probably be to either allow folders to be specified as structures or to have an empty horizontal and vertical blueprint built-in. That would probably be for another PR though.

Argmaster: First one may be viable, second one definitely is not an option. In general I don't think we should have dummy SBBs  like these. That is problematic topic though, because we can't make them impossible to create, so they will be used regardless of what we provide as an alternative. Also they will be much easier to create with #1580

codemob-dev: @IntegratedQuantum what do you think?

IntegratedQuantum: Why can't you just place it with the right orientation in the source blueprint?

We should definitely avoid needless blueprints, these are going to be quite expensive, particularly for LOD chunk generation.

codemob-dev: It's mostly just to put everything in one place, I can clean it up though.

blueprints/0.blp
Argmaster, line 1: This file should not be committed.

blueprints/1.blp
Argmaster, line 1: This file should not be committed.

blueprints/2.blp
Argmaster, line 1: This file should not be committed.

blueprints/3.blp
Argmaster, line 1: This file should not be committed.

blueprints/4.blp
Argmaster, line 1: This file should not be committed.

blueprints/5.blp
Argmaster, line 1: This file should not be committed.

blueprints/6.blp
Argmaster, line 1: This file should not be committed.

Argmaster (changes requested):
@IntegratedQuantum can we add `blueprints/` to `.gitignore`?

IntegratedQuantum:
We've noticed in the play session that mushroom caves are everywhere, which makes them less special than they should be.

So please lower the chance. I think putting the summed chance of all three biomes together to something like 0.1 should be good.

codemob-dev:
Got that, they should be pretty rare now.

assets/cubyz/biomes/cave/toadstool_big.zig.zon
```diff
@@ -0,0 +1,47 @@
...
+	.isCave = true,
+	.maxHeight = -512,
+	.minHeight = -48250,
+
+	.chance = 0.04,
+
+	.caves = -0.02,
+
+	.music = "cubyz:la_kubisa_caverna",
+	.stoneBlock = "cubyz:stone",
```
IntegratedQuantum: this is now `cubyz:slate`.
Also as you long as you don't explicitly want to enforce a certain stone type, it's based to just leave it empty. This might allow us to overwrite the stone based on other factors in the future.

assets/cubyz/sbb/mushroom/big/toadstool/base/0.zig.zon
```diff
@@ -0,0 +1,23 @@
+.{
+	.blueprint = "cubyz:mushroom/big/toadstool/base/0",
+	.children = .{
+		.white = .{
+			.{.structure = "cubyz:mushroom/big/toadstool/large_cap"},
+		},
+		.crimson = {
+			.{.structure = "cubyz:mushroom/big/toadstool/branch/red0", chance = 0.125},
+			.{.structure = "cubyz:mushroom/big/toadstool/branch/red1", chance = 0.125},
```
IntegratedQuantum, line 9: Nothing to change here, just a note to make things easier for you in the future: These numbers don't need to add up to 1.

IntegratedQuantum (changes requested):
One of the models contains an invalid branch block, we also noticed this during the last play session, and I guess I forgot to write it down:

Screenshot at 2025-08-14 16-22-02

assets/cubyz/biomes/cave/toadstool_big.zig.zon
IntegratedQuantum: Please also fix this in the other PRs.

codemob-dev:
> One of the models contains an invalid branch block, we also noticed this during the last play session, and I guess I forgot to write it down:

Weird, the model looks fine, I reexported it and it didn't show any diff. Maybe the rotation code is at fault?

codemob-dev:
Yeah, it works in generation:

image

IntegratedQuantum:
Maybe you accidentally have two variants of the same thing, one broken and one not?

IntegratedQuantum:
I think I've also seen it work in some cases.

IntegratedQuantum:
Sometimes it works
Screenshot at 2025-08-14 19-17-20
Sometimes it doesn't
Screenshot at 2025-08-14 19-17-33
Also noticed that this model has some unconnected branches too:
Screenshot at 2025-08-14 19-19-50

codemob-dev:
I'm not really sure if this is my fault, I don't have any other models and I have noticed that some branches don't connect when they are rotated.

IntegratedQuantum:
Could you investigate this further? does this also happens when rotating it with world edit commands, or only with structure placement?

codemob-dev:
> does this also happens when rotating it with world edit commands, or only with structure placement?

I can't reproduce it exactly with world edit commands, but the branch does lose all of its connections if I rotate it 180 or 270 degrees.

codemob-dev:
I am completely stumped as to why this could be happening, any ideas @IntegratedQuantum?

IntegratedQuantum:
I found it and fixed it

IntegratedQuantum (approved):

codemob-dev:
Nice!

> Also noticed that this model has some unconnected branches too:

I got this fixed too

PR #1715, opened 2025-07-23
Bolete mushroom caves

codemob-dev:
image

### DIFF
--- a/assets/cubyz/biomes/cave/bolete_big.zig.zon
+++ b/assets/cubyz/biomes/cave/bolete_big.zig.zon
@@ -8,7 +8,6 @@
 	.caves = -0.015,
 
 	.music = "cubyz:la_kubisa_caverna",
-	.stoneBlock = "cubyz:stone",
 	.ground_structure = .{
 		"2 to 3 cubyz:soil",
 	},


codemob-dev:
image

codemob-dev:
https://github.com/user-attachments/assets/cf2ffba1-ba8e-466e-8452-398f438143b3

codemob-dev:
It is ready for review.

careeoki:
copyImage
Bolete do not have gills, its what identifies them as such. You can also see this on the underside of the small bolete block.

codemob-dev:
huh, interesting. I'll modify the caps.

IntegratedQuantum:
In my opinion the caps should be larger, or the stems smaller, right now I think it's looking kind of weird, with the caps taking significantly less volume comapred to the stem.

codemob-dev:
image

I'm using penny bun mushrooms as reference, they often have a very wide stem compared to the cap.

Argmaster:
I have encountered an error?

```
Could not find child structure nor blueprint 'cubyz:mushroom/small/bolete/variant/8' for child resolution.
Could not find child structure nor blueprint 'cubyz:mushroom/small/bolete/variant/9' for child resolution
```

---

Firstly, similarily to #1716 , it is no longer sufficient to use branches to fake underside of the shroom cap. We need some way of making that spongey pipes etatable shrooms have.
Secondly, most of those legs are too fat, up to 5 blocks of diameter. IMO we don't have resolution sutable for making fat legs like that. Mushrooms tend to blend into one big mass of mushroom legs and that doesn't seem convincing to me. I do respect the gimmick of hidding inside a mushroom, but we did it once and we allowed once for mushroom legs to be fat, and now its enough. Both toadstools and glimmergils are quite bulky and I would prefer boletes to be more slim and for gods sake irregular - cap connection point should (almost) always be offset in XY plane relative to center of the leg by 1-3 blocks. 

image

image

image

Gimmicks are fun when they are rare and they make up minority of gameplay.

codemob-dev:
The mushrooms should look much more reasonable now.

Argmaster:
After recent play session with those mushrooms added as addon I think I am happy enough with current models.
@ikabod-kee @careeoki could you explicitly express your approval or request changes so we can progress towards merging this?

careeoki:
Yes, looks good to me

ikabod-kee:
Same

assets/cubyz/sbb/mushroom/big/bolete/base.zig.zon
```diff
@@ -0,0 +1,12 @@
+.{
+	.blueprint = "cubyz:mushroom/big/bolete/empty",
```
IntegratedQuantum, line 2: I'm not a fan of these empty intermediary structures.

I think we may have made a design mistake here:
Currently adding variants is O(n×m), if you have n bottom variants and m top variants, you need to list all top variants in the blueprint file of all bottom variants.
Which encourages adding another step in the hierarchy.

I think we should code this into the system. I'd suggest turning around the current format a bit to make this work, maybe something like this:
```zig
.{
    .blueprints = .{
        .{.id = "cubyz:mushroom/big/bolete/base/0", .chance = 1},
        .{.id = "cubyz:mushroom/big/bolete/base/1", .chance = 1, .(future)replacements = .{...}},
    },
    .children = .{
        .white = .{.structure = "cubyz:mushroom/big/bolete/cap"},
    },
}
```
This would also allow us to get rid of all of the repeated files.

@Argmaster what do you think about this?

Argmaster: I agree that those intermediate dummy sbbs have to be eliminated.

I need to think some more about the implication of such inversion.

Argmaster: I think this would mean that we cannot do #1580 as blueprint can no longer be linked one-to-one with SBB. Multiple SBB can use same blueprint for different reasons and detecting all of them will require re-checking all the SBBs (as changes will be applied dynamically). This was sort of a problem already, after #1500 changes, but with some constrains it was still possible.

Argmaster: So If I understand correctly this will have an effect of making SBBs correspond to directories, rather than to files.
It contains list of blueprints (files) and list of next sbbs (directories) to visit.

Argmaster: To make the example a bit more complicated, I would rather analyze bot/mid/top model, as 2 layer model removes some of the complexity.

## Example

bot:

```zig
.{
    .blueprints = .{
        .{.id = "cubyz:bot/0", .chance = 0.5},
        .{.id = "cubyz:bot/1", .chance = 0.5},
    },
    .children = .{
        .white = .{.structure = "cubyz:mid"},
    },
}
``` 

mid:

```zig
.{
    .blueprints = .{
        .{.id = "cubyz:mid/0", .chance = 0.5},
        .{.id = "cubyz:mid/1", .chance = 0.5},
    },
    .children = .{
        .white = .{.structure = "cubyz:top"},
    },
}
``` 

mid:

```zig
.{
    .blueprints = .{
        .{.id = "cubyz:top/0", .chance = 0.5},
        .{.id = "cubyz:top/1", .chance = 0.5},
    },
}
```

IntegratedQuantum: > So If I understand correctly this will have an effect of making SBBs correspond to directories, rather than to files.

Yes, if the files are ordered like in this PR, though technically it does allow to include blueprints from other directories.

As for #1580 I'm sure we can find a way to make that work either by just providing a dropdown of possible options, or maybe another way of selecting and editing the SBBs directly instead of through their blueprint structures.

I think for now it's best if we observe and follow the usage patterns, maybe we can even restrict it to only allow including blueprints from the child directory of the same name, in which case it wouldn't be ambiguous anymore what SBB a structure belongs to.

Argmaster: With this solution we can't complain when there are extra child blocks specified (ones that are not present in blueprint we selected)

IntegratedQuantum: Well, we could still complain if the color is not used in any of the blueprints specified.

Argmaster: I assume that replacements refer to #1412 in which case I dislike how one would have to repeat replacement spec for each of the blueprints listed. I think that it may be better to specify replacements as completely separate entity (eg. by `*.replacement.zig.zon` extension) so they are treated as yet another (autogenerated) blueprint, occupying another ID.

```zig
.{
    .replacements = .{...},
    .blueprint = "..."
}
```

IntegratedQuantum: > To make the example a bit more complicated, I would rather analyze bot/mid/top model, as 2 layer model removes some of the complexity.

I think in practice we will also see non-linear patterns, e.g. consider large and small caps, in this case you would need to resort to using separate colors:
```zig
.{
    .blueprints = .{
        .{.id = "cubyz:mid/0", .chance = 0.5},
        .{.id = "cubyz:mid/1", .chance = 0.5},
    },
    .children = .{
        .white = .{.structure = "cubyz:top_small"},
        .green = .{.structure = "cubyz:top_large"},
    },
}
```
And now the big question is, what if you have a bottom structure that fits both large and small tops?
That would require either copying the structure with different child colors(which could be done with replacements, maybe that's the best solution?), or making a new SBB that contain both kinds of the top structure, or we need to allow multiple child SBBs, which I'd like to avoid, since it's another alias table.

codemob-dev: Would something like this be possible?
```zig
.{
    .blueprints = .{
        .{.id = "cubyz:mid/0", .chance = 0.5},
        .{.id = "cubyz:mid/1", .chance = 0.5},
    },
    .children = .{
        .{.color = {.white, .red}, .structure = "cubyz:top_small"},
        .{.color = {.green, .red}, .structure = "cubyz:top_large"},
    },
}
```

Argmaster: > we need to allow multiple child SBBs

A list of possible structures for a child block you mean?

> And now the big question is, what if you have a bottom structure that fits both large and small tops?

The answer is quite likely "You don't". Proportions are unlikely to allow that.

What I noticed is that only blueprints are really reusable in this system. Reusing intermediate blocks often means generating structures which are very similar and that is not desirable outcome.
This inversion doesn't change that, still if you take intermediate (mid) structure part you are also taking everything after it.

IntegratedQuantum: > Would something like this be possible?

That would be no different internally, you'd still need to take a random number, and you'd still need another indirection in the internal data structure.

Argmaster: It would be good to ask @ikabod-kee what was his experience with trees, maybe it was different than mine.

ikabod-kee: My experience with making trees: I made sure there were no empty blueprints besides the generator itself. I also reuse blueprints when possible. I call them "general" blueprints that can be reused in different scenarios. Not the best way of doing things but meh

Argmaster: There is no such thing as generator in SBB system. The "generator" you are describing is exactly the usage pattern we want to get rid of ❤️ 
There are only structures which you can place in the world.

Argmaster: Thanks for your feedback, it gives me hope I was not all wrong with my experience 😜

ikabod-kee: It's kind of easy to trim the fat for these then. You just place the blueprint of a non-generator. You don't really need to use the generator in the middle of a blueprint.

Argmaster: > consider large and small caps, in this case you would need to resort to using separate colors:

Why? You could just have a single cap SBB which can use either large or small cap. That said to have small mushrooms, mixed mushrooms and big mushrooms you would need separate SBBs and they can't reuse intermediate SBBs either, as they are chained.

ikabod-kee: Yeah wait, couldn't you just use the same color sbb for caps? Unless they have seperate chances for things, it's redundant to use different colors.

But also this is a bit nitpicky maybe

ikabod-kee: I don't know about you, but I always had a designated "next in chain" color

IntegratedQuantum: > A list of possible structures for a child block you mean?

Yes

> The answer is quite likely "You don't". Proportions are unlikely to allow that.

Big and small was maybe a bad example, I think there are many cases where this may be useful, let's say you have a house and maybe you have a bunch of different options for chests and work stations, but maybe you don't want furnaces in this house, but in most of the other houses.

> What I noticed is that only blueprints are really reusable in this system. Reusing intermediate blocks often means generating structures which are very similar and that is not desirable outcome.
This inversion doesn't change that, still if you take intermediate (mid) structure part you are also taking everything after it.

I think this is a key insight. It's a bit of an inherent flaw of the tree structure, and also the reason why we cannot have rooms of random sizes.
Maybe the solution to both of these problems would be to introduce parameters or conditions that get passed down. This would also solving the cap size problem:
```zig
.{ // mid.zig.zon
    .blueprints = .{
        .{.id = "cubyz:mid/0", .chance = 0.5},
        .{.id = "cubyz:mid/1", .chance = 0.5},
    },
    .children = .{
        .{.color = .white, .structure = "cubyz:top", .parameters = .{.large}}, // or .parameters = .{.size = xyz}
        .{.color = .green, .structure = "cubyz:top", .parameters = .{.small}},
        .{.color = .red, .structure = "cubyz:top", .parameters = .{.small, .large}},
    },
}
.{ // top.zig.zon
    .blueprints = .{
        .{.id = "cubyz:top/0large", .condition = .large}, // or .condition = "size > xyz"
        .{.id = "cubyz:top/0small", .condition = .small},
        .{.id = "cubyz:top/1small", .condition = .small},
    },
}
```
This does however mean we'll need to deal with a turing-complete system.

Argmaster: > Maybe the solution to both of these problems would be to introduce parameters or conditions that get passed down.

No, that is too far. Next request you will get is error handling. We either accept limitations of declarative model or we should pick a proper imperative language instead. Declarative languages were not made for that purpose and we should not attempt to use a hammer as a screwdriver.

So I see ~~three~~ possible directions:
- we gamble which way of writing SBBs will be better - current or proposed - they are almost the same chaining methods, but can improve some important aspects of writing SBBs
- we can fragment SBBs further to hope we can reuse just a bit more code, but we enter the hell of shitload small files
- we can merge those SBBs into single file and stop attempting to reuse something that is not reusable (either deep nesting or local ID links `#<local-id>`)
- we get rid of declarative system and figure out some imperative one instead.

Argmaster: Local links/references is a declarative feature we may use to re-use some of the code (present in eg. [YAML](https://blog.daemonl.com/2016/02/yaml.html))

IntegratedQuantum: > we get rid of declarative system and figure out some imperative one instead.

I think we will need both, the current system is simple and efficient and covers most use-cases and has a low barrier to entry, but we cannot do e.g. dungeons with it (maybe it could do dungeon rooms, but even that is rather limited), which may end up as a mod-only thing.

> we gamble which way of writing SBBs will be better - current or proposed - they are almost the same chaining methods, but can improve some important aspects of writing SBBs

Well, that's why we are talking about it, so it isn't a gamble.
And unless anyone sees any major problems with the new system, I'd suggest we go with it, since it does make the process simpler since it no longer needs an extra text file for all blueprints with children. And it avoids the need for "empty" or "generator" SBBs.

Argmaster: > I think we will need both, the current system is simple and efficient and covers most use-cases and has a low barrier to entry, but we cannot do e.g. dungeons with it (maybe it could do dungeon rooms, but even that is rather limited), which may end up as a mod-only thing.

Well, it was never designed to handle such use cases to be honest. The most obvious problem is collisions between structures. This system doesn't provide a good way to annotate upper bound for structure size which is kind of necessary to avoid collisions between dungeons (It sucks when they collide, especially if they contain something more ambitious then few tunnels)

> Well, that's why we are talking about it, so it isn't a gamble. And unless anyone sees any major problems with the new system, I'd suggest we go with it, since it does make the process simpler since it no longer needs an extra text file for all blueprints with children. And it avoids the need for "empty" or "generator" SBBs.

It is a gamble in a sense that we have a very limited sample of use cases.
That said I don't see any obvious problems which didn't exist before, so sure, it might be consistently better.
Migration and re-implementation will require some effort though and we will see if any problems will arise as we go.
Does that mean #1714, #1715 and #1716 are blocked until then?

IntegratedQuantum: > Does that mean https://github.com/PixelGuys/Cubyz/pull/1714, https://github.com/PixelGuys/Cubyz/pull/1715 and https://github.com/PixelGuys/Cubyz/pull/1716 are blocked until then?

No, but any new PRs will be, to avoid having to do too much work when migrating. Either way it's better to do this sooner.

Do you think you could find the time to work on this during the next weeks? I know you are busy and it's okay to say no, I'm just asking because you wrote the original code and probably know it better than I do.

Argmaster: Yes, I want to work on that and I will figure out a way to find time.
I will try looking into it today in the evening to estimate what and how has to be changed.

codemob-dev: Does that mean that this is ready to merge?

IntegratedQuantum: after I review it

IntegratedQuantum (approved):
Everything seems to work now

PR #1716, opened 2025-07-23
Glimmergill mushroom caves

codemob-dev:
image

### DIFF
--- a/assets/cubyz/biomes/cave/glimmergill_big.zig.zon
+++ b/assets/cubyz/biomes/cave/glimmergill_big.zig.zon
@@ -8,7 +8,6 @@
 	.caves = -0.02,
 
 	.music = "cubyz:la_kubisa_caverna",
-	.stoneBlock = "cubyz:stone",
 	.ground_structure = .{
 		"2 to 3 cubyz:soil",
 	},


codemob-dev:
These are what I'm considering basing them off of:
[https://en.wikipedia.org/wiki/Lactarius_indigo](https://en.wikipedia.org/wiki/Lactarius_indigo)
image

codemob-dev:
https://github.com/user-attachments/assets/d0810e03-d68c-45f3-8044-255a66211b58

ikabod-kee:
I love the touch of water being collected in the inverted cap!

Argmaster:
@careeoki can we give cap blocks more visible edge? They are glowing so they have no shadow and they blend together hiding all the detail.

Something like this:

image

codemob-dev:
> Something like this:

How does it look with multiple blocks?

ikabod-kee:
> Something like this:

Show us how it tiles

Argmaster:
Alternative would be to add mushroom gills (non-cube) block with carpet rotation maybe.

Argmaster:
> > Something like this:
> 
> Show us how it tiles

It tiles like shit, this is why I am asking carrie to do that properly XD

ikabod-kee:
I think #191 would fix that :P  But that's WAY far ahead of us. So, it's really a future problem if you think about it.

codemob-dev:
> Alternative would be to add mushroom gills (non-cube) block with carpet rotation maybe.

Mushroom gills are radial, how would the texture work?

Argmaster:
Ehh, I don't know, the problem is that there is no difference between inside and outside of the cap, we got around it for #1714 because we could use branches, but it won't look good here and honestly we could use gills for all shroom types.
Since Cubyz is voxel game we can compromise on gills being axis aligned. I would prefer them to have custom model, but if it involves severe performance penalty having just white-ish carpet with lines or dots (for boletes) is good enough.
Let me do a prototype.

codemob-dev:
Small glimmergills:
image

Argmaster:
image

Ideally they would also have different models with gills sticking out of them.

codemob-dev:
That looks pretty good, how does rotation work with gills?

Argmaster:
Its just a carpet, so on horizontal faces it works well, on vertical faces rotation is fixed so it looks weird half of the time.

codemob-dev:
> on vertical faces rotation is fixed so it looks weird half of the time.

It could probably use a new rotation model, I believe 4 extra bits (2 for each vertical face) should fit.

ikabod-kee:
I feel like a glimmergill carpet would serve as a pretty neat building block, especially if you're making a house out of it.

Argmaster:
As for the general shape of the shroom, I would give them very short legs and tall/wide caps, following the reference image.
And pretty please kill symmetry whenever possible. If something can be crooked, it should be crooked.

image

image

image

image

gills should be glowy too, its my mistake that they are not.

codemob-dev:
Yeah, I'll try this.

codemob-dev:
image

codemob-dev:
To fix the visual edge issue, why not just have each face darkened or lightened slightly differently?

codemob-dev:
Here's a close up of one of the wet mushrooms:
image

codemob-dev:
And of course, a partially hollow mushroom for snale to hide in (:
image

careeoki:
> @careeoki can we give cap blocks more visible edge? They are glowing so they have no shadow and they blend together hiding all the detail.

I can't imagine any way of doing this that wouldn't look stupid and/or ruin the tiling. I think this isn't a texture problem and is a problem with how emission/lighting is handled

careeoki:
> Ideally they would also have different models with gills sticking out of them.

I have an idea for how this could look, I'll try it. 
Also note that bolete do not have gills, they (probably all mushroom caps) would benefit from something like minecraft's mushroom block placement

careeoki:
Here was my idea (no texture or anything), I'm not too crazy about it. I think I would just rather have an "inside" texture for the cap blocks. 
image
image

codemob-dev:
We should do this in a different PR.

careeoki:
Yeah, I still think the ones you built here look good.

Argmaster:
I still think its not clear what the actual shape of the shroom is because of lack of gills, but if this problem is not visible to others and there is no will to fix it now then I suppose I should refrain from further reviewing this PR.

Argmaster:
I don't think there are any pending requests / blocking dependencies for those mushroom models, so @ikabod-kee @careeoki could you explicitly express your approval or request changes so we can move on?

ikabod-kee:
Looks fine to me, I'll leave the review up to Carrie.

careeoki:
Looks good

PR #1726, opened 2025-07-28
Use AABBs for collision

### LINKED ISSUE #826: Use AABBs for collision
This would make it possible to have fences that you cannot jump over (which would still require #673) and it could also make collisions faster by having collision models have less detail.

codemob-dev:
Fixes #826


src/models.zig
```diff
@@ -195,6 +197,13 @@ pub const Model = struct {
 			self.allNeighborsOccluded = self.allNeighborsOccluded and self.isNeighborOccluded[neighbor];
 			self.noNeighborsOccluded = self.noNeighborsOccluded and !self.isNeighborOccluded[neighbor];
 		}
+		if(collision)|realCollision| {
+			self.collision = realCollision;
+		} else {
+			self.collision = main.globalAllocator.alloc(AABB, 1);
+			self.collision[0] = AABB {.min = @floatCast(self.min), .max = @floatCast(self.max)};
```
IntegratedQuantum: Please automatically create an accurate collision model from the mesh.

codemob-dev:
I now generate AABBs from the mesh. Everything seems to work apart from branches, which seem to behave weirdly.

codemob-dev:
This should be working.

src/models.zig
```diff
@@ -195,9 +198,165 @@ pub const Model = struct {
...
+								if(vec.dot(dir, quad.normalVec()) > 0) {
+									signed_intersections -= 1;
+								} else {
+									signed_intersections += 1;
+								}
+							}
+						}
+
+						if(signed_intersections != 0) {
+							grid[x][y][z] = true;
```
IntegratedQuantum: I don't think this check is good enough, It's assuming a property of the models that is not necessarily true.
Furthermore this will not work for zero-thickness models such as e.g. ladders, vines.
I think a better approach would be to voxelize the triangles, and fill all the hollow spaces.

src/models.zig
```diff
@@ -195,9 +198,165 @@ pub const Model = struct {
...
+							break;
+						}
+					}
+				}
+			}
+		}
+
+		// The theoretical maximum (I could come up with) is a 3d checkerboard pattern.
+		self.collision = main.globalAllocator.alloc(AABB, meshGridSize*meshGridSize*meshGridSize/2);
```
IntegratedQuantum: Allocating such a giant region only to shrink it is quite inefficient. Please just use a list instead.

src/models.zig
```diff
@@ -195,9 +198,165 @@ pub const Model = struct {
...
+								if(neighbor.isPositive()) {
+									boxMax += neighbor.relPos();
+								} else {
+									boxMin += neighbor.relPos();
+								}
+							}
+						}
+						setAll(&grid, boxMin, boxMax, false);
+
+						const minBlockX = @as(f32, @floatFromInt(boxMin[0]))/meshGridSize;
```
IntegratedQuantum: Please make user of vector operations instead of doing this element-wise

src/models.zig
```diff
@@ -195,9 +198,165 @@ pub const Model = struct {
...
+					if(!grid[x][y][z]) {
+						return false;
+					}
+				}
+			}
+		}
+		return true;
+	}
+
+	fn setAll(grid: *[meshGridSize][meshGridSize][meshGridSize]bool, min: Vec3i, max: Vec3i, value: bool) void {
```
IntegratedQuantum: Just use a pointer cast and a memset.

IntegratedQuantum (changes requested):

src/models.zig
codemob-dev: I was considering this, but because the triangles are typically on the intersection of multiple voxels finding which ones should be filled could be difficult. I could probably just move them half a voxel in the opposite of the normal though.

src/game.zig
```diff
@@ -52,176 +52,166 @@ pub const camera = struct { // MARK: camera
...
+		pub fn join(self: AABB, other: AABB) AABB {
+			return .{.min = @min(self.min, other.min), .max = @max(self.max, other.max)};
 		}
 
-		// Test axis a12
-		const a12 = Vec3d{f2[Z], 0, -f2[X]};
-		if(!test_axis(a12, v0, v1, v2, box_extents[X]*@abs(f2[Z]) + box_extents[Z]*@abs(f2[X]))) {
-			return false;
-		}
+		pub fn intersectsTriangle(self: AABB, triangle: [3]Vec3d) bool {
```
IntegratedQuantum: Could you move this to the place, and indentation it was originally in, to reduce the diff.

codemob-dev:
I can't figure out how to get this working. I'll leave the PR open if anyone wants to try and tackle it.

src/models.zig
```diff
@@ -203,38 +204,115 @@ pub const Model = struct {
...
+				const xf = (@as(f32, @floatFromInt(x)) + 0.5)/@as(f32, @floatFromInt(meshGridSize));
+
+				const zf = solveDepth(normal, v0, X, Y, Z, xf, yf);
+				if(zf < 0.0) continue;
+				const z: usize = @intFromFloat(zf * @as(f32, @floatFromInt(meshGridSize)));
+
+				if(z >= meshGridSize) continue;
+				 
+				const pos: [3]usize = .{x, y, z};
+				grid[pos[X]][pos[Y]][pos[Z]] = true;
```
IntegratedQuantum: Shouldn't this be the reverse operation? basically
```zig
const pos: [3]usize = .{x, y, z};
var realPos: [3]usize = undefined;
realPos[X] = pos[0];
realPos[Y] = pos[1];
realPos[Z] = pos[2];
grid[realPos[0]][realPos[1]][realPos[2]] = true;
```

codemob-dev: Yeah it probably should be

codemob-dev:
It still doesn't quite behave as expected, but I'm getting very close!

src/game.zig
```diff
@@ -52,11 +52,41 @@ pub const camera = struct { // MARK: camera
 };
 
 pub const collision = struct {
-	pub fn triangleAABB(triangle: [3]Vec3d, box_center: Vec3d, box_extents: Vec3d) bool {
+	pub const AABB = struct {
```
IntegratedQuantum: Should be `Aabb`, that's how we handle abbreviations

src/game.zig
```diff
@@ -52,11 +52,41 @@ pub const camera = struct { // MARK: camera
...
+		}
+
+		pub fn intersects(self: AABB, other: AABB) bool {
+			var i: usize = 0;
+			while(i < 3) : (i += 1) {
+				if(self.max[i] < other.min[i] or self.min[i] > other.max[i]) {
+					return false;
+				}
+			}
+			return true;
```
IntegratedQuantum: Why not use vector ops?
```suggestion
			return @reduce(.And, (self.max >= other.min)) and @reduce(.And, (self.min <= other.max));
```

src/game.zig
```diff
@@ -52,11 +52,41 @@ pub const camera = struct { // MARK: camera
...
+		}
+
+		pub fn join(self: AABB, other: AABB) AABB {
```
IntegratedQuantum: I don't think join is the right name for this.
Please inline this into its one use case, it is likely going to be removed from there anyways in the future.

src/game.zig
```diff
@@ -52,11 +52,41 @@ pub const camera = struct { // MARK: camera
...
+			return .{.min = @min(self.min, other.min), .max = @max(self.max, other.max)};
+		}
+	};
+
+	pub fn triangleAABB(aabb: AABB, triangle: [3]Vec3d) bool {
```
IntegratedQuantum: We don't need this anymore now right?

src/models.zig
```diff
@@ -40,6 +43,7 @@ const ExtraQuadInfo = struct {
 };
 
 const gridSize = 4096;
+const meshGridSize = 16;
```
IntegratedQuantum: The other one is also used for the mesh.
```suggestion
const collisionGridSize = 16;
```

src/models.zig
```diff
@@ -195,9 +200,236 @@ pub const Model = struct {
...
+	}
+
+	fn rasterize(triangle: [3]Vec3f, grid: *[meshGridSize][meshGridSize][meshGridSize]bool, normal: Vec3f) void {
+		var X: usize = undefined;
+		var Y: usize = undefined;
+		var Z: usize = undefined;
+
+		const v0 = triangle[0];
+		const v1 = triangle[1];
+		const v2 = triangle[2];
```
IntegratedQuantum: Instead of constantly multiplying and dividing by the meshGridSize, why not just do it here and only here?

src/models.zig
```diff
@@ -195,9 +200,236 @@ pub const Model = struct {
...
+	}
+
+	fn solveDepth(normal: Vec3f, v0: Vec3f, X: usize, Y: usize, Z: usize, u: f32, v: f32) f32 {
+		const nX = normal[X];
+		const nY = normal[Y];
+		const nZ = normal[Z];
+
+		if(@abs(nZ) < 1e-6) return 0.0;
+
+		const D = -(nX*v0[X] + nY*v0[Y] + nZ*v0[Z]);
```
IntegratedQuantum: Isn't this just
```suggestion
		const D = -vec.dot(v0, normal);
```

src/models.zig
```diff
@@ -195,9 +200,236 @@ pub const Model = struct {
...
+			const xEnd: f32 = @max(xa, xb);
+
+			const voxelXStart: usize = @intFromFloat(@max(xStart*@as(f32, @floatFromInt(meshGridSize)), 0.0));
+			const voxelXEnd: usize = @intFromFloat(@max(xEnd*@as(f32, @floatFromInt(meshGridSize)), 0.0));
+
+			for(voxelXStart..voxelXEnd) |x| {
+				if(x < 0 or x >= meshGridSize) continue;
+				const xf = (@as(f32, @floatFromInt(x)) + 0.5)/@as(f32, @floatFromInt(meshGridSize));
+
+				const zf = solveDepth(normal, v0, X, Y, Z, xf, yf);
```
IntegratedQuantum: Did you try if this function works for non-axis aligned faces?

src/models.zig
```diff
@@ -195,9 +200,236 @@ pub const Model = struct {
...
+
+			rasterize(triangle1, &hollowGrid, quad.normalVec());
+			rasterize(triangle2, &hollowGrid, quad.normalVec());
+		}
+
+		var grid: [meshGridSize][meshGridSize][meshGridSize]bool = .{.{.{true} ** meshGridSize} ** meshGridSize} ** meshGridSize;
+
+		const rawSize = 6*meshGridSize*meshGridSize - 12*meshGridSize + 8;
+		const queueSize = std.math.ceilPowerOfTwo(usize, rawSize) catch unreachable;
+		var floodfillQueue = main.utils.CircularBufferQueue(Vec3i).init(main.stackAllocator, queueSize);
```
IntegratedQuantum: You don't need to hit the exact size, the CircularBufferQueue can resize.

IntegratedQuantum (changes requested):
Also two important performance questions:
1. You should try and print the list sizes for a few models, particularly it would be important to check if that full block only takes one AABB, and that the stairs take 4 or less.
2. What is the debug performance now? Does it still take a significant portion of the loading time?

src/models.zig
codemob-dev: I probably could, but the tiny performance increase doesn't hurt.

src/models.zig
codemob-dev: Yeah, I didn't notice because of the indexing.

src/models.zig
IntegratedQuantum: Resizing is basically zero cost with the stack allocator here, and since you are not allocating anything in between.

src/models.zig
codemob-dev: No, I'll test if the torch model collides correctly.

IntegratedQuantum: You should try a model that is larger and not made from any axis aligned faces, like the grass

IntegratedQuantum: Also would allow crossing #495 off the list.

codemob-dev: Alright, I'll try that.

src/models.zig
codemob-dev: What would I initialize it at, 0?

IntegratedQuantum: 0 is not a power of 2. I usually just pick a random number, that's decently sized like 1024.

src/models.zig
codemob-dev: This does seem to work!

src/models.zig
codemob-dev: Yeah, I probably should've done this from the start.

codemob-dev:
> 1. You should try and print the list sizes for a few models, particularly it would be important to check if that full block only takes one AABB, and that the stairs take 4 or less.

I tested a few models, the chimney has 2 AABBs, cube has 1, and all stairs have 4 or less.

mods/cubyz/rotation/stairs.zig
```diff
@@ -235,6 +235,9 @@ pub fn createBlockModel(_: Block, _: *u16, _: ZonElement) ModelIndex {
 			}
 		}
 		const index = main.models.Model.init(quads.items);
+
+		std.debug.assert(index.model().collision.len <= 4);
```
IntegratedQuantum: Did you mean to commit this?

codemob-dev: No, I'm removing it.

mods/cubyz/rotation/stairs.zig
```diff
@@ -235,6 +235,7 @@ pub fn createBlockModel(_: Block, _: *u16, _: ZonElement) ModelIndex {
 			}
 		}
 		const index = main.models.Model.init(quads.items);
+
```
IntegratedQuantum: forgot a line

codemob-dev:
> 2. What is the debug performance now? Does it still take a significant portion of the loading time?

Model loading went from 7ms to 35ms, so there is only a 28ms difference total.

IntegratedQuantum:
Is that debug mode?

codemob-dev:
> Is that debug mode?

Yeah.

IntegratedQuantum:
Are you sure you measured the right thing? For me block loading went from 131 ms to 576 ms. half a second more.

codemob-dev:
> Are you sure you measured the right thing? For me block loading went from 131 ms to 576 ms. half a second more.

Yeah, it looks like I didn't.

IntegratedQuantum:
I think you should use bit operations instead of operating on individual booleans, I'm sure you can get a large speedup from this by doing up to 16 flood fill or greedy box operations at once.

codemob-dev:
> I'm sure you can get a large speedup from this by doing up to 16 flood fill or greedy box operations at once.

It depends on if flood fill and greedy box takes up that much time, I'll test what takes the longest.

codemob-dev:
It does look like rasterization takes the least amount of time, I'll try this optimization.

codemob-dev:
> I think you should use bit operations instead of operating on individual booleans

Should I assume the grid is 16x16x16, or allow it to be changed in the future?

IntegratedQuantum:
If you make sure to document your assumptions with an `assert(bitSizeOf(...) == ...)`, then you can have both.

codemob-dev:
> If you make sure to document your assumptions with an `assert(bitSizeOf(...) == ...)`, then you can have both.

It turns out you can do this:
```zig
const Uint = @Type(.{ .Int = .{
	.signedness = .unsigned,
	.bits = collisionGridSize,
}});
```

codemob-dev:
Nevermind, that doesn't work

codemob-dev:
Wait I'm stupid `Int` isn't capitalized it does work

IntegratedQuantum:
Also std.meta.Int is a shortcut

codemob-dev:
> I think you should use bit operations instead of operating on individual booleans

I tried this, and it didn't seem to change much, am I missing something or is the optimization not worth it?

IntegratedQuantum:
All you did is reduce the memory usage which doesn't matter since it's below the L1 cache size already.
The important trick is doing multiple operations at once, instead of iterating through one voxel at a time you need to iterate through entire columns at once using bitwise `or`, bitwise `and` and `@ctz`.
If you need a reference, lossy chunk compression does something similar:
https://github.com/PixelGuys/Cubyz/blob/90375b871a67e896ff22400914ae0f3753369154/src/server/storage.zig#L292-L321

codemob-dev:
I honestly can't figure out how to do this, bitwise operations are not my strong suit.

IntegratedQuantum:
I can do it for you, but I'd really recommend to learn this stuff, it's quite powerful, and many data structures and algorithms in Cubyz operate on entire columns of blocks at once using bitwise operations. Without this Cubyz couldn't run as efficient as it does.

codemob-dev:
The main problem I'm seeing is I can't find a good way to flood fill that would actually be faster using bitwise operations, the same goes for the greedy box algorithm.

IntegratedQuantum:
Here you go, would be nice if you could do some more testing of my variant.
I reduced it from ~470 ms to ~180 ms in debug mode, time is distributed like this:
39 ms → rasterizer | 75 ms → flood fill | 66 ms → greedy boxing
I think the flood fill part could still be improved, since the queue is the bottleneck, but I decided to keep the simpler but slower version.

codemob-dev:
Ohh, I see. Thanks for the help!

codemob-dev:
> I think the flood fill part could still be improved, since the queue is the bottleneck, but I decided to keep the simpler but slower version.

It probably doesn't need to be optimized any more, it barely takes any time.

src/game.zig
```diff
@@ -52,176 +52,56 @@ pub const camera = struct { // MARK: camera
...
-		if(!test_axis(a12, v0, v1, v2, box_extents[X]*@abs(f2[Z]) + box_extents[Z]*@abs(f2[X]))) {
-			return false;
-		}
-
-		// Test axis a20
-		const a20 = Vec3d{-f0[Y], f0[X], 0};
-		if(!test_axis(a20, v0, v1, v2, box_extents[X]*@abs(f0[Y]) + box_extents[Y]*@abs(f0[X]))) {
-			return false;
-		}
+	pub const Aabb = struct {
```
IntegratedQuantum: I would like to hear your reasons for renaming this from Box to Aabb.
I personally think that box is the more natural term for this.

codemob-dev: Aabb is the more technical term for it

IntegratedQuantum (changes requested):
The rasterization produces the wrong output for non-axis aligned models.
Here is what it does for the grass model currently (black is what I want, red is the actual model):

raster

Please try to fix this.

src/game.zig
IntegratedQuantum: Yeah, but why should we choose it over the term that anyone understands?

Argmaster: Boxes doesn't have to be axis aligned. This can be interpreted two ways - either we want to expose implementation details in name as a form of documentation or we don't want to do that to not constrain future changes. I would rather opt for latter.

codemob-dev:
image

I got a good visualization working for debugging

codemob-dev:
image
@IntegratedQuantum It seems to be working now

codemob-dev:
Here is the torch collision as well

image

IntegratedQuantum:
Could you also take a look at signs? They are the only block that is not aligned to the 16×16 grid. It appears that, because the faces are offset by 0.5 from the grid, the central face of the collision is missing on one side, which means it's divided quite iniefficiently.
Screenshot at 2025-08-11 15-51-41
Here you can see that it's not selecting the front collision box of the sign, because there is none. It appears the collision box is only generated for the back side of the sign.

codemob-dev:
> It appears the collision box is only generated for the back side of the sign

Weird, I'll have to investigate that.

codemob-dev:
@IntegratedQuantum It works now, I was using the wrong grid size in one of the calculations.

IntegratedQuantum:
found another edge case for fences:
Screenshot at 2025-08-11 19-45-29
I don't think the right column should be intersecting.
Could be a fundamental limitation of the rasterizer approach though.

codemob-dev:
> found another edge case for fences

Do you mean signs?

IntegratedQuantum:
yes

codemob-dev:
> I don't think the right column should be intersecting.

I checked, and it's just barely intersecting.

codemob-dev:
I don't think it really matters though

src/game.zig
```diff
@@ -393,7 +273,7 @@ pub const collision = struct {
 					volumeSum += gridVolume;
 
 					if(_block) |block| {
-						const collisionBox: Box = .{ // TODO: Check all AABBs individually
+						const collisionBox: Box = .{ // TODO: Check all Aabbs individually
```
IntegratedQuantum: unintended change

src/models.zig
```diff
@@ -195,9 +201,221 @@ pub const Model = struct {
...
+	}
+
+	fn solveDepth(normal: Vec3f, v0: Vec3f, X: usize, Y: usize, Z: usize, u: f32, v: f32) f32 {
+		const nX = normal[X];
+		const nY = normal[Y];
+		const nZ = normal[Z];
+
+		if(@abs(nZ) < 1e-6) return 0.0;
+
+		const D = -vec.dot(v0, normal);
```
IntegratedQuantum: (non-type) variables should start with lowercase letters. Also `D` is not a useful name.

src/models.zig
```diff
@@ -195,9 +201,221 @@ pub const Model = struct {
...
+
+		return (-(nX*u + nY*v + D))/nZ;
+	}
+
+	fn rasterize(triangle: [3]Vec3f, grid: *[collisionGridSize][collisionGridSize]CollisionGridInteger, normal: Vec3f) void {
+		var X: usize = undefined;
```
IntegratedQuantum: variables should start with a lowercase letter. I'd suggest calling it `xIndex` to avoid confusion with the x coordinate.

src/models.zig
```diff
@@ -195,9 +201,221 @@ pub const Model = struct {
...
+		if(y1 == y0) return x0;
+		return x0 + (x1 - x0)*(y - y0)/(y1 - y0);
+	}
+
+	fn solveDepth(normal: Vec3f, v0: Vec3f, X: usize, Y: usize, Z: usize, u: f32, v: f32) f32 {
+		const nX = normal[X];
+		const nY = normal[Y];
+		const nZ = normal[Z];
+
+		if(@abs(nZ) < 1e-6) return 0.0;
```
IntegratedQuantum: This cannot even happen, since you chose the direction such that normal[Z] is the largest component of the normal.

src/models.zig
```diff
@@ -195,9 +201,221 @@ pub const Model = struct {
...
+	fn rasterize(triangle: [3]Vec3f, grid: *[collisionGridSize][collisionGridSize]CollisionGridInteger, normal: Vec3f) void {
+		var X: usize = undefined;
+		var Y: usize = undefined;
+		var Z: usize = undefined;
+
+		const v0 = triangle[0]*@as(Vec3f, @splat(@floatFromInt(collisionGridSize)));
+		const v1 = triangle[1]*@as(Vec3f, @splat(@floatFromInt(collisionGridSize)));
+		const v2 = triangle[2]*@as(Vec3f, @splat(@floatFromInt(collisionGridSize)));
+
+		const absNormal = Vec3f{@abs(normal[0]), @abs(normal[1]), @abs(normal[2])};
```
IntegratedQuantum:
```suggestion
		const absNormal = @abs(normal);
```

src/models.zig
```diff
@@ -195,9 +201,221 @@ pub const Model = struct {
...
+		if(p0[1] > p2[1]) {
+			std.mem.swap(Vec2f, &p0, &p2);
+		}
+		if(p1[1] > p2[1]) {
+			std.mem.swap(Vec2f, &p1, &p2);
+		}
+
+		for(@intCast(voxelMin[Y])..@intCast(voxelMax[Y])) |y| {
+			if(y >= collisionGridSize) continue;
+			const yf = @as(f32, @floatFromInt(y)) + 0.5;
```
IntegratedQuantum: I think here lies the reason why your rasterizer doesn't work:
If `yf < p0[1]` or if `yf > p2[1]` then you test values outside the triangle.
Instead I'd suggest to clamp it to the min and max values so we never test points outside of the actual triangle.

src/models.zig
```diff
@@ -195,9 +201,221 @@ pub const Model = struct {
...
+
+			const xStart: f32 = @min(xa, xb);
+			const xEnd: f32 = @max(xa, xb);
+
+			const voxelXStart: usize = @intFromFloat(@max(@floor(xStart), 0.0));
+			const voxelXEnd: usize = @intFromFloat(@max(@ceil(xEnd), 0.0));
+
+			for(voxelXStart..voxelXEnd) |x| {
+				if(x < 0 or x >= collisionGridSize) continue;
+				const xf = @as(f32, @floatFromInt(x)) + 0.5;
```
IntegratedQuantum: Same here

src/models.zig
```diff
@@ -195,9 +201,221 @@ pub const Model = struct {
...
+	}
+
+	fn generateCollision(self: *Model, modelQuads: []QuadInfo) void {
+		var hollowGrid: [collisionGridSize][collisionGridSize]CollisionGridInteger = @splat(@splat(0));
+		const voxelSize: Vec3f = @splat(1.0/@as(f32, collisionGridSize));
+
+		for(modelQuads) |quad| {
+			var shift = Vec3f{0, 0, 0};
+			for(0..3) |i| {
+				if(@abs(quad.normalVec()[i]) == 1.0 and std.math.floor(quad.corners[0][i]*collisionGridSize) == quad.corners[0][i]*collisionGridSize) {
```
IntegratedQuantum: why not use `@floor`?

src/models.zig
```diff
@@ -195,9 +201,221 @@ pub const Model = struct {
...
+					const min = @as(Vec3f, @floatFromInt(boxMin))/@as(Vec3f, @splat(collisionGridSize));
+					const max = @as(Vec3f, @floatFromInt(boxMax))/@as(Vec3f, @splat(collisionGridSize));
+
+					collision.append(Box{.min = min, .max = max}) catch unreachable;
+				}
+			}
+		}
+
+		collision.shrinkAndFree(collision.items.len);
+		self.collision = collision.items;
```
IntegratedQuantum:
```suggestion
		self.collision = collision.toOwnedSlice();
```

src/models.zig
```diff
@@ -287,19 +286,19 @@ pub const Model = struct {
 
 			for(voxelXStart..voxelXEnd) |x| {
 				if(x < 0 or x >= collisionGridSize) continue;
-				const xf = @as(f32, @floatFromInt(x)) + 0.5;
+				const xf = std.math.clamp(@as(f32, @floatFromInt(x)) + 0.5, min[xIndex], max[xIndex]);
```
IntegratedQuantum: This should actually clamp between xStart and xEnd.

src/models.zig
```diff
@@ -195,9 +201,220 @@ pub const Model = struct {
...
+
+					const min = @as(Vec3f, @floatFromInt(boxMin))/@as(Vec3f, @splat(collisionGridSize));
+					const max = @as(Vec3f, @floatFromInt(boxMax))/@as(Vec3f, @splat(collisionGridSize));
+
+					collision.append(Box{.min = min, .max = max}) catch unreachable;
+				}
+			}
+		}
+
+		collision.shrinkAndFree(collision.items.len);
```
IntegratedQuantum: No longer needed, since you use shrinkAndFree.

src/models.zig
```diff
@@ -195,9 +201,220 @@ pub const Model = struct {
...
+			grid[elem.x][elem.y] = newValue;
+
+			if(elem.x != 0) floodfillQueue.pushBack(.{.x = elem.x - 1, .y = elem.y, .val = ~newValue});
+			if(elem.x != collisionGridSize - 1) floodfillQueue.pushBack(.{.x = elem.x + 1, .y = elem.y, .val = ~newValue});
+			if(elem.y != 0) floodfillQueue.pushBack(.{.x = elem.x, .y = elem.y - 1, .val = ~newValue});
+			if(elem.y != collisionGridSize - 1) floodfillQueue.pushBack(.{.x = elem.x, .y = elem.y + 1, .val = ~newValue});
+			floodfillQueue.pushBack(.{.x = elem.x, .y = elem.y, .val = ~newValue << 1 | ~newValue >> 1});
+		}
+
+		var collision: std.ArrayList(Box) = .init(main.globalAllocator.allocator);
```
IntegratedQuantum: Should use `main.List` instead of `std.ArrayList`

IntegratedQuantum:
also CI is failing

src/renderer.zig
```diff
@@ -1146,7 +1146,10 @@ pub const MeshSelection = struct { // MARK: MeshSelection
 	pub fn render(projectionMatrix: Mat4f, viewMatrix: Mat4f, playerPos: Vec3d) void {
 		if(main.gui.hideGui) return;
 		if(selectedBlockPos) |_selectedBlockPos| {
-			drawCube(projectionMatrix, viewMatrix, @as(Vec3d, @floatFromInt(_selectedBlockPos)) - playerPos, selectionMin, selectionMax);
+			const block = mesh_storage.getBlockFromRenderThread(_selectedBlockPos[0], _selectedBlockPos[1], _selectedBlockPos[2]) orelse return;
+			for(block.mode().model(block).model().collision) |aabb| {
+				drawCube(projectionMatrix, viewMatrix, @as(Vec3d, @floatFromInt(_selectedBlockPos)) - playerPos, @floatCast(aabb.min), @floatCast(aabb.max));
+			}
```
IntegratedQuantum: looks like you accidentally committed your debug visualization

src/models.zig
```diff
@@ -195,9 +201,219 @@ pub const Model = struct {
...
+					const min = @as(Vec3f, @floatFromInt(boxMin))/@as(Vec3f, @splat(collisionGridSize));
+					const max = @as(Vec3f, @floatFromInt(boxMax))/@as(Vec3f, @splat(collisionGridSize));
+
+					collision.append(Box{.min = min, .max = max});
+				}
+			}
+		}
+
+		collision.shrinkAndFree(collision.items.len);
+		self.collision = collision.items;
```
IntegratedQuantum: Again, please use toOwnedSlice, which is a safer helper function for shrinkAndFree+copying the items pointer.

IntegratedQuantum (approved):

PR #1756, opened 2025-08-11
Vine assets

careeoki:
for #1742 #1030

Willow and mahogany vine textures + items. This has newer willow vine textures compared to the ones in #1742

image


codemob-dev:
ooh nice, I like that!

IntegratedQuantum (approved):

PR #1760, opened 2025-08-14
Chest Animations

OneAvargeCoder193:
Things that need to be done before this is ready for review:

- [x] Simplify the chest shaders
- [x] Maybe separate the callback stuff into its own pr
- [x] ~Find a way to change the chest model on server side when the chest closes~
- [x] Add the rest of the chest textures
- [x] Make sure it works in multiplayer


assets/cubyz/shaders/block_entity/chest.vert
```diff
@@ -0,0 +1,113 @@
+#version 460
```
IntegratedQuantum: These shaders look they are almost an exact copy of the chunk shaders.

This is unmaintainable (and also unreviewable). Please find a way to reuse the same shader, you can for example use preprocessor options to distinguish the things that are different.

src/block_entity.zig
```diff
@@ -282,10 +367,149 @@ pub const BlockEntityTypes = struct {
...
+
+					main.renderer.chunk_meshing.faceBuffers[0].ssbo.bind(main.renderer.chunk_meshing.faceBuffers[0].binding);
+					main.renderer.chunk_meshing.lightBuffers[0].ssbo.bind(main.renderer.chunk_meshing.lightBuffers[0].binding);
+					c.glActiveTexture(c.GL_TEXTURE0);
+					main.blocks.meshes.blockTextureArray.bind();
+					c.glActiveTexture(c.GL_TEXTURE1);
+					main.blocks.meshes.emissionTextureArray.bind();
+					c.glActiveTexture(c.GL_TEXTURE2);
+					main.blocks.meshes.reflectivityAndAbsorptionTextureArray.bind();
+					c.glDrawElementsInstancedBaseVertexBaseInstance(c.GL_TRIANGLES, @intCast(6*faceData.items.len), c.GL_UNSIGNED_INT, null, 1, allocation.start*4, chunkAllocation.start);
```
IntegratedQuantum: There is a lot of stuff going on in this render function.

You are likely not going to be the last person who will want to render block models, so please extract all common functionality into a helper function usable by everyone.

src/renderer/chunk_meshing.zig
```diff
@@ -24,7 +24,7 @@ const gpu_performance_measuring = main.gui.windowlist.gpu_performance_measuring;
 
 const mesh_storage = @import("mesh_storage.zig");
 
-var pipeline: graphics.Pipeline = undefined;
+pub var pipeline: graphics.Pipeline = undefined;
```
IntegratedQuantum: Why is this public now?

src/block_entity.zig
IntegratedQuantum: While you are at it, maybe you can also cover the item display model use case?
If you do, this could be a good thing to extract into a separate PR.

IntegratedQuantum (changes requested):

src/block_entity.zig
OneAvargeCoder193: Do you mean like changing how the items are rendered?

OneAvargeCoder193: Which file should this be extracted to?

IntegratedQuantum: Where would you look for it first?
Maybe renderer.zig?

IntegratedQuantum: > Do you mean like changing how the items are rendered?

They are rendered with the blocks shader currently, so it probably wouldn't be a big change.

OneAvargeCoder193: I think I'll do the item things after this is merged because I'm not very good at extracting stuff it seems.

assets/cubyz/shaders/chunks/chunk_vertex.vert
```diff
@@ -97,6 +100,9 @@ void main() {
 
 	position += vec3(quads[quadIndex].corners[vertexID][0], quads[quadIndex].corners[vertexID][1], quads[quadIndex].corners[vertexID][2]);
 	position *= voxelSize;
+#ifdef ENTITY
+	position = (modelMatrix*vec4(position - vec3(1), 1)).xyz + vec3(1);
```
IntegratedQuantum: the - 1 and + 1 are changing the center of rotation. In my opinion this belongs in the model matrix.

src/renderer.zig
```diff
@@ -330,6 +359,81 @@ pub fn renderWorld(world: *World, ambientLight: Vec3f, skyColor: Vec3f, playerPo
 	gpu_performance_measuring.stopQuery();
 }
 
+pub fn renderBlock(projMatrix: Mat4f, modelMatrix: Mat4f, block: blocks.Block, lighting: u32, ambientLight: Vec3f, playerPosition: Vec3d) void {
```
IntegratedQuantum: lighting needs to be sampled for each corner using the same ambient occlusion code as for the regular meshes. Otherwise the animated part of the mesh looks out of place due to missing ambient occlusion.

src/renderer/chunk_meshing.zig
```diff
@@ -1262,6 +1262,13 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
 		}
 		self.mutex.lock();
 		self.chunk.data.setValue(chunk.getIndex(x, y, z), newBlock);
+
+		if(newBlock.blockEntity()) |blockEntity| {
+			var reader = main.utils.BinaryReader.init(blockEntityData);
+			blockEntity.updateClientData(.{_x, _y, _z}, self.chunk, .{.update = &reader}) catch |err| {
```
IntegratedQuantum: I deliberately removed the update on create event in #1739 because it promotes the wrong design choices.

You should make a system that is able to handle uninitialized block entities in any state, instead of relying on the fact that everything is initialized correctly upfront, which is not happening.

IntegratedQuantum (changes requested):

assets/cubyz/shaders/chunks/chunk_vertex.vert
OneAvargeCoder193: The reason this is done is so that the origin is actually where the block is, since the block is at 1, 1, 1.

IntegratedQuantum: Well technically it's also scaled by voxelSize at that point.
How about making the transform act before the chunk offset is added and all these other transformations are done?

OneAvargeCoder193: It shouldn't matter since renderBlock always has a voxelSize of 1, and it needs to be done to the += result otherwise i get nonsensical results when rotating.

IntegratedQuantum: Hmm, I see. Well in that case, I guess the best you can do is add a comment to explain why this is here.

src/renderer.zig
OneAvargeCoder193: How do i do that though? Would i have to do it for every single quad?

IntegratedQuantum: yes, for every single vertex. So it's probably best to do it in this function as well.

src/renderer/chunk_meshing.zig
OneAvargeCoder193: I think this code doesn't need to be here anymore since the chests are by default full blocks.

src/renderer.zig
OneAvargeCoder193: I can't do it the way its done for signs, because that requires the lid actually be in the world. I need to somehow calculate it

careeoki:
image
The latch plane is z fighting. it also has no backface
image
and i got it stuck open after spamming it a bunch

careeoki:
image

and i crashed after continuing to spam open and close chests

OneAvargeCoder193:
Weird, it shouldn't be z fighting at all.

careeoki:
Made a PR for the textures on your fork

OneAvargeCoder193:
Thanks!

OneAvargeCoder193:
> image
> and i crashed after continuing to spam open and close chests

I don't know if I am able to fix this because it seems to be an issue where onClose is called and the client side inventory is closed but the gui is still rendering

OneAvargeCoder193:
Or maybe it is something else i need to look into it

assets/cubyz/blocks/textures/chest/baobab/latch.png
IntegratedQuantum, line 1: The latch and front of the open chest top should not be drawn as separate faces.

src/Inventory.zig
```diff
@@ -448,8 +448,8 @@ pub const Sync = struct { // MARK: Sync
 			return null;
 		}
 
-		pub fn getInventoryFromId(serverId: InventoryId) Inventory {
-			return inventories.items[@intFromEnum(serverId)].inv;
+		pub fn getServerInventoryFromId(serverId: InventoryId) ServerInventory {
+			return inventories.items[@intFromEnum(serverId)];
```
IntegratedQuantum, line 456: I guess I forgot this earlier, please add an assertion that the mutex is locked to the start of the function.

src/block_entity.zig
```diff
@@ -237,23 +260,37 @@ pub const BlockEntityTypes = struct {
...
+			inv.inv.toBytes(writer);
 		}
-		pub fn onStoreServerToClient(_: BlockEntityIndex, _: *BinaryWriter) void {}
 		pub fn onInteract(pos: Vec3i, _: *Chunk) EventStatus {
 			if(main.KeyBoard.key("shift").pressed) return .ignored;
 
+			blk: {
+				var block = main.renderer.mesh_storage.getBlockFromRenderThread(pos[0], pos[1], pos[2]) orelse break :blk;
+				block.data |= 4;
+				main.renderer.mesh_storage.updateBlock(.{.x = pos[0], .y = pos[1], .z = pos[2], .newBlock = block, .blockEntityData = &.{}});
```
IntegratedQuantum: Now what happens if the client received an update from the server that the chest was broken, just before you do this update here?
Note that mesh updates from this function are not processed directly but in the next frame.

IntegratedQuantum (changes requested):

src/renderer.zig
```diff
@@ -54,11 +54,30 @@ var fakeReflectionUniforms: struct {
...
+	lodDistance: c_int,
+	zNear: c_int,
+	zFar: c_int,
+} = undefined;
 
 pub var activeFrameBuffer: c_uint = 0;
 
 pub const reflectionCubeMapSize = 64;
-var reflectionCubeMap: graphics.CubeMapTexture = undefined;
+pub var reflectionCubeMap: graphics.CubeMapTexture = undefined;
```
IntegratedQuantum: this doesn't need to be public anymore

src/server/world.zig
```diff
@@ -1204,7 +1204,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		baseChunk.mutex.lock();
 		const currentBlock = baseChunk.getBlock(x, y, z);
 		if(oldBlock != null) {
-			if(oldBlock.? != currentBlock) {
+			if(oldBlock.?.typ != currentBlock.typ) {
```
IntegratedQuantum: cmpxchg must only succeed if the blocks are exactly the same. Otherwise race conditions can occur (e.g. two players break a torch from the same block, now you may end up with a duped torch)

src/renderer/chunk_meshing.zig
```diff
@@ -1196,7 +1196,7 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
 		self.mutex.lock();
 		const oldBlock = self.chunk.data.getValue(chunk.getIndex(x, y, z));
 
-		if(oldBlock == newBlock) {
+		if(oldBlock.typ == newBlock.typ and blockEntityData.len != 0) {
```
IntegratedQuantum: Can you explain why you added the `!= 0` check? Data of length 0 may still be valid, and rejecting it here may be unexpected. If you want to encode a null value, then you should make it optional.

src/renderer.zig
```diff
@@ -330,6 +359,103 @@ pub fn renderWorld(world: *World, ambientLight: Vec3f, skyColor: Vec3f, playerPo
...
+	model.appendInternalQuadsToList(&faceData, main.stackAllocator, block, 1, 1, 1, false);
+	for(main.chunk.Neighbor.iterable) |neighbor| {
+		model.appendNeighborFacingQuadsToList(&faceData, main.stackAllocator, block, neighbor, 1 + neighbor.relX(), 1 + neighbor.relY(), 1 + neighbor.relZ(), false);
+	}
+
+	for(faceData.items) |*face| {
+		face.position.lightIndex = 0;
+	}
+
+	var allocation: graphics.SubAllocation = .{.start = 0, .len = 0};
```
IntegratedQuantum: Is this even used?

Please remove all the unused code in this function before I review further.

IntegratedQuantum (changes requested):

src/Inventory.zig
OneAvargeCoder193: In order to do this I had to unlock the inventory mutex when onFirstOpen and onLastClose callbacks were called, and re lock them afterwards, otherwise there would be a mutex locked in one place a function was called, but not in the other.

src/renderer.zig
OneAvargeCoder193: Isn't this passed into chunk_meshing.faceBuffers[0].uploadData?

src/Inventory.zig
IntegratedQuantum: that is likely not thread safe. I really don't like your attitude towards multithreading of "doing random things until the assertions no longer fail". There is likely a good reason why the mutex was locked in one place while the function was called. By unlocking it you may cause undesired data races.

src/renderer.zig
IntegratedQuantum: yes, that's where the allocation is created. But where is it used?

OneAvargeCoder193: The allocation is used inside uploadData when it calls glMapBufferRange

IntegratedQuantum: yes, that's where the allocation is created. But where is it used?

OneAvargeCoder193: I'm confused now, this is the exact same as how it is done inside generateBlockTexture, it creates the allocation variable and passes it into uploadData, without using it anywhere afterwards.

OneAvargeCoder193: Nevermind, i found where its used

OneAvargeCoder193: > yes, that's where the allocation is created. But where is it used?

Its passed into the call for c.glDrawElementsInstancedBaseVertexBaseInstance

IntegratedQuantum: look again, it is being used in generateBlockTexture.

IntegratedQuantum: > Its passed into the call for c.glDrawElementsInstancedBaseVertexBaseInstance

Ok, I didn't see that.
Either way the data you upload here is incorrect, the lightIndex is set to zero, and we don't want the same light values for all faces.

src/block_entity.zig
OneAvargeCoder193: I'm not exactly sure what to do in this situation

IntegratedQuantum: Don't put it in the same queue, instead you could update the block directly. Then the server update which is in the queue would still correctly break the block afterwards.

OneAvargeCoder193:
> and i got it stuck open after spamming it a bunch

Are you still able to reproduce this?

src/renderer.zig
OneAvargeCoder193: I think I fixed this.

IntegratedQuantum:
There are still some issues here, likely because my testing chest happened to be on a chunk border:

https://github.com/user-attachments/assets/66ac4722-0135-476d-b938-6e54c58d3fb7

IntegratedQuantum:
Also could you maybe split the PR into smaller steps, I really do not have the patience to review (and rereview) a 500 line PR right now.

OneAvargeCoder193:
> Also could you maybe split the PR into smaller steps, I really do not have the patience to review (and rereview) a 500 line PR right now.

I'm able to remove the renderBlockLit function from this pr, but a lot of the lines are from assets and the chest rotation mode, which I don't see how to split off.

BoySanic:
CI failed and there are merge conflicts. Please revisit when you have time!

Wbjitscool:
Whats causing the conflicts?

OneAvargeCoder193:
> Whats causing the conflicts?

old pr and i need to basically redo this now

Wbjitscool:
ah okay

Issue #1467, opened 2025-05-18
Plateau Biome


ikabod-kee:
![Image](https://github.com/user-attachments/assets/4c7e2c8f-c32e-4b18-90ce-abd43b498dcd)

I was in Development VC with others, checking out my old Cubyz versions, when we came across the plateau biome I made a while ago. They want it in. We'll need #843 for this.

IntegratedQuantum:
We have plateau biomes now

Issue #843, opened 2024-12-21
Ridges/cliffs parameter for biomes


ikabod-kee:
Sort of like when you set interpolation to ".none," except within just the confines of the biome and not the edge.

![image](https://github.com/user-attachments/assets/03e3320f-3f2e-4d16-955d-9db48826b242)

codemob-dev:
> Sort of like when you set interpolation to ".none," except within just the confines of the biome and not the edge.

How would this work?

codemob-dev:
This might not give exactly the same effect, but what if we allowed different post-processing functions to the heightmap, and we use [a smooth rounding function](https://www.desmos.com/calculator/vqqlqh4u5y) to get cliff-like shapes?

codemob-dev:
or maybe [this](https://www.desmos.com/calculator/vqqlqh4u5y)?

PR #1780, opened 2025-08-20
Heightmap operations

codemob-dev:
Adds heightmap operations which act as a way to generate more unique terrain by reusing heightmaps for different purposes.

Also adds a `stepped` heightmap operation and reworks the ferrock mountains to use it:
image


codemob-dev:
It will also probably help with #1467 and #843

codemob-dev:
Also may help generate better buttes: #1108

codemob-dev:
Could maybe help with #344?

codemob-dev:
Performance comparison (debug mode) of loading a world:

With PR:
image

Without PR:
image

IntegratedQuantum:
Some thoughts from my side:
- If it only "maybe helps", then I think that is a sign that this is not the right approach. Sure it does give more freedom to modders, but it doesn't seem to be the right kind of freedom. In my mind the right solution would be to produce better inputs (e.g. per-biome custom noise maps, maybe even loaded from a file) instead of trying to fix the final output, which is going to be a lot more limited.
- the interpolation between biomes is problematic, its influence is limited to a 32×32 area which is insufficient if you want to do anything fancy with it. (also from a quick glance it doesn't seem like your implementation would even work correctly)
Interpolation between the other biome parameters is already done during climategen, ideally this should also be done for any other effects.
- Making this a mod seems to be overkill. Splines would allow using this in addons as well.
- measuring performance in debug mode is not useful here, most terrain is generated in release modes.

codemob-dev:
> per-biome custom noise maps, maybe even loaded from a file

So instead of generating noise maps, biomes could use a premade file? Wouldn't this make biomes more repetitive?

> the interpolation between biomes is problematic, its influence is limited to a 32×32 area which is insufficient if you want to do anything fancy with it.

Yeah, that is a big problem that I did want to try to fix.

> Making this a mod seems to be overkill. Splines would allow using this in addons as well.

Splines have much less control. The `stepped` operation uses the mod function and a few other things that splines can't do. If we do go with using the mod system, it could make sense to allow biomes to specify parameters in the heightmap operations similar to rotation modes, and have a spline operation type that could have a similar use.

> measuring performance in debug mode is not useful here, most terrain is generated in release modes.

Ah, ok, I couldn't remember how to run it in `releaseFast`, so I just went with debug.

codemob-dev:
> but it doesn't seem to be the right kind of freedom

Why not? It opens a lot of new possibilities for unique terrain.

IntegratedQuantum:
> Wouldn't this make biomes more repetitive?

No, it would make all biomes that use this specific map more repetitive, so if you have something like a 2k noise map and layer different noise maps on top of each other with random offsets it will hardly be noticable to the player.

Furthermore note that it actually works surprisingly well. Here is an example of a 32×32 noise map, and just the interpolation of the biome base height is good enough to add some variation. So I think something like a 2k×2k noise map would not be noticable at all, unless you have some very recognizable patterns in it.
Screenshot_at_2024-11-19_22-07-04

And lastly we could still allow the option to generate the noise, so it would be repetitive for the expensive noise patterns (e.g. erosion), but not for the cheap ones.

> Splines have much less control

Splines can approximate any function on a fixed interval to an arbitrary precision, including mod.

> I couldn't remember how to run it in releaseFast

Double-click the run script. It's equally as good to measure in ReleaseSafe.

> Why not? It opens a lot of new possibilities for unique terrain.

This is the wrong question. The right question would be "How do we actually solve the underlying problems/limitations?"
And if, and only if, the answer to that question entails heightmap operations, then I'd be happy to add it.
But for now I don't see how heightmap operations help here, other than being a temporary bandaid, it's slightly better than what we have now, but I don't think that's good enough and I don't think we should waste our effort into an insufficient solution.

The same thing happened with #1179 where we now have an infinitely better solution than the originally proposed.

codemob-dev:
Yeah, I see what you mean, I'll see if I can make a system to do that. Also, can I see how you loaded and handled that noise map?

IntegratedQuantum:
It's an old screenshot, I think I only just generated a repeated seed or something like that instead of actually loading anything.
When I tried it I couldn't think of an easy solution for
> the interpolation between biomes is problematic, its influence is limited to a 32×32 area which is insufficient if you want to do anything fancy with it.

which is why I didn't investigate further at the time.

IntegratedQuantum:
Oh and I also was thinking about rotating the heightmaps randomly, which would have made the interpolation even more complicated.

IntegratedQuantum:
Because I just found this in-game, here is an excellent example of the types of patterns that you cannot produce with heightmap operations:

Screenshot at 2025-08-22 17-04-43

codemob-dev:
Neat!

PR #1781, opened 2025-08-21
Fix an accidentall memcpy in NoiseBasedVoronoi

codemob-dev:
Helps with #107, but NoiseBasedVoronoi is still slow.


IntegratedQuantum (approved):
I have looked into this, and it seems that it only happens in ReleaseSafe mode (which explains why I didn't catch this when I wrote this originally).
I'm guessing llvm doesn't like the fact that the panic function may have side effect, so it just makes a copy just in case the original is modified by the panic function.

PR #1793, opened 2025-08-29
Random blocks in structures

codemob-dev:
Allows specifying a selection of multiple blocks in simple structures:
image
image

### DIFF
--- a/src/server/terrain/simple_structure_utils.zig
+++ b/src/server/terrain/simple_structure_utils.zig
@@ -1,0 +1,58 @@
+const std = @import("std");
+const main = @import("main");
+const ZonElement = main.ZonElement;
+const AliasTable = main.utils.AliasTable;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+pub const BlockSelector = struct {
+	blocks: AliasTable(Entry),
+
+	const Entry = struct {
+		block: main.blocks.Block,
+		chance: f32,
+	};
+
+	pub fn parse(allocator: NeverFailingAllocator, zon: ZonElement, defaultBlock: []const u8) BlockSelector {
+		var items: []Entry = undefined;
+		switch(zon) {
+			.string, .stringOwned => {
+				items = allocator.alloc(Entry, 1);
+				items[0] = .{
+					.block = main.blocks.parseBlock(zon.as([]const u8, "")),
+					.chance = 1.0,
+				};
+			},
+			.array => {
+				items = allocator.alloc(Entry, zon.array.items.len);
+				for (0.., items) |i, *item| {
+					const element = zon.array.items[i];
+					if(element == .string or element == .stringOwned) {
+						item.* = .{
+							.block = main.blocks.parseBlock(element.as([]const u8, "")),
+							.chance = 1.0,
+						};
+					} else if(element == .object) {
+						item.* = .{
+							.block = main.blocks.parseBlock(element.get([]const u8, "block", "")),
+							.chance = element.get(f32, "chance", 1.0),
+						};
+					}
+				}
+			},
+			else => {
+				items = allocator.alloc(Entry, 1);
+				items[0] = .{
+					.block = main.blocks.parseBlock(defaultBlock),
+					.chance = 1.0,
+				};
+			}
+		}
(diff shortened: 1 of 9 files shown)


IntegratedQuantum (changes requested):
I would like to know the performance impact of this, both in absolute numbers and relative to the previous cost of generating these particular models.

codemob-dev:
The highlighted block here is the query for the random block.

image

The master branch for reference:
image

codemob-dev:
Compared to everything else, the performance impact is negligible.

IntegratedQuantum:
> Maybe the next person wants to have a striped pattern

Another thing to add to the list of things we'd want: random selection per biome/structure, like currently for the crystal caves.

Argmaster:
Ok, I want this very badly for jungle biome.
I think overloading those structures with more features doesn't make sense. We should allow randomized blocks because we can do that without much changes in configuration because we have Patterns (#1237) and I would advise to use them for consistency.
Realistically there will always be something to add to the generator, if not stripes, then crosses, circles or whatever, we can't exhaust the options. We already have SBBs to handle stuff that is more specific.

src/server/terrain/simple_structure_utils.zig
```diff
@@ -0,0 +1,58 @@
+const std = @import("std");
+const main = @import("main");
+const ZonElement = main.ZonElement;
+const AliasTable = main.utils.AliasTable;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+pub const BlockSelector = struct {
```
Argmaster: This is already implemented in `src/blueprint.zig` through `Pattern` struct,

codemob-dev: Huh, I forgot that existed.

codemob-dev: I got this implemented although we now have this monstrosity as the specifier:
image

IntegratedQuantum: Hmm, I don't like this. Those numbers should be zon.

codemob-dev: This is using the pattern parser but I can make a new one. If I do, should it be in `simple_structure_utils` or should it be with the rest of the `Pattern` code?

IntegratedQuantum: Probably in the Pattern struct like the other parser.

IntegratedQuantum: Also @Argmaster should we maybe also make the pattern parsing for commands zon-based?

codemob-dev: I think we should have a better UI for commands first in that case, we don't want to end up like minecraft where more complicated commands have to be written in a separate text editor.

IntegratedQuantum: Well, honestly I'd even argue we shouldn't have any commands, and only UI.

codemob-dev: With dedicated servers, a UI isn't an option.

IntegratedQuantum: yes it is, just join the server :​P

codemob-dev: I got the zon parser working now.

IntegratedQuantum:
> Realistically there will always be something to add to the generator, if not stripes, then crosses, circles or whatever, we can't exhaust the options. We already have SBBs to handle stuff that is more specific.

Well I pointed out stripes in particular because I think it would nice to have a consistent solution for both structures and the biome underground (→stoneBlock).
This would benefit everyone (→stripes in world edit, stripes in structures, random blocks in the terrain) and allow us to extend this with more options in the future.
And sure we have SBBs for some of the options, but giving more options to randomize SBBs is going to help make them less repetitive.

So I'd like both of you @codemob-dev @Argmaster to rethink this.

codemob-dev:
Perhaps we need a more abstracted pattern system that encompasses stripes, maybe ground structure, and other things we might want to add.

codemob-dev:
I'm closing this as I would have to rewrite most of this to implement the required features.

PR #1808, opened 2025-09-01
Oxygen bar icons

careeoki:
image
edit, new shape


Argmaster:
My mind may be cursed.
And I am very sorry to say that.
But those lungs look ambiguous.

careeoki:
🤨hmmmmmmmmmmmmmmmmmmmmmm..... well, i can make the lungs 1 pix taller
image

ikabod-kee:
I like the strange theming with organs. It's fun

IntegratedQuantum:
I wonder how it would look if we could depict a deflated lung for the no-oxygen case.

IntegratedQuantum (approved):
I guess we can merge this

Issue #1477, opened 2025-05-20
Alias `arena.allocator()` in all files that use it directly


Argmaster:
With "managed" data structures `arena.allocator()` appeared only during construction of data structures, but with "unmanaged" variants it started appearing everywhere. We should alias `arena.allocator()` as `arena` and rename current `arena` variable to `arenaAllocator`.

- [ ] `src/items.zig`
- [ ] `src/blocks.zig`
- [ ] `src/graphics.zig`
- [ ] `src/server/terrain/biomes.zig`

psyomara:
#1861 refactors them all.

PR #1824, opened 2025-09-06
Recipe rewrite

### LINKED ISSUE #1348: Allow tags in recipes
## Description

Currently recipes can only use `BaseItem` as input and output. It would be nice to have a system that would allow using tags to specify input and output of recipe to avoid duplicating all the recipes for all the wood types and alike. In UI Input/output specified via tag would be displayed like normal one, but would cycle through all the items matching given tag, changing e.g. every second.
Just to be clear: Idea is to not generate all the variants of eg. log to panks recipe, but to have one recipe that can take any log type and spits out correct plank type.

## Advantages:
- Reduced length of crafting list.
- Less boilerplate in recipe definitons.

## Problems:
- how to map log type to correct plank type?

## Alternatives:
- Allow tags just in recipe definitions and automatically generate all recipes based on those tags.

codemob-dev:
Fixes #1348

There is still a lot of cleaning up, but all the functionality is there.


codemob-dev:
It appears to be working, it's review time now.

assets/cubyz/recipes/special_recipes.zig.zon
```diff
@@ -1,47 +1,11 @@
 .{
 	.{
-		.inputs = .{"cubyz:birch_planks", "cubyz:iron_ingot"},
+		.inputs = .{".planks", "cubyz:iron_ingot"},
```
IntegratedQuantum: Do we really need tags for this?

Can't this also be `{mod}:{type}_planks` or (once these files are better sorted) `{mod}:planks/{type}`?

src/items.zig
```diff
@@ -23,6 +23,8 @@ const Vec3f = vec.Vec3f;
 const modifierList = @import("tool/modifiers/_list.zig");
 const modifierRestrictionList = @import("tool/modifiers/restrictions/_list.zig");
 
+const recipe_parser = @import("recipe_parser.zig");
```
IntegratedQuantum: If you move these things into a separate file, then I think `items/recipes.zig` would be more fitting, so we can later move all other recipe related functionality there.

src/items.zig
```diff
@@ -1207,50 +1209,17 @@ pub fn registerTool(assetFolder: []const u8, id: []const u8, zon: ZonElement) vo
...
-	for(inputs, 0..) |inputZon, i| {
-		const input = try parseRecipeItem(inputZon);
-		recipe.sourceItems[i] = input.item.?.baseItem;
-		recipe.sourceAmounts[i] = input.amount;
-	}
-	return recipe;
-}
-
 pub fn registerRecipes(zon: ZonElement) void {
+	var tempAllocator: main.heap.NeverFailingArenaAllocator = .init(main.globalAllocator);
```
IntegratedQuantum: Please use `defer deinit` here instead of resetting at the end

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+					.sourceItems = list.allocator.alloc(BaseItemIndex, 1),
+					.sourceAmounts = list.allocator.alloc(u16, 1),
+					.resultItem = recipe.sourceItems[0],
+					.resultAmount = recipe.sourceAmounts[0],
+				};
+				reversedRecipe.sourceItems[0] = recipe.resultItem;
+				reversedRecipe.sourceAmounts[0] = recipe.resultAmount;
+				list.append(reversedRecipe);
+			} else {
+				return error.InvalidReversibleRecipe;
```
IntegratedQuantum: Generally it's more readable to have the error path first, with an else branch it is harder to see what the actual condition is.
```zig
if(recipe.sourceItems.len == 1) {
	...
} else {
	return error.InvalidReversibleRecipe;
}
```
↓
```zig
if(recipe.sourceItems.len != 1) {
	return error.InvalidReversibleRecipe;
}
...
```

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+		inputCombos.deinit();
+		inputCombos = newInputCombos;
+	}
+	return inputCombos;
+}
+
+pub fn parseRecipe(allocator: NeverFailingAllocator, zon: ZonElement, list: *main.List(Recipe)) !void {
+	const inputs = zon.getChild("inputs").toSlice();
+	const recipeItems = allocator.alloc(ZonElement, inputs.len + 1);
+	@memmove(recipeItems[0..inputs.len], inputs);
```
IntegratedQuantum: If newly allocated memory aliases with another slice then that would be a sign of a big problem
```suggestion
	@memcpy(recipeItems[0..inputs.len], inputs);
```

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+	recipeItems[inputs.len] = zon.getChild("output");
+	defer allocator.free(recipeItems);
```
IntegratedQuantum: The `defer` should always be directly after the allocation.

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+
+	const itemCombos = try generateItemCombos(allocator, recipeItems);
+	defer itemCombos.deinit();
+	for(itemCombos.items) |itemCombo| {
+		defer allocator.free(itemCombo);
```
IntegratedQuantum: If an error happens, then this is leaking memory. You need to use a separate `for` loop in a `defer` block.

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+		const parsedInputs = itemCombo[0 .. itemCombo.len - 1];
+		const output = itemCombo[itemCombo.len - 1];
```
IntegratedQuantum: Why encode the data like this instead of storing it with the type system as a
`struct {output:T, input:ListUnmanaged(T)}`?

src/items.zig
IntegratedQuantum: Generally I also find it questionable to use a temporary arena here, especially when the data appears to be freed inside the function anyways.
Can this not just use `main.stackAllocator`?

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+		for(inputCombos.items) |combo| {
+			allocator.free(combo);
+		}
+		inputCombos.deinit();
+		inputCombos = newInputCombos;
+	}
+	return inputCombos;
+}
+
+pub fn parseRecipe(allocator: NeverFailingAllocator, zon: ZonElement, list: *main.List(Recipe)) !void {
```
IntegratedQuantum: Generally I think the right pattern is to use `main.stackAllocator` for all temporary allocations and explicitly pass in the allocators used for the result.

Here you do the opposite. For the result the implicitly passed `list.allocator` is used, whereas an allocator for temporary allocations is passed in.

This is confusing, since you take a mostly random allocator for the result data.
Furthermore taking an arbitrary allocator for local allocations, means you have made a commitment to support an arbitrary allocator, when in fact you are currently leaking memory, which is only allowed with an arena.

Now I agree that an arena could help simplify the allocation process here, but I'd prefer one arena per function so it's immediately clear that this is a local arena, and inheriting from the stackAllocator since it is more efficient that way. Furthermore, when using such an arena, you should not bother at all with freeing the memory, since it just makes the code harder to read.

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+			for(parsedItems.items) |item| {
+				const newInputs = allocator.dupe(ItemStack, inputs);
+				newInputs[startIndex] = item.item;
+				newInputCombos.append(newInputs);
+				newKeyList.append(item.keys);
+			}
+		}
+
+		for(keyList.items) |*keys| {
+			keys.deinit();
```
IntegratedQuantum: deinit should always be done with a defer, before the loop (right after the list creation). This is currently leaking memory on error.

IntegratedQuantum: same below

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+		}
+		itemPairs.append(.{
+			.item = .{
+				.item = .{.baseItem = item},
+				.amount = amount,
+			},
+			.keys = keys.clone() catch unreachable,
+		});
+	} else {
+		if(std.mem.eql(u8, "{mod}:branch/leafy/{type}", id)) {
```
IntegratedQuantum: Why is there a hardcoded string here?

src/recipe_parser.zig
```diff
@@ -0,0 +1,264 @@
...
+		keyList.deinit();
+	}
+	for(startingParsedItems.items) |item| {
+		const inputs = allocator.alloc(ItemStack, recipe.len);
+		inputs[0] = item.item;
+		inputCombos.append(inputs);
+		keyList.append(item.keys);
+	}
+	while(remainingItems.len > 1) {
+		remainingItems = remainingItems[1..];
```
IntegratedQuantum: I think the standard library has an iterator type for this.

IntegratedQuantum (changes requested):
I have not yet read everything. Before I continue my review, please apply my suggestion regarding arena allocation from the long comment.

src/recipe_parser.zig
codemob-dev: Mostly because the logic for the item parser is easier that way

IntegratedQuantum:
you forgot to commit the file

codemob-dev:
Oh, whoops

src/recipe_parser.zig
codemob-dev: That was old debugging stuff I forgot to remove, I got that now.

src/items/recipes.zig
```diff
@@ -0,0 +1,240 @@
...
+				newInputCombos.append(newInputs);
+				newKeyList.append(item.keys);
+			}
+		}
+		keyList = newKeyList;
+		inputCombos = newInputCombos;
+	}
+	const newInputCombos: main.List([]ItemStack) = .initCapacity(allocator, inputCombos.items.len);
+
+	return newInputCombos;
```
IntegratedQuantum: This seems wrong, why are you returning an uninitialized list here?

codemob-dev: That was a mistake, I haven't finished yet.

src/recipe_parser.zig
codemob-dev: I set this up now.

assets/cubyz/recipes/special_recipes.zig.zon
codemob-dev: Yeah, I guess that makes sense.

src/recipe_parser.zig
codemob-dev: There isn't one for exactly what I'm doing, but I simplified the logic a bit to make it work.

src/recipe_parser.zig
codemob-dev: I could modify the argument and return value to use a struct, but that depends on what you think.

assets/cubyz/recipes/brick_recipes.zig.zon
```diff
@@ -1,114 +1,18 @@
...
-	},
-	.{
-		.inputs = .{"cubyz:glacite"},
-		.output = "cubyz:glacite_bricks",
-	},
-	.{
-		.inputs = .{"cubyz:glacite"},
-		.output = "2 cubyz:glacite_wall",
+		.inputs = .{"{item}.stone"},
+		.output = "2 {item}_wall",
```
Argmaster: A small comment, basically repeating what I said in the past - I think that having numbers in strings is a bad idea, we basically end up with nested declarative language. While interpolated string syntax is a neat idea, number prefixes are not. Instead I would suggest inputs and outputs be lists of dictionaries with Id and count like so:

```
.{
  .{
    .inputs = .{.{.id = "{item}.stone", .count = 1}},
    .outputs = .{.{.id = "{item}_wall", .count = 2}}
  }
}
```
This is more explicit but with interpolated strings you no longer have to manually create large numbers of recipes so it shouldn't be a huge issue while improving clarity greatly (syntax highlighting, clear separation of counts and IDs etc.)


Unless that was already rejected  :sweat_smile:

codemob-dev: Yeah, I was going to do that but this PR is already big enough.

IntegratedQuantum: Yes, please do it in a separate PR.

assets/cubyz/recipes/brick_recipes.zig.zon
```diff
@@ -1,114 +1,18 @@
 .{
 	.{
-		.inputs = .{"cubyz:marble"},
-		.output = "cubyz:marble_tile",
+		.inputs = .{"{item}.stone"},
```
IntegratedQuantum: I'm still not convinced that we need tags.
For now I think we can afford to just craft everything into tiles if present. And in the future I think we should have a `stone/` folder. The current file way the files are ordered is really difficult to work with and a hierarchical would be beneficial in many ways.

Are there any cases where you do think a tag would be needed, even after the assets are ordered in a better way?

assets/cubyz/blocks/terracotta.zig.zon
```diff
@@ -1,5 +1,5 @@
 .{
-	.tags = .{.mineable},
+	.tags = .{.mineable, .stone},
```
IntegratedQuantum: Tags should be placed based on what it is, not based on what makes recipe creation more convenient (if you would really want that then e.g. a "tileable" tag would make more sense).

src/recipes.zig
```diff
@@ -0,0 +1,234 @@
...
+	if(std.mem.indexOfScalar(u8, id, ' ')) |index| blk: {
+		amount = std.fmt.parseInt(u16, id[0..index], 0) catch break :blk;
+		id = id[index + 1 ..];
+		id = std.mem.trim(u8, id, &std.ascii.whitespace);
+	}
+	var itemPairs: main.List(ItemKeyPair) = .initCapacity(allocator, 1);
+	var iterator = std.mem.splitScalar(u8, id, '.');
+	id = iterator.next().?;
+	var tags: main.List(Tag) = .init(allocator);
+	defer tags.deinit();
```
IntegratedQuantum: Please don't use the passed in allocator for purely local allocations. I'd suggest to also use another local arena in this function, especially given the long defer block below.

src/recipes.zig
```diff
@@ -0,0 +1,234 @@
...
+				});
+			}
+		}
+	}
+	return itemPairs;
+}
+
+fn generateItemCombos(allocator: NeverFailingAllocator, recipe: []ZonElement) !main.List([]ItemStack) {
+	var localArena: NeverFailingArenaAllocator = .init(main.stackAllocator);
+	var localAllocator = localArena.allocator();
```
IntegratedQuantum: based on #1477, the I think it would make more sense to swap the name.

src/recipes.zig
```diff
@@ -0,0 +1,234 @@
...
+					.keys = keys.clone() catch unreachable,
+				});
+			}
+		}
+	}
+	return itemPairs;
+}
+
+fn generateItemCombos(allocator: NeverFailingAllocator, recipe: []ZonElement) !main.List([]ItemStack) {
+	var localArena: NeverFailingArenaAllocator = .init(main.stackAllocator);
```
IntegratedQuantum: The arena needs to be deinited, which should be detected as a leak when you run it locally.

src/recipes.zig
```diff
@@ -0,0 +1,234 @@
...
+	var newInputCombos: main.List([]ItemStack) = .initCapacity(allocator, inputCombos.items.len);
+	for(inputCombos.items) |inputCombo| {
+		newInputCombos.append(allocator.dupe(ItemStack, inputCombo));
+	}
+	return newInputCombos;
+}
+
+pub fn parseRecipe(zon: ZonElement, list: *main.List(Recipe)) !void {
+	var localArena: NeverFailingArenaAllocator = .init(main.stackAllocator);
+	defer std.debug.assert(localArena.reset(.free_all));
```
IntegratedQuantum: Use deinit instead of reset

IntegratedQuantum (changes requested):
one more fundamental question before I start looking at the parsing code.

assets/cubyz/recipes/brick_recipes.zig.zon
codemob-dev: Now that I think about it, we might not. It depends on if we want to remove the extra flexibility.

IntegratedQuantum: I would suggest to remove it then. As @Argmaster said encoding more stuff in the string is not a good idea, so it would need to be rewritten anyways. So I think it would be easiest to just remove it and readd it when we think we actually need it.

Argmaster: The only problem with hierarchical only approach is that if at any point we encounter situation where something falls into multiple categories we will have a problem. Although this is just templates, no one stops us from having two templates. Hopefully we won't have any exponential recipe explosion.

src/items.zig
```diff
@@ -23,6 +23,8 @@ const Vec3f = vec.Vec3f;
 const modifierList = @import("tool/modifiers/_list.zig");
 const modifierRestrictionList = @import("tool/modifiers/restrictions/_list.zig");
 
+const recipes_zig = @import("recipes.zig");
```
IntegratedQuantum: Every file should be imported publicly in at least one place. Otherwise the CI has trouble finding tests and such.

Also I think it would make sense to put this into a new `items` folder, since it isn't imported in `main`.

src/items/recipes.zig
```diff
@@ -0,0 +1,217 @@
...
+	const inputs = zon.getChild("inputs").toSlice();
+	const recipeItems = arena.alloc(ZonElement, inputs.len + 1);
+	@memcpy(recipeItems[0..inputs.len], inputs);
+	recipeItems[inputs.len] = zon.getChild("output");
+
+	const itemCombos = try generateItemCombos(arena, recipeItems);
+	const reversible = zon.get(bool, "reversible", false);
+	for(itemCombos.items) |itemCombo| {
+		if(reversible and itemCombo.len != 2) {
+			return error.InvalidReversibleRecipe;
```
IntegratedQuantum: This check should be done outside the loop.
Otherwise this check is skipped if no items match.

src/items/recipes.zig
```diff
@@ -0,0 +1,217 @@
...
+		}
+		const parsedInputs = itemCombo[0 .. itemCombo.len - 1];
+		const output = itemCombo[itemCombo.len - 1];
+		const recipe = Recipe{
+			.sourceItems = main.stackAllocator.alloc(BaseItemIndex, parsedInputs.len),
```
IntegratedQuantum: Where is this freed? Please test for leaks and fix them.
Why is a long-lived allocation allocated with the stackAllocator anyways?

src/items/recipes.zig
```diff
@@ -0,0 +1,217 @@
...
+			recipe.sourceAmounts[i] = input.amount;
+		}
+		list.append(recipe);
+		if(reversible) {
+			var reversedRecipe = Recipe{
+				.sourceItems = main.stackAllocator.alloc(BaseItemIndex, 1),
+				.sourceAmounts = main.stackAllocator.alloc(u16, 1),
+				.resultItem = recipe.sourceItems[0],
+				.resultAmount = recipe.sourceAmounts[0],
+			};
```
IntegratedQuantum: Instead of duplicating the logic, how about making a helper function for constructing the recipe from an itemCombo.
Then here the entire for loop would basically just be
```zig
addRecipe(itemCombo, list);
if(reversible) {
    addRecipe(&.{itemCombo[1], itemCombo[0]}, list);
}
```

src/items/recipes.zig
```diff
@@ -0,0 +1,217 @@
...
+	if(idx == target.len) {
+		return newKeys;
+	} else {
+		return error.NoMatch;
+	}
+}
+
+const ItemKeyPair = struct {item: ItemStack, keys: std.StringHashMap([]const u8)};
+
+fn parseRecipeItem(allocator: NeverFailingAllocator, zon: ZonElement, keys: *const std.StringHashMap([]const u8)) !main.List(ItemKeyPair) {
```
IntegratedQuantum: The function is called `parseRecipeItem` but it returns many items.
Instead I'd suggest calling it `parseRecipeItemOptions`.

src/items/recipes.zig
```diff
@@ -0,0 +1,217 @@
...
+	var emptyKeys: std.StringHashMap([]const u8) = .init(arena.allocator);
+	const startingParsedItems = try parseRecipeItem(arena, recipe[0], &emptyKeys);
+	var inputCombos: main.List([]ItemStack) = .initCapacity(arena, startingParsedItems.items.len);
+	var keyList: main.List(std.StringHashMap([]const u8)) = .initCapacity(arena, startingParsedItems.items.len);
+	for(startingParsedItems.items) |item| {
+		const inputs = arena.alloc(ItemStack, recipe.len);
+		inputs[0] = item.item;
+		inputCombos.append(inputs);
+		keyList.append(item.keys);
+	}
```
IntegratedQuantum: Instead of special-casing the first iteration, I think it would be simpler to just start the iteration with one empty inputCombo and one empty map in the keyList.

src/items/recipes.zig
```diff
@@ -0,0 +1,217 @@
...
+	for(1.., recipe[1..]) |i, itemZon| {
+		var newKeyList: main.List(std.StringHashMap([]const u8)) = .init(arena);
+		var newInputCombos: main.List([]ItemStack) = .init(arena);
+
+		for(keyList.items, inputCombos.items) |*keys, inputs| {
+			const parsedItems = try parseRecipeItem(arena, itemZon, keys);
```
IntegratedQuantum: I don't like the complexity of this. E.g. consider the following recipe:
```zig.zon
.{
	.inputs = .{"{item}"},
	.output = "{item}_tile",
},
```
The first iteration will give you a list of all items in the game, and then for each one you call `parseRecipeItem`, which also needs to iterate through all items in the game.
Suddenly recipe parsing is O(n²) per recipe where n is the number of items. And it gets exponentially worse with more inputs.

Instead I'd suggest to parse all item candidates once in the beginning and then only attempt to merge sets here. This would bring it from O(n²) to O(n·m) where m is the number of items matching the second string.

Also instead of using hashmaps, it might be better to use lists of (optional) strings,
similar to how `Tag` turns a string into an integer you could turn the key into a sequential index. This would make merging the sets faster since you wouldn't need to do hashmap lookups everywhere.

IntegratedQuantum (changes requested):

Argmaster (commented):
I would recommend including unit tests for parsing part. Unit tests can significantly impact parser design and unless you nailed it on first try you may end up with either retrospectively added useless unit tests or a significant refactor in the future.
Alternatively you can design tests now, but merge them later. Only tricky part is that you need to know what you are doing.

IntegratedQuantum:
Yeah, TDD is a nice habbit to start, when it is applicable it is really easy and productive to work with, write a test, and then you just fix all the bugs until it works. Here however it is likely hard due to #1286

src/items/recipes.zig
codemob-dev: > The first iteration will give you a list of all items in the game, and then for each one you call `parseRecipeItem`, which also needs to iterate through all items in the game.

If there is only one valid item, it won't loop through every item in the game each time.

IntegratedQuantum: Oh, I didn't see that it inserts the key during parsing.
Honestly the code is kind of hard to read in my opinion.
I personally would still like to see a clearer cut between the string parsing and the pattern matching parts of the code. I think that would make it easier to understand, and more efficient since you wouldn't re-parse the same string 100 times.

codemob-dev:
Yeah, for the most part it's pretty easy to just open up a testing world with a bunch of items and check the available recipes.

IntegratedQuantum:
Please also include recipes from #1850

src/items/recipes.zig
codemob-dev: I started working on this, but it ends up requiring a lot of extra code. (Eg. a struct to handle the parsed string, a lot more helper functions.) I'm not sure there really is an easy way around the extra complexity. I had previously attempted to separate the pattern matching code, but this is as far as I could get before things started getting too complicated.

IntegratedQuantum: I think it's just a matter of where you choose the separating point. I was thinking about something like this:
```zig
const parsedItemsOptions = allocator.alloc(main.List(ItemKeyPair), recipe.len);
for(0..recipe.len) |i| {
    parsedItemsOptions[i] = try parseRecipeItemOptions(arena, itemZon); // Just the parsing, doesn't check previous key set
}

...
for(0..recipe.len) |i| {
	...

	for(keyList.items, inputCombos.items) |*keys, inputs| {
		const parsedItems = try filterRecipeItemOptions(arena, parsedItemsOptions[i], keys); // No parsing, only filters impossible matches
		for(parsedItems.items) |item| {
			...
		}
	}
	...
}
```

IntegratedQuantum: With this you should be able to keep most of the code you already have.

codemob-dev: Hmm, I'll try that.

codemob-dev: I did this, the performance is good enough that I'm not sure if sacrificing the readability of hashmaps is worth it though.

IntegratedQuantum: What is the performance? Did you measure how long it takes to load all recipes currently in debug mode?

codemob-dev: I haven't explicitly measured the performance, I can do that though.

codemob-dev:
For performance I get these results in debug:
```
[debug]: Spent 19ms loading 1 recipes
[debug]: Spent 58ms loading 10 recipes
[debug]: Spent 13ms loading 34 recipes
[debug]: Spent 627ms loading 4 recipes
```

codemob-dev:
The 1 recipe is cloth, the 10 recipes are wood_recipes, the 34 recipes are special_recipes, and the 4 recipes are brick recipes. The reason brick recipes take as long as they do is because they have to check every item in the game to see if they have a tile, brick, and wall counterpart while wood recipes only check in specific subdirectories.

IntegratedQuantum:
I guess it is fine for now then, but we should definitely do the stone asset refactor, moving them into a separate folder, soon.

codemob-dev:
In release mode I get these results:
```
[debug]: Spent 1ms loading 1 recipes
[debug]: Spent 4ms loading 10 recipes
[debug]: Spent 1ms loading 34 recipes
[debug]: Spent 67ms loading 4 recipes
```

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+		recipe.sourceItems[i] = input.item.?.baseItem;
+		recipe.sourceAmounts[i] = input.amount;
+	}
+	list.append(recipe);
+}
+
+pub fn parseRecipe(zon: ZonElement, list: *main.List(Recipe)) !void {
+	var arenaAllocator: NeverFailingArenaAllocator = .init(main.stackAllocator);
+	defer arenaAllocator.deinit();
+	const arena = arenaAllocator.allocator();
```
IntegratedQuantum: I think it would be nice to have an empty line after the arena preambles.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+	const inputs = zon.getChild("inputs").toSlice();
+	const recipeItems = arena.alloc(ZonElement, inputs.len + 1);
+	@memcpy(recipeItems[0..inputs.len], inputs);
+	recipeItems[inputs.len] = zon.getChild("output");
```
IntegratedQuantum: I think you could use `std.mem.concat` here

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+	}
+	var newInputCombos: main.List([]ItemStack) = .initCapacity(allocator, inputCombos.items.len);
+	for(inputCombos.items) |inputCombo| {
+		newInputCombos.append(allocator.dupe(ItemStack, inputCombo));
+	}
+	return newInputCombos;
+}
+
+pub fn addRecipe(itemCombo: []const ItemStack, list: *main.List(Recipe)) void {
+	const parsedInputs = itemCombo[0 .. itemCombo.len - 1];
```
IntegratedQuantum: At this point I would call them just inputs, since this function is not concerned with parsing, and for consistency with the `output` name.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+
+		for(keyList.items, inputCombos.items) |*keys, inputs| {
+			const parsedItems = try parseRecipeItemOptions(arena, pattern, keys);
+			for(parsedItems.items) |item| {
+				const newInputs = arena.dupe(ItemStack, inputs);
+				newInputs[i] = item.item;
+				newInputCombos.append(newInputs);
+				newKeyList.append(item.keys);
+			}
+		}
```
IntegratedQuantum, line 182: No need to change, but I've stumbled over this a few times and I wanted to at least mention it:
In my mind (which prefers recursive decent parsing) this would be better solved as a recursion without copying all the intermediate lists around. But I guess the performance is well enough, so no need to change.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+	if(idx == target.len) {
+		return newKeys;
+	} else {
+		return error.NoMatch;
+	}
+}
+
+const ItemKeyPair = struct {item: ItemStack, keys: std.StringHashMap([]const u8)};
+
+fn parseRecipeItemOptions(allocator: NeverFailingAllocator, itemStackPattern: ItemStackPattern, keys: *const std.StringHashMap([]const u8)) !main.List(ItemKeyPair) {
```
IntegratedQuantum: This isn't parsing anything anymore, so I'd suggest to use a different verb.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+					},
+					.keys = newKeys,
+				});
+			}
+		}
+	}
+	return itemPairs;
+}
+
+fn generateItemCombos(allocator: NeverFailingAllocator, recipe: []ZonElement) !main.List([]ItemStack) {
```
IntegratedQuantum: Also, since I mentioned this on discord, generally I'd avoid returning lists (especially given the whole unmanaged discussion #1181).
For the most part it would be simple as well to just return a slice, either as toOwnedSlice or with a dupe.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+			.item = .{
+				.item = .{.baseItem = item},
+				.amount = amount,
+			},
+			.keys = keys.clone() catch unreachable,
+		});
+	} else {
+		var iter = items.iterator();
+		while(iter.next()) |item| {
+			if(matchWithKeys(allocator, item.id(), pattern.items, keys) catch null) |newKeys| {
```
IntegratedQuantum:
```suggestion
			const newKeys = matchWithKeys(allocator, item.id(), pattern.items, keys) catch continue;
```

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+
+const ItemKeyPair = struct {item: ItemStack, keys: std.StringHashMap([]const u8)};
+
+fn parseRecipeItemOptions(allocator: NeverFailingAllocator, itemStackPattern: ItemStackPattern, keys: *const std.StringHashMap([]const u8)) !main.List(ItemKeyPair) {
+	const pattern = itemStackPattern.pattern;
+	const amount = itemStackPattern.amount;
+
+	var itemPairs: main.List(ItemKeyPair) = .initCapacity(allocator, 1);
+	if(pattern.items.len == 1 and pattern.items[0] == .literal) {
+		const item = BaseItemIndex.fromId(pattern.items[0].literal) orelse return itemPairs;
```
IntegratedQuantum: Should print an error if the item is not found.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+	const pattern = try parsePattern(allocator, id);
+	return .{
+		.amount = amount,
+		.pattern = pattern,
+	};
+}
+
+fn matchWithKeys(allocator: NeverFailingAllocator, target: []const u8, pattern: []const Segment, keys: *const std.StringHashMap([]const u8)) !std.StringHashMap([]const u8) {
+	var idx: usize = 0;
+	idx = 0;
```
IntegratedQuantum:
```suggestion
```

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+			.literal => |literal| {
+				if(literal.len + idx > target.len or !std.mem.eql(u8, target[idx .. idx + literal.len], literal)) {
+					return error.NoMatch;
+				}
+				idx += literal.len;
+			},
+			.symbol => |symbol| {
+				const endIndex: usize = if(i + 1 < pattern.len) blk: {
+					const nextSegment = pattern[i + 1];
+					if(nextSegment == .symbol) {
```
IntegratedQuantum: I think it would be reasonable to assert (by printing an error message and skipping the recipe) during parsing that the next segment has to be a literal. Otherwise the output would be too ambiguous.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+						return error.NoMatch;
+					}
+					break :blk std.mem.indexOfPos(u8, target, idx, nextSegment.literal) orelse {
```
IntegratedQuantum: This is ignoring one possibly relevant case: What if the literal exists multiple times in the id?
The simplest case would be `/`, consider the following hypothetical, generic torch recipe
```
.inputs = .{"{mod}:planks/{name}", "torch_addon:{fuelPath}/{fuelName}}
.output = "torch_addon:{fuelName}_torch"
```
Now imagine there is a fuel named `torch_addon:ores/rare/sodium` and a `torch_addon:sodium_torch`. This parser would not be able to match these, since it would attempt to substiture `fuelPath=ores` and `fuelName=rare/sodium` which does not exist as a torch. It would however also need to try a second substitution `fuelPath=ores/rare` and `fuelName=sodium`

I think either you should handle these edge cases, or we need to find a way to define and clearly communicate (ideally with an error message) to the user what this system can and can't match.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+}
+
+const ItemStackPattern = struct {
+	amount: u16,
+	pattern: main.List(Segment),
+};
+
+fn parseItemZon(allocator: NeverFailingAllocator, zon: ZonElement) !ItemStackPattern {
+	var id = zon.as([]const u8, "");
+	id = std.mem.trim(u8, id, &std.ascii.whitespace);
```
IntegratedQuantum: Is trimming really needed? In my opinion we shouldn't allow any whitespaces here.
I think this is a remnant from when recipes were a custom format.

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+	var amount: u16 = 1;
+	if(std.mem.indexOfScalar(u8, id, ' ')) |index| blk: {
+		amount = std.fmt.parseInt(u16, id[0..index], 0) catch break :blk;
+		id = id[index + 1 ..];
+		id = std.mem.trim(u8, id, &std.ascii.whitespace);
```
IntegratedQuantum: same here

src/items/recipes.zig
```diff
@@ -0,0 +1,214 @@
...
+			const endIndex = std.mem.indexOfScalarPos(u8, pattern, idx, '{') orelse pattern.len;
+			segments.append(.{.literal = pattern[idx..endIndex]});
+			idx = endIndex;
+		}
+	}
+	var newSegments: main.List(Segment) = .init(allocator);
+	for(segments.items) |segment| {
+		switch(segment) {
+			.literal => |literal| {
+				newSegments.append(.{.literal = allocator.dupe(u8, literal)});
```
IntegratedQuantum: Since these are substrings from the string from the recipe zon, which outlives all of the code here, I don't think we need to dupe this.

IntegratedQuantum (changes requested):

src/items/recipes.zig
codemob-dev: Yeah, probably. I always think of std.mem.concat as a string concatenator and I never think of using it for other things.

src/items/recipes.zig
codemob-dev: This is some top quality code right here

src/items/recipes.zig
codemob-dev: I'm usually the other way around, I prefer not to use recursion.

src/items/recipes.zig
codemob-dev: How would I even check that? What if the error appears accidentally? I feel like this is the kind of thing that is much easier to express via documentation.

IntegratedQuantum: Well, it depends on how you define what are valid inputs.
The simplest rule would be to only allow multiple tags if separated by a `:`. This would be quite limiting, but very easy to define errors.

But honestly I think the best solution would be to just handle it by generating and checking multiple matches.

Either way expressing restriction through the documentation is not good enough in my opinion. These are subtle mistakes, and most people tend to not read documentation careful enough (or not at all). We should protect addon creators from bugs instead saying "haha, it's your fault, you should have read the documentation"

assets/cubyz/recipes/brick_recipes.zig.zon
```diff
@@ -1,114 +1,18 @@
 .{
 	.{
-		.inputs = .{"cubyz:marble"},
-		.output = "cubyz:marble_tile",
+		.inputs = .{"{item}"},
+		.output = "{item}_tile",
```
IntegratedQuantum: I honestly think we should avoid fully generic recipes (at least in our own config), except from cases like torches where we want to cover modded planks.
Otherwise this may result in unexpected behavior.

So for these stone recipes I think we should write `cubyz:` in front of them. Same for most of the wood recipes.

Automation is problematic in these cases, e.g. what if a mod has an extra special stone that requires a special table (#1734) to turn into tiles?

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+const NeverFailingArenaAllocator = main.heap.NeverFailingArenaAllocator;
+const ItemStack = items.ItemStack;
+const Tag = main.Tag;
+const Recipe = items.Recipe;
+const BaseItemIndex = items.BaseItemIndex;
+const Block = main.blocks.Block;
+
+const Segment = union(enum) {literal: []const u8, symbol: []const u8};
+
+fn parsePattern(allocator: NeverFailingAllocator, pattern: []const u8) !main.List(Segment) {
```
IntegratedQuantum: Please also replace the other list return types.

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+		}
+		keyList = newKeyList;
+		inputCombos = newInputCombos;
+	}
+	var newInputCombos: main.List([]ItemStack) = .initCapacity(allocator, inputCombos.items.len);
+	defer newInputCombos.deinit();
+	for(inputCombos.items) |inputCombo| {
+		newInputCombos.append(allocator.dupe(ItemStack, inputCombo));
+	}
+	return newInputCombos.toOwnedSlice();
```
IntegratedQuantum:
```suggestion
	const newInputCombos = allocator.alloc([]ItemStack, inputCombos.items.len);
	for(inputCombos.items, 0..) |inputCombo, i| {
		newInputCombos[i] = allocator.dupe(ItemStack, inputCombo);
	}
	return newInputCombos;
```

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+			});
+		}
+	}
+	return itemPairs;
+}
+
+fn generateItemCombos(allocator: NeverFailingAllocator, recipe: []ZonElement) ![][]ItemStack {
+	var arenaAllocator: NeverFailingArenaAllocator = .init(main.stackAllocator);
+	defer arenaAllocator.deinit();
+	var arena = arenaAllocator.allocator();
```
IntegratedQuantum:
```suggestion
	const arena = arenaAllocator.allocator();
```

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+	if(idx == target.len) {
+		return newKeys;
+	} else {
+		return error.NoMatch;
+	}
+}
+
+const ItemKeyPair = struct {item: ItemStack, keys: std.StringHashMap([]const u8)};
```
IntegratedQuantum: The usage of ItemStack seems not quite right, I'd suggest to use `item: BaseItemIndex` and `amount` separately to avoid packing and unpacking the `?Item`.

IntegratedQuantum: Also I think at this size it would be best to expand this into multiple lines.

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+
+fn findRecipeItemOptions(allocator: NeverFailingAllocator, itemStackPattern: ItemStackPattern, keys: *const std.StringHashMap([]const u8)) !main.List(ItemKeyPair) {
```
IntegratedQuantum: Again, please don't return Lists if you don't need a List.

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+	const pattern = itemStackPattern.pattern;
+	const amount = itemStackPattern.amount;
+
+	var itemPairs: main.List(ItemKeyPair) = .initCapacity(allocator, 1);
```
IntegratedQuantum, line 143: Initializing with 1 is kind of unoptimal for the else branch, I'd suggest moving this into the else branch without capacity and do the if branch differently, see comment below.

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+	if(pattern.items.len == 1 and pattern.items[0] == .literal) {
+		const item = BaseItemIndex.fromId(pattern.items[0].literal) orelse return error.ItemNotFound;
+		itemPairs.append(.{
```
IntegratedQuantum:
```suggestion
		return allocator.dupe(ItemKeyPair, &.{
```

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+	} else {
+		var iter = items.iterator();
+		while(iter.next()) |item| {
+			const newKeys = matchWithKeys(allocator, item.id(), pattern.items, keys) catch |err| {
+				if(err != error.NoMatch) {
+					return err;
+				} else {
+					continue;
+				}
+			};
```
IntegratedQuantum: I really don't like seeing such long catch blocks.
I think an optional would generally work better to notify about non-existence, instead of adding a special error code.

I think something like this should be possible:
```zig
const newKeys = (try matchWithKeys(allocator, item.id(), pattern.items, keys)) orelse continue;
```

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+		switch(segment) {
+			.literal => |literal| {
+				newSegments.append(.{.literal = literal});
+			},
+			.symbol => |symbol| {
+				newSegments.append(.{.symbol = symbol});
+			},
+		}
+	}
+	return newSegments;
```
IntegratedQuantum: This whole block is now just a very elaborate copy operation.
```suggestion
	return allocator.dupe(Segment, segments.items);
```

src/items/recipes.zig
```diff
@@ -0,0 +1,218 @@
...
+	var arenaAllocator: NeverFailingArenaAllocator = .init(main.stackAllocator);
+	defer arenaAllocator.deinit();
+	const arena = arenaAllocator.allocator();
+
+	var segments: main.List(Segment) = .init(arena);
+	var idx: usize = 0;
+	while(idx < pattern.len) {
+		if(pattern[idx] == '{') {
+			idx += 1;
+			const endIndex = std.mem.indexOfScalarPos(u8, pattern, idx, '}') orelse return error.UnclosedBrackets;
```
IntegratedQuantum: You can be a little more specific
```suggestion
			const endIndex = std.mem.indexOfScalarPos(u8, pattern, idx, '}') orelse return error.UnclosedBraces;
```

IntegratedQuantum (changes requested):

src/items/recipes.zig
codemob-dev: Annoyingly, that doesn't seem to be possible because of the `errdefer` in the function.

IntegratedQuantum: Why is this a problem? `return err` would also trigger `errdefer`

codemob-dev: Returning null wouldn't.

IntegratedQuantum: Oh right, now I see what you mean.

IntegratedQuantum: Could you then at least shorten the expression a bit, so it doesn't take 5 lines?
```
if(err == error.NoMatch) continue;
return err;
```

src/items/recipes.zig
codemob-dev: I ended up making an `ItemWithAmount` struct, it might be useful elsewhere.

src/items/recipes.zig
codemob-dev: Should I handle the rare case where both versions are valid? (`ores/rare`+`sodium` and `ores`+`rare/sodium`)

codemob-dev: The main reason against it is that it would add extra, probably unnecessary, complexity.

IntegratedQuantum: > Should I handle the rare case where both versions are valid?

I don't see how this could make sense in any real circumstance.

However, I wouldn't be opposed to it if it leads to a simpler implementation.

IntegratedQuantum:
Please rebase, fix conflicts, and update the new convenience functions for the arenas.

codemob-dev:
> Please rebase, fix conflicts, and update the new convenience functions for the arenas.

Done.

src/items/recipes.zig
codemob-dev: I've been thinking about this for a while, and I honestly have no clue how to do this.

IntegratedQuantum: I would just do it like this:
Just add the same item multiple times with different key-value mappings.
Then afterwards you just need to remove all duplicate recipes.

codemob-dev: I believe I got this working now!

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+	}
+	return newInputCombos;
+}
+
+pub fn addRecipe(itemCombo: []const ItemWithAmount, list: *main.List(Recipe)) void {
+	const inputs = itemCombo[0 .. itemCombo.len - 1];
+	const output = itemCombo[itemCombo.len - 1];
+	const recipe = Recipe{
+		.sourceItems = main.globalAllocator.alloc(BaseItemIndex, inputs.len),
+		.sourceAmounts = main.globalAllocator.alloc(u16, inputs.len),
```
IntegratedQuantum: Since this is freed in a completely different file, I think it would be better to pass the allocator as a function argument.
This reduces the amount of work when migrating to a different allocator and reduces the risk for accidental leaks when doing so.

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+					.amount = amount,
+				},
+				.keys = match,
+			});
+		}
+	}
+	return itemPairs.toOwnedSlice();
+}
+
+fn generateItemCombos(allocator: NeverFailingAllocator, recipe: []ZonElement) ![][]ItemWithAmount {
```
IntegratedQuantum: Please do a bit more const correctness
```suggestion
fn generateItemCombos(allocator: NeverFailingAllocator, recipe: []const ZonElement) ![]const []const ItemWithAmount {
```

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+	const arena = main.stackAllocator.createArena();
+	defer main.stackAllocator.destroyArena(arena);
+
+	var inputCombos: main.List([]ItemWithAmount) = .initCapacity(arena, 1);
```
IntegratedQuantum:
```suggestion
	var inputCombos: main.List([]const ItemWithAmount) = .initCapacity(arena, 1);
```

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+				const newInputs = arena.dupe(ItemWithAmount, inputs);
+				newInputs[i] = item.item;
+				newInputCombos.append(newInputs);
+				newKeyList.append(item.keys);
+			}
+		}
+		keyList = newKeyList;
+		inputCombos = newInputCombos;
+	}
+	const newInputCombos = allocator.alloc([]ItemWithAmount, inputCombos.items.len);
```
IntegratedQuantum:
```suggestion
	const newInputCombos = allocator.alloc([]const ItemWithAmount, inputCombos.items.len);
```

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+}
+
+const ItemWithAmount = struct {
+	item: BaseItemIndex,
+	amount: u16,
+};
+
+const ItemKeyPair = struct {item: ItemWithAmount, keys: std.StringHashMap([]const u8)};
+
+fn findRecipeItemOptions(allocator: NeverFailingAllocator, itemStackPattern: ItemStackPattern, keys: *const std.StringHashMap([]const u8)) ![]ItemKeyPair {
```
IntegratedQuantum:
```suggestion
fn findRecipeItemOptions(allocator: NeverFailingAllocator, itemStackPattern: ItemStackPattern, keys: *const std.StringHashMap([]const u8)) ![]const ItemKeyPair {
```

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+		id = id[index + 1 ..];
+	}
+	const pattern = try parsePattern(allocator, id);
+	return .{
+		.amount = amount,
+		.pattern = pattern,
+	};
+}
+
+fn matchWithKeys(allocator: NeverFailingAllocator, target: []const u8, pattern: []const Segment, keys: *const std.StringHashMap([]const u8)) ![]std.StringHashMap([]const u8) {
```
IntegratedQuantum:
```suggestion
fn matchWithKeys(allocator: NeverFailingAllocator, target: []const u8, pattern: []const Segment, keys: *const std.StringHashMap([]const u8)) ![]const std.StringHashMap([]const u8) {
```

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+	for(0.., pattern) |i, segment| {
+		switch(segment) {
+			.literal => |literal| {
+				if(literal.len + idx > target.len or !std.mem.eql(u8, target[idx .. idx + literal.len], literal)) {
+					return error.NoMatch;
+				}
+				idx += literal.len;
+			},
+			.symbol => |symbol| {
+				var indexes: main.List(usize) = .init(allocator);
```
IntegratedQuantum:
```suggestion
				var endIndices: main.List(usize) = .init(allocator);
```

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+				} else {
+					indexes.append(target.len);
+				}
+				if(indexes.items.len == 0) {
+					return error.NoMatch;
+				}
+				if(indexes.items.len == 1) {
+					if(newKeys.get(symbol)) |value| {
+						if(!std.mem.eql(u8, target[idx..indexes.items[0]], value)) {
+							return error.NoMatch;
```
IntegratedQuantum: This check should be done in both branches, and thus I'd suggest putting it into the while loop above, before even adding the candidate to the indices list.

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+					idx = indexes.items[0];
+				} else {
+					defer newKeys.deinit();
+					var newKeyPairs: main.List(std.StringHashMap([]const u8)) = .init(allocator);
+					defer newKeyPairs.deinit();
+					for(indexes.items) |endIndex| {
+						if(matchWithKeys(allocator, target[endIndex..], pattern[i + 1 ..], &newKeys)) |newKeyMatches| {
+							newKeyPairs.appendSlice(newKeyMatches);
+							allocator.free(newKeyMatches);
+						} else |_| {}
```
IntegratedQuantum: Please don't ignore errors.

Again proper testing would have helped you find problem with this.

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+					} else {
+						newKeys.put(allocator.dupe(u8, symbol), allocator.dupe(u8, target[idx..indexes.items[0]])) catch unreachable;
+					}
+					idx = indexes.items[0];
+				} else {
+					defer newKeys.deinit();
+					var newKeyPairs: main.List(std.StringHashMap([]const u8)) = .init(allocator);
+					defer newKeyPairs.deinit();
+					for(indexes.items) |endIndex| {
+						if(matchWithKeys(allocator, target[endIndex..], pattern[i + 1 ..], &newKeys)) |newKeyMatches| {
```
IntegratedQuantum: This is missing an equivalent to `newKeys.put` somewhere.

In fact this branch has negative test coverage, you cannot even test it inside the game.
Please add some test cases to test the parsing/matching functions. Since they take string arguments it should be easy to write tests for them.

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const NeverFailingArenaAllocator = main.heap.NeverFailingArenaAllocator;
+const Tag = main.Tag;
+const Recipe = items.Recipe;
+const BaseItemIndex = items.BaseItemIndex;
+const Block = main.blocks.Block;
+
+const Segment = union(enum) {literal: []const u8, symbol: []const u8};
+
+fn parsePattern(allocator: NeverFailingAllocator, pattern: []const u8) ![]Segment {
```
IntegratedQuantum:
```suggestion
fn parsePattern(allocator: NeverFailingAllocator, pattern: []const u8) ![]const Segment {
```

src/items/recipes.zig
```diff
@@ -0,0 +1,226 @@
...
+	defer segments.deinit();
+	var idx: usize = 0;
+	while(idx < pattern.len) {
+		if(pattern[idx] == '{') {
+			if(segments.items.len > 0 and segments.items[segments.items.len - 1] == .symbol) {
+				return error.AmbiguousSymbols;
+			}
+			idx += 1;
+			const endIndex = std.mem.indexOfScalarPos(u8, pattern, idx, '}') orelse return error.UnclosedBraces;
+			if(idx == endIndex) return error.EmptyBrackets;
```
IntegratedQuantum:
```suggestion
			if(idx == endIndex) return error.EmptyBraces;
```

IntegratedQuantum (changes requested):
Please add testing for the newly added edge case handling. I have found some problems in the review, but neither of us can guarantee that it works (and keeps working in the future) without having some unit testing here.

src/items/recipes.zig
codemob-dev: This is intentional, the errors are just used instead of optionals to allow `errdefer` to be used.

src/items/recipes.zig
codemob-dev: Fixed, I'm working on tests now.

codemob-dev: I got some decent tests, not sure if it is enough though.

src/items/recipes.zig
```diff
@@ -0,0 +1,254 @@
...
+}
+
+test "pattern parsing" {
+	try std.testing.expectError(error.AmbiguousSymbols, parsePattern(main.heap.testingAllocator, "cubyz:{a}{b}"));
+	try std.testing.expectError(error.EmptyBraces, parsePattern(main.heap.testingAllocator, "{}"));
+	try std.testing.expectError(error.UnclosedBraces, parsePattern(main.heap.testingAllocator, "cubyz:{foo"));
+
+	const pattern = try parsePattern(main.heap.testingAllocator, "foo:{bar}/{baz}");
+	defer main.heap.testingAllocator.free(pattern);
+	try std.testing.expectEqual(4, pattern.len);
```
IntegratedQuantum: This is a very weak condition. A test with a high false-positive chance is rather useless.
Instead please check the values of the output as well.

src/items/recipes.zig
```diff
@@ -0,0 +1,254 @@
...
+	const pattern = try parsePattern(arena, "foo:{bar}/{baz}");
+
+	var keys: std.StringHashMap([]const u8) = .init(arena.allocator);
+
+	const newKeys = try matchWithKeys(arena, "foo:1/2/3", pattern, &keys);
+	try std.testing.expectEqual(2, newKeys.len);
+
+	var matched = false;
+	for(newKeys) |keyPairs| {
+		if(matchWithKeys(arena, "foo:3/1/2", try parsePattern(arena, "foo:{baz}/{bar}"), &keyPairs) catch null) |_| {
```
IntegratedQuantum: Again, this test condition is very weak. Please check if the resulting keymap has the correct values for baz and bar.

src/items/recipes.zig
```diff
@@ -0,0 +1,254 @@
...
+test "pattern matching" {
+	const arena = main.heap.testingAllocator.createArena();
+	defer main.heap.testingAllocator.destroyArena(arena);
+
+	const pattern = try parsePattern(arena, "foo:{bar}/{baz}");
+
+	var keys: std.StringHashMap([]const u8) = .init(arena.allocator);
+
+	const newKeys = try matchWithKeys(arena, "foo:1/2/3", pattern, &keys);
+	try std.testing.expectEqual(2, newKeys.len);
```
IntegratedQuantum, line 260: same here

src/items/recipes.zig
```diff
@@ -0,0 +1,254 @@
...
+	try std.testing.expectError(error.AmbiguousSymbols, parsePattern(main.heap.testingAllocator, "cubyz:{a}{b}"));
+	try std.testing.expectError(error.EmptyBraces, parsePattern(main.heap.testingAllocator, "{}"));
+	try std.testing.expectError(error.UnclosedBraces, parsePattern(main.heap.testingAllocator, "cubyz:{foo"));
+
+	const pattern = try parsePattern(main.heap.testingAllocator, "foo:{bar}/{baz}");
+	defer main.heap.testingAllocator.free(pattern);
+	try std.testing.expectEqual(4, pattern.len);
+}
+
+test "pattern matching" {
```
IntegratedQuantum, line 243: please test the error paths in this function as well.

src/items/recipes.zig
```diff
@@ -0,0 +1,254 @@
...
+	const arena = main.heap.testingAllocator.createArena();
+	defer main.heap.testingAllocator.destroyArena(arena);
```
IntegratedQuantum: Don't test with an arena, if the function interface accepts an arbitrary allocator.

IntegratedQuantum (changes requested):

src/items/recipes.zig
```diff
@@ -0,0 +1,254 @@
...
+	for(itemCombos) |itemCombo| {
+		addRecipe(allocator, itemCombo, list);
+		if(reversible) {
+			addRecipe(allocator, &.{itemCombo[1], itemCombo[0]}, list);
+		}
+	}
+}
+
+test "pattern parsing" {
+	try std.testing.expectError(error.AmbiguousSymbols, parsePattern(main.heap.testingAllocator, "cubyz:{a}{b}"));
```
Argmaster, line 231: IMO those cases should be in separate tests. Unit test should consist of preparation, assertion and teardown. Multiple assertions mean you won't instantly know what feature doesn't work unless you look into stack trace.

src/items/recipes.zig
codemob-dev: I had done that before, but deallocating it is a pain because each key and value is allocated individually with the arena.

IntegratedQuantum: The alternative is of course to make it part of the interface, and only allow passing an arena through an assertion.
Anything else is an easy source for bugs.

src/items/recipes.zig
IntegratedQuantum: I think one test-block per function is fine. Having to look at the trace is likely the first thing you will do anyways in case of failure.

Argmaster: Unit tests do not test functions, they test features / behaviors. Here you can clearly see 4 separate test cases, 3 negative and 1 positive. Moreover, similarly to functions, tests should be single purpose with same side effects of violating this rule. Here it is not a huge issue, its just 4 test cases, but if there is no strict guideline to follow you will surely end up with tests which cover anything between 1 to tens of behaviors. This means that you will see no clear difference between 1%, 10% or 50% of tests failing since they can vary in size so much. They will vary in size without allowing them to cover multiple features / behaviors at once, this will only exponentially increase the resulting mess.

Possibly I have fucked that up in the past to 😅

IntegratedQuantum: Yes I agree it makes sense to group test blocks by testing one feature/behavior.
But I think in the end it's a matter of definition, what is a feature/behavior?
I personally would group all the error cases together into one testing block, testing the behavior that the function produces errors during parsing when it encounters illegal syntax.

src/items/recipes.zig
codemob-dev: There are none, it either matches or it doesn't.

src/items/recipes.zig
codemob-dev: I removed some unneeded dupe calls and it works now.

src/items/recipes.zig
codemob-dev: @Argmaster I did a little reorganization, what do you think now?

Argmaster: Every time you calla function and assert the result you get a single behavior and hence a single test. You may need multiple assertions per call to test complex objects. In here we clearly have multiple arrange-assert statements
Anyway, none of what I am saying is binding @codemob-dev
For sources of my opinions I can link the following:
- https://youtu.be/EZ05e7EMOLM?si=QHGIJmcS3QVAEQZo
- https://youtu.be/u5senBJUkPc?si=oJ7814GAmYPZJgQ8

src/items/recipes.zig
Argmaster: Well, then why do you have `try matchWithKeys`?

codemob-dev: To assert that it will match.

IntegratedQuantum: Look, if the function signature contains an error, then I will call it an error.
Naming aside though, please test all possible cases.

Argmaster: > To assert that it will match.

Very well, now make test where you assert that it doesn't 🐌

codemob-dev: This is done now.

src/items/recipes.zig
```diff
@@ -0,0 +1,286 @@
...
+
+fn matchWithKeys(allocator: NeverFailingAllocator, target: []const u8, pattern: []const Segment, keys: *const std.StringHashMap([]const u8)) ![]const std.StringHashMap([]const u8) {
+	var idx: usize = 0;
+	var newKeys = keys.clone() catch unreachable;
+	errdefer newKeys.deinit();
+
+	for(0.., pattern) |i, segment| {
+		switch(segment) {
+			.literal => |literal| {
+				if(literal.len + idx > target.len or !std.mem.eql(u8, target[idx .. idx + literal.len], literal)) {
```
IntegratedQuantum: There actually is a helper function for this:
```suggestion
				if(!std.mem.startsWith(u8, target[idx..], literal)) {
```

src/items/recipes.zig
```diff
@@ -0,0 +1,286 @@
...
+					return error.NoMatch;
+				}
+				idx += literal.len;
+			},
+			.symbol => |symbol| {
+				var endIndices: main.List(usize) = .init(allocator);
+				defer endIndices.deinit();
+				if(newKeys.get(symbol)) |value| {
+					if(idx + value.len >= target.len or !std.mem.eql(u8, target[idx .. idx + value.len], value)) {
```
IntegratedQuantum: Same here
```suggestion
					if(!std.mem.startsWith(u8, target[idx..], value)) {
```

IntegratedQuantum (changes requested):
Two more small changes, then we are ready to go here.

IntegratedQuantum (approved):

PR #1857, opened 2025-09-21
Add crimson wasteland biome

codemob-dev:
image
image
image


assets/cubyz/biomes/red_plains.zig.zon
```diff
@@ -0,0 +1,35 @@
...
+	.minHeightLimit = 7,
+	.minHeight = 22,
+	.maxHeight = 40,
+	.maxHeightLimit = 50,
+	.smoothBeaches = true,
+	.minRadius = 256,
+	.maxRadius = 320,
+	.roughness = 1,
+	.hills = 15,
+	.chance = 0.2,
```
IntegratedQuantum: Given that it is mostly barren, I'd prefer if this was one of the super-rare biome, with 0.01 chance like the modern art biome.

IntegratedQuantum (changes requested):
I would suggest using stalagmites instead of (or additional to) boulders, I think they would make the terrain look more spiky and dangerous.

assets/cubyz/biomes/red_plains.zig.zon
codemob-dev: Can do

codemob-dev:
ooh yeah, that could work

IntegratedQuantum:
Also about the name, "red_plains" seems kind of mundane.
Maybe "crimson_wasteland" would be better?

codemob-dev:
NeonBeach also suggested `land_of_the_dead`

IntegratedQuantum:
> land_of_the_dead

Hmm, I don't know, I think that would imply some kind of zombie living there which is not the case.

ikabod-kee:
Strange that chalk would occur in a molten sort of biome, considering it's a sedimentary rock.

I will also say, this biome looks very bland.

codemob-dev:
> Strange that chalk would occur in a molten sort of biome, considering it's a sedimentary rock.

This is a video game, we don't need to be constrained by real life physics.

> I will also say, this biome looks very bland.

It *is* a wasteland. The new spikes help out a bit for this though.
image

ikabod-kee:
Perhaps it could use patches of red, crimson, and black? Replace the chalk with obsidian, or basalt?

codemob-dev:
Perhaps, I'll try that.

ikabod-kee:
I think it could also benefit from changes in elevation, like the savannah

codemob-dev:
It looks a lot better now.

image

ikabod-kee:
Yeah, there we go!

codemob-dev:
I did some reworking of the magma rocks and black pits

image

careeoki:
Basalt might look better than black chalk. would add some more texture variety too

codemob-dev:
Perhaps for the stalagmites?

codemob-dev:
It does look pretty good
image

careeoki:
I would still like to see how it looks on the ground. Generally try not to use _too_ too many different blocks in 1 biome

codemob-dev:
It looks pretty good, I do like the contrast it adds with the black pits.

image

careeoki:
Yeah that looks nice

assets/cubyz/biomes/crimson_wasteland/base.zig.zon
```diff
@@ -0,0 +1,62 @@
...
+	.minRadius = 256,
+	.maxRadius = 320,
+	.roughness = 1,
+	.hills = 15,
+	.chance = 0.02,
+	.validPlayerSpawn = false,
+	.stoneBlock = "cubyz:obsidian",
+	.structures = .{
+		.{
+			.id = "cubyz:stalagmite",
```
IntegratedQuantum, line 21: Just a suggestion, but given that the underground is made from obsidian, I think it would be fun to have some rare obsidian stalagmites as well.

assets/cubyz/biomes/crimson_wasteland/black_pit.zig.zon
```diff
@@ -0,0 +1,17 @@
+.{
```
IntegratedQuantum: Are the pits supposed to be flat now?
Screenshot at 2025-09-22 17-54-48
If so then I wouldn't call them pits.

IntegratedQuantum: Also at that point, maybe they could just be large ground patches that are prioritized to be placed after the others?

IntegratedQuantum (changes requested):

assets/cubyz/biomes/crimson_wasteland/black_pit.zig.zon
codemob-dev: No, they are supposed to be pits. I must have messed up the height.

assets/cubyz/biomes/crimson_wasteland/base.zig.zon
codemob-dev: Hmm, that could be neat.

codemob-dev:
It should be fixed now.

assets/cubyz/biomes/crimson_wasteland/black_pit.zig.zon
IntegratedQuantum: I think it might be better to do it like the other pit, giving it a height offset rather than a fixed height.

codemob-dev: I was trying that, but it kept generating bits of water and it didn't look all that great.

IntegratedQuantum: Hmm still doesn't really look like pits to me in some places.
Oh how about setting `.interpolation = .none`? That might give it a more well-defined edge.

codemob-dev: I'm not sure, it seems to highlight the low-res grid.
image

IntegratedQuantum: That is odd. Usually the noise makes it lot less worse. Does it looks so everywhere?

IntegratedQuantum: We really need #213

codemob-dev: > Does it looks so everywhere?

More or less, yes.

codemob-dev: I upped the height of the entire biome to give more space for the pits, and that seemed to fix the issue.

assets/cubyz/biomes/crimson_wasteland/base.zig.zon
```diff
@@ -0,0 +1,69 @@
...
+	.minHeightLimit = 7,
+	.minHeight = 42,
+	.maxHeight = 60,
+	.maxHeightLimit = 65,
+	.smoothBeaches = true,
+	.minRadius = 256,
+	.maxRadius = 320,
+	.roughness = 1,
+	.hills = 15,
+	.chance = 0.02,
```
IntegratedQuantum: Didn't we agree on 0.01? Could you explain why you made it higher again?

assets/cubyz/biomes/crimson_wasteland/magma_rock.zig.zon
```diff
@@ -0,0 +1,28 @@
+.{
```
IntegratedQuantum, line 1: one more suggestion for the magma rocks: `.interpolation = .linear` could help give it a more spiky shape, making it fit better with the general style of the biome.

assets/cubyz/biomes/crimson_wasteland/black_pit.zig.zon
```diff
@@ -0,0 +1,17 @@
...
+	.properties = .{},
+	.minRadius = 8,
```
IntegratedQuantum: Any radius less than 16 is meaningless, and just makes more biomes disappear. Please start at 16 and adjust the chances accordingly, anything else is just wasteful.

assets/cubyz/biomes/crimson_wasteland/base.zig.zon
codemob-dev: I was doing some flying around, and the biome was almost impossible to find even in creative mode.

assets/cubyz/biomes/crimson_wasteland/magma_rock.zig.zon
codemob-dev: Yeah, that helped a lot!

assets/cubyz/biomes/crimson_wasteland/base.zig.zon
IntegratedQuantum: That's kind of the point. You are not supposed to look for it and find it.
The purpose of these biomes is to catch you by surprise. They are supposed to reward you for exploring, or for climbing that mountain and seeing it in the distance. "Huh, I have never seen such a biome before, let me go check it out"
The goal in my opinion is that you can fly around and find one rare biome. And from I have seen so far equivalent biomes, like e.g. the glass forest, are not as rare as you may think, especially as we keep adding more of them.

IntegratedQuantum:
Please also put the biome into the `rare` folder as discussed in https://github.com/PixelGuys/Cubyz/pull/1860#issuecomment-3324549163

codemob-dev:
Done.

assets/cubyz/biomes/rare/crimson_wasteland/black_pit.zig.zon
```diff
@@ -0,0 +1,17 @@
...
+	.chance = 0,
+	.minHeight = 5,
+	.maxHeight = 10,
+	.roughness = 3,
+	.hills = 3,
+	.stoneBlock = "cubyz:chalk/black",
+	.parentBiomes = .{
+		.{
+			.id = "cubyz:rare/crimson_wasteland/base",
+			.chance = 48,
```
IntegratedQuantum: I think you forgot to turn the chance here after increasing the radius, the biome is looking rather patchy with all these pits:
Screenshot at 2025-09-24 21-00-06

codemob-dev: Fixed.

IntegratedQuantum (approved):

codemob-dev:
Is there a reason this hasn't been merged yet?

IntegratedQuantum:
I was waiting for CI completion and forgot

PR #1868, opened 2025-09-24
fixed incorrect parentBiome ids

NeonBeach:
correction of parentBiome ids in #1860 by @careeoki


IntegratedQuantum (approved):
Shows that we really need more testing in the CI, particularly with #1097

PR #1899, opened 2025-10-05
Fancy fern model

careeoki:
Makes them more fern-y
image
image-1


IntegratedQuantum (approved):
A variant of this could also be interesting for flowers, imagine a little flower in the center.

PR #1914, opened 2025-10-07
Shift to reverse workbench cycle

Blobicycake:
While holding shift, tool types on the workbench menu will cycle backwards.


OneAvargeCoder193:
This code to select items is used a lot in other places, for example to select the difficulty in the world creation menu, could you also add this in those places?

IntegratedQuantum (changes requested):
In my opinion you should go for the right solution (#1920) instead of implementing some niche placeholder that most people would likely miss

PR #1916, opened 2025-10-08
Cactus SBBs

careeoki:
image
There are 3 types of cactus:

Regular
image

Young
image

and Old Growth
image
(inspired by awesome old cacti like this:)
Saguaro_with_many_arms


assets/cubyz/sbb/cactus/arms.zig.zon
```diff
@@ -0,0 +1,9 @@
+.{
+	.blueprints = .{
+		.{.id = "cubyz:cactus/arm/0"},
+		.{.id = "cubyz:cactus/arm/1"},
+		.{.id = "cubyz:cactus/arm/2"},
+		.{.id = "cubyz:cactus/arm/3"},
+		.{.id = "cubyz:cactus/arm/4", .chance = 0.5},
```
IntegratedQuantum, line 7: I'd suggest to give the arms a null option so not all cacti have the same 4 sides.
Screenshot at 2025-10-10 21-31-18
Furthermore maybe it would make sense to have them at a different height. I think it looks kind of odd that they all are at the same height.

careeoki: Added null and offset the arms of 4_arms base

Issue #1901, opened 2025-10-05
You can place a block on yourself when bridging diagonally


OneAvargeCoder193:
### Steps to Reproduce and Observed Behavior

If you are facing diagonally on a block, crouching and holding two keys to move to the corner, you can place a block on the top face of the block, teleporting you.

PR #1931, opened 2025-10-08
Fix being able to place blocks in your own characters hitbox while on the edge of a block

eve-v0:
fixes https://github.com/PixelGuys/Cubyz/issues/1901
The problem was being able to place blocks in yourself while on the edge of the block,
it allowed insanely fast building upwards, as fast as you can click to place blocks as being inside a block teleports you up

fix was removing " - @as(Vec3d, @splat(0.00005)) "  - this part decreased the characters hitbox during collision check

I guess its intended function was to fix possible false positives from floating point precision error while placing blocks,
after the change i did some testing with placing blocks on the edge of characters hitbox but didn't experience any problems

### DIFF
--- a/src/renderer.zig
+++ b/src/renderer.zig
@@ -971,7 +971,7 @@ pub const MeshSelection = struct { // MARK: MeshSelection
 	}
 
 	fn canPlaceBlock(pos: Vec3i, block: main.blocks.Block) bool {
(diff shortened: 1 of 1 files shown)


eve-v0:
by "placing blocks on the edge of characters hitbox" i meant testing for false positives due to removal of decrasing characters collision, fix by itself works

PR #1944, opened 2025-10-09
Allow sbbs to snap to the floor and ceiling (and also allow offsetting them)

### LINKED ISSUE #1408: Snapping SBB Children

OneAvargeCoder193:
closes #1408


OneAvargeCoder193:
With this pr we can also have paths that follow the ground

OneAvargeCoder193:
I converted this to a draft because the snapping is causing some errors

PR #1963, opened 2025-10-11
Dynamic nametag scaling based on player distance

eve-v0:
Font size changes from 48px to 16px from 1 - 100 blocks
Tested on different resolutions

### DIFF
--- a/src/entity.zig
+++ b/src/entity.zig
@@ -178,7 +178,7 @@ pub const ClientEntityManager = struct {
 			var buf = graphics.TextBuffer.init(main.stackAllocator, ent.name, .{.color = 0}, false, .center);
 			defer buf.deinit();
 			const size = buf.calculateLineBreaks(48/dstclamped*heightRef, 1024);
-			buf.render(xCenter - size[0]/2, yCenter - size[1], 48/dstclamped*heightRef);
+			buf.renderTextWithoutShadow(xCenter - size[0]/2, yCenter - size[1], 48/dstclamped*heightRef);
 		}
 	}
 


src/entity.zig
```diff
@@ -165,6 +165,8 @@ pub const ClientEntityManager = struct {
 
 			const distance: f64 = std.math.sqrt(pos3d[0]*pos3d[0] + pos3d[1]*pos3d[1] + pos3d[2]*pos3d[2]);
 			const dstclamped: f32 = @as(f32, @floatCast((std.math.clamp(distance, 0, 160) + 80)/80));
+			const heightRef: f16 = (@as(f16, @floatFromInt(main.Window.height))/1080.0);
+			// 1080p = 16px
```
IntegratedQuantum, line 169: Please use the gui scale instead of computing a worse version of it yourself.

eve-v0: i think using gui scale is a bad idea
1. gui scale increases from 0.5 to 1 to 2 to 3... so its not smooth
2. gui scale is calculated also using screen width but player size reference is only dependent on screen height

using gui scale would make different resolutions have unpredictable and not reasonable results
for example playing on a thinner resolutions despite the size of everything in the world being the same the text could be too small

eve-v0:
@IntegratedQuantum small text rendering for nametags after disabling text shadows looks much better, do you think that anti aliasing for text rendering is needed or is this enough?

IntegratedQuantum:
Well, it is at least readable now.
@ikabod-kee I would like to hear your opinion on this.

ikabod-kee:
I agree that the text should only have integer scaling, as to not have artifacts and to keep it readable.

ikabod-kee:
If that issue is already fixed, then I'll have to see how it looks in-game. Could you send a video of it happening?

ikabod-kee:
It also may be worth using a different font for names so that it scales better*. Not sure though.

Edit: smoother*

eve-v0:
@ikabod-kee the minimal text size is currently 16px and after disabling shadow rendering for text it looks and scales very well on any distance, its now easily readable even for weird fonts, i would love to send you the video but because of the lower resolution of the video and video compression it looks worse than it really is,
image

if you want to see for yourself just join a server on unstable unstable.cubyz.ddns.net:3000

eve-v0:
i think the only thing that could be better is antialiasing but not just for text for the whole game

IntegratedQuantum:
Closed in favor of #1993 which doesn't have any aliasing problems from scaling the text

PR #1970, opened 2025-10-12
/gamemode command to allow specifying game mode with a single letter 's' , 'c'

eve-v0:


Argmaster:
This is something that is supposed to be solved by auto completion.

copygirl:
I agree. Quick gamemode switching functionality like Minecraft has (hold F3+F4) is a better solution than introducing shorthand commands.

PR #1973, opened 2025-10-12
Reduce reflectivity of glass

careeoki:
To stop glass from being way too glowy.
image
### Before
image

### After
image


ikabod-kee:
Maybe glass should have some transparent texture on it, like slight smudges/residue.

careeoki:
> Maybe glass should have some transparent texture on it, like slight smudges/residue.

It already has exactly that.

ikabod-kee:
It does??

IntegratedQuantum (changes requested):
I think this makes a lot of sense for the brighter colors.
However at the same time it seems to make the other colors appear quite dark. So instead of making them all less reflective, I'd suggest to make the darker colors (indigo, crimson, brown, etc.) more reflective. I think this also matches with real world intution (if you look through a glass pane into the dark night, then it appears to be more reflective than at day time)

careeoki:
Done, also decreased the reflectivity of white glass a little more

IntegratedQuantum (approved):
Yeah, I think it looks better now. Also puts white glass closer to its original purpose.

PR #1977, opened 2025-10-13
Command and argument autocomplete

eve-v0:
currently the functionality is very basic
and just autocompletes command word
the plan is to

modify all command files to have them show their command and available arguments

take a command list from combining all commands from files instead of a preset array

check if thing to autocomplete is after a command so it should be an argument
take it from argument list of the command before so it matches the command

disable autocomplete for whole if the first character of the input is not a slash

make the autocomplete possibly server sided allowing autocomplete for custom server sided command on unmoded client

any suggestions please write a comment
and if someone knows a better way to get type's "List" length than this  input.currentString.items[0..input.currentString.items.len].len

### DIFF
--- a/src/gui/windows/chat.zig
+++ b/src/gui/windows/chat.zig
@@ -223,10 +223,13 @@ pub fn autoComplete(_: usize) void {
 		const newLen = start + comp.len + 1; 
 		if (newLen > buf.len) return;
 
-		@memcpy(buf[0..start], lineSlice[0..start]);
-        @memcpy(buf[start..start+comp.len], comp);
+		@memcpy(buf[0 .. start], lineSlice[0..start]);
+        @memcpy(buf[start .. start + comp.len], comp);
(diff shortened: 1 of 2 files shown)


Argmaster:
We need a more standardized way to communicate available options between client and server and we definitely don't want to list them manually.
I think it should wait for https://github.com/PixelGuys/Cubyz/pull/1425 so it can be built on top of declarative system instead of writing stuff by hand.

eve-v0:
@Argmaster i see doAutoComplete in the parser so i guess there will a proper place to write command and argument autocompletion there?
another thing so you mean the system should be created when there is ability for server upon client joining to send them all available commands and arguments so autocomplete would work for custom server sided commands?

Argmaster:
I am not sure at the moment when the communication should take place, but definitely there will have to be an exchange every now and then because some information (eg. Player names) can not be obtained only once. This, it must be possible for a client to query the server.

PR #1993, opened 2025-10-14
Better name bar rendering

bagggage:
Reduces visual noise when many players are connected. Makes it possible to estimate the distance of any given player from you.

- now text size is relative to world units;
- added fadeout on long distances.

Alternative to #1963.

image_2025-10-14_10-17-29
image_2025-10-14_10-17-57
image_2025-10-14_10-19-07
image_2025-10-14_10-19-42

### DIFF
--- a/src/entity.zig
+++ b/src/entity.zig
@@ -153,7 +153,6 @@ pub const ClientEntityManager = struct {
 		mutex.lock();
 		defer mutex.unlock();
 
-		// this value is relative to the fov-y and window size
 		const screenUnits = @as(f32, @floatFromInt(main.Window.height))/1024;
 		const fontBaseSize = 128.0;
 		const fontMinScreenSize = 16.0;


src/entity.zig
```diff
@@ -169,11 +176,17 @@ pub const ClientEntityManager = struct {
 			const xCenter = (1 + projectedPos[0]/projectedPos[3])*@as(f32, @floatFromInt(main.Window.width/2));
 			const yCenter = (1 - projectedPos[1]/projectedPos[3])*@as(f32, @floatFromInt(main.Window.height/2));
 
-			graphics.draw.setColor(0xff000000);
+			const maxDistanceSquare = 128.0*128.0;
+			const minAlpha = 0x28;
+			const trancparency = 1 - vec.lengthSquare(pos3d)/maxDistanceSquare;
```
thecreare: I think `trancparency` is a typo, should probably be `transparency`

bagggage: Yes, thanks

copygirl:
It should be noted that this is an alternative to #1963.

IntegratedQuantum:
@ikabod-kee could you also check this one out, if it looks better than the other one.

ikabod-kee:
> @ikabod-kee could you also check this one out, if it looks better than the other one.

I need to figure out how to connect to my own server

codemob-dev:
> I need to figure out how to connect to my own server

On your other game make sure to change your username then connect to `127.0.0.1`

ikabod-kee (approved):
I really like this implementation. Approved!

IntegratedQuantum:
> On your other game make sure to change your username

Or create a testing mode world

src/entity.zig
```diff
@@ -169,11 +176,17 @@ pub const ClientEntityManager = struct {
 			const xCenter = (1 + projectedPos[0]/projectedPos[3])*@as(f32, @floatFromInt(main.Window.width/2));
 			const yCenter = (1 - projectedPos[1]/projectedPos[3])*@as(f32, @floatFromInt(main.Window.height/2));
 
-			graphics.draw.setColor(0xff000000);
+			const maxDistanceSquare = 128.0*128.0;
+			const minAlpha = 0x28;
+			const transparency = 1 - vec.lengthSquare(pos3d)/maxDistanceSquare;
```
IntegratedQuantum: Instead of having a fixed cutoff, how about using some kind of logistic curve, or something like `e^(-x²)`?

src/entity.zig
```diff
@@ -169,11 +176,17 @@ pub const ClientEntityManager = struct {
...
+			const alpha: u32 = @intFromFloat(std.math.clamp(transparency*0xff, minAlpha, 0xff));
```
IntegratedQuantum: Should we really stop the fading? I think it would be nice to have it fade away in some logarithmic manner, getting slower and slower the further you are away, but still showing a notable difference between 1k and 10k or 10k and 100k.

bagggage: Sounds not bad to me, I will try it.

bagggage: I found it quite difficult to find a good logarithmic function that would show a noticeable difference over long distances (1k-10k), while still having enough difference over short distances (0-100 blocks).

Some progress with `ln`:
image
`y [0,255]` - alpha channel;
`x [0,inf)` - distance in blocks;

~100 blocks:
Screenshot From 2025-11-08 15-48-14

~200 blocks:
Screenshot From 2025-11-08 15-47-48

~500 blocks:
Screenshot From 2025-11-08 15-47-27

~1000 blocks:
Screenshot From 2025-11-08 15-46-38

~6000 blocks:
Screenshot From 2025-11-08 15-51-06

IntegratedQuantum: That does seem rather complex.
The two things I would try are `255/log(x + 10)` and `255 - constant*log(x + 1)`. Did you try these?

bagggage: I'll check

bagggage: Found some similar functions that look decent, but the `255 - constant*log(x + 1)` variant works better at long distances, while `constant_a/log(x + constant_b)` decays too slowly and would probably be too subtle. I'll try option 2 (the purple one) in the game.

*Using `x^2` because the distance is calculated with `vec.lengthSquare`*

Long distances:
image

Short distances:
image

bagggage: I found a better variant: `255 - 38*(log(x + 1) - 80)`, added the constant subtraction because the name was becoming transparent too quickly at short distances. I've made the changes - you can check it in the game.

image

IntegratedQuantum:
Looks good to me.
@ikabod-kee are you also happy with this?

ikabod-kee:
Yes 👍

ikabod-kee:
Actually I haven't checked XD give me some time

ikabod-kee:
Huh.. it's actually kind of more janky now. The shadows seem to wiggle as you move further away, and the names don't fully disappear. Additionally the scaling is now off?

ikabod-kee:
Huh wait, I think it might've been bugged because of this being an old version. My fov was weird.

bagggage:
> Huh.. it's actually kind of more janky now

Nothing has changed in the scaling code since last time. I only adjusted the transparency. Scaling is limited to prevent the text from becoming too small and unreadable.

bagggage:
> Also the scaling bug seemed to be an issue with this version specifically.

Is this something I can fix, or shouldn't I worry about it?

ikabod-kee:
Actually, wait wait wait

ikabod-kee:
https://github.com/user-attachments/assets/a5ee623b-afe1-411b-9d11-38d9025fcfd4

Okay, I think the text scaling (from far away) should be a consistent size.
I also realize that this isn't just a "this version" issue. The text size changes based on your FOV.

ikabod-kee:
So yes, sadly this is something you must fix 😔

bagggage:
Yes, the size depends on the FOV - that's what makes it "tied" to the screen size of 3D objects. I think the issue is that the initial size is too large with a small FOV value. There are some minimum world size parameters in place to make the text appear larger when the game is running in fullscreen mode compared to windowed mode. There might be a bug or oversight there - I'll look into it.

ikabod-kee:
Any update on this?

bagggage:
Yes, I simplified the algorithm - now the size doesn't depend on FOV, which fixes the issue. I also slightly adjusted the values, so the overall appearance remains consistent with the previous version (though the text might now appear slightly smaller in fullscreen mode compared to the old version).

Let me know what you think.

ikabod-kee:
I will check it out shortly! Thank you!

ikabod-kee:
https://github.com/user-attachments/assets/a5435b39-6840-45e4-a033-8c38112af517

There seems to be an issue! At minimum size, the letters are not scaled pixel-perfectly. Notice the misalignments.

bagggage:
The position is not pixel perfect, but the font size itself is still 18 pixels, give me a minute

bagggage:
Long story short, 18 pixels for a 16-pixel pixel font was a poor choice. As for the position, I tried making it pixel-aligned on screen, but... it looks horribly jittery in motion, so I decided to just go with a minimum size of 16 pixels.

bagggage:
Alright, I got confused with the positioning - I’m not even sure anymore if my change affected anything at all - but I think it’s not that important. The main thing is that the minimum size now matches the font.

ikabod-kee:
> Long story short, 18 pixels for a 16-pixel pixel font was a poor choice

Yeah, now imagine the trouble we're gonna go through replacing it in the future XD

Anyways, cool, let me check it out.

src/entity.zig
```diff
@@ -153,13 +153,19 @@ pub const ClientEntityManager = struct {
 		mutex.lock();
 		defer mutex.unlock();
 
+		// this value is relative to the fov-y and window size
```
IntegratedQuantum: outdated comment, goes to show why it's usually better to put this information into variable names.

PR #1994, opened 2025-10-14
Turn up the teleport search radius

codemob-dev:
Turns the teleport search radius up to 24576 (x1.5). No longer do you have to teleport to random locations to find your biome!


Argmaster:
Well, no matter how high you raise it, unless it is the size of the world, it won't be enough

codemob-dev:
> Well, no matter how high you raise it, unless it is the size of the world, it won't be enough

Since this fixes most issues with teleporting I think it makes sense for this to be the default.

codemob-dev:
Fair enough.

Issue #1990, opened 2025-10-14
Pull items from inventory when pick-blocking (middle click)


careeoki:
### Description

For survival mode. 
Example: you pick-block slate, which you have in your inventory. it moves the slate to your selected hotbar slot.

PR #1997, opened 2025-10-14
Improved MMB item picking

eve-v0:
Fix https://github.com/PixelGuys/Cubyz/issues/1990
item picking will work for items in inventory

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -587,19 +587,17 @@ pub const Player = struct { // MARK: Player
 							}
 							break :blk selectedSlot;
 						};
-						std.log.info("slot for found item: {any}" ,.{targetSlot});
 						
 						var carried: Inventory = undefined;
 						carried = Inventory.init(main.globalAllocator, 1, .normal, .{.hand = main.game.Player.id}, .{});
+						defer carried.deinit(main.globalAllocator);
 
-						//inventory.deposit(@intCast(slotIdx), carried, inventory.getAmount(slotIdx));
 						inventory.takeHalf(@intCast(slotIdx), carried);
 						inventory.takeHalf(@intCast(slotIdx), carried);
-						std.log.info("carried: {any}",.{carried});
 						
 						inventory.depositOrSwap(@intCast(targetSlot), carried);
 						inventory.depositOrSwap(@intCast(slotIdx), carried);
-						carried.deinit(main.globalAllocator);
+						
 						return;	
 					}
 				}


src/game.zig
```diff
@@ -568,13 +568,32 @@ pub const Player = struct { // MARK: Player
...
-			for(0..12) |slotIdx| {
+			for(0..30) |slotIdx| {
 				if(std.meta.eql(inventory.getItem(slotIdx), item)) {
 					if(isCreative()) {
 						inventory.fillFromCreative(@intCast(slotIdx), item);
 					}
-					selectedSlot = @intCast(slotIdx);
-					return;
+					if(slotIdx <= 12)
+					{
```
copygirl: Bracket placement doesn't look consistent with existing code.

src/game.zig
```diff
@@ -568,13 +568,32 @@ pub const Player = struct { // MARK: Player
...
+							if(inventory.getItem(selectedSlot) == null) break :blk selectedSlot;
+							// Look for an empty slot
+							for(0..12) |slotId| {
+								if(inventory.getItem(slotId) == null) {
+									break :blk slotId;
+								}
+							}
+							break :blk selectedSlot;
+						};
+					std.log.info("slot for found item: {any}" ,.{targetSlot});
```
copygirl: Debug print left in the code?

src/game.zig
```diff
@@ -568,13 +568,32 @@ pub const Player = struct { // MARK: Player
...
+					}
+					
+
```
copygirl: Empty lines?

eve-v0:
@copygirl this is a draft for a reason

copygirl:
> this is a draft for a reason

Yes, so people can give feedback.

**edit:** I see my feedback was marked resolved without being resolved.
I don't think this is how that feature is supposed to be used.

eve-v0:
I am trying to figure out solution and that's why theres for example debug code print left in so your feedback to remove is just annoying and its all irrevelant anyways because this is outdated
formatting is a last thing to do
i would appricate your feedback if it was actually about the issue i try to resolve or the way i try to do it and not formatting

copygirl:
I've commented on the code as it is presented in the PR, so it's not outdated. I simply did not see that the code is currently a no-op. (A comment could've clarified.) That doesn't change the fact that formatting may not be accepted as is, and the debug-print should be removed before the PR is merged, as such is isn't irrelevant. And I'm not sure how to phrase this right, but there was no reason to get annoyed.

As for perhaps more appropriate feedback, I'm not familiar with inventory management, but the code currently has a bunch of hardcoded values. It'd benefit from making it more obvious that `< 12` are hotbar slots, and above are inventory.

src/game.zig
```diff
@@ -568,13 +568,32 @@ pub const Player = struct { // MARK: Player
...
 			// Check if there is already a slot with that item type
-			for(0..12) |slotIdx| {
+			for(0..30) |slotIdx| {
 				if(std.meta.eql(inventory.getItem(slotIdx), item)) {
 					if(isCreative()) {
 						inventory.fillFromCreative(@intCast(slotIdx), item);
 					}
-					selectedSlot = @intCast(slotIdx);
-					return;
+					if(slotIdx <= 12)
```
copygirl: Logic error? Should likely be `<` if inventory slots are from `0..12` (exclusive).
```suggestion
					if(slotIdx < 12)
```

PR #2008, opened 2025-10-14
Add a disabled state to the button

bagggage:
Adds a `disabled` field to the button struct.
Now any button can be easily disabled by setting `button.disabled = true`.

image_2025-10-15_00-44-01

### DIFF
--- a/src/gui/components/Button.zig
+++ b/src/gui/components/Button.zig
@@ -54,6 +54,7 @@ pub var buttonUniforms: struct {
 
 pos: Vec2f,
 size: Vec2f,
+disabled: bool = false,
 pressed: bool = false,
 hovered: bool = false,
 onAction: gui.Callback,
@@ -127,20 +128,22 @@ pub fn mainButtonPressed(self: *Button, _: Vec2f) void {
 pub fn mainButtonReleased(self: *Button, mousePosition: Vec2f) void {
 	if(self.pressed) {
 		self.pressed = false;
-		if(GuiComponent.contains(self.pos, self.size, mousePosition)) {
+		if(!self.disabled and GuiComponent.contains(self.pos, self.size, mousePosition)) {
 			self.onAction.run();
 		}
 	}
 }
 
 pub fn render(self: *Button, mousePosition: Vec2f) void {
-	const textures = if(self.pressed)
+	const textures = if(self.disabled)
+		normalTextures
+	else if(self.pressed)
 		pressedTextures
 	else if(GuiComponent.contains(self.pos, self.size, mousePosition) and self.hovered)
 		hoveredTextures
 	else
 		normalTextures;
-	draw.setColor(0xff000000);
+	draw.setColor(if(self.disabled) 0xa0000000 else 0xff000000);
 	textures.texture.bindTo(0);
 	pipeline.bind(draw.getScissor());
 	self.hovered = false;
@@ -151,7 +154,7 @@ pub fn render(self: *Button, mousePosition: Vec2f) void {
 		const lowerTexture = (textures.outlineTextureSize - Vec2f{1, 1})/Vec2f{2, 2}/textures.outlineTextureSize;
 		const upperTexture = (textures.outlineTextureSize + Vec2f{1, 1})/Vec2f{2, 2}/textures.outlineTextureSize;
 		textures.outlineTexture.bindTo(0);
-		draw.setColor(0xffffffff);
+		draw.setColor(if(self.disabled) 0xffa0a0a0 else 0xffffffff);
 		// Corners:
 		graphics.draw.boundSubImage(self.pos + Vec2f{0, 0}, cornerSize, .{0, 0}, cornerSizeUV);
 		graphics.draw.boundSubImage(self.pos + Vec2f{self.size[0], 0} - Vec2f{cornerSize[0], 0}, cornerSize, .{upperTexture[0], 0}, cornerSizeUV);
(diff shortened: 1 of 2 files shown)


IntegratedQuantum:
What is the motivation for this?

bagggage:
> What is the motivation for this?

Prevent the receipt of click events in a context where a temporary block or restriction is in effect for that event. Simply put, for example, the user cannot press "Join" to connect to a server as long as the IP address input field is empty, as this is a knowingly unsuccessful event.

The code for handling click events can become simpler now, as there is a kind of guarantee that if a button is pressed, the context is inherently correct.

Secondly, it provides quick visual feedback to the user about the unavailability of an action. Otherwise, for instance, in an invalid context, it would be necessary to provide feedback to the user after the click, such as displaying an error message or a message about invalid input (as in the example above), and this requires additional logic for message delivery and so on.

IntegratedQuantum:
Could you then please implement this for the `Join` button? It's easier to review things if I can test them.

bagggage:
> It's easier to review things if I can test them.

You can test this right now on the [server-list](/bagggage/Cubyz/tree/server-list) (#2017) branch. But if there are any objections, I can implement it for the `Join` button on the current branch.

BoySanic:
It'll be easier to review if you can implement it on a button that exists in the master branch, like the ``Join` button suggestion

bagggage:
> It'll be easier to review if you can implement it on a button that exists in the master branch, like the ``Join` button suggestion

Okay, I implemented some network address validation with disabling the `join` button in the `Multiplayer` window.

image

src/gui/windows/multiplayer.zig
```diff
@@ -85,6 +86,46 @@ fn copyIp(_: usize) void {
...
+		var iter = std.mem.splitScalar(u8, str, ':');
+		break :blk iter.first();
+	};
+	if(str.len > portLess.len) {
+		const port = str[portLess.len + 1 ..];
+		_ = std.fmt.parseUnsigned(u16, port, 10) catch return false;
+	}
+
+	// check if this is a valid domain name
+	const validDomain: bool = blk: {
```
IntegratedQuantum: I'd suggest to use `std.net.isValidHostName`

src/gui/windows/multiplayer.zig
```diff
@@ -85,6 +86,46 @@ fn copyIp(_: usize) void {
...
+	// min doamin name length is 2
+	if(str.len < 2) return false;
+
+	// check if port is valid
+	const portLess = blk: {
+		var iter = std.mem.splitScalar(u8, str, ':');
+		break :blk iter.first();
+	};
+	if(str.len > portLess.len) {
+		const port = str[portLess.len + 1 ..];
```
IntegratedQuantum: Cubyz ports are allowed to start with a `?` which signals a symmetric NAT connection.

src/gui/windows/multiplayer.zig
```diff
@@ -85,6 +86,46 @@ fn copyIp(_: usize) void {
...
 }
 
+fn isCorrectNetAddress(str: []const u8) bool {
+	// min doamin name length is 2
+	if(str.len < 2) return false;
+
+	// check if port is valid
+	const portLess = blk: {
+		var iter = std.mem.splitScalar(u8, str, ':');
+		break :blk iter.first();
```
IntegratedQuantum: How about using iter.rest() for the port string?

src/gui/components/Button.zig
```diff
@@ -127,20 +128,22 @@ pub fn mainButtonPressed(self: *Button, _: Vec2f) void {
...
+	const textures = if(self.disabled)
+		normalTextures
+	else if(self.pressed)
 		pressedTextures
 	else if(GuiComponent.contains(self.pos, self.size, mousePosition) and self.hovered)
 		hoveredTextures
 	else
 		normalTextures;
-	draw.setColor(0xff000000);
+	draw.setColor(if(self.disabled) 0xa0000000 else 0xff000000);
```
IntegratedQuantum, line 146: This should be a separate texture, to give more artistic control.

IntegratedQuantum (changes requested):

src/gui/windows/multiplayer.zig
bagggage: Oh, I didn't notice it before

src/gui/windows/multiplayer.zig
```diff
@@ -85,6 +86,28 @@ fn copyIp(_: usize) void {
...
+		if(portPart.len == 0) return false;
+
+		const port = if(portPart[0] == '?') portPart[1..] else portPart;
+		_ = std.fmt.parseUnsigned(u16, port, 10) catch return false;
+	}
+
+	if(std.net.isValidHostName(address)) return true;
+
+	// check if this is a valid IP address
+	_ = std.net.Address.parseIp(address, 0) catch return false;
```
IntegratedQuantum: valid hostname is also true for valid IPv4 addresses I believe. So I don't think this is actually needed.

IntegratedQuantum (changes requested):

src/gui/windows/multiplayer.zig
bagggage: But what about IPv6?

IntegratedQuantum: Isn't supported, therefor the button should be gray for those as well.

bagggage: Okay, I just found that it doesn't recognize IPv6 as a valid address anyway

IntegratedQuantum:
superseded by #3203

Issue #1574, opened 2025-06-01
Typing cursor is really hard to see on high resolution monitors


RanPix:
![Image](https://github.com/user-attachments/assets/f4a57801-fecf-4910-a873-cc3b7ff711ed)

PR #2010, opened 2025-10-14
More visible TextInput cursor

MaxHobbyDev:
Closes: #1574

Increase the visibility of the Cursor in the input text boxes by calculating a thickness based on the font size.
I also made the cursor blink.

Gif:
Normal Cursor

Large Resolution:
image

Smaller:
image

I did also play with another style of underscore cursor which I like but it's probably better to keep the more common style. (No blinking)
Underscore Cursor

### DIFF
--- a/src/gui/components/TextInput.zig
+++ b/src/gui/components/TextInput.zig
@@ -522,7 +522,7 @@ pub fn render(self: *TextInput, mousePosition: Vec2f) void {
 		}
 
 		const milliTime = std.time.milliTimestamp();
-		if(self.lastBlinkTime + blinkDurationMs <= milliTime) {
+		if(milliTime -% self.lastBlinkTime > blinkDurationMs) {
 			self.lastBlinkTime = milliTime;
 			self.showCusor = !self.showCusor;
 		}


src/gui/components/TextInput.zig
```diff
@@ -514,7 +519,17 @@ pub fn render(self: *TextInput, mousePosition: Vec2f) void {
 			draw.setColor(0x440000ff);
 			self.textBuffer.drawSelection(textPos, @min(selectionStart, cursor), @max(selectionStart, cursor));
 		}
-		draw.setColor(0xff000000);
-		draw.line(cursorPos, cursorPos + Vec2f{0, 16});
+
+		const milliTime = main.game.world.?.milliTime;
```
IntegratedQuantum: This time is inaccurate. It does not show the time of the last frame.
I'd suggest to instead just use std.time directly.

src/gui/components/TextInput.zig
```diff
@@ -34,6 +34,9 @@ textSize: Vec2f = undefined,
 scrollBar: *ScrollBar,
 onNewline: gui.Callback,
 optional: OptionalCallbacks,
+blinkDurationMs: i64 = 500.0,
```
IntegratedQuantum: Should be constant in my opinion.

MaxHobbyDev:
> When I click in the text input it sometimes seems to start in the invisible phase which is kind of confusing.

Thanks for pointing this out.

I did notice that when the text is highlighted the cursor is hard to see but I think that can be a different issue/PR:
image

src/gui/components/TextInput.zig
```diff
@@ -514,7 +520,17 @@ pub fn render(self: *TextInput, mousePosition: Vec2f) void {
 			draw.setColor(0x440000ff);
 			self.textBuffer.drawSelection(textPos, @min(selectionStart, cursor), @max(selectionStart, cursor));
 		}
-		draw.setColor(0xff000000);
-		draw.line(cursorPos, cursorPos + Vec2f{0, 16});
+
+		const milliTime = std.time.milliTimestamp();
+		if(self.lastBlinkTime + blinkDurationMs <= milliTime) {
```
IntegratedQuantum: We don't want bugs/crashes on integer overflow, so please use arithmetic that is aware of it:
```suggestion
		if(milliTime -% self.lastBlinkTime > blinkDurationMs) {
```

IntegratedQuantum (changes requested):

PR #2021, opened 2025-10-15
Fix empty crafting window being too large

### LINKED ISSUE #2020: Empty crafting window size not being set / too large

When you don't have any items and attempt to open the crafting window for the first time that session, its dimensions are much too large:

Image

copygirl:
Fixes #2020.

I opted to go for the simplest fix for now: Simply setting up the window with an empty list of crafting recipes.

Perhaps a more ideal solution would be for there to be a text saying "no recipes found" or so.
Let me know if I should address this with this PR or if that's wasted effort.

### DIFF
--- a/src/gui/windows/inventory_crafting.zig
+++ b/src/gui/windows/inventory_crafting.zig
@@ -124,30 +124,22 @@ fn findAvailableRecipes(list: *VerticalList) bool {
 }
 
 fn refresh() void {
+	const oldScrollState = if(window.rootComponent) |oldList| oldList.verticalList.scrollBar.currentState else 0;
 	const list = VerticalList.init(.{padding, padding + 16}, 300, 8);
 	const recipesChanged = findAvailableRecipes(list);
-	const hasRootComponent = (window.rootComponent != null);
-	if(!recipesChanged and hasRootComponent) {
+	if(!recipesChanged and window.rootComponent != null) {
 		list.deinit();
 		return;
 	}
-
-	const oldRootComponent = window.rootComponent;
-	defer if(oldRootComponent) |*comp| comp.deinit();
-
-	if(list.children.items.len == 0) {
-		list.deinit();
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
> Perhaps a more ideal solution would be for there to be a text saying "no recipes found" or so.

Yes that would be useful. I've seen that some players are confused by the empty window.

IntegratedQuantum:
Though I would suggest the more explicit "No craftable recipes found"

copygirl:
Turns out my assumptions were wrong, the initial code did not work.

- `refresh` is called more often than I thought, not just (somehow) when the window needed to update.
- `findAvailableRecipes` returns whether the recipes have changed since last call, I believe.
- However, it also returns `false` when the window is created initially, so we need to special-case that. That's what `hasRootComponent` is for. If no root component has been added yet, the `Label` *will* be added.

This is what the crafting window looks like as a result:

image

src/gui/windows/inventory_crafting.zig
```diff
@@ -123,18 +124,30 @@ fn findAvailableRecipes(list: *VerticalList) bool {
...
-	if(window.rootComponent) |*comp| {
-		comp.deinit();
+
+	const oldRootComponent = window.rootComponent;
+	defer if(oldRootComponent) |*comp| comp.deinit();
+
+	if(list.children.items.len == 0) {
+		list.deinit();
+		const label = Label.init(.{padding, padding + 16}, 120, "No craftable\nrecipes found", .center);
+		window.rootComponent = label.toComponent();
```
IntegratedQuantum: I think you could simplify the logic a lot if you just do `list.add(label)` here, instead of using the label as the root component directly.

copygirl: Whelp, yeah, that simplifies the function (and changes) a whole lot.

IntegratedQuantum (approved):

PR #2024, opened 2025-10-16
Implement culling in VerticalList.render

### LINKED ISSUE #1991: My framerate halves when UI is open

Image
Standing still, framerate hovering at ~30 fps

Image
Standing still, UI open, framerate hovering at ~15fps

BoySanic:
Fixes #1991

Items that do not fall within the bounds of the guiwindow are not
rendered.

Whether a item is within view or not is determined by calculating the
item's position in the list and the item's position offset by the scrollbar's
position

If that offset value is < 0, skip rendering
If that offset value is > self.maxHeight, skip rendering.

### DIFF
--- a/src/gui/components/VerticalList.zig
+++ b/src/gui/components/VerticalList.zig
@@ -139,7 +139,7 @@ pub fn render(self: *VerticalList, mousePosition: Vec2f) void {
 		const itemYPos = child.pos()[1];
 		const adjustedYPos = itemYPos + shiftedPos[1];
 
-		if(adjustedYPos + child.size()[1] < 0 or adjustedYPos > self.maxHeight + child.size()[1]) {
+		if(adjustedYPos + child.size()[1] < 0 or adjustedYPos - child.size()[1] > self.maxHeight + child.size()[1]) {
 			continue;
 		}
 		child.render(mousePosition - shiftedPos);


BoySanic:
Before this change, ~100 messages would halve my framerate from 850 to ~400 
After this change, it stays steady at ~850

BoySanic:
Should be fixed now.

I was calculating an average of item height, which doesn't work because not all items in a VerticalList have the same height.

So I'm now using the child's size()[1] as its height and adding to `currYPos` in the for loop to calculate an item's real YPos within the clip rect.

I then had to add the childHeight to both conditions of the early continue if statement to pad it by 1 item's height so that the next message is rendered before it's fully in view.

Wunka:
I have no experience with the codebase but by looking a little bit at it. Does `child.pos()[1] + shiftedPos[1]` or `child.pos()[1] + child.size()[1] + shiftedPos[1]` not give you all the data you need to compare it to the visible list and if it is inside it? In `add`  this is all already calculated or not?

IntegratedQuantum (changes requested):
it also appears to break in the controls window

BoySanic:
Issue in controls window should be fixed.

I also investigated what @Wunka suggested and I did indeed overcomplicate my solution.

I appreciate you pitching in :)

BoySanic:
Do you mean in the controls window, or chat?

I thought I caused a problem in controls too until I checked in `master` and it has the same behavior of cutting off the bottommost control option.

IntegratedQuantum:
I saw it only in chat.

BoySanic:
Master:
image


Separate issue. I did not cause a regression I think if this is what you're talking about.

IntegratedQuantum:
I mean it removes it before it disappears below the edge of the container:
Screenshot at 2025-10-17 20-08-25

BoySanic:
This doesn't happen on my machine, so I'm not sure.

image

As I scroll I can watch the text disappear outside of the clip rect and I never see it stop rendering, just disappear from view.

BoySanic:
https://github.com/user-attachments/assets/85f67cfc-e8ef-4438-a174-5549ffb284ce

Wunka:
For me it happens too but not with multi line messages.

BoySanic:
I think I see what's going on. One sec, I'll update it

BoySanic:
Should work now. I wasn't subtracting the child item's height from the shifted position before.

IntegratedQuantum (approved):
It appears to work now, thank you

PR #2032, opened 2025-10-17
Fix item tooltip obstructing items and overflowing

Bluperman949:
Currently, the item tooltip has a few problems:
- It renders directly under the mouse
- It obstructs the item being hovered over
- The yellow border can overflow past the edges of the screen
- The tooltip flips vertically when near the bottom of the screen, which I find distracting

This implements the following fixes:
- Horizontally offset the tooltip so the item underneath is visible
- Vertically aligned the tooltip's first line with the mouse's position so it feels more centered without being unpredictable
- Made the tooltip butt up against the bottom of the screen instead of vertically flipping
- Cleaned up the handling of border/padding offsets so the tooltip never escapes the edges of the screen

https://github.com/user-attachments/assets/b6e5fb26-c06b-4d34-b298-f2e48f5089f2

### DIFF
--- a/src/gui/gui.zig
+++ b/src/gui/gui.zig
@@ -769,26 +769,29 @@ pub const inventory = struct { // MARK: inventory
 				const tooltip = item.getTooltip();
 				var textBuffer = graphics.TextBuffer.init(main.stackAllocator, tooltip, .{}, false, .left);
 				defer textBuffer.deinit();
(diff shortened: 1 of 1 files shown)


Bluperman949:
Don't mind me, just horribly fumbling with git for some reason

IntegratedQuantum (approved):
Looks good, thank you

Issue #1396, opened 2025-05-03
Tree Revamp Tracking


ikabod-kee:
### **Oak**
- [ ] **Tall Oaks** (Plains)
- [ ] **Oak Trees** (Forest, Plains) [need to be redone with new logs]
- [ ] **Small Oaks** (Prairie, Mountains, Limestone Mountains, etc)
### **Birch**
- [ ] **Thin Birches** (Thin Birch Forest, Birch Forest)
- [ ] **Birch Trees** (Birch Forest, Forest) [need to be redone with new logs]
### **Pine**
- [x] **Pine Trees** (Taiga)
- [x] **Tall Pines** (Taiga)
### **Mahogany**
- [ ] **Tall Mahogany** (Jungle)
- [ ] **Mahogany Trees** (Jungle)
### **Baobab**
- [x] **Baobab Trees** (Savannah, Limestone Mountains)
- [ ] **Acacia-styled Baobabs** (Savannah)
### **Willow**
- [ ] **Willow Trees** (Swamp)
- [ ] **Wetland Willows** (Wetlands)
### **General / Substitutional**
- [x] **Cacti** (Desert)
- [ ] **Bushes** (Bushlands, Mountains, etc)
- [ ] **Small Bushes** (Bushlands, Mountains, Prairie, etc)
- [ ] **Saplings**
- [ ] **Dead Trees**
- [ ] **Fallen Trees**
- [x] **Giant Mushrooms**
- [ ] **Autumn Trees** (Block replacements will be used)

Argmaster:
Can we have sapling-like/young trees too? Like made only with branches, 3-5 blocks tall?

IntegratedQuantum:
I have updated the PR descritpion, are tall pines also considered done with #1377 ?

ikabod-kee:
Yes. I'll check them off.

codemob-dev:
@ikabod-kee You should probably look over this again, there are a few things that are done but not checked off.

ikabod-kee:
Am I missing any now?

codemob-dev:
I don't think so, it looks like everything is there.

IntegratedQuantum:
I think now is also the time to again think about adding resin #1325 to trees.

ikabod-kee:
> I think now is also the time to again think about adding resin [#1325](https://github.com/PixelGuys/Cubyz/issues/1325) to trees.

About that... only legacy logs can have resin applied to it, and only in the up-right rotation

PR #2042, opened 2025-10-17
Baobab SBBs

codemob-dev:
Progress towards #1396

Adds 1 new biome, baobab_shrublands made primarily of the new grandidieri baobabs. Replaces all usages of baobab simple trees with the new SBBs.


IntegratedQuantum:
It does not fix #1895 because the blocks still have the same configuration error, and will still appear when you have multiple trees overlap.

assets/cubyz/biomes/savannah/baobab_shrublands.zig.zon
```diff
@@ -0,0 +1,86 @@
+.{
```
IntegratedQuantum, line 1: I would propose putting this into the savannah subfolder, since it's kind of related.

codemob-dev: Makes sense

codemob-dev:
> I looked around in the biomes a bit and I think the base savannah could deserve a small chance at some of those older trees from the shrublands. It is weird that all of them are young.

Fixed.

IntegratedQuantum:
Please add some roots
Screenshot at 2025-10-19 20-15-26

codemob-dev:
Done!

image

PR #2059, opened 2025-10-18
improve logging

cn-cn-cn:


cn-cn-cn:
CI failure seems unrelated?

src/main.zig
```diff
@@ -213,6 +213,25 @@ pub const std_options: std.Options = .{ // MARK: std_options
...
+	const addr = first_trace_address orelse @returnAddress();
+	std.log.err("This is a bug, please report it on the issue tracker.\n----8<---- start of panic\npanic: {s}\nerror return trace: {?f}", .{
+		msg,
+		@errorReturnTrace(),
+	});
+
+	var trace_buf: [1 << 20]u8 = undefined;
+	var fbw = std.io.Writer.fixed(&trace_buf);
+	std.debug.dumpCurrentStackTraceToWriter(addr, &fbw) catch {
+		std.log.err("failed to dump stack trace", {});
```
copygirl:
```suggestion
		std.log.err("failed to dump stack trace", .{});
```

copygirl (changes requested):
This doesn't look right. Could be what's causing the CI failure.

cn-cn-cn:
Ha, I tested the failure path but not that one. Thanks!

src/main.zig
```diff
@@ -213,6 +213,25 @@ pub const std_options: std.Options = .{ // MARK: std_options
 	}.logFn,
 };
 
+pub fn panicToLog(msg: []const u8, first_trace_address: ?usize) noreturn {
+	const addr = first_trace_address orelse @returnAddress();
+	std.log.err("This is a bug, please report it on the issue tracker.\n----8<---- start of panic\npanic: {s}\nerror return trace: {?f}", .{
```
IntegratedQuantum: This is not entirely true, since panic is also called when running out of memory.
I'd suggest "If you think this is a bug, please ..."

src/main.zig
```diff
@@ -213,6 +213,25 @@ pub const std_options: std.Options = .{ // MARK: std_options
...
+		msg,
+		@errorReturnTrace(),
+	});
+
+	var trace_buf: [1 << 20]u8 = undefined;
```
IntegratedQuantum: This is bigger than the actual logging buffer, I'd suggest to use the same size for both.

src/main.zig
IntegratedQuantum: Also what is that ----8<---- supposed to mean?

cn-cn-cn: Thanks. Would you be fine with "This is probably a bug"? Also, on second thought, I should add a link.

`8<` is intended to resemble scissors, as in, "cut here". If you didn't see it, users probably won't either, do you have a suggestion for making it clearer? The `----` is just a separator. Both are there to tell the user what the important data is.

codemob-dev: Oh, now I see it!

IntegratedQuantum: > 8< is intended to resemble scissors, as in, "cut here".

Ideally the user should just send the whole log file.

src/main.zig
```diff
@@ -249,27 +271,27 @@ fn deinitLogging() void {
 	}
 }
 
-fn logToFile(comptime format: []const u8, args: anytype) void {
-	var buf: [65536]u8 = undefined;
-	var fba = std.heap.FixedBufferAllocator.init(&buf);
-	const allocator = fba.allocator();
+fn logToWriter(writer: *std.Io.Writer, comptime format: []const u8, args: anytype) std.Io.Writer.Error!void {
```
IntegratedQuantum: You should remove the error return, since it's ignored everywhere anyways.
These log functions are very big contributors to compile time.

src/main.zig
```diff
@@ -213,6 +213,28 @@ pub const std_options: std.Options = .{ // MARK: std_options
...
+		@errorReturnTrace(),
+	});
+
+	var trace_buf: [log_buffer_size - "stack trace: ".len]u8 = undefined;
+	var fbw = std.io.Writer.fixed(&trace_buf);
+	std.debug.dumpCurrentStackTraceToWriter(addr, &fbw) catch {
+		std.log.err("failed to dump stack trace", .{});
+	};
+	std.log.err("stack trace: {s}", .{fbw.buffered()});
+	std.log.err("end of panic\n", .{});
```
IntegratedQuantum: I don't think this is needed anymore

src/main.zig
```diff
@@ -213,6 +213,28 @@ pub const std_options: std.Options = .{ // MARK: std_options
 	}.logFn,
 };
 
+/// The maximum size a log message can have.
+const log_buffer_size = 64 << 10;
+
+pub fn panicToLog(msg: []const u8, first_trace_address: ?usize) noreturn {
+	const addr = first_trace_address orelse @returnAddress();
+	std.log.err("This is probably a bug, if you think so please make and issue and add the entire log: <https://github.com/pixelguys/cubyz/issues/new?template=bug.yml>\n----8<---- start of panic\npanic: {s}\nerror return trace: {?f}", .{
```
IntegratedQuantum: Please also remove the starting scissors and start of panic.

src/main.zig
```diff
@@ -213,6 +213,28 @@ pub const std_options: std.Options = .{ // MARK: std_options
...
+	});
+
+	var trace_buf: [log_buffer_size - "stack trace: ".len]u8 = undefined;
+	var fbw = std.io.Writer.fixed(&trace_buf);
+	std.debug.dumpCurrentStackTraceToWriter(addr, &fbw) catch {
+		std.log.err("failed to dump stack trace", .{});
+	};
+	std.log.err("stack trace: {s}", .{fbw.buffered()});
+	std.log.err("end of panic\n", .{});
+	std.debug.defaultPanic(msg, addr);
```
IntegratedQuantum: This prints it again. Is there a way to avoid printing the panic twice?

IntegratedQuantum (changes requested):
I also noticed that there are a bunch of ANSI escape characters in the output files, maybe you can filter them? or configure it not to print them?

src/main.zig
BoySanic:
```suggestion
	std.log.err("This is probably a bug. If you think so, please make an issue and add the entire log: <https://github.com/pixelguys/cubyz/issues/new?template=bug.yml>\n----8<---- start of panic\npanic: {s}\nerror return trace: {?f}", .{
```

Tiny suggestion, but and -> an and slight punctuation changes

src/main.zig
cn-cn-cn: I will measure compile time and report back.

src/main.zig
cn-cn-cn: Thanks! Will apply.

src/main.zig
cn-cn-cn: Sorry, "this"? You mean the end of panic?

src/main.zig
cn-cn-cn: You can silence the `err` log level in `std.Options`, this isn't true for stderr, but since the log function always logs to stderr, I will remove it and assume that all error logs are included.

cn-cn-cn:
About the ANSI escapes, I will not filter them in this PR because I have another one coming up that adds ANSI escape filtering in other contexts.

src/main.zig
IntegratedQuantum: > You can silence the err log level in std.Options

Yes, but it's under our control, and I won't approve any PR that silences it.

src/main.zig
IntegratedQuantum: yes.

cn-cn-cn:
It turns out that there is no need for ANSI filtering, and the need for configurable ANSI formatting is now [expressed in a comment](https://github.com/ziglang/zig/commit/b64535e3c8770db9bf314fc14d1b2b276b864f72#diff-7194f8244f143a85a7a746c039e71c572a156dda1130e629c7655e9dcdb0ff4aR49).

src/main.zig
cn-cn-cn: image

Measuring compile time found no significant difference.

src/main.zig
```diff
@@ -213,6 +214,50 @@ pub const std_options: std.Options = .{ // MARK: std_options
...
+const log_buffer_size = 64 << 10;
+
+fn dumpStackTraceNoAnsi(writer: *std.Io.Writer, start_addr: ?usize) !void {
+	if(builtin.target.cpu.arch.isWasm()) {
+		if(builtin.os.tag == .wasi) {
+			try writer.writeAll("Unable to dump stack trace: not implemented for Wasm\n");
+		}
+		return;
+	}
+	// TODO: update when cubyz' zig gets updated to b64535e, which
```
IntegratedQuantum, line 226: Please also make an issue for it.

src/main.zig
```diff
@@ -213,6 +214,50 @@ pub const std_options: std.Options = .{ // MARK: std_options
 	}.logFn,
 };
 
+/// The maximum size a log message can have.
+const log_buffer_size = 64 << 10;
+
+fn dumpStackTraceNoAnsi(writer: *std.Io.Writer, start_addr: ?usize) !void {
+	if(builtin.target.cpu.arch.isWasm()) {
```
IntegratedQuantum: Please no code for platforms that will never be supported.

src/main.zig
```diff
@@ -213,6 +214,50 @@ pub const std_options: std.Options = .{ // MARK: std_options
...
+	var trace_buf: [log_buffer_size - "stack trace: ".len]u8 = undefined;
+	var fbw = std.io.Writer.fixed(&trace_buf);
+	dumpStackTraceNoAnsi(&fbw, addr) catch {};
+	std.log.err("stack trace: {s}", .{fbw.buffered()});
+
+	@breakpoint();
+	@trap();
+}
+
+pub const panic = std.debug.FullPanic(panicToLog);
```
IntegratedQuantum: Can you put this line next to the other std_options please?

src/main.zig
```diff
@@ -249,27 +294,27 @@ fn deinitLogging() void {
...
+	try writer.writeAll(string);
+}
 
-	const string = std.fmt.allocPrint(allocator, format, args) catch format;
-	defer allocator.free(string);
-	(logFile orelse return).writeAll(string) catch {};
-	(logFileTs orelse return).writeAll(string) catch {};
+fn logToFile(comptime format: []const u8, args: anytype) void {
+	{
+		var writer = (logFile orelse return).writerStreaming(&.{});
```
IntegratedQuantum: How does the streaming writer work? If it's unbuffered it may produce an uncomfortable amount of syscalls.

cn-cn-cn: The only difference between `writerStreaming` and `writer` is that `writer` tries to change the file position in the writer only, while `writerStreaming` uses and changes the actual file position of the file descriptor. In this case, I used `writerStreaming`, because we don't have a global writer, so the file position is handled by the OS.

The writer is unbuffered because we already do our own buffering: `logToWriter` writes everything to a buffer `buf` first, then writes it to `writer`, using as few syscalls as possible (1 in the general case, but `writeAll` will write the whole buffer instead of returning a short write). This is inherited from the logging code that was there before, and I don't know what the original reasoning behind this was. Equivalent (or better, as the original has an arbitrary limit for log message length) functionality can be achieved by using a buffered writer and flushing.

src/main.zig
cn-cn-cn: #2153

src/main.zig
IntegratedQuantum: Yeah, that is probably a good idea. Could you make an issue or PR for that then?

cn-cn-cn: Do you mind if I add it to this PR, since it's already changing a lot of logging code?

src/main.zig
IntegratedQuantum: +15% does seem like a lot to me. It might be worth taking a closer look here to get more significant measurements.
When I measured compile time the margin of error never was this big. Also what are you even measuring at 200s? debug compile time should be around 10-20 s.

IntegratedQuantum: Either way you should remove the error return, since it isn't needed.

src/main.zig
IntegratedQuantum: go for it

src/main.zig
cn-cn-cn: I can't quite explain the added 15% on master, I assume that's due to added code and will measure again with the last commit from `panic-to-log` and in `Debug` mode.

If that shows it to have a large effect on compile time, I will of course remove it, but other than performance, what benefit does removing the error return have over making that function more reusable?

IntegratedQuantum: It simplifies the interface.

cn-cn-cn: I've removed the error return.

For some reason, the compile after all the changes in this PR, except adding back the error return for `logToWriter`, still seems to be faster.

image

src/main.zig
```diff
@@ -71,8 +72,14 @@ fn cacheStringImpl(comptime len: usize, comptime str: [len]u8) []const u8 {
 fn cacheString(comptime str: []const u8) []const u8 {
 	return cacheStringImpl(str.len, str[0..].*);
 }
+
+const log_buffer_size = 64 << 10;
 var logFile: ?std.fs.File = undefined;
```
IntegratedQuantum: I just noticed that this is probably a mistake and should be initalized with `null`
Could you look into this as well please?

src/main.zig
```diff
@@ -233,43 +284,47 @@ fn initLogging() void {
...
 	};
+	logFileTsWriter = logFileTs.?.writer(&logFileTsBuffer);
 
 	supportsANSIColors = std.fs.File.stdout().supportsAnsiEscapeCodes();
 }
 
 fn deinitLogging() void {
 	if(logFile) |_logFile| {
+		logFileWriter.interface.flush() catch {};
+		logFileBuffer = undefined;
```
IntegratedQuantum: I understand for the writer, but there is nothing gained by setting the byte buffer to undefined.

IntegratedQuantum (changes requested):

src/main.zig
cn-cn-cn: `logFile` is always initialized in `initLogging`, but `logFileTs` and `supportsANSIColors` are not. I've changed all three to be initialized in the definition, either to `null` or `false`. Personally, I would like to put all the logging-related state in a struct following the "subsystem" pattern in Cubyz' codebase, but I assume that wouldn't be your preferred style?

src/main.zig
```diff
@@ -233,43 +283,45 @@ fn initLogging() void {
...
 	}
 }
 
-fn logToFile(comptime format: []const u8, args: anytype) void {
-	var buf: [65536]u8 = undefined;
-	var fba = std.heap.FixedBufferAllocator.init(&buf);
-	const allocator = fba.allocator();
+fn logToWriter(writer: *std.Io.Writer, comptime format: []const u8, args: anytype) void {
+	writer.print(format, args) catch {};
+	writer.flush() catch {};
```
IntegratedQuantum, line 320: If the buffered writer is shared between all threads, then you need a mutex.

cn-cn-cn: I was about to propose that but wanted to ask if it was somehow already serialized because the original didn't have a mutex either.

IntegratedQuantum: Well the original relied on atomicity of syscalls.

cn-cn-cn: `writeAll` isn't atomic and can produce torn writes, since it has to do as many writes as necessary to output the whole slice. But thanks, I will add a mutex.

src/main.zig
```diff
@@ -289,6 +296,10 @@ fn initLogging() void {
 }
 
 fn deinitLogging() void {
+	std.debug.assert(logMutex.tryLock()); // not thread-safe
+	defer logMutex = undefined;
```
IntegratedQuantum, line 300: This is problematic. There actually already are log messages after this is called (during allocator deiniting). We don't want to deadlock (in case undefined happens to have the state of being locked) in that case.

Furthermore there is always the chance for a panic after this is called, with undefined this would then cause a deadlock in the panic handler.

IntegratedQuantum: Generally I think we shouldn't lock these functions, since any panic inside here would then cause a deadlock in the panic handler.

IntegratedQuantum (changes requested):

src/main.zig
IntegratedQuantum: Also to allow getting any logs in such deadlock situations, we should probably move the log to stdout before the file logging.

cn-cn-cn: Thanks, will implement.

> Generally I think we shouldn't lock these functions, since any panic inside here would then cause a deadlock in the panic handler.

Would you be okay if it was unlocked immediately after checking that no other thread held the log mutex? That way we still get an `unreachable` when we want and don't deadlock.

IntegratedQuantum: Yeah that would be fine.
But honestly, since this happens (and must happen) before/after basically everything else anyways and thus there are no other threads, I don't really get the point of doing this.

Issue #1853, opened 2025-09-20
Per blocktype blockentity data


OneAvargeCoder193:
You might want a container similar to a chest, but with a different amount of slots. Currently, you would need to copy and paste the entire Chest block entity code and the window gui code, and then increase the amount of slots. It would be nice to be able to just have a way to specify `.slotCount = number of slots` somewhere in the zon and have it automatically do all that for you.

OneAvargeCoder193:
Closed as a copy of #1551

PR #2063, opened 2025-10-18
Block Properties refactor

sebat2:
## Summary
Implemented a sorted array system for rarely-used block properties, significantly reducing memory usage.

This is the first step toward addressing #1853.

## Main changes
- Added generic `SortedBlockProperties<size, DataType>` type with binary search for efficient access
- Moved rarely-used properties (`allowOres`, `gui`, `tickEvent`, `touchFunction`, `blockEntity`) from full [65536] arrays to sorted arrays with ~100-300 entries
- Organized all block properties into `BlockProps` structure for better code readability
- Added helper functions `isSortedProp()` and `resetSortedProperties()` for automatic handling of sorted properties

## Performance impact
**Memory savings:** ~5 properties × (65536 - 100) entries × type size ≈ 320 KB with current configuration, with potential for further optimization.

**Access performance:** O(log n) instead of O(1), but with n ≤ 300, the difference is negligible in practice.

## Future work
Next steps for #1853:
- Move most functions from block_entity.zig vtable to their own sorted BlockProps fields
- Possibly remove block_entity entirely in favor of organizing entity blocks as separate files (one file per entity block type)

---
*Also added .vscode and saves to .gitignore*

### DIFF
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -54,16 +54,15 @@ pub const Ore = struct {
 // MARK: Sorted Block Properties
 fn SortedBlockProperties(comptime sortedBlockSize: usize, comptime DataType: type) type {
 	const Cmp = struct {
-			fn less(target: u32, candidate: u32) std.math.Order {
-			if (target == candidate) return .eq;
-			if (target < candidate) return .lt;
+		fn less(target: u32, candidate: u32) std.math.Order {
+			if(target == candidate) return .eq;
+			if(target < candidate) return .lt;
 			return .gt;
 		}
 	};
 
 	// if the block id is in the array, then the property value is true, otherwise false.
-	if(DataType == bool)
-	{
+	if(DataType == bool) {
 		return struct {
 			const Self = @This();
 
@@ -83,32 +82,21 @@ fn SortedBlockProperties(comptime sortedBlockSize: usize, comptime DataType: typ
 				return true;
 			}
 
-			pub fn addBlockProperty(self: *Self, blockId: u32, propVal: bool) void
-			{
-				if(propVal == false)
-				{
+			pub fn addBlockProperty(self: *Self, blockId: u32, propVal: bool) void {
+				if(propVal == false) {
 					return;
 				}
 
-				if (self.allocatedSize + 1 > sortedBlockSize) {
+				if(self.allocatedSize + 1 > sortedBlockSize) {
 					@panic("Failed to add block property. Consider increasing the size of the array");
 				}
 
-				const slice = self.idxLookup[0 .. self.allocatedSize];
+				const slice = self.idxLookup[0..self.allocatedSize];
 
-				const insertIdx = std.sort.lowerBound(
-					u32, 
-					slice,
-					blockId, 
-					Cmp.less
-				);
+				const insertIdx = std.sort.lowerBound(u32, slice, blockId, Cmp.less);
 
-				std.mem.copyBackwards(
-					u32,
-					self.idxLookup[(insertIdx + 1) .. self.allocatedSize + 1], 
-					self.idxLookup[insertIdx .. self.allocatedSize]
-				);
-				
+				std.mem.copyBackwards(u32, self.idxLookup[(insertIdx + 1) .. self.allocatedSize + 1], self.idxLookup[insertIdx..self.allocatedSize]);
+
 				self.idxLookup[insertIdx] = blockId;
 
 				self.allocatedSize += 1;
@@ -118,8 +106,7 @@ fn SortedBlockProperties(comptime sortedBlockSize: usize, comptime DataType: typ
 				self.allocatedSize = 0;
 			}
 		};
-	}
-	else {
+	} else {
 		return struct {
 			const Self = @This();
 
@@ -128,8 +115,8 @@ fn SortedBlockProperties(comptime sortedBlockSize: usize, comptime DataType: typ
 			data: [sortedBlockSize]DataType = undefined,
 
 			fn compare(target: u32, candidate: u32) std.math.Order {
-				if (target == candidate) return .eq;
-				if (target < candidate) return .lt;
+				if(target == candidate) return .eq;
+				if(target < candidate) return .lt;
 				return .gt;
 			}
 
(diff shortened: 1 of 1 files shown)


src/blocks.zig
```diff
@@ -51,43 +51,190 @@ pub const Ore = struct {
...
+			idxLookup: [sortedBlockSize]u32 = undefined,
+			data: [sortedBlockSize]DataType = undefined,
+
+			fn compare(target: u32, candidate: u32) std.math.Order {
+				if(target == candidate) return .eq;
+				if(target < candidate) return .lt;
+				return .gt;
+			}
+
+			pub fn getBlockSortedIdx(self: *const Self, blockId: u32) ?usize {
```
Argmaster, line 123: This doesn't need to be public.

.gitignore
```diff
@@ -18,3 +19,4 @@ test.png
 !/assets/backgrounds/default_background.png
 /assets/cubyz/music
 /assets/cubyz/fonts
+/saves/*
```
Argmaster, line 22: This should no longer be needed?
Also god damn that missing trailing lf

.gitignore
```diff
@@ -3,6 +3,7 @@ logs/
 zig-out/
 zig-cache/
 .zig-cache/
+.vscode/
```
Argmaster, line 6: IDE configurations are strictly forbidden from being mentioned anywhere in the repo.

src/blocks.zig
```diff
@@ -51,43 +51,190 @@ pub const Ore = struct {
...
+	pub var density: [maxBlockCount]f32 = undefined;
+	pub var terminalVelocity: [maxBlockCount]f32 = undefined;
+	pub var mobility: [maxBlockCount]f32 = undefined;
+
+	/// ------------------------------------------------- Sorted Block Properties
+	/// These properties are rarely used, so to save memory we use sorted arrays
+	/// with ~100-300 entries instead of creating arrays with maxBlockCount (65536) entries.
+	/// Magic Value - Increase it if you need to store more block properties.
+	/// Consider creating a separate variable with a comment for a specific property if only one needs a larger size.
+	pub const maxSortedBlockProperties: usize = 100;
```
Argmaster, line 227: Magic value is unfortunately unacceptable, as you don't know how many of them will be added by add-ons. I don't think the scale would be so huge that falling to array storage makes sense, but it should at least be resizeable to meet the demand in worst case scenario.

src/blocks.zig
```diff
@@ -51,43 +51,190 @@ pub const Ore = struct {
...
+
+	pub var sortedAllowOres: SortedBlockProperties(maxSortedBlockProperties, bool) = .{};
```
Argmaster, line 229: The name is overqualified, it duplicates the information present in the type.

src/blocks.zig
```diff
@@ -51,43 +51,190 @@ pub const Ore = struct {
...
+	// TODO: Tick event is accessed like a milion times for no reason. FIX IT
+	pub var sortedTickEvent: SortedBlockProperties(maxSortedBlockProperties, TickEvent) = .{};
```
Argmaster, line 231: The name is overqualified, it duplicates the information present in the type.

src/blocks.zig
```diff
@@ -51,43 +51,190 @@ pub const Ore = struct {
...
+	pub var sortedTouchFunction: SortedBlockProperties(maxSortedBlockProperties, *const TouchFunction) = .{};
```
Argmaster, line 232: The name is overqualified, it duplicates the information present in the type.

src/blocks.zig
```diff
@@ -51,43 +51,190 @@ pub const Ore = struct {
...
+	pub var sortedBlockEntity: SortedBlockProperties(maxSortedBlockProperties, *BlockEntityType) = .{};
+
+	/// GUI that is opened on click.
+	pub var sortedGui: SortedBlockProperties(maxSortedBlockProperties, []u8) = .{};
```
Argmaster, line 236: The name is overqualified, it duplicates the information present in the type.

src/blocks.zig
```diff
@@ -51,43 +51,190 @@ pub const Ore = struct {
...
+		const sortedProp = &@field(BlockProps, decl.name);
+
+		if(comptime isSortedProp(@TypeOf(sortedProp.*))) {
+			std.log.info("Cleared \'{s}\' from {d} entries", .{decl.name, sortedProp.allocatedSize});
+			sortedProp.clear();
+		}
+	}
+}
+
+const BlockProps = struct {
```
Argmaster, line 190: What is the purpose of that struct?

.gitignore
psznm: @Argmaster I am sorry, what?

1. Can you link us to where that is written? I just checked CONTRIBUTING.md and I could not find anything related.
2. I hope you can see the ridiculousness of that statement. Because you are forbidden to add project configuration to repo, you are also forbidden to make git ignore it?

src/blocks.zig
sebat2: Code organization. Grouping all block properties in one struct makes the code more readable than having them floating in the global namespace. Also, I'll probably make this struct public in later changes when I'm splitting most of the blocks.zig and block_entity.zig code into different files.

.gitignore
sebat2: @psznm Exactly my thoughts on it.

IntegratedQuantum (changes requested):
This adds a lot of complexity for saving 320 kiB (compare too the many gigabytes used by the rest of the game) while also likely degrading performance.
It would have to save at least 100× more memory to be even close to being reasonable.

For #1853 why not just use a single pointer per block type (ideally put into the BlockEntity struct) that points to the block entity data?

.gitignore
codemob-dev: > Because you are forbidden to add project configuration to repo, you are also forbidden to make git ignore it?

It's pretty easy to do this yourself without adding it to gitignore, there is some more info here: https://github.com/PixelGuys/Cubyz/issues/1851

psznm: > It's pretty easy to do this yourself without adding it to gitignore, there is some more info here: [#1851](https://github.com/PixelGuys/Cubyz/issues/1851)

That is nice. I didn't know that was even possible. It would be useful to copy the comment from zig .gitignore to this .gitignore.

Still, the statement @Argmaster  made without any sort of explanation or suggestion really rubbed me the wrong way. I really think just a little bit more effort to help those who don't know goes a long way.

Argmaster: Yeah, @QuantumDeveloper we should include that in contribution guidelines. That was "knowledge of the elders" and I didn't exactly remember where to look for that discussion.

sebat2:
@IntegratedQuantum 
The performance impact from binary search is extremely low, especially with low n values. Also, by using less memory we might fit more data into L1 cache, so I wouldn't worry about performance at all. For me the only downside is implementation complexity, but it makes adding new properties much less of an issue, which I need for #1853.

For #1853, I don't want to use the monolithic 'single pointer per block type' structure like we currently have for block entities. Instead, I want a more modular approach where each block can have unique functionality composed from smaller pieces.

My idea is to define block functionality in .zon files without hardcoding the block itself:

Example:
# chest/_defaults.zig.zon
```
.{
    .tags = .{.choppable, .wood},
    .blockHealth = 10,
    .drops = .{
        .{.items = .{.auto}},
    },
    .model = "cubyz:cube",
    .rotation = "cubyz:planar",
    
    // Instead of .blockEntity = .chest, we use modular properties:
    .inventory = .{
        .size = 30,
        .type = .items  // Could be .blocks, .items, .energy, or .liquids (useful for mods)
    },
    .onInteract = .{
        .interact = .openInventory,
        .type = .items
    }
}
```
In this example, .inventory would automatically register all the functions it needs into BlockProps, but by itself it wouldn't allow opening the inventory. For that we'd use the .onInteract property. This way we can easily create new functional blocks by reusing existing code and combining different properties.

src/blocks.zig
Argmaster: This is not C and those variables are not in the global namespace, but rather in the module scope of the module dedicated for this data. Following this relocation logic you should move all the methods in this file into that struct, as they are one big family and then the code would be equivalent to a file struct in a file struct. Files are from Zig perspective equivalent to structs.
Therefore past structure of this file should be interpreted as a struct with only static fields and static methods.

Argmaster:
You cannot completely remove `.blockEntity = chest` because this part of the system must be moddable. Additionally you cannot define full list of params used by block entities, because the system has to be moddable.

There are some properties we will need in the future, eg. related to explosions but as of now there are only few we know about. Mods may require more, but your solution doesn't provide modding support for adding new props (past one didn't do that either)

Modding is different than add-ons - modding adds code, add-ons add assets.

sebat2:
I want to create a system where you can define new parameters in code and register them with the block_manager at startup, *( compile time #1528 ) so modding will be possible.

Argmaster:
A way to save a bit of memory that was briefly considered was using ListUnmanaged, but the problem is that we are using Arena allocator, so that arena will aggregate unused memory on realloc, resulting in 1.5 memory usage compared to necessary capacity.

codemob-dev:
> I want to create a system where you can define new parameters in code and register
them with the block_manager at startup.

Check out rotation modes for how to do this kind of thing at compile-time.

sebat2:
> k out rotation modes for how to do this kind of thing at compile-time.

Do we really want to compile mods with the game ?

codemob-dev:
Yes, see https://github.com/PixelGuys/Cubyz/issues/1528

sebat2:
> Yes, see #1528

Ok, works for me, but overall this isn't really the subject of this review, so maybe we can talk about this when I start working on what I've proposed ;D ? 

I know it has to support modding, so don't worry - I have that in mind.

Argmaster:
> Possibly remove block_entity entirely in favor of organizing entity blocks as separate files (one file per entity block type)

https://github.com/PixelGuys/Cubyz/pull/1637

Argmaster:
> Move most functions from block_entity.zig vtable to their own sorted BlockProps fields

This would aggregate a lot of logic in one place and solves none of the problems with block entities we have now.

PR #2068, opened 2025-10-19
Improve particle light sampling

RanPix:
checks if the particle has changed the block position and then sample

(requires #2066)


RanPix:
there is a problem where when i break the block and spawn the particles it probably samples the light before the light values were updated and some of them are black

PR #2071, opened 2025-10-19
Remove another thing that needs to be removed because it doesn't do anything (idk why it exists still, its really old or smth) the thing is a random item called unobtainable that is defined with a struct but isn't actually used anywhere so it doesn't do anything

codemob-dev:


kachek47:
Is it possible this is a kind of default failsafe object that the game can give the player during glitches/errors, just so the game can give the player SOMETHING to prevent hard crashes or unexpected game behavior? If so, it should be kept in.

Argmaster:
Codemob literally asked Quantum and he said to remove it :P

Issue #1966, opened 2025-10-12
Rope block/entity


d3viousrodent:
### Description

<img width="1600" height="1600" alt="Image" src="https://github.com/user-attachments/assets/0b4d5566-deb6-44bb-a5e9-c58403d3a4bb" />

Discord post: https://discord.com/channels/443805812390100992/702167480654757991/1426752048019542050

I noticed caves can be very steep and difficult to traverse vertically at times, especially in The Root. So I believe adding a rope would benefit exploration and lessen the tedium of descending large caves. It's functionality, appearance, limitations and recipe are not set in stone and this post is only to get an idea going — though, I have suggestions. As shown in the image above, a length limit would allow it to be helpful without being overpowered. I also think the player should be able to craft it early game. However it is implemented — I'm sure it will benefit the game.

copygirl:
This was actually my first foray into creating something for Cubyz.

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/7b28b721-4b64-418e-bcd7-0633d85a4d1c" />

(I stole the texture from the unused bellows, so it's just temporary.)

Unfortunately, Cubyz does not yet have proper climbing mechanics, which is why I opened #1937. There is a silly workaround where I just made ropes act like a liquid. Though even that is relying on a bug, because it affects the entire voxel space rather than just the collision box of the rope, but that means it can be solid and you can "swim" on the side of the rope. It's not possible to get up a block that only has a rope attached to its side this way, so that's why the screenshot has extra rope sticking out beyond the ledges. Perhaps this could be solved by hardcoding a simplified climbing property for now?

Attaching the rope to the ceiling would work, but it'd be silly to do so on the side of blocks, so I'm thinking it'd make sense to have a metal anchors of sorts to visually attach it to blocks. Though maybe only when attached to the side. Whether this should require crafting a specific item, I'm not certain.

Maybe it should also be possible to place rope horizontally, either manually (for decoration?) or automatically by placing two anchors on the same axis and then clicking with sufficient rope, if they're not too far away from each other.

In the far future it might be interesting to consider entity-based ropes that would sag when placed horizontally. It would also allow the anchors to be placed in any two spots. Rope cannons could fire anchors, allowing them to be placed at a distance.

I made my temporary crafting recipe use leaves, as they don't currently have any use, and are pretty plentiful. In the future we should consider using vines, or require processing of fibers in some way.

d3viousrodent:
[...]

PR #2077, opened 2025-10-19
Patches to allow for rope placement/breaking

copygirl:
Related issue with some discussion: #1966

Asking for feedback on how to allow for this type of functionality.

I want to implement ropes that can be placed by extending them downwards, up to a certain maximum length. And in reverse, they can be pulled up / picked up by breaking the top-most rope repeatedly. Perhaps in the near future, this could be extended to work on ladders that could do similar both upwards and downwards.

### Changes

- Introduce a new `RotationMode.onBlockInteract` function which is called when a block is right clicked, without requiring a block entity. This function returns `true` when the default interaction (and thus placement logic) should be skipped.
- This new function is called in `triggerOnInteractBlockFromRenderThread`, which is also where the block entity's `onInteract` function is called.
- Update `UpdateBlock.BlockDropLocation` to include an `offset`. This defaults to `.{0, 0, 0}` and allows a dropped block to be offset by any amount. With ropes, this allows the player to break a distant rope but drop the item on the top rope that's actually being broken.
- Make a bunch of fields and some methods in `MeshSelection` public.
    - In `triggerOnInteractBlockFromRenderThread` we need `posBeforeBlock` and `lastPos`.
    - In the rope's `onBlockBreaking`, we currently need `selectionFace/min/max`.
    - `onBlockInteract` uses `canPlaceBlock`, and could use `updateBlockAndSendUpdate`, but uses its custom version with zero offset right now.

### Questions

- Should the block entity `onInteract` be called from the default `onBlockInteract` implementation?
- The parameters passed to `onBlockInteract` are things that I personally needed. Should there be more/less?
- In the rope's `onBlockBreaking` implementation I ***unlock the inventory mutex*** to be able to break the remote block. This is because `MeshSelection.breakBlock` locks the mutex, and only after everything is done, does it unlock, and then runs the default block breaking logic. Please advice on how this should be handled.
    - By the way, the mutex should really be unlocked using `defer` in this function.

### Notes

This includes files in the `assets/ropes/` and `mods/ropes/` folders. They are meant to be included for the sake of being able to test the changes. If desired, the rope could also be included within the game. Otherwise I'll remove them before the PR is merged.
- The textures are a work-in-progress based on the bellows texture.
- I hope to include a "rope anchor" model that is placed with the first placed rope.
- Cubyz doesn't have climbing functionality, so the ropes actually act as a liquid currently.
- It was suggested to make the rope model less thick. Thoughts?

### DIFF
--- a/mods/ropes/rotation/rope.zig
+++ b/mods/ropes/rotation/rope.zig
@@ -109,24 +109,7 @@ pub fn generateData(_: *main.game.World, _: Vec3i, _: Vec3f, _: Vec3f, _: Vec3i,
 	return false;
 }
 
-pub fn onBlockInteract(
-	/// The block being interacted with.
-	block: Block,
-	/// The position of the interacted block.
-	pos: Vec3i,
-	/// The face, or normal, pointing away from the block.
-	face: Vec3i,
-	/// The relative position within the block being clicked, between {0,0,0} and {1,1,1}.
-	rel: Vec3f,
-	/// The player's inventory.
-	inv: Inventory,
-	/// The currently selected slot in the player's inventory.
-	slot: u32,
-	/// The item that's currently being held, if any.
-	item: ?Item,
-) bool {
-	_ = .{ face, rel, item }; // unused
-
+pub fn onBlockInteract(block: Block, pos: Vec3i, _: Vec3i, _: Vec3f, inv: Inventory, slot: u32, _: ?Item) bool {
 	// Check to see if the rope orientation is up/down.
 	if(block.data != RopeOrientation.z.toData()) return false;
 
--- a/src/renderer/mesh_storage.zig
+++ b/src/renderer/mesh_storage.zig
@@ -185,10 +185,10 @@ pub fn triggerOnInteractBlockFromRenderThread(x: i32, y: i32, z: i32) EventStatu
 	const mesh = node.mesh.load(.acquire) orelse return .ignored;
 	const block = mesh.chunk.getBlock(x & chunk.chunkMask, y & chunk.chunkMask, z & chunk.chunkMask);
 
-	const pos: Vec3i = .{ x, y, z };
+	const pos: Vec3i = .{x, y, z};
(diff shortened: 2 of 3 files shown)


careeoki:
[Here's some 2 pix rope](https://github.com/careeoki/Cubyz/tree/rope):
image
image
It looks more ropey to me. And the unique thickness would be fun for building.
This model does break the "climbing", though.

Some other thoughts:
- crafting rope with grass vegetation might make more sense
- rope should be non-solid so you can climb up and down 1 block spaces

ikabod-kee:
This does bring up a question. Should we have ropes be a valid way of getting to the Root?
You know, honestly, if you can manage to get down there, I guess you deserve whatever is coming. XD

IntegratedQuantum:
> Introduce a new RotationMode.onBlockInteract

This should be done in the onInteract callback introduced in #2144

> The parameters passed to onBlockInteract are things that I personally needed. Should there be more/less?

Add more only when they are needed.

> And in reverse, they can be pulled up / picked up by breaking the top-most rope repeatedly.

I know what you wanted to here, but it just feels weird to me to break something and there are no obvious visual changes to the block in question. For placement it feels right, but somehow breaking just doesn't.

I think it would be nicer if it were to break the whole rope at once, which obviously isn't possible yet. So until then I'd suggest to just break the selected block for now.
Furthermore the "unlock the inventory mutex" part seems problematic, it is going to be inviting odd race conditions. So again please wait until there is a fix for #1744

> It was suggested to make the rope model less thick. Thoughts?

I think carrie's smaller rope looks nicer.

IntegratedQuantum:
> This does bring up a question. Should we have ropes be a valid way of getting to the Root?

This is exactly what we want https://github.com/PixelGuys/Cubyz/blob/master/docs/GAME_DESIGN_PRINCIPLES.md#problems-have-multiple-solutions

copygirl:
Admittedly, my focus has gone elsewhere. If someone would like to adopt this, feel free.

PR #2078, opened 2025-10-20
Palm Trees

careeoki:
Adds palm, a new woodtype, which can be found on beaches. It could also be used for oasis biomes, tropical biomes, and i know @Argmaster wanted palm wood for his jungle. 
image
image
image
image

### DIFF
--- a/assets/cubyz/biomes/beach/warm/base.zig.zon
+++ b/assets/cubyz/biomes/beach/warm/base.zig.zon
@@ -31,7 +31,7 @@
 			.id = "cubyz:sbb",
 			.structure = "cubyz:tree/palm/coconut",
 			.placeMode = .degradable,
-			.chance = 0.003,
+			.chance = 0.0015,
 		},
 	},
 }


careeoki:
>     * add a new beach biome variant that is wider and has more common palm trees.

What's the proper way to do this? I tried to add another transitionbiome with a lower chance, but it seems to get rid of the ocean_shelf.

image
image

IntegratedQuantum:
Hmm, I think that may only be possible if you extend the shelf to 3 or 4, which may not be such a good idea.
Alternatively you could also make a new ocean biome  with wider beaches, but that seems too hacky in my opinion.

careeoki:
Yeah, that fixes it. It's a lot bigger but I don't think it looks bad, usually beaches have a large flat/shallow area before quickly dropping off into the ocean. And it can make some neat little islands sometimes
image
image

assets/cubyz/biomes/beach/warm/base.zig.zon
```diff
@@ -27,5 +27,11 @@
 			.depth = 1,
 			.smoothness = 0.1,
 		},
+		.{
+			.id = "cubyz:sbb",
+			.structure = "cubyz:tree/palm/coconut",
+			.placeMode = .degradable,
+			.chance = 0.003,
```
IntegratedQuantum: I think it should be rarer still.

assets/cubyz/blocks/branch/palm.zig.zon
```diff
@@ -0,0 +1,37 @@
+.{
```
IntegratedQuantum, line 1: Should be branch/palm

IntegratedQuantum (changes requested):
In my opinion the new beaches and shelf are a bit too wide. Maybe 2 and 3 wide would be enough?

careeoki:
> In my opinion the new beaches and shelf are a bit too wide. Maybe 2 and 3 wide would be enough?

Yeah that looks fine, done

assets/cubyz/biomes/beach/warm/base.zig.zon
careeoki: Done

assets/cubyz/blocks/branch/palm.zig.zon
careeoki: Fixed

PR #2095, opened 2025-10-21
Implement Chisel as a proper tool

### LINKED ISSUE #625: Integrate the chisel into the tool crafting system.
#594 would be great for this.

### LINKED ISSUE #2140: Trying to fully destroy a block with a chisel causes it to flicker, which can make you teleport.

Since you are put inside of the block for some time, you can make it teleport you. this only works in survival because you break it in creative, and it only works from 2 directions (I think).

https://github.com/user-attachments/assets/57d79975-0278-498f-99e5-1cf2c5650259

BoySanic:
- Replaces logic that checks for a chisel item for logic that checks for a chisel tool
- Adds chisel tool via chisel.zig.zon

TODO:

- [x] ~~Fix break animation so that it displays only on the 1/8th of the block that's being broken~~ ~~Ended up disabling the animation for the chisel to match current behavior.~~ Fixing the animation will be in a separate issue.
- [x] Fix chisel textures because they're completely broken right now
- [x] Fix chisel matrix as it likely doesn't entirely make sense
- [x] Make chisels take durability damage when breaking parts of a block

Fixes #625
fixes #2140

### DIFF
--- a/src/rotation.zig
+++ b/src/rotation.zig
@@ -11,8 +11,8 @@ const vec = main.vec;
 const Vec3i = vec.Vec3i;
 const Vec3f = vec.Vec3f;
 const Mat4f = vec.Mat4f;
-
 const ZonElement = main.ZonElement;
+
 const list = @import("rotation");
 
 pub const RayIntersectionResult = struct {


src/items.zig
```diff
@@ -822,7 +822,7 @@ pub const Tool = struct { // MARK: Tool
 		return self.texture.?;
 	}
 
-	fn id(self: *Tool) []const u8 {
+	pub fn id(self: *Tool) []const u8 {
```
IntegratedQuantum: Just use tool.type.id().
We should probably remove this function.

mods/cubyz/rotation/stairs.zig
```diff
@@ -290,10 +290,11 @@ pub fn rayIntersection(block: Block, item: ?main.items.Item, relativePlayerPos:
...
 	if(item) |_item| {
 		switch(_item) {
-			.baseItem => |baseItem| {
-				if(std.mem.eql(u8, baseItem.id(), "cubyz:chisel")) { // Break only one eigth of a block
+			.tool => |tool| {
+				if(std.mem.eql(u8, tool.id(), "cubyz:chisel")) { // Break only one eigth of a block
 					currentData.data |= closestRay(.bit, currentData.*, relativePlayerPos, playerDir);
-					if(currentData.data == 255) currentData.* = .{.typ = 0, .data = 0};
+					if(currentData.data == 255)
+						currentData.* = .{.typ = 0, .data = 0};
```
IntegratedQuantum: I don't like this. Multiline if's should have braces.

mods/cubyz/rotation/stairs.zig
```diff
@@ -306,7 +307,8 @@ pub fn onBlockBreaking(item: ?main.items.Item, relativePlayerPos: Vec3f, playerD
 pub fn canBeChangedInto(oldBlock: Block, newBlock: Block, item: main.items.ItemStack, shouldDropSourceBlockOnSuccess: *bool) RotationMode.CanBeChangedInto {
 	if(oldBlock.typ != newBlock.typ) return RotationMode.DefaultFunctions.canBeChangedInto(oldBlock, newBlock, item, shouldDropSourceBlockOnSuccess);
 	if(oldBlock.data == newBlock.data) return .no;
-	if(item.item != null and item.item.? == .baseItem and std.mem.eql(u8, item.item.?.baseItem.id(), "cubyz:chisel")) {
+	if(item.item != null and item.item.? == .tool and std.mem.eql(u8, item.item.?.tool.id(), "cubyz:chisel")) {
+		item.item.?.tool.durability -= 1;
```
IntegratedQuantum: This will desync client and server. Please use the existing mechanism to communicate durability changes.

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,65 @@
...
+	.parameters = .{
+		.{
+			.source = .massDamage,
+			.destination = .damage,
+			.matrix = .{
+				1.5, 1.0, 0x0, 0x0, 0x0,
+				1.0, 0.5, 0.5, 0x0, 0x0,
+				0x0, 0.5, 0.5, 0x0, 0x0,
+				0x0, 0x0, 0x0, 0.1, 0x0,
+				0x0, 0x0, 0x0, 0x0, 0.1,
```
IntegratedQuantum: I think mass damage should be distributed equally, due to how you'd swing a chisel

src/items.zig
BoySanic: I'll push up my changes shortly but I've refactored so we can remove it altogether.

mods/cubyz/rotation/stairs.zig
BoySanic: I think I left the line break in by mistake. I'll fix that.

assets/cubyz/tools/chisel.zig.zon
BoySanic: I set all the active slots to 0.5

assets/cubyz/items/chisel.zig.zon
```diff
@@ -1,4 +0,0 @@
-.{
```
IntegratedQuantum, line 1: If you remove items please add a migration, otherwise this will cause an error when opening an old world.

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,65 @@
+.{
+	.blockTags = .{.mineable, .choppable},
```
IntegratedQuantum: Maybe we should add a new tag for chiselable blocks?

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,65 @@
...
+	.disabled = .{
+		0, 0, 1, 1, 1,
+		0, 0, 0, 1, 1,
+		1, 0, 0, 1, 1,
+		1, 1, 1, 0, 1,
+		1, 1, 1, 1, 0,
+	},
+	.optional = .{
+		0, 0, 0, 0, 0,
```
IntegratedQuantum: You can make one or two more slots optional, either make only the tip required or make it like the shover.

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,65 @@
...
+			.factor = 1.0,
+			.method = .average,
+		},
+		.{.source = .swingSpeed, .destination = .swingSpeed, .matrix = .{
+			0.5, 0.5,  0x0, 0x0, 0x0,
+			0.5, 0.25, 0.1, 0x0, 0x0,
+			0x0, 0.1,  0.1, 0.1, 0x0,
+			0x0, 0x0,  0.1, 0.0, 0x0,
+			0x0, 0x0,  0x0, 0x0, 0.0,
+		}, .factor = 1.0, .method = .average},
```
IntegratedQuantum: Please format this like the others.

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,65 @@
...
+			},
+			.factor = 1.0,
+			.method = .average,
+		},
+		.{.source = .swingSpeed, .destination = .swingSpeed, .matrix = .{
+			0.5, 0.5,  0x0, 0x0, 0x0,
+			0.5, 0.25, 0.1, 0x0, 0x0,
+			0x0, 0.1,  0.1, 0.1, 0x0,
+			0x0, 0x0,  0.1, 0.0, 0x0,
+			0x0, 0x0,  0x0, 0x0, 0.0,
```
IntegratedQuantum: All slots should contribute to swing time.

mods/cubyz/rotation/stairs.zig
```diff
@@ -276,8 +276,8 @@ fn closestRay(comptime typ: enum {bit, intersection}, block: Block, relativePlay
 pub fn rayIntersection(block: Block, item: ?main.items.Item, relativePlayerPos: Vec3f, playerDir: Vec3f) ?RayIntersectionResult {
 	if(item) |_item| {
 		switch(_item) {
-			.baseItem => |baseItem| {
-				if(std.mem.eql(u8, baseItem.id(), "cubyz:chisel")) { // Select only one eighth of a block
+			.tool => |tool| {
+				if(std.mem.eql(u8, tool.type.id(), "cubyz:chisel")) { // Select only one eighth of a block
```
IntegratedQuantum: I still don't like that it's hardcoded, but I guess we can take one step at a time.

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,65 @@
...
+			.source = .hardnessDamage,
+			.destination = .damage,
+			.matrix = .{
+				0.5, 0.5, 0x0, 0x0, 0x0,
+				0.5, 0.1, 0.0, 0x0, 0x0,
+				0x0, 0.0, 0.0, 0x0, 0x0,
+				0x0, 0x0, 0x0, 0.0, 0x0,
+				0x0, 0x0, 0x0, 0x0, 0.0,
+			},
+			.factor = 1.0,
```
IntegratedQuantum, line 54: I'd suggest to make the damage numbers lower, to encourage using higher-end chisels.

src/renderer.zig
```diff
@@ -1089,6 +1085,10 @@ pub const MeshSelection = struct { // MARK: MeshSelection
 					if(currentBlockProgress < 1) {
 						mesh_storage.removeBreakingAnimation(lastSelectedBlockPos);
 						if(currentBlockProgress != 0) {
+							if(isTool and std.mem.eql(u8, stack.item.?.tool.type.id(), "cubyz:chisel")) {
```
IntegratedQuantum: I don't understand what this is trying to do. Could you explain it please?

BoySanic: I was trying to disable the block break animation if you're using a chisel.

If you remove the check for the chisel here you get a block breaking animation over the full block, which looks strange when you're just breaking 1/8th of the block with the chisel

BoySanic: The current chisel behavior does this possibly unintentionally by meeting the block damage needed to break it before it can display the animation. So no animation is displayed.

I just tried to match that behavior as it seemed more complex to try to figure out displaying the block break animation scaled to 1/8th of the block

IntegratedQuantum: I think it's better to leave the breaking animation broken rather than special-casing the behavior of the chisel.

mods/cubyz/rotation/stairs.zig
BoySanic: We could generalize it so that a tool can have a "affects sub-blocks" tag?

I'm not sure what other tools might do chisel-like things, but that's my suggestion to fix the hard-coded behavior is to have it depend on a tag on the tool instead of a tool id.

src/renderer.zig
BoySanic: Maybe a separate issue to handle texture scaling for the block break animation in future? I can revert this change to bring back the block breaking animation. It is a change from current chisel behavior, even if that behavior is unintentional.

mods/cubyz/rotation/stairs.zig
Argmaster: Vtables?

codemob-dev: Not item vtables again!

assets/cubyz/items/chisel.zig.zon
BoySanic: I don't think there is currently a way to migrate an item to a tool. At least, with the current migration system.

The migrations that exist right now can do item to item migrations. 

I gave it a couple tries, the first being `.old = "chisel", .new = "chisel"` which was skipped because the items were the same.

So I renamed the tool to `chisell` to try to see if that would get it to work.
But, since the `chisell` is a tool, it does not find an item.

BoySanic: Maybe this needs a temporary code change at some point during inventory load to replace instances of the chisel item with a hard-coded iron + wood chisel tool?

BoySanic: Actually, I'm probably over-complicating it.

Maybe we can just delete the item from their inventory. They probably got enough uses out of that old chisel anyways.

BoySanic: Doesn't seem like we currently support deleting items via migrations.
It would just update the palette. So there should be a replacement item for the Chisel. However, the tool shares its name, and when it doesn't share its name it doesn't seem to be able to set the item to the tool version. (I get an integer overflow when I try to load an old world with a chisel in it)


The only choices seem to be:

1. Add a item to tool migration
2. Make a code change to be removed with a new issue similar to #480. Probably during inventory load, to replace the chisel item with a hard-coded chisel tool. (Hard-coded, meaning, populate a crafting grid and create the tool then set the item in the inventory.

src/renderer.zig
IntegratedQuantum: Current chisel behavior is a crutch. There is no reason to preserve it.
And yeah please make an issue for the breaking animation to match the selection box.

assets/cubyz/items/chisel.zig.zon
BoySanic: Just to document this: I tried migrating the chisel item to the iron_ingot item. Which appeared to work initially. Except that when you attempt to stack those iron onto an existing stack of iron, that initial stack of iron is seemingly deleted.

I think it's fine if the resultant iron cannot be combined with the correct iron. However, I don't think it should be possible to delete potentially 119 items just because you added more items to the stack.

For clarification, on the client side, it displays as if you increased the count of items on the stack. Then when you close the inventory (a chest in my case) and reopen it, the large stack has been replaced with the migrated item's smaller stack instead.

assets/cubyz/tools/chisel.zig.zon
BoySanic: In my latest changes, .chiselable is now the tag chisel uses.

It's implicitly added to blocks if the rotation mode is stairs via `getBlockTags` on rotation.

Future rotations can add an arbitrary number of tags to blocks using those rotations.

assets/cubyz/tools/chisel.zig.zon
BoySanic: This should be fixed. Idk what is wrong with my text editor but it won't obey disabling the autoformatter and Cubyz' `./compiler/zig/zig fmt` formats this file like this.

codemob-dev: Do you use vscode? I know it has a setting to turn off auto formatting.

BoySanic: Neovim :sweat_smile: 

It's been great except that I can't seem to figure out how to configure zls to not auto format on save.

If it gets too cumbersome I might end up on vscode again though, we'll see.

mods/cubyz/rotation/stairs.zig
BoySanic: I've implemented it now so that it uses the `chiselable` tag to determine if the equipped tool can target sub-blocks like this.

assets/cubyz/tools/chisel.zig.zon
BoySanic: RE format:

The CI wants to format the file in the way we don't want... Which is out of my control.

Argmaster: Add a trailing comma.

BoySanic: That seemed to work. Thanks!

assets/cubyz/items/chisel.zig.zon
IntegratedQuantum: I think we should block this until #2159 is fixed. Please reinvestigate the behavior after this is fixed.

assets/cubyz/tools/chisel.zig.zon
IntegratedQuantum: still looks the same.

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,71 @@
...
+		1, 0, 0, 1, 1,
+		1, 1, 1, 0, 1,
+		1, 1, 1, 1, 0,
+	},
+	.optional = .{
+		0, 0, 0, 0, 0,
+		0, 1, 1, 0, 0,
+		0, 1, 1, 0, 0,
+		0, 0, 0, 1, 0,
+		0, 0, 0, 0, 1,
```
IntegratedQuantum: bottom one should always be mandatory

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,71 @@
+.{
+	.blockTags = .{.chiselable},
+	.disabled = .{
+		0, 0, 1, 1, 1,
+		0, 0, 0, 1, 1,
+		1, 0, 0, 1, 1,
+		1, 1, 1, 0, 1,
+		1, 1, 1, 1, 0,
```
IntegratedQuantum: Since the base is pretty thick, maybe it would make sense to allow morre slots in the base?
`1, 1, 1, 0, 0` and the same for the last column.

mods/cubyz/rotation/stairs.zig
```diff
@@ -4,19 +4,22 @@ const main = @import("main");
...
 const vec = main.vec;
 const Mat4f = vec.Mat4f;
+const Tag = main.Tag;
 const Vec2f = vec.Vec2f;
 const Vec3f = vec.Vec3f;
 const Vec3i = vec.Vec3i;
 const ZonElement = main.ZonElement;
 
 var modelIndex: ?ModelIndex = null;
+const blockTag = "chiselable";
```
IntegratedQuantum: You can add it to the Tag enum and thus make it accessible as `.chiselable`

mods/cubyz/rotation/stairs.zig
```diff
@@ -276,9 +279,12 @@ fn closestRay(comptime typ: enum {bit, intersection}, block: Block, relativePlay
 pub fn rayIntersection(block: Block, item: ?main.items.Item, relativePlayerPos: Vec3f, playerDir: Vec3f) ?RayIntersectionResult {
 	if(item) |_item| {
 		switch(_item) {
-			.baseItem => |baseItem| {
-				if(std.mem.eql(u8, baseItem.id(), "cubyz:chisel")) { // Select only one eighth of a block
-					return closestRay(.intersection, block, relativePlayerPos, playerDir);
+			.tool => |tool| {
+				const tags = tool.type.blockTags();
+				for(tags) |tag| {
+					if(std.mem.eql(u8, tag.getName(), blockTag)) {
```
IntegratedQuantum: Never compare tags by name, the whole point of using an enum is to allow comparing by enum value.

mods/cubyz/rotation/stairs.zig
```diff
@@ -306,8 +315,14 @@ pub fn onBlockBreaking(item: ?main.items.Item, relativePlayerPos: Vec3f, playerD
...
 	if(oldBlock.data == newBlock.data) return .no;
-	if(item.item != null and item.item.? == .baseItem and std.mem.eql(u8, item.item.?.baseItem.id(), "cubyz:chisel")) {
-		return .yes; // TODO: Durability change, after making the chisel a proper tool.
+	if(item.item != null and item.item.? == .tool) {
+		return .{.yes_costsDurability = 1};
 	}
 	return .no;
 }
+
+pub fn getBlockTags(allocator: NeverFailingAllocator) ?[]Tag {
```
IntegratedQuantum: Woah, why is this doing allocation, just put it in a comptime slice `&.{}`

src/blocks.zig
```diff
@@ -99,8 +99,24 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
...
+	for(block_tags) |tag| {
+		_blockTags[size][index] = tag;
+		index += 1;
+	}
+	if(rotation_tags) |tags| {
+		for(tags) |tag| {
+			_blockTags[size][index] = tag;
+			index += 1;
+		}
+	}
```
IntegratedQuantum: Use std.mem.concat

src/rotation.zig
```diff
@@ -124,6 +125,9 @@ pub const RotationMode = struct { // MARK: RotationMode
 			}
 			return .no;
 		}
+		pub fn getBlockTags(_: NeverFailingAllocator) ?[]Tag {
+			return null;
```
IntegratedQuantum: Should return an empty slice `&.{}` instead of null.

src/rotation.zig
```diff
@@ -161,6 +165,8 @@ pub const RotationMode = struct { // MARK: RotationMode
 	onBlockBreaking: *const fn(item: ?main.items.Item, relativePlayerPos: Vec3f, playerDir: Vec3f, currentData: *Block) void = &DefaultFunctions.onBlockBreaking,
 
 	canBeChangedInto: *const fn(oldBlock: Block, newBlock: Block, item: main.items.ItemStack, shouldDropSourceBlockOnSuccess: *bool) CanBeChangedInto = DefaultFunctions.canBeChangedInto,
+
+	getBlockTags: *const fn(allocator: NeverFailingAllocator) ?[]Tag = DefaultFunctions.getBlockTags,
```
IntegratedQuantum: This is the signature I would like to see
```suggestion
	getBlockTags: *const fn() []const Tag = DefaultFunctions.getBlockTags,
```

IntegratedQuantum (changes requested):

mods/cubyz/rotation/stairs.zig
BoySanic: I don't remember the exact reason I ran into problems but initially I was having problems during compile when I tried to create a slice containing the tag.

But I tried it now with `&.{}` and it did work, so that's how it is now.

assets/cubyz/tools/chisel.zig.zon
```diff
@@ -0,0 +1,71 @@
...
+			.source = .massDamage,
+			.destination = .damage,
+			.matrix = .{
+				0.5, 0.5, 0x0, 0x0, 0x0,
+				0.5, 0.5, 0.5, 0x0, 0x0,
+				0x0, 0.5, 0.5, 0x0, 0x0,
+				0x0, 0x0, 0x0, 0.5, 0.5,
+				0x0, 0x0, 0x0, 0.5, 0.5,
+			},
+			.factor = 1.0,
```
IntegratedQuantum: Also apply the stuff from #2031 please

src/rotation.zig
```diff
@@ -6,12 +6,12 @@ const chunk = @import("chunk.zig");
...
 const main = @import("main");
 const ModelIndex = main.models.ModelIndex;
+const Tag = main.Tag;
 const vec = main.vec;
 const Vec3i = vec.Vec3i;
 const Vec3f = vec.Vec3f;
 const Mat4f = vec.Mat4f;
 const ZonElement = main.ZonElement;
-
 const list = @import("rotation");
```
IntegratedQuantum, line 16: I like to keep new lines between groups of aliases.

src/blocks.zig
```diff
@@ -99,8 +99,11 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
 	_mode[size] = rotation.getByID(zon.get([]const u8, "rotation", "cubyz:no_rotation"));
 	_blockHealth[size] = zon.get(f32, "blockHealth", 1);
 	_blockResistance[size] = zon.get(f32, "blockResistance", 0);
+	const rotation_tags = _mode[size].getBlockTags();
+	const block_tags = Tag.loadTagsFromZon(main.worldArena, zon.getChild("tags"));
```
IntegratedQuantum: For temporary data please use the stackAllocator, we shouldn't needlessly leak stuff into the arena.

mods/cubyz/rotation/stairs.zig
```diff
@@ -306,8 +318,12 @@ pub fn onBlockBreaking(item: ?main.items.Item, relativePlayerPos: Vec3f, playerD
...
-	if(item.item != null and item.item.? == .baseItem and std.mem.eql(u8, item.item.?.baseItem.id(), "cubyz:chisel")) {
-		return .yes; // TODO: Durability change, after making the chisel a proper tool.
+	if(item.item != null and item.item.? == .tool) {
+		return .{.yes_costsDurability = 1};
 	}
 	return .no;
 }
+
+pub fn getBlockTags() []const Tag {
+	return &.{Tag.chiselable};
```
IntegratedQuantum:
```suggestion
	return &.{.chiselable};
```

mods/cubyz/rotation/stairs.zig
```diff
@@ -276,9 +277,15 @@ fn closestRay(comptime typ: enum {bit, intersection}, block: Block, relativePlay
...
 	if(item) |_item| {
 		switch(_item) {
-			.baseItem => |baseItem| {
-				if(std.mem.eql(u8, baseItem.id(), "cubyz:chisel")) { // Select only one eighth of a block
-					return closestRay(.intersection, block, relativePlayerPos, playerDir);
+			.tool => |tool| {
+				const tags = tool.type.blockTags();
+				for(tags) |tag| {
+					switch(tag) {
+						.chiselable => {
```
IntegratedQuantum: For one case just use `if` instead of switch, since it's less verbose

careeoki:
@BoySanic I made a PR on your fork with a new chisel texture
image

BoySanic:
> @BoySanic I made a PR on your fork with a new chisel texture image

I love it! I've merged it :)

BoySanic:
Things we can split off as separate issues:

1. Breaking the last 1/8th of the block causes that block to drop. Should that behavior remain?
2. Block breaking texture should only display on the part of the block that has been selected.

IntegratedQuantum:
> Things we can split off as separate issues:

Please make and link the issues then.

> Breaking the last 1/8th of the block causes that block to drop. Should that behavior remain?

Yes, that is reasonable now that the chisel is a proper tool. Previously I didn't allow it because it would have allowed you to break blocks without having the required damage for it.

src/rotation.zig
```diff
@@ -6,12 +6,13 @@ const chunk = @import("chunk.zig");
...
 const main = @import("main");
 const ModelIndex = main.models.ModelIndex;
+const Tag = main.Tag;
 const vec = main.vec;
 const Vec3i = vec.Vec3i;
 const Vec3f = vec.Vec3f;
 const Mat4f = vec.Mat4f;
-const ZonElement = main.ZonElement;
 
+const ZonElement = main.ZonElement;
```
IntegratedQuantum, line 14: You put the newline in the wrong place

IntegratedQuantum (changes requested):

IntegratedQuantum:
this also fixes #2140

src/rotation.zig
IntegratedQuantum: since it was only a small thing and I want this in the play session, I fixed it myself.

PR #2105, opened 2025-10-22
Add grass block recipes

thecreare:
Adds recipes for each grass block, they can be crafted with their corresponding ground and vegetation type.

- 4x `soil` + `grass_vegetation` = 4x `grass`
- 4x `mud` + `lush_grass_vegetation` = 4x `lush_grass`
- 4x `dirt` + `dry_grass_vegetation` = 4x `dry_grass`
- 4x `permafrost` + `cold_grass_vegetation` = 4x `cold_grass`

image

### DIFF
--- a/assets/cubyz/recipes/grass_recipes.zig.zon
+++ b/assets/cubyz/recipes/grass_recipes.zig.zon
@@ -1,0 +1,18 @@
+.{
+	.{
+		.inputs = .{"4 cubyz:soil", "cubyz:grass_vegetation"},
+		.output = "4 cubyz:grass",
+	},
+	.{
+		.inputs = .{"4 cubyz:mud", "cubyz:lush_grass_vegetation"},
+		.output = "4 cubyz:lush_grass",
+	},
+	.{
+		.inputs = .{"4 cubyz:dirt", "cubyz:dry_grass_vegetation"},
+		.output = "4 cubyz:dry_grass",
+	},
+	.{
+		.inputs = .{"4 cubyz:permafrost", "cubyz:cold_grass_vegetation"},
+		.output = "4 cubyz:cold_grass",
+	},
+}


ikabod-kee:
I'm all for this, but why 4?

careeoki:
I imagine you did 4 soil + 1 grass vegetation so it's cheaper. But consider the following for a 1 soil + 1 grass vegetation recipe:

Would it be a good idea for grass blocks to drop their soil + a matching grass vegetation? That way you wouldn't be losing grass vegetation if you have to break and place a grass block. My worry is that it could be annoying for it to drop two different items, but if you're digging up grass, you probably already have a slot with grass vegetation, so it might not be so bad?

thecreare:
> why 4?

Vegetation is sometimes pretty sparse and having to harvest an entire biome's worth for a handful of grass blocks didn't seem reasonable to me. The number itself is pretty arbitrary though

> Would it be a good idea for grass blocks to drop their soil + a matching grass vegetation?

I agree losing vegetation is annoying, I think this could work but I also feel like it would defeat the point of grass not dropping itself. Like it might be better to make grass blocks directly drop themselves instead? I don't know what would be best

Argmaster:
What is the logic behind grass not dropping itself beyond the fact its a m*necraft thing?
Its not like you cant collect grass irl:

image

ikabod-kee:
I think the logic is of course less inventory clutter. Maybe grass vegetation could be crafted into grass seeds?

IntegratedQuantum:
I think instead of adding recipes we should fundamentally rethink our approach to grass here.
Grass dropping itself is one option, grass spreading is another. Following @Argmaster's suggestion Grass could also work more like a carpet (requires #1821 if want flowers to still grow on top). There was also the idea for grass seeds like in terraria.

du82:
I think grass spreading is a better idea

Argmaster:
> I think grass spreading is a better idea

Could you present some arguments please?

careeoki:
The problem with grass spreading is that many many biomes mix grass blocks and their soil types for more variation. Grass spreading would eventually make all the soil disappear, leaving a flat ground of grass

Argmaster:
I, personally (as if my opinions mattered), would prefer grass as full blocks, since terraforming (especially for big builds) requires a lot of time and clicking even without needing to double click per block or wait for grass to spread.

careeoki:
Grass should be a full block. Even post-orevolution i can imagine there would still be limitations.

In terms of QOL, being able to have grass blocks in the inventory is the best, especially for building like ARG said.

I tried "grass blocks drop grass vegetation" for a few minutes and it feels reasonable. Essentially it just saves you 1 slot compared to "grass blocks drop themselves" or "grass blocks drop grass seeds." It does let you gather a lot of grass vegetation very quickly

For an alternative for grass spreading, we could have a "grass starter" item that when you right click a soil block, it creates an AoE which spreads grass in a 16 block diameter, then stops when all the grass is grown. Crafted with grass plants and sulfur.
M*necraft's botania mod has a similar item.

ikabod-kee:
I think a grass block dropping 1 grass seed and 1 base block would be best for reduced clutter.

The grass seed would be uniform. You could even equip a building accessory that allows you to automatically place seeds when placing dirt, converting it instantly to grass blocks.

Plus, this allows us to add grass variants to more blocks without technically adding new survival blocks!

ikabod-kee:
Idea for Renewability:
Grass will go up or down one stage (#1409) over time.
Vegetation --> 1 grass seed.

ikabod-kee:
>  Grass will go up or down one stage (#1409) over time.

I think for building purposes, it should only grow if placed on its corresponding grass block, or perhaps some sort of farmland.

ikabod-kee:
I do not think this should be merged. Item vtables are coming soon, which will allow for grass seeds. 👍

PR #2111, opened 2025-10-22
Cold beaches (and gravel retexture)

careeoki:
Makes cold oceans use unique gravel beaches instead of the sandy beaches. Pairs great with #2078 so you don't get palms next to a glacier ;-)
Also migrates ocean biomes to ocean folder
image
Its wide variant is a rocky beach
image

And a gravel retexture. Why I feel this is necessary:
- The current gravel is very harsh and noisy. In large amounts, _like on the cold beaches_, it can be a bit much.
- It's a year old now and I know I can do better than that texture.

gravel com


codemob-dev:
I like it, it also looks a bit more shoverable now

IntegratedQuantum (approved):

PR #2128, opened 2025-10-24
Allow placing on and breaking of fluid & gas blocks when holding their items

thecreare:
Makes building with liquids and gasses behave the same as normal blocks when their corresponding item is held. Applies to all `.fluid` and `.air` tagged blocks. I disabled breaking in survival because digging a hole in a lake with a water block seemed weird.

https://github.com/user-attachments/assets/5c06ede1-6ff5-4820-b2f0-3bb23e3b9ad3


codemob-dev:
> I disabled breaking in survival because digging a hole in a lake with a water block seemed weird.

In survival it should be impossible to obtain water blocks anyway.

thecreare:
Makes sense, should I remove that check?
My thought was some servers do allow players to get water blocks (ashframe) and it would be weird there, though I also understand the argument of not making decisions based on modified servers

codemob-dev:
Because we don't have buckets yet, being able to break water with water seems like a pretty elegant solution. (As a side note, when I was implementing water placeable blocks I accidentally made it so they could break water lol)

PR #2136, opened 2025-10-26
Make the Seed Configurable

H41ogen:
This PR adds an input field to the world creation GUI that allows the user to input a seed. If the user leaves this field blank, then the seed is randomly generated. If the user enters a string that cannot be parsed to a `u64`, it will be hashed.
- - -
Screenshot_20251025_230211

### DIFF
--- a/src/server/world.zig
+++ b/src/server/world.zig
@@ -31,14 +31,6 @@ pub const Settings = struct {
 	allowCheats: bool = false,
 	testingMode: bool = false,
 	seed: u64 = undefined,
-
-	pub fn load(self: *Settings, path: []const u8, allocator: main.heap.NeverFailingAllocator, wio: WorldIO) !void {
-		var gamerules = files.cubyzDir().readToZon(allocator, try std.fmt.allocPrint(allocator.allocator, "saves/{s}/gamerules.zig.zon", .{path})) catch ZonElement.initObject(allocator);
-		self.defaultGamemode = std.meta.stringToEnum(main.game.Gamemode, gamerules.get([]const u8, "default_gamemode", "creative")) orelse .creative;
-		self.allowCheats = gamerules.get(bool, "cheats", true);
-		self.testingMode = gamerules.get(bool, "testingMode", false);
-		self.seed = try wio.loadWorldSeed();
-	}
 };
 
 fn findValidFolderName(allocator: main.heap.NeverFailingAllocator, name: []const u8) []const u8 {
@@ -352,7 +344,7 @@ const ChunkManager = struct { // MARK: ChunkManager
 	pub fn init(world: *ServerWorld, settings: ZonElement) !ChunkManager { // MARK: init()
 		const self = ChunkManager{
 			.world = world,
-			.terrainGenerationProfile = try server.terrain.TerrainGenerationProfile.init(settings, world.settings.seed),
+			.terrainGenerationProfile = try server.terrain.TerrainGenerationProfile.init(settings, world.seed),
 		};
 		entityChunkHashMap = .init(main.globalAllocator.allocator);
 		server.terrain.init(self.terrainGenerationProfile);
@@ -428,7 +420,7 @@ const ChunkManager = struct { // MARK: ChunkManager
 		const biomeMap = terrain.CaveBiomeMap.CaveBiomeMapView.init(main.stackAllocator, ch.super.pos, ch.super.width, 32);
 		defer biomeMap.deinit();
 		for(server.world.?.chunkManager.terrainGenerationProfile.generators) |generator| {
-			generator.generate(server.world.?.settings.seed ^ generator.generatorSeed, ch, caveMap, biomeMap);
+			generator.generate(server.world.?.seed ^ generator.generatorSeed, ch, caveMap, biomeMap);
 		}
 		if(pos.voxelSize != 1) { // Generate LOD replacements
 			for(ch.super.data.palette()) |*block| {
@@ -505,7 +497,7 @@ const WorldIO = struct { // MARK: WorldIO
 		const worldData = ZonElement.initObject(main.stackAllocator);
 		defer worldData.deinit(main.stackAllocator);
 		worldData.put("version", worldDataVersion);
-		worldData.put("seed", self.world.settings.seed);
+		worldData.put("seed", self.world.seed);
 		worldData.put("doGameTimeCycle", self.world.doGameTimeCycle);
 		worldData.put("gameTime", self.world.gameTime);
 		worldData.put("spawn", self.world.spawn);
@@ -536,7 +528,10 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 
 	tickSpeed: std.atomic.Value(u32) = .init(12),
 
-	settings: Settings = undefined,
+	defaultGamemode: main.game.Gamemode = undefined,
+	allowCheats: bool = undefined,
+	testingMode: bool = undefined,
+	seed: u64 = undefined,
 
 	path: []const u8,
 	name: []const u8 = &.{},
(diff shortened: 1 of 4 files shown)


src/gui/windows/save_creation.zig
```diff
@@ -46,9 +47,25 @@ fn testingModeCallback(enabled: bool) void {
 	testingMode = enabled;
 }
 
+fn getWorldSeed(seedStr: []const u8) u64 {
+	if(seedStr.len == 0)
+	{
```
cn-cn-cn: If I'm reading cubyz' code correctly, this is preferred:

```zig
if(condition) {
    code;
}
```

over

```zig
if(condition)
{
    code;
}
```

notcancername (changes requested):

codemob-dev:
Related to: #199

src/gui/windows/save_creation.zig
```diff
@@ -46,9 +47,23 @@ fn testingModeCallback(enabled: bool) void {
 	testingMode = enabled;
 }
 
+fn getWorldSeed(seedStr: []const u8) u64 {
+	if(seedStr.len == 0) {
+		return std.crypto.random.int(u64);
```
OneAvargeCoder193: cubyz uses its own random number generator, found in main.random.
`return main.random.nextInt(u64, &main.seed);`

H41ogen: i fixed it, is that all i have to do to get re-reviewed?

BoySanic: Looks like you're still using `std.crypto.random.int`.

I would instead use the existing code that creates the seed for the world.
https://github.com/PixelGuys/Cubyz/blob/0c5f3dd4392a9c22c6c62c64f05ae034bfc947da/src/server/world.zig#L561

BoySanic: https://github.com/H41ogen/Cubyz/blob/2b218bae06cfcc53ad00cd79a01e5f7d27ce52fc/src/server/world.zig#L562

Seeing as you never changed this to begin with, I doubt your seed input actually set the seed.

H41ogen: oh, interesting. i'll fix that, sorry

H41ogen: Turns out it's a lot more complicated than i thought. should i close this PR and make another when i finish?

BoySanic: You could convert it to draft if you wanted. But there's no harm in leaving it open while you continue to work on it.

cn-cn-cn: > Seeing as you never changed this to begin with, I doubt your seed input actually set the seed.

@BoySanic, you're mistaken. It does actually set the seed in the save file, with which the value set in `init` gets overwritten:

https://github.com/pixelguys/Cubyz/blob/dda5fcbcf284ccdb13e6f6a436d1b7420a7469e4/src/server/world.zig#L597

https://github.com/pixelguys/Cubyz/blob/dda5fcbcf284ccdb13e6f6a436d1b7420a7469e4/src/server/world.zig#L476

cn-cn-cn: > I would instead use the existing code that creates the seed for the world.

The existing code is lacking because it depends directly on the time, making the search space to reconstruct a seed from a generated world much smaller.

BoySanic: > The existing code is lacking because it depends directly on the time

This is an unrelated change to making the seed configurable

cn-cn-cn: > This is an unrelated change to making the seed configurable

That's fair. It's not clear what the intended seed generation scheme is, the `init` functions uses the time, but as far as I can tell, this seed is never actually used and replaced by 48 bits from Cubyz' RNG.

H41ogen:
turns out that the only thing that needed fixing was the spawnpoint not depending on the seed.

src/server/world.zig
IntegratedQuantum, line 1: Please also remove the alternative paths used to generate or parse the world seed currently.

It might make sense to adjust ServerWorld.init to use the new `WorldSettings` struct internally and for parsing.

src/gui/windows/save_creation.zig
```diff
@@ -46,9 +47,23 @@ fn testingModeCallback(enabled: bool) void {
 	testingMode = enabled;
 }
 
+fn getWorldSeed(seedStr: []const u8) u64 {
+	if(seedStr.len == 0) {
+		return main.random.nextInt(u64, &main.seed);
+	} else {
+		return std.fmt.parseInt(u64, seedStr, 0) catch {
+			return std.hash.Wyhash.hash(0, seedStr);
```
IntegratedQuantum: Please explain why you chose this hash.

IntegratedQuantum (changes requested):

src/gui/windows/save_creation.zig
H41ogen: because it hashes a string to a u64. what do you think would work better?

src/server/world.zig
H41ogen: i can try this. my logic behind the current solution was just to make as little changes as possible.

H41ogen: > It might make sense to adjust ServerWorld.init to use the new `WorldSettings` struct internally and for parsing.

can you explain why? I'm confused why it wouldn't make sense to parse the seed in save_creation so it can be stored always as a u64 instead of sometimes a []const u8 and sometimes a u64.

src/gui/windows/save_creation.zig
H41ogen: also, from the github of wyhash:
image

src/server/world.zig
H41ogen: never mind, i see why.

src/gui/windows/save_creation.zig
IntegratedQuantum: also from the same github:
Screenshot at 2025-11-13 16-35-25

H41ogen: rapidhash isnt available in zig though (as far as i know)

BoySanic: It won't be included in zig for now, apparently:
https://github.com/ziglang/zig/commit/a0d16829215f56fe42548f1014cba0f2df5a1609

src/gui/windows/save_creation.zig
```diff
@@ -47,8 +48,10 @@ fn testingModeCallback(enabled: bool) void {
 }
 
 fn createWorld(_: usize) void {
-	const worldName = textInput.currentString.items;
-	const worldSettings: main.server.world_zig.WorldSettings = .{.gamemode = gamemode, .allowCheats = allowCheats, .testingMode = testingMode};
+	const worldName = nameInput.currentString.items;
+	const worldSeed = seedInput.currentString.items;
+
+	const worldSettings: main.server.world_zig.WorldSettings = .{.gamemode = gamemode, .allowCheats = allowCheats, .testingMode = testingMode, .seed = worldSeed};
```
IntegratedQuantum: The number of settings will likely grow further in the future. Now would be a good time to split it onto multiple lines.

src/server/world.zig
IntegratedQuantum: I think you misunderstood me. By parsing I meant the process of parsing the seed from the existing file, and not your "parseSeed" function (which should probably be named "chooseSeed", since half of it is generating a value and not just parsing it).

My idea was to replace the old `loadWorldSeed` with a call to `WorldSettings.init` that would then load the entire settings to the WorldSettings struct.

Furthermore please remove all old code that initialized the seed to a random value.

src/gui/windows/save_creation.zig
IntegratedQuantum: alright, let's keep Wyhash then.

H41ogen:
i thought i did remove all the old code that randomized the seed. where else is it being done?

IntegratedQuantum:
Oh yeah, sorry I didn't see that.

src/server/world.zig
H41ogen: > My idea was to replace the old `loadWorldSeed` with a call to `WorldSettings.init` that would then load the entire settings to the WorldSettings struct.

Could you elaborate a bit more? I'm confused. Do you want me to have a worldSettings variable to live in the ServerWorld struct and store world settings? Why is the seed a world setting? And is it possible this could be a separate PR?

H41ogen: never mind, I got bored and did all of it, i think.

src/server/world.zig
```diff
@@ -26,11 +26,21 @@ const Palette = main.assets.Palette;
...
-pub const WorldSettings = struct {
-	gamemode: Gamemode = .creative,
+pub const Settings = struct {
+	defaultGamemode: Gamemode = .creative,
 	allowCheats: bool = false,
 	testingMode: bool = false,
+	seed: u64 = undefined,
+
+	pub fn load(self: *Settings, path: []const u8, allocator: main.heap.NeverFailingAllocator, wio: WorldIO) !void {
+		var gamerules = files.cubyzDir().readToZon(allocator, try std.fmt.allocPrint(allocator.allocator, "saves/{s}/gamerules.zig.zon", .{path})) catch ZonElement.initObject(allocator);
```
IntegratedQuantum: I didn't think of this, it being in two separate files is kind of problematic. So I guess scratch the parsing for now (until we combine the files). The other changes are good though, I just don't like this parsing function because it reopens and reparses the files.

IntegratedQuantum: allocator usage is also problematic in here.

IntegratedQuantum (changes requested):

src/server/world.zig
H41ogen: done.

Issue #2132, opened 2025-10-25
World Creation rework


YhormXavs:
<img width="400" height="282" alt="Image" src="https://github.com/user-attachments/assets/ba0da234-d082-4ee8-ba76-6bde841daa31" />

0. Button to create a new world wich is now on the world list menu.
1. Buttons to switch the world creation submenu.
2. World Name(Maybe add format support idk)
3. Sub-menu scroll bar
4. Here would go all the content that the submenus cover to configure the newly created world.
5. Sub-menus. Now, here's the real stuff.

**Sub-menu 01: Generation**:
- **World Type**: Vanilla, Continents, Islands, Single Biome, Superflat, Blank, WorldTypes added by addons/mods, etc...
- **Structure generation**: V/F
- **Dungeon Generation**: V/F //This applies to all structures with the "dungeon" tag both vanilla and modded
- **Structure Distance(Config per Structure)**
- **Structure Rarity(Config per Structure)**
- **Structure Black/Whitelist**
- **Biome Black/Whitelist**

**Sub-menu 02: Difficulty**:
- **Damage to players:** A multiplier to all the damage that the player suffers
- **Damage to mobs:** A multiplier to all the damage that mobs suffers
- **Darkness kills players:** V/F _//Standing in the darkness(idk how lighting work in this game, but let suposse that this acts when light = 0). for 60s instakill the player._
- **Natural Regeneration:** V/F
- **Hunger multiplier:** (0-200%) _//How fast the hunger bar drains_
- **Healing Multiplier:** (0-200%) _// How fast the player renegerates a health point._
- **Unlimited lifes**: V/F _//If its false, the gamerule down applies_
- **Lifes**: Instead of having a game rule that determines whether a player can respawn only once per world, I think it's better to let the player decide how many times they can respawn. (0-999)
- **Death penalties**: Since this is still being debated, this should be a checklist of all the punishments that come with dying (apart from losing a life). But here are some of my favorites:
1. _Drop all items except tools/armor/weapon/trinkets_
2. _Drop the entire inventory_
3. _**% Energy Bar:** What percentage should this bar be at when respawning? (0-100%)_
4. _**% Health Bar:** What percentage should this bar be at when respawning? (0-100%)_
5. _**Effects upon respawning:** A check list of all available potion effects that will be applied when respawning_
6. _**Forgive Players:** V/F Should neutral mobs forgive dead players?_

**Sub-menu 03: Gamerules**:
[...]

PR #2145, opened 2025-10-27
(WIP) New World/Save Creation GUI

H41ogen:
I'm reworking the GUI for world creation, and this PR is just here so I can get feedback as I work on it. Inspired by #2132

https://github.com/user-attachments/assets/0929c3ee-6c4e-42ea-b0b1-e1b00911f25a


skrwo:
I think the "add-ons", "generation settings" and "game rules" should be tabs instead, so players can see what they can change without scrolling through all pages

H41ogen:
> I think the "add-ons", "generation settings" and "game rules" should be tabs instead, so players can see what they can change without scrolling through all pages

i agree, and i actually didnt think of that, however currently the only way to do something similar would be to just display a lot of round buttons at the top, which i dont personally think works well :(

H41ogen:
Progress update! Next up is the generation settings i think
- - -
image

Wunka:
> > I think the "add-ons", "generation settings" and "game rules" should be tabs instead, so players can see what they can change without scrolling through all pages
> 
> i agree, and i actually didnt think of that, however currently the only way to do something similar would be to just display a lot of round buttons at the top, which i dont personally think works well :(

Maybe if the SelectableList from PR: #2053 is merged you could use a horizontal version of it? That could look good.

Else I realy like how this looks, but for this to be maybe later an reviewable PR I would maybe recommend to just have the general structure here (maybe with the Generation tab for the seed and game rules for the other options but nothing more) and have the rest of the tabs in seperate PR's.

With the flatworld option and the test world option of #2288 maybe having the general structure be merged soon could be good. So we don't add to much to the menu before this rewrite is merged.

(The change to the Selectablelist could then be done later if needed and when the SelectableList is merged.)

Issue #2162, opened 2025-10-29
Can't drop items through crosshair


thecreare:
### Steps to Reproduce and Observed Behavior

All windows block items from being thrown when clicking anywhere within them, the crosshair should be an exception to this

PR #2164, opened 2025-10-29
Add clickThrough property to GuiWindow

thecreare:
Allows windows to not sink clicks, fixes https://github.com/PixelGuys/Cubyz/issues/2162
Should this also apply to the chat window?

### DIFF
--- a/src/gui/GuiWindow.zig
+++ b/src/gui/GuiWindow.zig
@@ -64,6 +64,7 @@ closeable: bool = true,
 isHud: bool = false,
 
 shiftClickableInventory: ?main.items.Inventory = null,
+clickThrough: bool = false,
 
 /// Called every frame.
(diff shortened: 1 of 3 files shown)


Argmaster:
So, I don't know if this was blessed by @IntegratedQuantum or not but I would like to point out that this is a workaround which special cases very specific behavior. Instead I think it should be implemented through storing windows and components in a tree hierarchy and doing an event (as in [event](https://www.w3schools.com/js/js_events.asp) propagation (as in [propagation](https://www.geeksforgeeks.org/javascript/what-is-event-propagation-capturing-bubbling/) ) through them to have a robust solution that won't require changes the second we want movement to be possible when UI is open or to stop character from shifting when text input is active.

IntegratedQuantum:
I have made a more robust implementation, based on what @Argmaster suggested, in #2511

PR #2174, opened 2025-10-31
Vertical Hotbar

H41ogen:
There's no way to enable it yet, but it's there.
- - - 
image


codemob-dev:
> There's no way to enable it yet, but it's there.

You should probably add that to the settings (or maybe in a new button on the window during the reorder hud menu), or mark this as draft until then.

H41ogen:
> > There's no way to enable it yet, but it's there.
> 
> You should probably add that to the settings (or maybe in a new button on the window during the reorder hud menu), or mark this as draft until then.

i can't find a good place for it in the settings. maybe under graphics, but that seems to be for rendering settings. i could add it there until there's a way to add more buttons when reordering the hud.

Wunka:
Why did you close this @H41ogen? With #2183 you can even add ad nice button for rotating which is only visible when you reorder HUD

H41ogen:
It still has the problem of too-skinny windows behaving weird. for example, you can't put the vertical hotbar all the way to the right.

PR #2190, opened 2025-11-01
Fill partial stacks before placing in an empty slot

### LINKED ISSUE #2120: Shift-clicking items into storage arbitrarily places items in a slot in the inventory
I think it would be better UX if the partial item stacks tried to combine with other partial item stacks within the storage before starting to occupy another slot in storage.

I wouldn't say it's a bug necessarily so I didn't file it as such.

BoySanic:
Fixes #2120

The approach is to first loop the inventory only for slots with at least some of source item to attempt filling those slots first.

~~Then, if remainingAmount > 0, do the same loop again but to find an empty slot to place the remainder into.~~
EDIT:
I found a more efficient way. During the existing loop, find the first empty slot and mark that as the `selectedEmptySlot`.

Then, distribute the blocks into all slots with some of the item in it.

Then, after the loop, place the remaining amount (if any) in the `selectedEmptySlot` from earlier.

### DIFF
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -1644,7 +1644,7 @@ pub const Command = struct { // MARK: Command
 			if(self.amount > sourceStack.amount) return;
 
 			var remainingAmount = self.amount;
-			var selectedEmptySlot: ?i32 = null;
+			var selectedEmptySlot: ?u32 = null;
 			for(self.dest._items, 0..) |*destStack, destSlot| {
 				if(destStack.item == null and selectedEmptySlot == null) {
 					selectedEmptySlot = @intCast(destSlot);
@@ -1663,7 +1663,7 @@ pub const Command = struct { // MARK: Command
 			}
 			if(remainingAmount > 0 and selectedEmptySlot != null) {
 				cmd.executeBaseOperation(allocator, .{.move = .{
-					.dest = .{.inv = self.dest, .slot = @intCast(selectedEmptySlot.?)},
+					.dest = .{.inv = self.dest, .slot = selectedEmptySlot.?},
 					.source = self.source,
 					.amount = remainingAmount,
 				}}, side);


BoySanic:
Item stacks in chest before shift clicking any:
image


Shift clicked the 59 stack:
image

Shift clicked the 60 stack:
image

Shift clicked the 120 stack:
image

BoySanic:
https://github.com/user-attachments/assets/fba96e2a-cde2-463a-b592-4beb5da5b987

src/Inventory.zig
```diff
@@ -1645,8 +1645,12 @@ pub const Command = struct { // MARK: Command
 			if(!self.dest.canHold(.{.item = sourceStack.item, .amount = self.amount})) return;
 
 			var remainingAmount = self.amount;
+			var selectedEmptySlot: i32 = -1;
 			for(self.dest._items, 0..) |*destStack, destSlot| {
-				if(std.meta.eql(destStack.item, sourceStack.item) or destStack.item == null) {
+				if(destStack.item == null and selectedEmptySlot == -1) {
```
IntegratedQuantum: Use optional, instead of random number. This also likely crashes if there is no empty slot.

IntegratedQuantum (changes requested):

src/Inventory.zig
BoySanic: Done :+1:

src/Inventory.zig
```diff
@@ -1657,6 +1661,13 @@ pub const Command = struct { // MARK: Command
 					if(remainingAmount == 0) break;
 				}
 			}
+			if(remainingAmount > 0 and selectedEmptySlot != null) {
+				cmd.executeBaseOperation(allocator, .{.move = .{
+					.dest = .{.inv = self.dest, .slot = @intCast(selectedEmptySlot.?)},
```
IntegratedQuantum: Please use a type that avoids at least one of the intCasts.

BoySanic: Why is `destSlot` a usize anyway?

BoySanic: `destSlot` is just the index provided by the for loop, I guess. Just strange to me that it's a `usize` and not an int of some kind.

PR #2199, opened 2025-11-03
fix issue with verticle list relative positions

Wunka:
While implement culling in VerticalList.render #2024 the culling was only tested on the chat from what I know.

While making a example for #2065 I noticed a bug while showing the list.
Problem was `shiftedPos` is relative to the parent component, different than all other variables used in the if.
I did not create an issue because I knew the problematic code and was able to make a quick fix. I can also make an issue if wanted.

### DIFF
--- a/src/gui/components/VerticalList.zig
+++ b/src/gui/components/VerticalList.zig
@@ -139,7 +139,7 @@ pub fn render(self: *VerticalList, mousePosition: Vec2f) void {
 		const itemYPos = child.pos()[1];
 		const adjustedYPos = itemYPos + shiftedPos[1] - self.pos[1];
 
-		if(adjustedYPos + child.size()[1] < 0 or adjustedYPos > self.maxHeight) {
+		if(adjustedYPos + 2*child.size()[1] < 0 or adjustedYPos - child.size()[1] > self.maxHeight) {
 			continue;
 		}
 		child.render(mousePosition - shiftedPos);


codemob-dev:
Have you tested this to make sure nothing breaks?

Wunka:
> Have you tested this to make sure nothing breaks?

I tested with a VerticleList with a few labels in the World creation menu and with the the chat. I just spammed a few messages and also multi line messages.

While writing the changes I had print statements which just gave which and why each element was ignored.

BoySanic:
This seems to work in my testing. Items are fully out of view before they're culled.

For what it's worth, the culling was tested on chat and on the various settings menus.
In what way did it break in your dropdown list?

Wunka:
> This seems to work in my testing. Items are fully out of view before they're culled.
> 
> For what it's worth, the culling was tested on chat and on the various settings menus. In what way did it break in your dropdown list?

image

There should be the creative option.

Here Where I moved it under the allow cheats checkbox

image
 
... another bug now to find out if its from master, the selectable list or me ... Not today I think

Wunka:
> I think it might might sense to leave some margin before cutting things off. E.g. with the state of this PR, the bottom shadow pixels of the text are culled early.

I added a margin of the childs own size. So unless something renders more than double its size it should be culled correctly

IntegratedQuantum (approved):
that works now, thank you

PR #2216, opened 2025-11-05
Generalized vtables

### LINKED ISSUE #2211: Generalize comptime vtables hashmaps
We use them so often (rotation modes, block entities, future comptime modding things) it would be nice to have a set of utility functions to generate the required structs and methods.

codemob-dev:
Fixes #2211

This is a prototype, I would like feedback on if this is the right direction to go in or not.


assets/cubyz/blocks/chest/_defaults.zig.zon
codemob-dev, line 1: Once this gets moved to mods, the capitalized struct won't be a problem.

src/block_entity.zig
```diff
@@ -528,38 +475,29 @@ pub const BlockEntityTypes = struct {
 	};
 };
 
-var blockyEntityTypes: std.StringHashMapUnmanaged(BlockEntityType) = .{};
```
codemob-dev, line 541: I just realized this was called `blockyEntityTypes` lol. How did that get through review?

BoySanic: image

/shrug

But there's also a BlockEntityTypes struct:
https://github.com/PixelGuys/Cubyz/blob/64b5813cab66d4a759a4c6c537d889edca013045/src/block_entity.zig#L186 

Which you've renamed in this branch

codemob-dev:
I made the hashmap comptime which simplifies the interface a bit.

Argmaster:
When you're on it, could you implement inheritance too?

Argmaster:
jk

codemob-dev:
> When you're on it, could you implement inheritance too?

*Technically* this is one-level inheritance

IntegratedQuantum (changes requested):
The big and important question you should ask before doing something like this is:
How bad are the error messages (e.g. on missing or incorrect functions)?

codemob-dev:
> How bad are the error messages (e.g. on missing or incorrect functions)?

When removing the `onLoadClient` function from the chest I get this error:

```
src/utils/meta.zig:28:8: error: VTable missing field 'onLoadClient'
       @compileError("VTable missing field '" ++ field.name ++ "'");
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/utils/meta.zig:8:53: note: called at comptime here
  const map: std.StaticStringMap(VTable) = createMap();
```

src/rotation.zig
```diff
@@ -139,37 +44,135 @@ pub const RotationMode = struct { // MARK: RotationMode
...
 
-	onBlockBreaking: *const fn(item: ?main.items.Item, relativePlayerPos: Vec3f, playerDir: Vec3f, currentData: *Block) void = &DefaultFunctions.onBlockBreaking,
+	canBeChangedInto: *const fn(oldBlock: Block, newBlock: Block, item: main.items.ItemStack, shouldDropSourceBlockOnSuccess: *bool) CanBeChangedInto,
 
-	canBeChangedInto: *const fn(oldBlock: Block, newBlock: Block, item: main.items.ItemStack, shouldDropSourceBlockOnSuccess: *bool) CanBeChangedInto = DefaultFunctions.canBeChangedInto,
+	getBlockTags: *const fn() []const Tag,
+};
 
-	getBlockTags: *const fn() []const Tag = DefaultFunctions.getBlockTags,
+pub const DefaultRotationMode = struct {
```
IntegratedQuantum: I personally prefer the use of default values, as it was before.
This makes it more readable to the user what the default values are, since they are in the expected location.

You can extract them using `@typeInfo()`.

src/utils.zig
```diff
@@ -8,6 +8,7 @@ const NeverFailingAllocator = main.heap.NeverFailingAllocator;
 
 pub const file_monitor = @import("utils/file_monitor.zig");
 pub const VirtualList = @import("utils/virtual_mem.zig").VirtualList;
+pub const meta = @import("utils/meta.zig");
```
IntegratedQuantum: I think this is useful enough to have it in the top-level.
I'd suggest to make a separate PR first that moves relevant stuff (e.g. function pointer casts) there.

assets/cubyz/blocks/chest/_defaults.zig.zon
```diff
@@ -6,5 +6,5 @@
 	},
 	.model = "cubyz:cube",
 	.rotation = "cubyz:planar",
-	.blockEntity = .chest,
+	.blockEntity = "cubyz:chest",
```
IntegratedQuantum: The id changes could be moved to a separate PR.

src/blocks.zig
```diff
@@ -387,7 +387,7 @@ pub const Block = packed struct { // MARK: Block
 		return _onInteract[self.typ];
 	}
 
-	pub inline fn mode(self: Block) *RotationMode {
+	pub inline fn mode(self: Block) *const RotationMode {
```
IntegratedQuantum, line 425: The const correctness changes could be moved to a separate PR.

IntegratedQuantum (changes requested):
I'd suggest to split out some things into other PRs.

src/utils.zig
codemob-dev: > I think this is useful enough to have it in the top-level.

You mean move it out of utils? I don't feel like we should sacrifice the organization here for a few less characters.

src/blocks.zig
codemob-dev: Alright, it will have to be merged before this though.

src/utils.zig
IntegratedQuantum: `utils` is literally the antithesis of organizations.
Plus std.meta is also at the top level, we should put it in the location where people would expect it to be.

src/blocks.zig
codemob-dev: #2354

src/utils.zig
codemob-dev: >  I'd suggest to make a separate PR first that moves relevant stuff (e.g. function pointer casts) there.

#2364

assets/cubyz/blocks/chest/_defaults.zig.zon
codemob-dev: I undid this and I'm renaming the structs instead, the id changes could probably be made when it is moved to mods.

codemob-dev:
Closing this as I haven't touched it in ages and it has too many conflicts. I might revisit it at some point.

PR #2232, opened 2025-11-07
Allow pressing ESC once to exit chat fully

### LINKED ISSUE #709: Allow pressing ESC once to exit chat fully

thecreare:
Removed the return that cancels GUI closing if there is a focused text box. This also applies to searching in the creative inventory and editing signs but I think its positive change in those cases too

fixes #709


Argmaster:
Please use title of the original issue as pull request title instead of its ID. 
At the same time please keep `fixes #709` in description body as it is now.

PR #2256, opened 2025-11-11
Implemented Mat3f in some scenarios where a 4D matrix is inefficient and unneeded

Auseawesome:
These changes server not only to improve performance but also readability, making it clearer what certain sections of code do by reducing the need to convert between 4d and 3d vectors in calculations.

Additionally added markers within the renderer to make it easier to find certain sections of code. Markers were already present but only on *some* of the structs.

My motivations for doing this were when I was reading through the code to try to understand how portions of it worked, the 4D matrices at times impacted readability and made it less clear what was happening, especially with an abundance of `vec.xyz()`. I feel as though this is a necessary change to work towards a more maintainable code base that is easier for other people to understand and adapt in future.

### DIFF
--- a/mods/cubyz/rotation/branch.zig
+++ b/mods/cubyz/rotation/branch.zig
@@ -9,7 +9,6 @@ const rotation = main.rotation;
 const Degrees = rotation.Degrees;
 const RotationMode = rotation.RotationMode;
 const vec = main.vec;
-const Mat4f = vec.Mat4f;
 const Vec2f = vec.Vec2f;
 const Vec3f = vec.Vec3f;
 const Vec3i = vec.Vec3i;
--- a/mods/cubyz/rotation/carpet.zig
+++ b/mods/cubyz/rotation/carpet.zig
@@ -10,7 +10,7 @@ const Degrees = rotation.Degrees;
 const RayIntersectionResult = rotation.RayIntersectionResult;
 const RotationMode = rotation.RotationMode;
 const vec = main.vec;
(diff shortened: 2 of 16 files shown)


Auseawesome:
After these most recent commits, a lot of rotations have been replaced with a Mat3f rotation. The end result is a Mat3f is used almost as much as a Mat4f, with what should become performance gains significant enough to warrant the addition of this new matrix type.

IntegratedQuantum (changes requested):
I'm not convinced that this is the best way to achieve this.
Wouldn't a `Mat4f.mulVec3` achieve similar readability results? This has the advantage of upholding the "Only one obvious way to do things" from the `zig zen`.

About the performance, don't forget about the first rule of optimization: Don't make claims about performance without measuring it. There is a good chance that any performance gains are already discovered by the compiler for 4d matrices, and even without that the code paths where you did apply them take up a small fraction of the loading time.

Auseawesome:
At least on the terrible laptop I was testing on I did notice some improvements. I will try to benchmark them at some point. My main reason for not adding a multiply vec3 method for mat4f was the additional overhead required for converting types that while minimal is ran many times when loading assets.

I'll get back to you with actual benchmarks ASAP.

Argmaster:
Hey @Auseawesome this PR seems to be inactive, do you plan to continue working on it?

Auseawesome:
I've been meaning to benchmark it but never managed to get Zig benchmarking properly working on my system.

I need to find time to try setting it up and hopefully will get results back.

Issue #1522, opened 2025-05-25
Event scheduling


Argmaster:
I am proposing same thing I described [here](https://github.com/PixelGuys/Cubyz/pull/1338#issuecomment-2891804511), basically events following Command Design Pattern: You declare a function and set of arguments you want it to be run with and you schedule it to be run after n server ticks in the future. Server should save which actions it was supposed to run when it closes if there are any not run so it can load them back next time it starts. Actions may require information about their position, so they are bound to a chunk and loaded with that chunk. 

This would allow us to implement fast leaf decay as an event that would be scheduled by breaking log block. That event would check direct neighbors of that log to check if they are leaves or logs and either destroy or keep them accordingly, possibly scheduling next event with higher radius. This would distribute the load of destroying the leaves for multiple ticks and would allow for batching light updates.

This could also be used to spawn torch particles every X ticks. This system is effectively a cousin of random ticks but more reliable when you want to enforce certain speed of actions, instead of relying on randomness to make it right.

PR #2265, opened 2025-11-12
Leaf decay

tillpp:
When breaking a block, leaves ask themself: should i vanish? (leaves look for a path to the log)
And if they do vanish, their direct neighbours ask themself the same question. (they get updated, too)

- also renamed ..._leaves to leaves/...
- added an .onUpdate,.onBreak Event/tag.
- a delayed Update queue, to spread the load onto multiple ticks
- added undecayable leaves, that drop when breaking the normal leaves

[video of leaf breaking](https://github.com/user-attachments/assets/531b4c1c-f8c0-4aad-a422-eeeb05cb1c64)

### DIFF
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -69,7 +69,6 @@ var _alwaysViewThrough: [maxBlockCount]bool = undefined;
 var _hasBackFace: [maxBlockCount]bool = undefined;
 var _blockTags: [maxBlockCount][]Tag = undefined;
 var _light: [maxBlockCount]u32 = undefined;
-
 /// How much light this block absorbs if it is transparent
 var _absorption: [maxBlockCount]u32 = undefined;
 


tillpp:
the formatter wants me to change that, but it looks less readable.
grafik
grafik

Auseawesome:
> the formatter wants me to change that, but it looks less readable. grafik grafik

You can always use comments to disable the formatter which should allow checks to pass without formatting being applied. This shouldn't be overused but is likely applicable in this situation to improve readability.

codemob-dev:
> the formatter wants me to change that, but it looks less readable.

For a lot of the examples here it's just because you forgot a trailing comma.

> You can always use comments to disable the formatter which should allow checks to pass without formatting being applied. This shouldn't be overused but is likely applicable in this situation to improve readability.

I don't think you can. Perhaps in this case we need the recipe pattern matching syntax for migrations?

tillpp:
> For a lot of the examples here

"a lot"???
there are only two examples

> For a lot of the examples here it's just because you forgot a trailing comma.

no, only in the lower given example. 
And i already had added the trailing comma:  https://github.com/PixelGuys/Cubyz/pull/2265/commits/c62e400a62b52cb6a84ef2dd8833c2d1fae0c174 

but this didnt help for the migration file or any .zig file, but I talked about this with  QD and he basicly said "nobody is going to read the full migration file anyway" , so 🤷

Argmaster:
> the formatter wants me to change that, but it looks less readable. grafik grafik

For future reference use `// zig-fmt: off` https://github.com/ziglang/zig/commit/d21a1922eb5d76b9b0d0611eaeb42c91f83234ab

BoySanic:
The demonstration video shows the leaves decaying very fast.
I think we'd want them to decay slower so that it isn't a rapid change from the player's perspective.

tillpp:
@BoySanic Yeah i personally also like it slower, but QD did say he wanted it to be nearly instantly

grafik

https://discord.com/channels/443805812390100992/895731509972058194/1436705659675611166

IntegratedQuantum:
I'd suggest to move the migrations to a separate PR, that's something I can review quickly, and it would reduce the risk for further conflicts.

tillpp:
> I'd suggest to move the migrations to a separate PR, that's something I can review quickly, and it would reduce the risk for further conflicts.

done: https://github.com/PixelGuys/Cubyz/pull/2278

tillpp:
since this PR also introduces BlockUpdates and **new events** (like .onBreak and .onUpdate), i think this PR might be a candidate for High Priority.

tillpp:
how do i rerun the tests? 
grafik

codemob-dev:
> how do i rerun the tests? grafik

It runs on push, so just push an empty commit

tillpp:
This PR has been open for a while now, and anticipating a death sentence for it everyday here is kind of pulling too much Focus and Energy of my Battery.
I would like to close this PR merged or not, or give it to someone else.

Argmaster:
Well, that's how things roll here - either you have time to wait for your PRs or they don't get merged. Nothing will change as long as Quantum wants to review every change in detail.

tillpp:
> Well, that's how things roll here - either you have time to wait for your PRs or they don't get merged. Nothing will change as long as Quantum wants to review every change in detail.

yeah i know that ..., and i unfortunatly i don't got the time or social battery for that, if you had read my last comment closely you would have noticed, that i never said that this PR has to be merged. it might be closed or someone else ,if they want to, can continue this or do whatever else with it.

Edit: TL;Dr: This PR is distracting me too much from other stuff. Like: if i want to continue on other projects, a part of my brain is constently on this PR

Edit 2: after a discussion on the discord, i have decided to continue this, but i might not fix issues immediatly

src/callbacks/block/server/_list.zig
```diff
@@ -1 +1,2 @@
 pub const replaceBlock = @import("replaceBlock.zig");
+pub const leafDecay = @import("leafDecay.zig");
```
IntegratedQuantum, line 1: Please keep them sorted alphabetically

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,113 @@
...
+		if(log.decayProhibitor()) {
+			return true;
+		}
+		// it is the same type of leave
+		// continue search!
+		else if(log.typ == leave.typ) {
+			const neighbourRange = 1; // 1 = leaves need path to log without air gab
+			for(0..neighbourRange*2 + 1) |offsetX| {
+				for(0..neighbourRange*2 + 1) |offsetY| {
+					for(0..neighbourRange*2 + 1) |offsetZ| {
```
IntegratedQuantum: Adding diagonal neighbors will result in a lot of pointless events. I'd suggest to only go through direct neighbors `for(chunk.Neighbors.iterable)`

IntegratedQuantum: @careeoki do you think you could make the palm trees not have purely diagonally connected leaf blocks? Or do you think we should support diagonal leaf decay, even if it's more expensive?

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,113 @@
...
+			continue;
+		checked[@as(usize, @intCast(index))] = true;
+
+		// get the (potential) log
+		const chunkPosition = main.chunk.ChunkPosition.initFromWorldPos(value, 1);
+		var chunk = world.getOrGenerateChunkAndIncreaseRefCount(chunkPosition);
+		chunk.mutex.lock();
+		const log = chunk.getBlock(value[0] & main.chunk.chunkMask, value[1] & main.chunk.chunkMask, value[2] & main.chunk.chunkMask);
+		chunk.mutex.unlock();
+		chunk.decreaseRefCount();
```
IntegratedQuantum: This whole block getting is quite expensive and verbose. Please invest some more time into figuring out why world.getBlock doesn't work.

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,113 @@
...
+		if(@max(@abs(x), @abs(y), @abs(z)) > checkRange)
+			continue;
+
+		// mark as checked
+		const arrayIndexX = x + checkRange;
+		const arrayIndexY = y + checkRange;
+		const arrayIndexZ = z + checkRange;
+		const index = (arrayIndexX*checkLength + arrayIndexY)*checkLength + arrayIndexZ;
+		if(checked[@as(usize, @intCast(index))])
+			continue;
```
IntegratedQuantum: Ideally this flag should be set before you even add the entry to the queue.

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,113 @@
...
+	queue.pushBack(Vec3i{wx, wy, wz + 1});
+
+	while(queue.popFront()) |value| {
+		// calc relative position
+		const x = @as(i32, @intCast(value[0])) - wx;
+		const y = @as(i32, @intCast(value[1])) - wy;
+		const z = @as(i32, @intCast(value[2])) - wz;
+
+		// out of range
+		if(@max(@abs(x), @abs(y), @abs(z)) > checkRange)
```
IntegratedQuantum: Why not use
```suggestion
		if(x*x + y*y + z*z > checkRange*checkRange)
```

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,113 @@
...
+	queue.pushBack(Vec3i{wx, wy - 1, wz});
+	queue.pushBack(Vec3i{wx, wy + 1, wz});
+	queue.pushBack(Vec3i{wx, wy, wz - 1});
+	queue.pushBack(Vec3i{wx, wy, wz + 1});
+
+	while(queue.popFront()) |value| {
+		// calc relative position
+		const x = @as(i32, @intCast(value[0])) - wx;
+		const y = @as(i32, @intCast(value[1])) - wy;
+		const z = @as(i32, @intCast(value[2])) - wz;
```
IntegratedQuantum: pointless casts
```suggestion
		const x = value[0] - wx;
		const y = value[1] - wy;
		const z = value[2] - wz;
```

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,113 @@
...
+	const wx = params.chunk.super.pos.wx + params.x;
+	const wy = params.chunk.super.pos.wy + params.y;
+	const wz = params.chunk.super.pos.wz + params.z;
+
+	main.items.Inventory.Sync.ServerSide.mutex.lock();
+	defer main.items.Inventory.Sync.ServerSide.mutex.unlock();
+
+	if(Server.world) |world| {
+		params.chunk.mutex.lock();
+		const leave = params.chunk.getBlock(wx & main.chunk.chunkMask, wy & main.chunk.chunkMask, wz & main.chunk.chunkMask);
```
IntegratedQuantum:
```suggestion
		const leaf = params.chunk.getBlock(wx & main.chunk.chunkMask, wy & main.chunk.chunkMask, wz & main.chunk.chunkMask);
```

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,113 @@
...
+		params.chunk.mutex.unlock();
+
+		// check if there is any log in the proximity?^
+		if(foundWayToLog(world, leave, wx, wy, wz))
+			return .ignored;
+		// no, there is no log in proximity
+		world.updateBlock(wx, wy, wz, main.blocks.Block.air);
```
IntegratedQuantum: Please use cmpXchgBlock, otherwise another thread may hypothetically change the block in the meantime.

Also I think the replaced block should be a parameter instead of being hardcoded to air.

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,113 @@
...
+
+		// trigger others leaves:
+		world.updateSurrounding(wx, wy, wz);
```
IntegratedQuantum: should be done automatically by cmpXchgBlock.

src/server/world.zig
```diff
@@ -1057,6 +1062,21 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		}
 		ChunkManager.mutex.unlock();
 
+		// event queue
+		for(0..10) |_| {
+			while(true) {
+				if(self.delayedUpdateQueue.popFront()) |event| {
```
IntegratedQuantum: Ideally it should always execute all events from the last frame, so instead of using a queue, use a list that gets exchanged with an empty list every update.
The current solution is going to result in very inconsistent update speeds depending on the situation.

src/server/world.zig
```diff
@@ -1057,6 +1062,21 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+					var ch = self.getOrGenerateChunkAndIncreaseRefCount(chunk.ChunkPosition.initFromWorldPos(event, 1));
+					defer ch.decreaseRefCount();
+					ch.mutex.lock();
+					const block = ch.getBlock(event[0] & chunk.chunkMask, event[1] & chunk.chunkMask, event[2] & chunk.chunkMask);
+					ch.mutex.unlock();
```
IntegratedQuantum: again, world.getBlock would be better.

src/server/world.zig
```diff
@@ -1237,8 +1257,34 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		for(userList) |user| {
 			main.network.Protocols.blockUpdate.send(user.conn, &.{.{.x = wx, .y = wy, .z = wz, .newBlock = newBlock, .blockEntityData = &.{}}});
 		}
+		if(oldBlock) |old| {
+			self.updateSurrounding(wx, wy, wz);
```
IntegratedQuantum: Why only do this when an oldBlock was supplied?
Behavior shouldn't magically change based on an unrelated parameter.

src/server/world.zig
```diff
@@ -1237,8 +1257,34 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+
+			// onBreak event
+			_ = old.onBreak().run(.{.block = old, .chunk = baseChunk, .x = wx & chunk.chunkMask, .y = wy & chunk.chunkMask, .z = wz & chunk.chunkMask});
```
IntegratedQuantum: I think this should only be called if the block was actually broken (and not just a state change, like e.g. connecting branches to their neighbors).

src/server/world.zig
```diff
@@ -1237,8 +1257,34 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		}
 		return null;
 	}
+	pub fn updateSurrounding(self: *ServerWorld, wx: i32, wy: i32, wz: i32) void {
+		// trigger updates:
+		const updateRange = 1;
+		for(0..updateRange*2 + 1) |offsetX| {
+			for(0..updateRange*2 + 1) |offsetY| {
+				for(0..updateRange*2 + 1) |offsetZ| {
```
IntegratedQuantum: Again, are diagonals really necessary?

assets/cubyz/blocks/leaves/undecayable/yellow.zig.zon
```diff
@@ -0,0 +1,9 @@
+.{
```
IntegratedQuantum: @careeoki @ikabod-kee I would like to hear your opinion on these.
I personally am quite unhappy with having yet another leaf variant.

Also related question: Should leaves maybe just always decay even when placed?

src/callbacks/block/server/leafDecay.zig
tillpp: should i wait for carees answer before fixing this?

assets/cubyz/blocks/leaves/undecayable/yellow.zig.zon
ikabod-kee: Wait, are you saying you don't like yellow leaves?

src/callbacks/block/server/_list.zig
tillpp: didnt know it should be alphabetically. fixed

src/server/world.zig
tillpp: >  The current solution is going to result in very inconsistent update speeds depending on the situation.

the inconsistent update speeds were kind of intentional, i was thinking: how about we have multiple updateQueues, one for high Priority, executing immediatly, and one for low Priority, updating inconsistently for stuff that isnt that important.

(Also: spreading the updates to multiple ticks, while being nearly instant, was btw your idea?)

src/server/world.zig
tillpp: i will again wait for @careeoki opinion on diagonal leaves

src/callbacks/block/server/leafDecay.zig
careeoki: I might be able to make it work for the axis aligned fronds, but I don't think it would look very good for the diagonal fronds. They would have to be very thick without diagonals.

If it's significantly slower then I would be OK with a compromise then.

assets/cubyz/blocks/leaves/undecayable/yellow.zig.zon
Crepestrom: placed leaves by the player should not decay,
but nateral ones should.

careeoki: Yeah, i would much prefer this as some kind of a state within the normal leaf block. It also raises the question if we eventually need undecaying leafy branches or something.

Leaves always decaying would be very annoying for building, like if you wanted to make a hedge or custom tree that doesn't have log blocks.

tillpp: but i imagine they would be useful for blueprinting.
you turn of the update gamerule (if we add that in the future) and place the decayable blocks
This way the blueprint knows which to decay and which not.

tillpp: (without having to use a clumsy /replace cmd)

ikabod-kee: On one hand, making placed leaves decay would be annoying.
On the other hand, you could use a branch to prevent them from decaying.

I'm a bit mixed.

ikabod-kee: Yeah, I think placed leaf blocks should not decay.

ikabod-kee: Actually, there is some gameplay application for them decaying. Leaves in Minecraft are often used as the perfect nerd-pole block.

So, why not just have the leaf decay apply to all blocks instead of just logs, but not having a log will make them decay into brown leaves. Experimental idea, yes.

ikabod-kee: You could have an item called "mulch" that can be applied to leaves to keep them from turning brown, and they themselves also act as logs.

ikabod-kee: Just thinking out loud. Feel free to chime in with your ideas.

ikabod-kee: Also, maybe instead of having a leaf decay, you could have a leaf-killing spray?

careeoki: I don't think we need to overcomplicate this. The player should always have more control over what they're building, without any arbitrary restrictions or caveats.

And for nerd poling, I don't think it's an issue, same reasons as why we closed the nerd poling issue. We are making a game where you can build wherever you want, so we have to design around that.

ikabod-kee: > I don't think we need to overcomplicate this. The player should always have more control over what they're building, without any arbitrary restrictions or caveats.

> And for nerd poling, I don't think it's an issue, same reasons as why we closed the nerd poling issue. We are making a game where you can build wherever you want, so we have to design around that.

You're right then, but I still don't like that there's so many leaf variants. We could possibly give them gravity when ticked instead of completely erasing them.

BoySanic: A hacky way to do this might be to make the drop for leaf blocks a variant of the leaf block that doesn't decay ;)

careeoki:
So after trying this ingame for a bit this is what I noticed:
- Breaking 1 leaf block can just destroy an entire tree sometimes
tree die
- Breaking 1 leaf block can cause a chain reaction, decaying all the leaves in an area. Oak SBBs seem most susceptible to this
- It would be nice if when leafy branches decay, they left behind regular branches. I don't know if that's in-scope of this PR though.

assets/cubyz/blocks/leaves/undecayable/yellow.zig.zon
careeoki: > You're right then, but I still don't like that there's so many leaf variants

Me neither, but I'm sure there's another way to do it other than them being seperate blocks.

tillpp: @BoySanic 
> A hacky way to do this might be to make the drop for leaf blocks a variant of the leaf block that doesn't decay ;)

this is how it is implemented right now
you break leaves/oak you get leaves/undecayable/oak

tillpp: but then 1/3 of all leave files would be an undecayable variant.
I mean we already use _opaque for every leave block, so i dont really see an issue with adding undecayables

ikabod-kee: The issue that I have personally is that they look the same in the creative inventory. There will definitely be people who mix the two up, whether that be when building normally, or creating trees out of SBBs.

ikabod-kee: Another issue is that the naturally generating dead leaf piles will completely die when you break one of their blocks.

tillpp:
> Breaking 1 leaf block can just destroy an entire tree sometimes

1.in the code, branches don't count as logs, 
it is this way, because i thought having to manually search for all the branches as the player would be to tedious.
this is something we should discuss, if we should count branches as logs

> Breaking 1 leaf block can cause a chain reaction, decaying all the leaves in an area. Oak SBBs seem most susceptible to this

2. Yeah the chain reaction was intentional (QD idea), such that old world with flying leave forest would just vanish after a click. but they should only decay if they can't path way in a 5 block area to a log block. can you send a recording of it?
EDIT: i think the cause is the same as in (1.), so making branches count as leaves would fix this

> It would be nice if when leafy branches decay, they left behind regular branches. I don't know if that's in-scope of this PR though.

EDIT:Nvm
~~3. this can be changed, if other want it this way too, but i personally kind of liked it. ( it was also made by accident )~~

assets/cubyz/blocks/leaves/undecayable/yellow.zig.zon
tillpp: one solution would be to change the dead leaf piles to be of the undecayable variant

Argmaster:
Ad. 1 yes, definitely. Otherwise jungle trees will be broken.
Ad. 2 I think she refers to situation where the trunk is still there, but the leaves are just further than 5 blocks.

tillpp:
with this commit, all of @careeoki  problems should be fixed

tillpp:
but it does make breaking the those sbbs trees really annoying imo

src/callbacks/block/server/leafDecay.zig
tillpp: i tried figuring it out, the the lower i got the more confused was i getting what meant what, i got to a place where no sunlight has touched the code in millenia, deep down

tillpp: works now, i dont know why

src/callbacks/block/server/leafDecay.zig
tillpp: yeah, but would make code less readable, but i can do it.

src/callbacks/block/server/leafDecay.zig
tillpp: oh that would be cool, 
addon devs can go wild on this one

src/server/world.zig
tillpp: when the block breaks the oldBlock is always supplied (inventory.zig line 1824), that's why i thought i didnt matter. 
Changed it.

src/server/world.zig
tillpp: I am not sure if `Command.UpdateBlock.run` at `if(side==.server)` would be the correct position for this, but i also dont see any other fitting place for it.
I added it there

tillpp:
This is btw how it would look like, if you set 	`.decayReplacement = "cubyz:dirt"` 
AddOn creator, go crazy!

[video](https://github.com/user-attachments/assets/4d0fc6ac-0961-4c2a-aea7-65cf767705cc)

src/server/world.zig
IntegratedQuantum: > Also: spreading the updates to multiple ticks, while being nearly instant, was btw your idea?

Yes, and it is still happening with my proposed solution, except it has more consistent update speed of one "layer" per tick.

src/callbacks/block/server/leafDecay.zig
IntegratedQuantum: @tillpp could you please measure the average performance of both variants?

tillpp: Breaking the same tree ( a yellow blob birk tree ) standing alone gives me the following results for "foundWayToLog" function that gets run for each leave:

old way: 85.37±39.36 μs
new way: 30.37±14.83 μs

EDIT: i meant μs not ms

src/server/world.zig
tillpp: fixed.

tillpp:
i just realized that removing the logBreak callback, was a mistake,
since the logs need a wider area to update when they break to also trigger leaves, that are not adjacent to them.

image you have Log: W, Leave: E
```
EEEEEEEEE
WEEWEEWEE
```
breaking the right wood would not destroy any leaves, since they are still in range of the wood in the middle and connected.
```
EEEEEEEEE
WEEWEE EE
```
but breaking the the middle wood gives you dangling leaves: (markes by lowecase e)
```
EEEEEeeee
WEE Ee ee
```
those dangling leaves should break, but they don't because when the middle wood block broke it just updated its neighbours, which didnt decay and update their neighbours because they were still in range of the Log on the left.


My solution: logs have a .onBreak=.{,type = .widerUpdate} property, giving them the same range of update as the foundWayToLog checkRange

src/callbacks/block/server/leafDecay.zig
IntegratedQuantum: 85 µs is quite a lot.

@careeoki how about a compromise: Decay only checks the nearest neighbor, but diagonal blocks are updated.
So it would still remove the whole palm tree when broken normally, but when you break one of the leaf blocks it may also decay additional parts of that branch even if the whole tree still stands.

assets/cubyz/blocks/leaves/undecayable/yellow.zig.zon
tillpp: this is the last last Unresolved issue right now.
looks like we haven't decided on which way we want to got.

But to be honest, i still don't really understand how i would make it with rotations, such that placing a Block by a human, would give it a different blockData, maybe with overriding updateData or modifyBlock, but i am not sure if they are not also called by the world for a different thing that is unrelated to the player.

This would limit in the future, what could be decayable, cause their blockData would already be occupied for the "placed by humans" boolean

src/callbacks/block/server/leafDecay.zig
careeoki: > but when you break one of the leaf blocks it may also decay additional parts of that branch even if the whole tree still stands.

I think i would prefer no diagonals over this. I could make it work.

IntegratedQuantum:
That is annoying, it means we would be returning to instantly breaking all of them in the first tick, since they are (usually) all in range.
Also in terms of gameplay I'd prefer if it only updates things that are directly related to the block you broke.

So how about a slightly different approach: Instead of using only the distance to the nearest log, how about also taking into account the distance to the original event source?
In that case the result would look like this:
```
EEEEEEEEE
WEEWEE EE
↓
EEEEE
WEEWE
```
This would generally also solve the problem of breaking a big tree in the middle of a dense forest, with current approach no leaves disappear, with my proposal you'd see a hole in the leave roughly matching the shape of the original tree. Additionally, if you manually punch a hole in the leaves, then you'd also get the satisfying effect of seeing some surrounding leaves decay.

```
EEEEEEEEE
WEEWEEWEE
↓ break
EEEEEEEEE
WEE EEWEE
↓ decay
EE   EEEE
WE   EWEE
```

```
WEEEE
WWEEE
WEEEE
↓ break
WEEEE
WWEEE
WEEE 
↓ decay
WEEEE
WWEE 
WEE  
```

MiniBaker007:
> Instead of using only the distance to the nearest log, how about also taking into account the distance to the original event source?

I don't really understand your proposed solution? What is the rule for whether or not a leaf should decay?

IntegratedQuantum:
The rule would basically be
```zig
if(distance(leaf, log) > distance(leaf, eventSource) or distance(leaf, log) > 5) decay();
```

tillpp:
> So how about a slightly different approach: Instead of using only the distance to the nearest log, how about also taking into account the distance to the original event source?

but wouldnt that mean, that 
`WEEEE` would become `WEE` if you break the leave on the right?.

tillpp:
i guess one solution would be to store the eventSource block type too.

To differentiate if a log or a leave was broken by the player

IntegratedQuantum:
> WEEEE would become WEE if you break the leave on the right?.

Yes. Is that bad?

tillpp:
> Yes. Is that bad?

i imagine it to be annoying for people building tree houses for example. you just placed a block next to a leave and it breaks the leaves instead.

MiniBaker007:
Ah I see, I kind of like that.

So it would be (?):
```
E E E E E E
W E E E E E
↓ break
E E E E E o
W E E E E E
↓ decay
E E E E . o
W E E E E .
↓ decay
E E E . . o
W E E E . .
↓ decay
E E E . . o
W E E . . .
[Done]

```
Except, how exactly would you pass around the event origin?   I think you would need to add an origin parameter to `cmpxchgBlock`. Is that acceptable? If not, I suppose you could also add another function in `ServerWorld ` that handles specifically updates with  an event origin.

But to me, this begs the question: Is this supposed to be essentially the start of a general tick update system? It certainly looks like that to me. But if it's the case, I think a lot more thought should be going into the implementation details. I'm not saying a lot of thought hasn't already gone into this, just that it's an important , and potentialy complicated,aspect of the engine, and I already see several existing/potential issues with the current implementation.

Maybe the tick update system could be moved into a separate PR? I do see the utility of implementing it together with something that makes use of it for testing, but is there something else that would propagate tick updates like leaf decay that's maybe simpler, so we don't mix questions of how leaves should decay and of how tick updates should work?

tillpp:
1. if you add code marks they will be displayed monospaced, and it is easier to read 

> Except, how exactly would you pass around the event origin? I think you would need to add an origin parameter to `cmpxchgBlock`. Is that acceptable? If not, I suppose you could also add another function in `ServerWorld ` that handles specifically updates with an event origin.

1. no need, i just make the delayedUpdateQueue  use a struct that stores:
- the position to be updated
- the source position
- the source block
(instead of just the position to be updated)
EDIT: i think i misunderstood your question. yeah that is an issue.


> But to me, this begs the question: Is this supposed to be essentially the start of a general tick update system? [..]. But if it's the case, I think a lot more thought should be going into the implementation details. I'm not saying a lot of thought hasn't already gone into this, just that it's an important , and potentialy complicated,aspect of the engine, 

2. The leaf decay is implemented by using a general tick update system.

How it works: there is a queue that stores all blocks that should be updated in the next tick.
You break a block, block adds all their neighbours to the soon to be updated queue.
When they get updated, they might update their own neighbours too. 


> I already see several existing/potential issues with the current implementation.

3. I would like to hear them.

> Maybe the tick update system could be moved into a separate PR? I do see the utility of implementing it together with something that makes use of it for testing, but is there something else that would propagate tick updates like leaf decay that's maybe simpler, so we don't mix questions of how leaves should decay and of how tick updates should work?

We can create an Issue for how this general tick update system should work. (But i don't think that there are that many different approaches to implementing a tick system that fit QD requirements)

IntegratedQuantum:
> i imagine it to be annoying for people building tree houses for example. you just placed a block next to a leave and it breaks the leaves instead.

Hmm, that is a good point. So maybe we should only update it when it actually comes from breaking a log or leaves?
Though I am starting to wonder if this isn't getting too complicated. Maybe we should return to your original proposal to spawn all the events in a radius with a distance-based timestamp.

tillpp:
Maybe making a seperate PR isnt a bad idea, for the tick updates

MiniBaker007:
> 1. no need, i just make the delayedUpdateQueue  use a struct that stores:
> 
> * the position to be updated
> * the source position
> * the source block

Yes, with the method I was describing, that would also be necessary anyway. I guess the question was not so much about `cmpxchgBlock` specifically, but whether or not we want the generic update queue to have an origin/source parameter. Maybe with some `anyopaque` black magic we could have a generic `params `field, though.

> 2. The leaf decay is implemented by using a general tick update system.

Yes, I understand what the current implementation is and how it works, I was just asking if we want the tick update system this implements to be more broadly used by everything else in the game that deals with tick updates.

> > I already see several existing/potential issues with the current implementation.
> 
> 3. I would like to hear them.

1. When destroying a block, we want to update the neighbors, but if we use the same function for decaying leaves, that will send a generic update to the neighbors, without the origin/source block context, also potentially duplicating the update for each block if we're adding updates to the queue somewhere else. That's why I suggested a separate function for leaf decay earlier.
2. I think the `delayedUpdateQueue `should probably be a set or something, instead of a circular buffer, because right now, if I'm not mistaken, each tick, a blocck could have multiple updates, which I don't think is desirable. And if you have update events with different params (e.g. with or without a source/origin, then you run into problems with race conditions. (Maybe these aren't problems that necessarily need to be solved now, but at least they're worth keeping in mind.)
3. Maybe this is more of an issue with the leafDecay callback specifically, but I suspect there's probably some issues with accessing blocks outside of chunk boundaries, like not calling `getOrGenerateChunkAndIncreaseRefCount` when looking for paths to a log. I'm not that familiar with that part of the code, but it seems to be a necessary step when  accessing neighbors. I also wonder if increasing and decreasing chunk ref counts every block update might be inefficient, but I don't really know too much about that.

> We can create an Issue for how this general tick update system should work. (But i don't think that there are that many different approaches to implementing a tick system)

Looks like maybe you changed your mind on that, but my point was that if it's something general that will be used by other aspects of the game, then it might make more sense to address it somewhere not specific to leaf decay. And if we're happy wiht the tick update system, then that can be merged sooner so that other PRs can use it as well.

tillpp:
>3. Maybe this is more of an issue with the leafDecay callback specifically, but I suspect there's probably some issues with accessing blocks outside of chunk boundaries, like not calling getOrGenerateChunkAndIncreaseRefCount when looking for paths to a log. I'm not that familiar with that part of the code, but it seems to be a necessary step when accessing neighbors. I also wonder if increasing and decreasing chunk ref counts every block update might be inefficient, but I don't really know too much about that.

funnily enough, this is how it used to be, until  it was said to be verbose and expensive, and that i should use world.getBlock instead.  (so you can't blame that on me)
grafik

MiniBaker007:
> funnily enough, this is how it used to be, until it was said to be verbose and expensive, and that i should use world.getBlock instead. (so you can't blame that on me) 

Well, I do kind of agree, I think this logic should be performed at a higher level, though I'm not sure exactly how. I guess you could store chunks that have already been accessed and check that list first or something.

Also:
> I think the delayedUpdateQueue should probably be a set or something, instead of a circular buffer.

Maybe *in addition of, not instead.

tillpp:
I came up with the following Solution:
having just a priorityQueue for the updates, making timed ticks possible.
so you say which block and when it should be updated

Advantages:
- simpler
- leaves break in layers.

Disadvantage:
- leaves don't immediately get destroyed when the log next to them is broken
- Blocks might be updated multiple times,

> - Blocks might be updated multiple times,

 We could avoid this like @MiniBaker007 said with a Set, the problem is: the same block could be updated for different times, which could lead to problem with timing.  (Technically we could still have a Set that resets each tick , but how likely is it, that the same block will be updated at the exact same tick? I would guess that this Optimization is not worth the effort, and if i am wrong, we can still implement it in a future PR)
 
>  - leaves don't immediately get destroyed when the log next to them is broken

I could still implement this in addition but as you said  ``Though I am starting to wonder if this isn't getting too complicated.`` it would make the code more complex, because i would have to pass the eventSource through different functions.

Argmaster:
#1522

tillpp:
@Argmaster oh it is very similar to this, but one problem i see with #1522  is that, it could lead to race conditions,
Let's say i break a log, but destroy a leave and replace it with another block before it decayed, then would still decay.
To solve this, we could save which blocktype the block had, when we scheduled the block, and so on.

I think this(#2265 ) approach is simpler and gives similar results.

Argmaster:
> To solve this, we could save which blocktype the block had, when we scheduled the block, and so on.

Doesn't sound like a big issue to me.

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+	return false;
+}
+pub fn run(_: *@This(), params: main.callbacks.ServerBlockCallback.Params) main.callbacks.Result {
+	const wx = params.chunk.super.pos.wx + params.x;
+	const wy = params.chunk.super.pos.wy + params.y;
+	const wz = params.chunk.super.pos.wz + params.z;
+
+	main.items.Inventory.Sync.ServerSide.mutex.lock();
```
Argmaster: So, there is a dependency between this and inventory system, I assume the dependency is implicit - we don't explicitly share any resources, but there is something somewhere that we share and we have to lock it?

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,94 @@
...
+	defer main.items.Inventory.Sync.ServerSide.mutex.unlock();
+
+	if(Server.world) |world| {
+		if(world.getBlock(wx, wy, wz)) |leaf| {
+			// check if there is any log in the proximity?^
+			if(foundWayToLog(world, leaf, wx, wy, wz))
+				return .ignored;
+
+			// no, there is no log in proximity
+			_ = world.cmpxchgBlock(wx, wy, wz, leaf, Block{.typ = leaf.decayReplacement(), .data = 0});
```
Argmaster: Data should be naturalDefault or whatever it is called (look up in rotations)

src/server/world.zig
```diff
@@ -640,6 +646,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 			updateRequest.region.decreaseRefCount();
 		}
 		self.regionUpdateQueue.deinit();
+		self.updateSystem.deinit();
```
Argmaster: So, if server crashes all the update backlog goes to hell?

src/server/UpdateSystem.zig
```diff
@@ -0,0 +1,67 @@
...
+		self.* = UpdateSystem{
+			.queue = .init(main.globalAllocator.allocator, void{}),
+		};
+		return self;
+	}
+	pub fn deinit(self: *UpdateSystem) void {
+		self.queue.deinit();
+		main.globalAllocator.destroy(self);
+	}
+	pub fn add(self: *UpdateSystem, position: Vec3i, inTicks: u32) void {
```
Argmaster: inTicks -> delayTicks

tillpp:
> > To solve this, we could save which blocktype the block had, when we scheduled the block, and so on.
> 
> Doesn't sound like a big issue to me.

every time i rewrite the system, someone throws in a new idea for me to rewrite the system again.
I will just wait until you guys have gathered all of your idea together for a final rewrite,  instead of me adjusting it everyday.

src/Inventory.zig
```diff
@@ -1829,6 +1829,22 @@ pub const Command = struct { // MARK: Command
...
+				} else {
+					// onBreak event
+					const baseChunk = main.server.world.?.getOrGenerateChunkAndIncreaseRefCount(.{
+						.wx = self.pos[0] & ~@as(i32, main.chunk.chunkMask),
+						.wy = self.pos[1] & ~@as(i32, main.chunk.chunkMask),
+						.wz = self.pos[2] & ~@as(i32, main.chunk.chunkMask),
+						.voxelSize = 1,
+					});
+					defer baseChunk.decreaseRefCount();
+					_ = self.oldBlock.onBreak().run(.{
```
Argmaster: I was under impression that this branch would also be triggered for additive rotation modes, eg torches, no?

src/server/world.zig
```diff
@@ -546,6 +548,9 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 
 	biomeChecksum: i64 = 0,
 
+	// TODO: do this per chunk not world.
```
Argmaster: No new Todos allowed. Make an issue, if you really want you can leave the issue link instead.

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,94 @@
...
+	// init array to mark already searched blocks.
+	const checkRange = logBreak.widerUpdaterRange;
+	const checkLength = checkRange*2 + 1;
+	var checked: [checkLength*checkLength*checkLength]bool = undefined;
+	for(0..checkLength*checkLength*checkLength) |i| {
+		checked[i] = false;
+	}
+
+	// queue for breath-first search
+	var queue = main.utils.CircularBufferQueue(Vec3i).init(main.globalAllocator, 32);
```
Argmaster: Should use stack allocator

src/server/world.zig
tillpp: It is marked as a TODO,
but for current LeaveDecay, yes, it isnt that important right now.

src/callbacks/block/server/leafDecay.zig
tillpp: Why do they use globalAllocator? world.zig, that is were i copied that line from:
grafik

src/callbacks/block/server/leafDecay.zig
tillpp: I used it because replaceBlock.zig also used it:
grafik

Modifiyng a block does require locking the inventory:

grafik

So this is not my issue, can be a seperate PR or something like "seperate inventory from blockbreaking" or something like that, idk

src/callbacks/block/server/leafDecay.zig
tillpp: do you mean naturalStandart? it is a member variable of rotationMode, i dont have an instance of a RotationMode anywhere. I can only solve this ugly, without a major rewrite on how to access block properties. 

just like it was solved here in blocks.zig
grafik

src/server/world.zig
tillpp: I followed QDs advice to write more TODOs, 
but that  advice was given 4 years ago, didn't know about the new rules .
solved.

src/Inventory.zig
tillpp: I was under impression that the upper if condition only returns not null on failure.
grafik

src/server/UpdateSystem.zig
tillpp: hardest problem in CS, naming things

tillpp:
> > To solve this, we could save which blocktype the block had, when we scheduled the block, and so on.
> 
> Doesn't sound like a big issue to me.

1.My point was, that it would require a lot of rewrite (in fact i think the entire PR could be started from scratch)
without having any advantage, that couldnt be solved with blockData.

2. Even having disadvantages, like race conditions, that can be mitigated with my idea of storing the block when the event was triggered, but that doesnt fix it completly:  the user can break a block and place the exact same block again, but the eventtimer wouldnt be reset.

for example: a plant might be broken, before the "growth" event hit, then the exact same plant would gets planted without resetting  the eventTimer, making it grow faster then intended. 
There are ways to prevent this, (forexample: removing the events from the queue when the block was broken) but they are expensive.

so i sugguest we leave it like it is

src/server/world.zig
tillpp: i originally had the plan to have two update systems, one for important stuff, and one to be thrown away if the world is closed and run slower if the load gets to heavy.

src/callbacks/block/server/leafDecay.zig
MiniBaker007: Because the data for those need to persist beyond the end of the function's scope. But, in the findPathToLog function,  you don't need the queue anymore after you exit the function.

tillpp: fixed: https://github.com/PixelGuys/Cubyz/pull/2265/commits/b5b68c412c07b6305b715d90e67d7c18625921b5

src/callbacks/block/server/leafDecay.zig
IntegratedQuantum: the problem is #1744

IntegratedQuantum:
> I think this(https://github.com/PixelGuys/Cubyz/pull/2265 ) approach is simpler and gives similar results.

From looking at it, the big advantage that I would see in #1522 is that it allows batching events (you could have one leafDecay event that checks an entire shell of leaves at once, instead of spawning 100s of separate events). Furthermorre it seems that it would generalize nicer to other types of events (e.g. a worldedit command could be implemented to spawn one event per chunk, allowing it to be parallelized)

Either way, we would still need a dedicated block update system, since any block change can trigger changes of its neighbors (rotation mode changes would also be captured like this)
This system should be separate to allow optimizing it. A block update is a dumb action and doesn't need any data besides its 15 bit position relative to the chunk it's in.

So I think we should split this PR into exactly these two cases, one PR for the leaf decay as dumb block updates (→basically what you had before implementing the time mechanics), and one PR for the more complex system to fix the leaf decay edge case when breaking logs.

tillpp:
> So I think we should split this PR into exactly these two cases, one PR for the leaf decay as dumb block updates (→basically what you had before implementing the time mechanics

what about the old problem, that we would have to widen the update Range for logBreaks?
Or do we just let it stay like this for now, (i.e. logBreak potentially breaking all leaves instantly nearby and only blocks futher away decaying in layers)
until the event complex system has been implemented in a seperate PR?

IntegratedQuantum:
> what about the old problem, that we would have to widen the update Range for logBreaks?

That's what I meant by "the leaf decay edge case when breaking logs". Please fix it separately. It's honestly not the biggest issue, considering that the player can just retrigger the decay when breaking another leaf block.

So far all the controversy has been about the fix to that, and we could have the first half of the system merged much sooner without it. Plus it makes reviewing easier when there isn't a 500 line PR that tries to implement multiple things.

tillpp:
> So far all the controversy has been about the fix to that, and we could have the first half of the system merged much sooner without it. Plus it makes reviewing easier when there isn't a 500 line PR that tries to implement multiple things.

YOU GUYS kept complaining and keep wanting me to change stuff that could have been seperate PRs, and then change your mind again.
So DONT blame me.

first you guys didnt want the timed update idea, so i did the immediate blockupdate system
Then you guys wanted to have the leaves closes to the log to break.
Nope to complex, changed mind, rather having to want the first idea again (`` Maybe we should return to your original proposal to spawn all the events in a radius with a distance-based timestamp.``)
then i implement that, oh now you guys want me to add (https://github.com/PixelGuys/Cubyz/issues/1522) and make it more complex again.
Nope lets keep it simple again and split it into two PR.

> Plus it makes reviewing easier when there isn't a 500 line PR that tries to implement multiple things.

At this point, i am the last person anyone here can blame for this situation tbh.


> the leaf decay edge case when breaking logs

for some reason i misread that line

IntegratedQuantum:
> So DONT blame me.

I'm not trying to blame you at all, I'm just stating the facts. At the end of the day it's no ones fault, it's just that this topic turned out to be more complex than anticipated and required looking at multiple design approaches and their edge cases.

The only thing you could be ~~credited~~ blamed for is giving us a chance to get this important feature in the next version

src/blocks.zig
```diff
@@ -69,6 +69,13 @@ var _alwaysViewThrough: [maxBlockCount]bool = undefined;
 var _hasBackFace: [maxBlockCount]bool = undefined;
 var _blockTags: [maxBlockCount][]Tag = undefined;
 var _light: [maxBlockCount]u32 = undefined;
+
+/// is this block decayable?
+var _onBreak: [maxBlockCount]ServerBlockCallback = undefined;
+var _onUpdate: [maxBlockCount]ServerBlockCallback = undefined;
+var _decayProhibitor: [maxBlockCount]bool = undefined;
+var _decayReplacement: [maxBlockCount]u16 = undefined;
```
IntegratedQuantum: The replacement should be part of the callback data, since it isn't needed for everything. Also it should be a full block, and not just the type (this would allow for some sort of progressive decay)

src/blocks.zig
```diff
@@ -69,6 +69,13 @@ var _alwaysViewThrough: [maxBlockCount]bool = undefined;
 var _hasBackFace: [maxBlockCount]bool = undefined;
 var _blockTags: [maxBlockCount][]Tag = undefined;
 var _light: [maxBlockCount]u32 = undefined;
+
+/// is this block decayable?
```
IntegratedQuantum: I think this comment is outdated.

assets/cubyz/blocks/leaves/undecayable/yellow.zig.zon
IntegratedQuantum: > i still don't really understand how i would make it with rotations, such that placing a Block by a human, would give it a different blockData, maybe with overriding updateData or modifyBlock, but i am not sure if they are not also called by the world for a different thing that is unrelated to the player.

You can do it in `generateData`, updateData is called when a neighbor changes. generateData is only called upon a player action.

src/Inventory.zig
```diff
@@ -1829,6 +1829,22 @@ pub const Command = struct { // MARK: Command
 					const actualBlock = main.server.world.?.getBlockAndBlockEntityData(self.pos[0], self.pos[1], self.pos[2], &writer) orelse return;
 					main.network.Protocols.blockUpdate.send(user.?.conn, &.{.init(self.pos, actualBlock, writer.data.items)});
 					return error.serverFailure;
+				} else {
+					// onBreak event
```
IntegratedQuantum: Should it be called `onPlayerBreak`, if it is exclusive to player inventory actions?

Also why not call it for other break events? After all the block is still broken, just not by the player.

src/Inventory.zig
```diff
@@ -1829,6 +1829,22 @@ pub const Command = struct { // MARK: Command
...
+					const baseChunk = main.server.world.?.getOrGenerateChunkAndIncreaseRefCount(.{
```
IntegratedQuantum: Please don't generate a new chunk.

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,96 @@
...
+
+	// init array to mark already searched blocks.
+	const checkRange = 5;
+	const checkLength = checkRange*2 + 1;
+	var checked: [checkLength*checkLength*checkLength]bool = undefined;
+	for(0..checkLength*checkLength*checkLength) |i| {
+		checked[i] = false;
+	}
+
+	// queue for breath-first search
```
IntegratedQuantum:
```suggestion
	// queue for breadth-first search
```

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,96 @@
...
+		// get the (potential) log
+		if(world.getBlock(value[0] + wx, value[1] + wy, value[2] + wz)) |log| {
+			// it is a log
+			// end search.
+			if(log.decayProhibitor()) {
+				return true;
+			}
+			// it is the same type of leaf
+			// continue search!
+			else if(log.typ == leaf.typ) {
```
IntegratedQuantum: I don't like this. There is a random else, three lines away from the corresponding if.

Also generally early-continues/breaks/returns are more readable, so I'd suggest
```suggestion
			if(log.typ != leaf.typ) continue;
			// it is the same type of leaf
			// continue search!
```

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,96 @@
...
+				for(main.chunk.Neighbor.iterable) |offset| {
+					// relative position
+					const X = value[0] + offset.relX();
```
IntegratedQuantum: Please follow the naming convention: Variable names should be lowerCamelCase
```suggestion
					const x = value[0] + offset.relX();
```

IntegratedQuantum: Also why not use a vector?
```zig
const relativePosition = value + offset.relPos();
if(vec.lengthSqr(relativePosition)...

...getIndexInArray(relativePosition...
```

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,96 @@
...
+
+	if(Server.world) |world| {
+		if(world.getBlock(wx, wy, wz)) |leaf| {
+			// check if there is any log in the proximity?^
+			if(foundWayToLog(world, leaf, wx, wy, wz))
+				return .ignored;
+
+			// no, there is no log in proximity
+			var replacement = Block{.typ = leaf.decayReplacement(), .data = 0};
+			replacement.data = replacement.mode().naturalStandard;
```
IntegratedQuantum:
```suggestion
			const replacement = self.decayReplacement;
```

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,96 @@
...
+
+			_ = world.cmpxchgBlock(wx, wy, wz, leaf, replacement);
+
+			return .handled;
```
IntegratedQuantum: Only return handled if the cmpxchgBlock succeeded

src/server/world.zig
```diff
@@ -546,6 +548,8 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 
 	biomeChecksum: i64 = 0,
 
+	updateSystem: *UpdateSystem,
```
IntegratedQuantum: There should be one queue per EntityChunk

src/server/world.zig
IntegratedQuantum: I think it's fine not to safe block updates. This will only cause issues in rare cases, and the issues aren't that severe honestly, since a block update can be easily reinitiated by the player.

It is more important for the long-term timed events.

src/server/world.zig
```diff
@@ -1237,8 +1245,20 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		for(userList) |user| {
 			main.network.Protocols.blockUpdate.send(user.conn, &.{.{.x = wx, .y = wy, .z = wz, .newBlock = newBlock, .blockEntityData = &.{}}});
 		}
+
+		self.updateSurrounding(wx, wy, wz);
```
IntegratedQuantum: This is very vague
```suggestion
		self.triggerNeighborBlockUpdates(wx, wy, wz);
```

src/server/UpdateSystem.zig
```diff
@@ -0,0 +1,46 @@
+const std = @import("std");
+
+const main = @import("main");
+const ZonElement = main.ZonElement;
+const vec = main.vec;
+const Vec3i = vec.Vec3i;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+pub const UpdateSystem = struct {
+	queue: main.utils.CircularBufferQueue(Vec3i),
```
IntegratedQuantum: Please use a ListUnmanaged and swap it every tick.

Also this being a global resource accessible from multiple threads, it needs a mutex.

```zig
self.mutex.lock();
const list = self.list;
defer list.deinit(main.globalAllocator);
self.list = .{};
self.mutex.unlock();
for(list.items) ...
```

src/server/UpdateSystem.zig
```diff
@@ -0,0 +1,46 @@
...
+
+	pub fn init() *UpdateSystem {
+		const self = main.globalAllocator.create(UpdateSystem);
```
IntegratedQuantum: This allocation is pointless, just store it by value.

src/Inventory.zig
tillpp: I think i misunderstood your comment:
https://github.com/PixelGuys/Cubyz/pull/2265#discussion_r2532197333

tillpp: fixed

src/callbacks/block/server/leafDecay.zig
tillpp: true

tillpp: also: the else is completly useless in this situation anyway because the first condition returns anyway

src/callbacks/block/server/leafDecay.zig
tillpp: 1.a relic, in a previous version, there used to be a cast here, so i had to use 3 different variables, and since lowercase xyz was already used i used XYZ instead.

i can fix that.

2. isn't  lengthSqrt expensive?
EDIT: oops nvm, misread "sqr" as "sqrt"

src/server/world.zig
tillpp: I think it would make more sense in the ServerChunk? or even Chunk?

IntegratedQuantum: EntityChunk is the one intended for server-side simulation work.

src/server/UpdateSystem.zig
tillpp: Out of curiousity:
I assume i should use ListUnmanaged because it has O(1) for appending? 
Or was there another reason aside from time-complexity?

IntegratedQuantum: You should use a list because it is a simpler data structure than a queue, since it doesn't have to do any wrapping in the underlying array, and we don't need a queue if we swap the data structure (which reduces congestion on the lock).
Complexity in both cases is O(1) amortized.

src/server/world.zig
tillpp: ok done.

src/blocks.zig
```diff
@@ -415,6 +441,10 @@ pub const Block = packed struct { // MARK: Block
 		return _opaqueVariant[self.typ];
 	}
 
+	pub inline fn decayProhibitor(self: Block) bool {
```
IntegratedQuantum: Please put them in the same order as their corresponding arrays above.

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,100 @@
...
+	// queue for breadth-first search
+	var queue = main.utils.CircularBufferQueue(Vec3i).init(main.stackAllocator, 32);
+	defer queue.deinit();
+
+	queue.pushBack(Vec3i{0, 0, 0});
+	checked[getIndexInCheckArray(Vec3i{0, 0, 0}, checkRange)] = true;
+
+	while(queue.popFront()) |value| {
+		// get the (potential) log
+		if(world.getBlock(value[0] + wx, value[1] + wy, value[2] + wz)) |log| {
```
IntegratedQuantum: Use wrapping arithmetic in world space to avoid integer overflow at the world border
```suggestion
		if(world.getBlock(value[0] +% wx, value[1] +% wy, value[2] +% wz)) |log| {
```

src/callbacks/block/server/leafDecay.zig
```diff
@@ -0,0 +1,100 @@
...
+
+	if(params.block.mode() == main.rotation.getByID("cubyz:decayable")) {
+		if(params.block.data != 0)
+			return .ignored;
+	} else {
+		std.log.err("Expected {s} to have cubyz:decayable rotation", .{params.block.id()});
+	}
+
+	main.items.Inventory.Sync.ServerSide.mutex.lock();
+	defer main.items.Inventory.Sync.ServerSide.mutex.unlock();
```
IntegratedQuantum: Mutexes should be locked around the tightest possible scope to avoid congestion. In this case that would be the cmpxchgBlock call.

src/server/UpdateSystem.zig
```diff
@@ -0,0 +1,53 @@
+const std = @import("std");
+
+const main = @import("main");
+const ZonElement = main.ZonElement;
+const vec = main.vec;
+const Vec3i = vec.Vec3i;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+pub const UpdateSystem = struct {
```
IntegratedQuantum: On terminology, I think we should generally refer to this as block updates, in a game there are many things that can be updated, so I think we should be more specific.
Unless you plan to extend this to other update code, in which case this should also include handling random tick updates.

src/server/UpdateSystem.zig
```diff
@@ -0,0 +1,53 @@
+const std = @import("std");
```
IntegratedQuantum: Only use UpperCamelCase in file names when the file is used as a struct.
In this case I think we should just make the file a struct instead of having a needless extra namespace inside of it.

src/server/UpdateSystem.zig
```diff
@@ -0,0 +1,53 @@
...
+pub const UpdateSystem = struct {
+	list: main.ListUnmanaged(Vec3i) = .{},
+	mutex: std.Thread.Mutex = .{},
+
+	pub fn init() UpdateSystem {
+		return .{};
+	}
+	pub fn deinit(self: *UpdateSystem) void {
+		self.mutex.lock();
+		defer self.mutex.unlock();
```
IntegratedQuantum: In deinit there should be no other thread locking the mutex.
A better practice is to set `self.mutex = undefined;` to actually catch this issue, instead of locking the lock which hides the issue.

src/server/UpdateSystem.zig
```diff
@@ -0,0 +1,53 @@
...
+		const list = self.list;
+		defer list.deinit(main.globalAllocator);
+		self.list = .{};
+		self.mutex.unlock();
+
+		// handle events
+		for(list.items) |event| {
+			const x = event[0] & main.chunk.chunkMask;
+			const y = event[1] & main.chunk.chunkMask;
+			const z = event[2] & main.chunk.chunkMask;
```
IntegratedQuantum: I'd suggest to instead use `struct{x: u5, y: u5, z: u5}`. This allows us to assert at the callsite that no one inserted positions outside the chunk, and avoids the extra operations here.

src/callbacks/block/server/leafDecay.zig
tillpp: but would that make sense logically? I mean it would take the value of the other side of the world, but the other side of the world isnt visible?

IntegratedQuantum: the plan is to have the world wrap around, see also #310

src/server/UpdateSystem.zig
tillpp: ok done.

src/blocks.zig
```diff
@@ -394,7 +414,15 @@ pub const Block = packed struct { // MARK: Block
...
-
+	pub inline fn onBreak(self: Block) ServerBlockCallback {
+		return _onBreak[self.typ];
+	}
+	pub inline fn onUpdate(self: Block) ServerBlockCallback {
+		return _onUpdate[self.typ];
+	}
+	pub inline fn decayProhibitor(self: Block) bool {
+		return _decayProhibitor[self.typ];
+	}
```
IntegratedQuantum, line 424: Again please put these functions in the same order as the array declarations above (or in this case I think it would make more sense to move the array declarations below the onInteract)

src/server/world.zig
```diff
@@ -1237,8 +1249,41 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+					.z = z,
+				});
+			}
+		}
+		self.triggerNeighborBlockUpdates(wx, wy, wz);
+
 		return null;
 	}
+	pub fn triggerNeighborBlockUpdates(self: *ServerWorld, wx: i32, wy: i32, wz: i32) void {
+		// trigger updates:
```
IntegratedQuantum: pointless comment
```suggestion
```

IntegratedQuantum (changes requested):
Two more small things, then we can merge this.

src/blocks.zig
```diff
@@ -69,10 +69,14 @@ var _alwaysViewThrough: [maxBlockCount]bool = undefined;
 var _hasBackFace: [maxBlockCount]bool = undefined;
 var _blockTags: [maxBlockCount][]Tag = undefined;
 var _light: [maxBlockCount]u32 = undefined;
+
```
IntegratedQuantum: unintended line change. light and absorption belong together and should not be separated by a line.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):
Alright, time to break everyone's spawn atmosphere

src/callbacks/block/server/leafDecay.zig
codemob-dev: > 1.a relic, in a previous version, there used to be a cast here, so i had to use 3 different variables, and since lowercase xyz was already used i used XYZ instead.

relX, relY, relZ

PR #2268, opened 2025-11-12
Updated main import to aid type detection

Auseawesome:
Changed @import("main") to @import("main.zig") to make consistent with other imports and significantly improve type detection within editor.


Auseawesome:
Nevermind, this seems to cause issues that aren't present for me locally.

PR #2276, opened 2025-11-13
Added replaceable to foliage and corrected spelling everywhere

### LINKED ISSUE #2267: Foliage should be replaceable

right now it’s really annoying to place blocks over small foliage since you have to break it all manually

Image

We should update the block assets to have .replaceable = true so they can be replaced by blocks the player builds.

Auseawesome:
Fixes #2267

The refactor of replaceable's spelling is necessary because it will facilitate easy adding of this property on things in the future and avoid confusion as to why this behaviour was not working. (See issue to see how it was described to add .replaceable however it was spelled as .replacable in code leading to confusion when I went to solve this issue)


Auseawesome:
Due to the hitboxes, while this tag works, this is currently not a fully suitable solution.

Either the hitbox needs to be resized or replaceable needs to be modified so right clicking on the hitbox replaces rather than placing adjacent to to.

Auseawesome:
I've marked this as ready for review as the changes in it correctly add the tag. The additional functionality for blocks with a replaceable tag can be achieved in a separate PR.

H41ogen:
It seems this was never moved out of the WIP category.

Argmaster:
How is that still not moved to ready for review...
@Auseawesome please resolve conflicts tho.

Auseawesome:
Unsure how it was never marked as ready. Brought it up multiple times. I'll fix the conflicts now.

Argmaster:
I will try to bring it up if another review stream happens.

IntegratedQuantum (approved):

PR #2277, opened 2025-11-13
Make hovering over an item highlight all other items of that type

### LINKED ISSUE #2271: Hovering over an item should highlight all items of that type
### Description

This would help with organization and inventory management.

codemob-dev:
Fixes #2271 
This will conflict with #2247, although it should be an easy fix.


IntegratedQuantum (changes requested):
I'm not convinced that this is a good idea. I personally find it a bit weird in-game.
@ikabod-kee please try this.

ikabod-kee:
It could use more contrast, but I actually really like this change.

I could see this working better if maybe the moused-over slot was outlined or white rather than just a darker blue.

ikabod-kee:
So: give the moused-over slots a white highlight, while the others get the old blue highlight.

ikabod-kee:
image
Getting this error when running

codemob-dev:
> image Getting this error when running

Fixed

SpellDigger:
this is straight up counterintuintive and u cant even justify it, when hovering over a itemslot one would only expect the itemslot they are hovering on to be highlighted, this would just create people asking if its a bug

it also looks wierd like quantum said, we really need more justification thats pretty convincing on this

codemob-dev:
> u cant even justify it

1. It highlights items that you already have if you go to craft something
2. It makes it easier to count how many of a specific item you have

ikabod-kee:
Perhaps having it be under a toggle/hold thing when pressing Ctrl, Shift, or Alt would work

codemob-dev:
> Perhaps having it be under a toggle/hold thing when pressing Ctrl, Shift, or Alt would work

Since it doesn't actually obstruct anything on the screen (it isn't clutter) I don't think that should be necessary; maybe we could have a toggle in settings for it?

ikabod-kee:
Yeah, a toggle in the settings sounds nice. I do actually like the feature.

ikabod-kee:
Also, it seems like some checks are failing. I'll check out the new changes when I can

codemob-dev:
> Also, it seems like some checks are failing. I'll check out the new changes when I can

Yeah, I'll fix that in a bit.

Argmaster:
As an alternative to toggle, you could also make the highlight of other slots of same type less visible (can I get a screenshot? It should have been included in the description!)

H41ogen:
maybe other items should only be highlighted if some key is held down, or if you hover over an item for a second or two

codemob-dev:
Closing this as I haven't really been motivated to work on it anymore. I might try it again at some other point though!

PR #2278, opened 2025-11-13
Renamed leaves_ to leaves/

tillpp:
QD wanted me to seperate the leafDecay PR (
https://github.com/PixelGuys/Cubyz/pull/2265/#issuecomment-3528826289)
So this PR is only for renaming the files from _leaves to leaves/...

### DIFF
--- a/assets/cubyz/blocks/branch/leafy/_defaults.zig.zon
+++ b/assets/cubyz/blocks/branch/leafy/_defaults.zig.zon
@@ -9,7 +9,4 @@
 		.shellModel = "cubyz:cube",
 		.textureSlotOffset = 6,
 	},
-	.onUpdate = .{
-		.type = .leafBreak,
-	},
 }


IntegratedQuantum:
Please leave out the undecayable variants for now.

careeoki:
I think it might be better if it was `leaves/oak`, `leaves/opaque/oak`, `leaves/undecaying/oak`, matching the branch folder

IntegratedQuantum:
Yeah, that would also allow using _defaults more effectively.

tillpp:
> Please leave out the undecayable variants for now.

ok, done,
then the undecayables will be included in the leafDecay PR?
or in a third seperate PR?

IntegratedQuantum:
> then the undecayables will be included in the leafDecay PR?

Yes, there we'll also make the decision whether they even should exist as such.

tillpp:
ok, i think this is ready for a review now

IntegratedQuantum (approved):
Before continuing on #2265 please also refactor these files and put all common fields into _defaults

careeoki:
Um, this breaks all the tree sbbs

image

IntegratedQuantum:
It's because of #2005

(the one time where I don't double check someone's PR this happens...)

IntegratedQuantum:
@tillpp how could you even test leaf decay when this breaks so badly?

careeoki:
Also pine leaves were missed, probably because they're called pine needles
image

tillpp:
> Also pine leaves were missed, probably because they're called pine needles

yeah, i wasn't sure if i should rename them into leaves/pine, since they have needles in their name?

> how could you even test leaf decay when this breaks so badly?

there were only round trees in the biomes, that i tested.

careeoki:
> yeah, i wasn't sure if i should rename them into leaves/pine, since they have needles in their name?

Yes, they should be in leaves folder, then eventually when we have language support we can call them "Pine Needles" if we like

PR #2282, opened 2025-11-14
Adding a setSpawn command

Crepestrom:
This request will add a command that allows players to set their spawnpoint when they die.
This spawnpoint is saved in the players folder allong with the other player data.

If hooked up properly it can also later be used to make blocks that allow players to set their spawnpoint.

it updates the server, game, network, world, and Inventory.zig
also the commands folder for the set spawn command.

This is mostly for under the hood improvements to add blocks that can Set Spawn

- The Spawnpoitns are saved in the player files in the world along with the other playerdata
- Spawnpoint is saved, changed, and set even across different worlds and sessions

Go to ([link](https://github.com/PixelGuys/Cubyz/pull/2374)) Instead this is depreciated

### DIFF
--- a/Inventory.zig
+++ b/Inventory.zig
@@ -1,0 +1,2212 @@
+const std = @import("std");
+
+const main = @import("main");
+const BaseItem = main.items.BaseItem;
+const Block = main.blocks.Block;
+const Item = main.items.Item;
+const ItemStack = main.items.ItemStack;
+const Tool = main.items.Tool;
+const utils = main.utils;
+const BinaryWriter = utils.BinaryWriter;
+const BinaryReader = utils.BinaryReader;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const vec = main.vec;
+const Vec3d = vec.Vec3d;
+const Vec3f = vec.Vec3f;
+const Vec3i = vec.Vec3i;
+const ZonElement = main.ZonElement;
+const Neighbor = main.chunk.Neighbor;
+const BaseItemIndex = main.items.BaseItemIndex;
+const ToolTypeIndex = main.items.ToolTypeIndex;
+
+const Gamemode = main.game.Gamemode;
+
+const Side = enum {client, server};
+
+pub const InventoryId = enum(u32) {_};
+
+pub const Callbacks = struct {
+	onUpdateCallback: ?*const fn(Source) void = null,
+	onFirstOpenCallback: ?*const fn(Source) void = null,
+	onLastCloseCallback: ?*const fn(Source) void = null,
+};
+
+pub const Sync = struct { // MARK: Sync
+
+	pub const ClientSide = struct {
+		pub var mutex: std.Thread.Mutex = .{};
+		var commands: main.utils.CircularBufferQueue(Command) = undefined;
+		var maxId: InventoryId = @enumFromInt(0);
+		var freeIdList: main.List(InventoryId) = undefined;
+		var serverToClientMap: std.AutoHashMap(InventoryId, Inventory) = undefined;
+
+		pub fn init() void {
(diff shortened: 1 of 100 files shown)


src/server/command/setSpawn.zig
```diff
@@ -0,0 +1,42 @@
...
+		return;
+	}
+	if(z == null) {
+		z = source.player.pos[2];
+	}
+	x = std.math.clamp(x.?, -1e9, 1e9); // TODO: Remove after #310 is implemented
+	y = std.math.clamp(y.?, -1e9, 1e9);
+	z = std.math.clamp(z.?, -1e9, 1e9);
+
+	main.items.Inventory.Sync.setSpawn(source, .{x.?, y.?, z.?});
```
OneAvargeCoder193: You shouldn't send spawn points through the inventory system. A simple call to main.network.Protocols.genericUpdate.sendSpawnPoint should be good enough. The Inventory.Sync system should only be used for things that need to be synchronized.

Crepestrom: oh ok i was copy and pasting a lot of code

Crepestrom:
did not have a chance to test it with multiplayer yet either though

Crepestrom:
it should be finished now

Crepestrom:
Updated to the main branch rechecked and it still works

src/server/command/setSpawn.zig
Crepestrom: well since it changes the players spawn point on the cleint side and user it does need to be synced
clientside so when the player dies its in the correct spot
and serverside so when the client disconnects their spawnpoint is saved

OneAvargeCoder193: Check how the tp command is done, it deosn't go through the Inventory system but instead through network.zig

Crepestrom: it functions more similar to the gamemode command since it has to save data when the player leaves the game
therefore it does need to be run through inventory

Crepestrom: the tp command changes the players position
position seems to be synced to the server from the client through chunk request in the server.zig
In the end players Spawn position has to be synced anyway

Crepestrom: Health and damage is also synced through the inventory.zig

Crepestrom: Is there a reason to why it should not sync through the inventory system?

OneAvargeCoder193: > Health and damage is also synced through the inventory.zig

> Is there a reason to why it should not sync through the inventory system?

Health and damage is synced through the inventory system as it needs to be synced between each player, and if it gets desynced then that would be really bad. The inventory system should only be used when the data needs to be synced between multiple players, though that is a bit of a simplification.

Crepestrom: ah ok do you know which system should i use to use to sync it then?

OneAvargeCoder193: > ah ok do you know which system should i use to use to sync it then?

You don't need to sync it, just create a new genericUpdate packet in network.zig and pass it through there

Crepestrom: ok i will check it but i have moved it into a generic update

Crepestrom: will check to make sure it still functions

Crepestrom: nope it does not either save properly anymore or load properly anymore

Crepestrom: i think loading is broken now

Crepestrom: ah whatever i willl just revert it

Crepestrom:
eh damnit i ruined it

Crepestrom:
ill open a new one i saved a copy of the right version

PR #2291, opened 2025-11-14
Cobblestone and slate building block retextures

careeoki:
Usually I'd make a gif, but I wasn't able to get it to display the colors correctly.
image
image
Cobblestone is the biggest change here. Old one was really high contrast, overly dithered, and still clinging on to the yellow tint from ancient times. 
The new one should fit better with other (relatively) recent textures. It also has more unique shading, with the highlights closer to the middle for a kinda faceted look.
image
The change to bricks, tiles, and walls is more subtle, just reducing the noisiness for a more polished stone appearance and giving it some better shading.


ikabod-kee:
Carrie never misses

codemob-dev:
I like it, it helps distinguish it from m-word game

du82:
I preferred the older more random looking cobblestone, the new one looks more patterned and I'm not a huge fan

ikabod-kee (approved):
Love these changes.

careeoki:
This was never added to the Github Project thing?

PR #2298, opened 2025-11-15
Hide usernames when GUI is hidden

thecreare:
I think this change is beneficial because GUI is most often hidden for screenshots/panoramas and usernames look bad in those

Before change:
image

After change:
image


Wunka:
I like this. 
Why did you make this a draft?

thecreare:
> Why did you make this a draft?

I mentioned I had these changes stashed for a while waiting until I had fewer open PRs and I was told to upload them as a draft if I was worried about that.

PR #2310, opened 2025-11-16
move Parsing of common Arguments for commands to one place

Wunka:
I have noticed that different commands use different implementations for the position parsing logic.
In my opinion its best if all use the same logic and have a certain center point where this logic lies.
The particle command seems as the best one to me.
This is a draft because I am not certain where exactly to put them and if with this PR I should already change the commands who use other methods of parsing the position. 

It could also be an idea to make the pos1 and pos2 worldedit commands use also positions so that you don't have to perfectly stand on the wanted position.

### DIFF
--- a/src/server/command/particles.zig
+++ b/src/server/command/particles.zig
@@ -32,7 +32,7 @@ fn parseArguments(source: *User, args: []const u8) anyerror!void {
 	var split = std.mem.splitScalar(u8, args, ' ');
 	const particleId = split.next() orelse return error.TooFewArguments;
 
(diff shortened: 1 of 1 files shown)


bagggage:
I think the problem runs a bit deeper - even now, every command parses everything on its own, leading to duplicated implementations and sometimes even inconsistent behavior for the same things (as in this case). Each command can also format and print error messages however it wants, with no standardized behavior.

To fix this, we probably need some kind of command caller that automatically parses the command and its arguments based on types and generates error messages automatically, invoking the command only if argument parsing succeeds. But that's a topic for a separate issue)

PR #2321, opened 2025-11-20
fix projection matrix not updating on window resize

gurtophagus:
projection matrix was not updated on window resize leading to stretched rendering in the main menu and the game world. example:

this: 

https://github.com/user-attachments/assets/2ded6189-da88-44d3-a616-e39b201168d3

is now fixed as:

https://github.com/user-attachments/assets/5e163657-4f47-4db6-8dae-62f9461b7bae

including the menu and all.
i checked logically and via debug prints that it only updates if the window resizes or user updates the fov in the graphic settings

### DIFF
--- a/src/renderer.zig
+++ b/src/renderer.zig
@@ -132,12 +132,13 @@ var lastFov: f32 = 0;
 pub fn updateFov(fov: f32) void {
 	if(lastFov != fov) {
 		lastFov = fov;
+		game.projectionMatrix = Mat4f.perspective(std.math.degreesToRadians(fov), @as(f32, @floatFromInt(lastWidth))/@as(f32, @floatFromInt(lastHeight)), zNear, zFar);
 	}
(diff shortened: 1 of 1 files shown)


BoySanic:
This seems fine to me.

The root cause here was that in #2004 we changed the behavior so that it doesn't update `game.ProjectionMatrix` every frame.
This fix seems simple enough to me as it updates the projection matrix on framebuffer changes as it used to.

The one thing that bugs me which I'm not sure there's an elegant fix for is that `updateFov` kinda does what `updateViewport` does except only when Fov changes, and `updateViewport` is only called when the framebuffer changes (or you're taking a panoramic screenshot).

We could reduce code duplication (one line) by re-combining these into one function and maybe moving the check for fov changes outside of the function. 

Not necessary, though.

BoySanic:
On second thought recombining isn't great either since it'd do extra work every frame having to run the rest of `updateViewport`

IntegratedQuantum (approved):

PR #2322, opened 2025-11-20
fix: Item drop/throw physics

tacf:
Some small tweaks to fix the item drop/throw physics

- tweak: Changed throw vertical position to Player eye position
- tweak: Added const to represent X, Y and Z values when accessing vector positions.
- tweak: Item velocity
- fix: Ground detection for friction application
- fix: Item collision with blocks

Velocity was tuned down.
Attempted to apply same strategy to detect when item is onGround and apply proper friction.
src/itemdrop.zig:371 seemed to be improperly using max value of the bounding box which seems to be the cause of items clipping passing through blocks when dropped.

As as small suggestion, added 3 constants to represent X, Y, Z to vec.zig. The intention is to improve code redeability. This obviously is optional and i'll revert it if this is not something you with to pursue. The purpose is not to push toward an immediate change but to introduce the ideia in order to be adopted as the code changes.

### DIFF
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -1475,7 +1475,8 @@ pub const Command = struct { // MARK: Command
 				if(_items[0].item != null) {
 					if(side == .server) {
 						const direction = vec.rotateZ(vec.rotateX(Vec3f{0, 1, 0}, -user.?.player.rot[0]), -user.?.player.rot[2]);
-						main.server.world.?.dropWithCooldown(_items[0], user.?.player.pos, direction, 20, main.server.updatesPerSec*2);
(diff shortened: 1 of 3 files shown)


codemob-dev:
There is already a lineup of PRs that accomplishes this: #2121 and #2223. Always search for existing issues and PRs before starting a new one.

In the future, don't put multiple unrelated changes in one PR. The item throw position, XYZ constants, and item physics changes should have been separated. Also, when creating a PR that fixes an issue, add `fixes #1234` to the PR description so it links the issue.

> As as small suggestion, added 3 constants to represent X, Y, Z to vec.zig. The intention is to improve code redeability.

IMO `vel.*[vec.X]` is less readable than `vel.*[0]`.

PR #2324, opened 2025-11-21
Split Hotbar from Inventory

### LINKED ISSUE #2311: Should the hotbar be separate from the inventory?
This would simplify a decent amount of code and fix #2253. It would also help prevent inventory slot mappings from getting too unwieldy, especially with armor and accessories (#604)

Wunka:
Closes: #2311 

I have renamed the inventory to mainInventory and added hotbar and am now just going error by error to get this to work.
I started this in a draft to get a little bit of input how this can be done good.

While we can make somethings work with just getting the mainInventory and hotbar, the Inventory based instructions like depositToAny or depositOrDrop are things were I would like some input. Currently I have done an optional fallback parameter but maybe there are better ways?
Maybe this system should be first done good in a different PR
(I have not yet implemented the fallback only added the parameter)

If we do the parameter there is the question of how should we best serialize this?
I am currently just exploiting that the fallback should not be the same as the dest Inventory.

There is also the Question of how the priority system should work. 
Example if I shift click item A into my inventory from a chest:
We now search the hotbar. If we find A the item from the chest goes also there.
But what if we don't find A or there is not enough space for the items from the chest. 
Should the mainInventory be searched first for A or should we fill empty spaces inside the hotbar first?
So either:
search(hotbar), search(mainInv), emptySpaces(hotbar), emptySpaces(mainInv)
or
search(hotbar), emptySpaces(hotbar), search(mainInv), emptySpaces(mainInv)

In my opinion the first option is better but maybe others have different opinions.

### DIFF
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -130,8 +130,8 @@ pub const ServerSide = struct { // MARK: ServerSide
 				if(self.managed == .internallyManaged) {
 					if(self.inv.type.shouldDepositToUserOnClose()) {
 						const playerMainInventory = getInventoryFromSource(.{.playerMainInventory = user.id}) orelse @panic("Could not find player main inventory");
-						sync.ServerSide.executeCommand(.{.depositOrDrop = .{.dest = playerMainInventory, .source = self.inv, .dropLocation = user.player.pos}}, null);
-						// TODO: add Hotbar after depositOrDrop is updated to slice
+						const playerHotbar = getInventoryFromSource(.{.playerHotbar = user.id}) orelse @panic("Could not find player hotbar");
+						sync.ServerSide.executeCommand(.{.depositOrDrop = .init(&.{playerHotbar, playerMainInventory}, self.inv, user.player.pos)}, null);
 					}
 					inventoryCreationMutex.lock();
 					defer inventoryCreationMutex.unlock();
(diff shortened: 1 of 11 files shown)


Wunka:
The only thing missing now in my opinion is migrations but I don't really know currently how I would do that.
The rest should be done. 
I have tested a few things like crafting, putting something into a chest shfiting from the chest with different scenarios like full hotbar, not full hotbar but item in mainInventory. 

I in some lines I don't like the how the code looks like example: `for([2]?Inventory{self.dest, self.fallback}) |_destInv| {` so if someone knows a better way  / more beautiful way please tell me (without just coping the entire code based on the inventory)

IntegratedQuantum (changes requested):
I think this PR is too big in its current form. I'd suggest to split off independent things and adapt them to be less hardcoded for just the hotbar.

In the future we probably want to allow crafting from chest ingredients (like in Terraria) and shift clicking/depositing to more than one other inventory (→ accessory slots, multiple open chests, "deposit to nearby chests" like in terraria).

> There is also the Question of how the priority system should work.

It should first got into the inventory that already has it, if all inventories have it, then it should start filling up the first one, even if the item exists partially in the others.
so
if(search(A)) emptySpaces(A)
if(search(B)) emptySpaces(B)
emptySpaces(A)
emptySpaces(B)

> Currently I have done an optional fallback parameter but maybe there are better ways?

Ideally it should be a slice of arbitrary length.

Issue #2331, opened 2025-11-22
branch problem: should they decay too? or not?


tillpp:
Should branches act like logs and prevent leaves from breaking? (current implementation)
Which could lead to very annoying searches for branches in leaves.

Or we let them decay too. 

Or only leafy branches into non-leafy branches.

Discuss!

Argmaster:
They should decay like leaves, because they are hard to track down. If we don't decay them, leaf blobs will remain as they are, especially for the SBB trees. They look nice, but we have to compensate for their poor cutting down experience. Alternative is giving up visuals which sucks in a game revolving around huge render distance.

Although, branches are somewhere in-between  - they also have to preserve attached leaves if the branches are connected to logs, otherwise large trees (like ancient sapele from jungle PR) may not work as expected, as they can have leaves like 8 blocks away from the trunk.

tillpp:
One solution could be a two tiered decay system.
Leaves either require Logs or Branches
and Branches require Logs

careeoki:
This is my ideal branch decay:

- when a tree's branches are no longer connected to a log, decay the branches.
- when a thin tree (made of only branches) is no longer connected to the ground, decay the whole tree. 
- when a thin tree is broken in the middle of it's trunk, decay the branches that are no longer connected to the ground (the floating half of the tree)
- block updates should not break thin trees

branches "being connected" means the actual branch connections from its rotation mode, it should not care about adjacent blocks it is not physically connected to

I think #2328  is very much needed before this, so that you don't lose the wood you would get from the branches when they decay.

PR #2334, opened 2025-11-23
Branch decay [Simple Implementation]

tillpp:
Introduces a two tiered system of decay.
- Leaves depend on either logs or branches.
- And branches depend on logs.

This means that breaking the logs of a tree is already enough for the whole tree to decay.
This would fix https://github.com/PixelGuys/Cubyz/issues/2331#issuecomment-3566930054

Downsides:
The log search algorithm is more expensive now, since it compares tags instead of the decayProhibitor attribute

Things i also had to introduce:
- Callbacks have a deinit function now.
- the 7th bit of the cubyz:branch data is used for storing the was-this-placed-by-human boolean

### DIFF
--- a/src/callbacks/block/server/decay.zig
+++ b/src/callbacks/block/server/decay.zig
@@ -11,8 +11,6 @@ const Branch = main.rotation.list.@"cubyz:branch";
 
 decayReplacement: blocks.Block,
 prevention: []const main.Tag,
-branchRotation: *const main.rotation.RotationMode,
-decayRotation: *const main.rotation.RotationMode,
 
 pub fn init(zon: ZonElement) ?*@This() {
 	const result = main.worldArena.create(@This());
@@ -35,9 +33,6 @@ pub fn init(zon: ZonElement) ?*@This() {
 			result.prevention = prevention.items;
 		}
 	}
-	// branchRotation
-	result.branchRotation = main.rotation.getByID("cubyz:branch");
-	result.decayRotation = main.rotation.getByID("cubyz:decayable");
 	return result;
 }
 fn getIndexInCheckArray(relativePosition: Vec3i, checkRange: comptime_int) usize {
@@ -72,7 +67,8 @@ fn foundWayToLog(self: *@This(), world: *Server.ServerWorld, leaf: Block, wx: i3
 	queue.pushBack(Vec3i{0, 0, 0});
 	checked[getIndexInCheckArray(Vec3i{0, 0, 0}, checkRange)] = true;
 
-	const sourceIsBranch = leaf.mode() == self.branchRotation;
+	const branchRotation = main.rotation.getByID("cubyz:branch");
+	const sourceIsBranch = leaf.mode() == branchRotation;
 
 	while(queue.popFront()) |value| {
 		// get the (potential) log
@@ -84,7 +80,7 @@ fn foundWayToLog(self: *@This(), world: *Server.ServerWorld, leaf: Block, wx: i3
 
 			// it is the same type of leaf? continue search! (Don't do it for branches. We've got isConnected instead!)
 			if(!sourceIsBranch and log.typ != leaf.typ) continue;
-			if(sourceIsBranch and log.mode() != self.branchRotation and !log.viewThrough()) return true;
+			if(sourceIsBranch and log.mode() != branchRotation and !log.viewThrough()) return true;
 			const branchData = Branch.BranchData.init(log.data);
 
 			for(main.chunk.Neighbor.iterable) |offset| {
@@ -111,10 +107,10 @@ pub fn run(self: *@This(), params: main.callbacks.ServerBlockCallback.Params) ma
 	const wy = params.chunk.super.pos.wy + params.blockPos.y;
 	const wz = params.chunk.super.pos.wz + params.blockPos.z;
 
-	if(params.block.mode() == self.decayRotation) {
+	if(params.block.mode() == main.rotation.getByID("cubyz:decayable")) {
 		if(params.block.data != 0)
 			return .ignored;
-	} else if(params.block.mode() == self.branchRotation) {
+	} else if(params.block.mode() == main.rotation.getByID("cubyz:branch")) {
 		const bd = Branch.BranchData.init(params.block.data);
 		if(bd.placedByHuman)
 			return .ignored;


tillpp:
Next time, i should probably create a new Branch from origin, sorry

careeoki:
I really really like how this feels, it makes SBB trees so much nicer to chop down. It's satisfying to watch the branches go away as you chop through the trunk of a tree.
I feel this really needs #2328 though since you lose out on a lot of wood without the branches.

Only problem I can think of is that little trees made of only branches can be chopped down by breaking one block, but honestly I don't mind if some trees are easier to chop.

tillpp:
> Only problem I can think of is that little trees made of only branches can be chopped down by breaking one block, but honestly I don't mind if some trees are easier to chop.

It might be some effort, but in those cases we could just use the non-decaying branches, but we would have to edit all little trees for that

ikabod-kee:
I think that branches should only decay if they aren't connected to a block, so a little sapling wouldn't wither away if you just placed a block next to it. Plus free realistic expectations!

tillpp:
> I think that branches should only decay if they aren't connected to a block, so a little sapling wouldn't wither away if you just placed a block next to it. Plus free realistic expectations!

that means two branches next to each other never decay, even if their logs are gone, because the branches are still connected to each other

Argmaster:
You would have to trace branch directions and consider only blocks branch is connected to.

ikabod-kee:
Yeah. That sounds reasonable.

ikabod-kee:
> You would have to trace branch directions and consider only blocks branch is connected to.

Actually that sounds pretty performant, as you'd only need to check for connections.

tillpp:
> You would have to trace branch directions and consider only blocks branch is connected to.

optimization now implemented.

tillpp:
~~Bug:  ( **that i will fix in the future, not right now, got no time today. But i write it down to not forget it** ) 
Branches look for branches of the same type while searching for the log,
The problem is that leafy branches are considered a different type of block then normal branches.~~

Fixed

careeoki:
In this situation I would expect the lower half of the tree to remain, since it's still connected to the ground.
thin birch
Also, placing a block next to these just kills the tree.

careeoki:
By the way, you missed `cactus_arm`, `palm/branch`, and `candy_cane/branch`

tillpp:
> In this situation I would expect the lower half of the tree to remain, since it's still connected to the ground. Also, placing a block next to these just kills the tree.

Yeah this is caused by the same thing that caused the little trees to decay.
Again, I think the only viable Solution for this I see, is to edit all structures with branches but no logs in them and to set them to non-decayable, (if we want to maintain the same flexibility for addOns)
I don't exactly know where all the non-SBB files are.
grafik

tillpp:
> candy_cane/branch

i think they should never decay, because they have no log Block that they can connect too

Edit: same for palm
Edit2: added candy_cane decay and palm decay,see below why

tillpp:
Hmm, we could also change the algorithm such that for branches it doesn't look for logs anymore, but for any solid blocks instead.
(And i think this was probably the implicit expectation you had in mind)

And i think i even like that a little bit more.

But that also means loosing some flexibility for addOns makers. (They can not decide which block type should prevent decaying for branches, but on the otherhand, that's not that important anyway, since the branch connectivy already holds that information )

tillpp:
We still got issues for large thin birch trees. 
Because the ground is too far away from the branches on top of the tree, any update above 5 blocks above grounds would trigger a branch decay.
Which means, their SBB has to manually be set to undecayable branches. (This can be done in a seperate PR)

MiniBaker007:
> You would have to trace branch directions and consider only blocks branch is connected to.

I don't know if this is what @Argmaster meant here, but this is how I understood it, and what I think would be a nicer approach:

Branches can (or maybe must) depend on a neighbor. So, in the branch rotation data, in addition to the 6 bits required for each possible connection, there would be 3 additional bits* for defining which one of its neighbors a branch depends on, if any. A valid neighbor for a branch to depend on would be either a full block or a branch which connects to it.

With this approach, decaying branches would just require a simple local update rule. You could allow branches to have no dependency and thus never decay, although I think that would just make things unnecessarily complicated, without much benefit.

_*3 bits for 2^3 = 8 possibilities; we need 6 or 7._

tillpp:
> I think that would just make things unnecessarily complicated

Yup
and in addition: branches can depend on more then one direction. (for example: the branches that spawn at the ground, without leaves and trees in frosty biomes)

MiniBaker007:
> the branches that spawn at the ground, without leaves and trees in frosty biomes

I'm not sure what exactly you're referring to, and I don't really understand how a branch would need to depend on more than one neighbor, unless you had two branches coming together and merging into one?

Edit: If you mean like the dead branches on the ground, I'm not sure exactly how they are generated, but I suppose it would be a good example of where having no dependencies (i.e. never decay) would be desirable.

tillpp:
> > the branches that spawn at the ground, without leaves and trees in frosty biomes
> 
> I'm not sure what exactly you're referring to, and I don't really understand how a branch would need to depend on more than one neighbor, unless you had two branches coming together and merging into one?

this: ( in this case i found one that had 2 connection points to the ground, but there are bigger ones)
grafik

MiniBaker007:
> Edit: If you mean like the dead branches on the ground, I'm not sure exactly how they are generated, but I suppose it would be a good example of where having no dependencies (i.e. never decay) would be desirable.

Editted my comment just before I saw your response.

tillpp:
yeah, anyway way to complicated without any benefit,

tillpp:
Ok, just to clarify: 
There are alternative approaches to implementing this.

And to keep this PR clean of any of those alternatives that would require a >>**complete rewrite**<< of this PR,
i would like you to discuss it here:

https://github.com/PixelGuys/Cubyz/issues/2331

To keep this PR only about this implementation.
And if we find a better idea for implementation, we create a new PR for that and close this one or keep them both alive for comparisons.  (I will rename this branch soon, to differentiate them)
That way, it is also less confusing, which comments refer to the old and which one to the new implementation.

tillpp:
I think the Merge might have broken some stuff,

going to fix that when i have time ,not now

src/callbacks/block/server/decay.zig
```diff
@@ -7,17 +7,32 @@ const vec = main.vec;
...
 		result.decayReplacement = main.blocks.parseBlock(blockname);
 	} else result.decayReplacement = main.blocks.Block.air;
-
+	// prevention
+	result.prevention = .{};
+	if(zon.getChildOrNull("prevention")) |tagNames| {
+		if(tagNames == .array) {
+			for(tagNames.array.items) |value| {
+				const tagName = value.as(?[]const u8, null) orelse continue;
+				result.prevention.append(main.globalAllocator, main.Tag.find(tagName));
```
IntegratedQuantum: Please use the worldArena, then you don't need a deinit function.

assets/cubyz/blocks/candy_cane/branch.zig.zon
```diff
@@ -1,5 +1,8 @@
 .{
 	.tags = .{.choppable, .wood},
+	.onUpdate = .{
+		.type = .Decay,
+	},
```
IntegratedQuantum: I'm not sure if they should decay. They are really not like normal trees.

assets/cubyz/blocks/glimmergill_branch.zig.zon
```diff
@@ -1,5 +1,8 @@
 .{
-	.tags = .{.choppable, .cuttable, .mushroom},
+	.tags = .{.choppable, .cuttable, .mushroom, .branch},
+	.onUpdate = .{
+		.type = .Decay,
+	},
```
IntegratedQuantum: same for all mushrooms

assets/cubyz/blocks/baobab_branch.zig.zon
```diff
@@ -1,5 +1,8 @@
 .{
-	.tags = .{.choppable, .wood},
+	.tags = .{.choppable, .wood, .branch},
+	.onUpdate = .{
+		.type = .Decay,
```
IntegratedQuantum: please follow the naming convention: It should be `decay`

IntegratedQuantum (changes requested):

assets/cubyz/blocks/candy_cane/branch.zig.zon
tillpp: yeah, I only did this, because carrie requested it: https://github.com/PixelGuys/Cubyz/pull/2334#issuecomment-3572398718

tillpp: @careeoki what do you think now?

tillpp:
Don't merge this jet, there are two issues:

first: @Argmaster s optimization https://github.com/PixelGuys/Cubyz/pull/2334#issuecomment-3569107279 that i implemented in https://github.com/PixelGuys/Cubyz/pull/2334/commits/b42d73e6f2dd7788e4a374cb52e227bedd0fc8a5 is in principle good, but it works under the assumptions, that branches are not faulty. Like for example issue: https://github.com/PixelGuys/Cubyz/issues/2368 . Because of that, branches only decay half way or never.

second: when breaking a thin birch tree, only the middle part decays, not the branches on top, i haven't figured out yet why, investigation on going.. (or when i have time),

assets/cubyz/blocks/candy_cane/branch.zig.zon
careeoki: They don't have to decay if we don't want them to. My main concern was palm branch and cactus arms

tillpp:
Found the cause of issue 2:
The trunk of the thin birch trees are connected at the top to the leaves. And since leaves are full blocks, the branches don't decay, a side effect of : https://github.com/PixelGuys/Cubyz/pull/2334#issuecomment-3568821479 

> I think that branches should only decay if they aren't connected to a block, so a little sapling wouldn't wither away if you just placed a block next to it. Plus free realistic expectations!

I see two possible solutions:
1. I remove @ikabod-kee idea and revert back to my original idea. This would also solve the first issue. but we would have to tell small bushes SBBs to use undecayable branches. 

 or 
 2.letting this be a problem, that future PRs can solve by editing all SBB and deconnecting all branches from leaves.

what do you guys think? should i use solution 1 or 2

careeoki:
> letting this be a problem, that future PRs can solve by editing all SBB and deconnecting all branches from leaves.

I don't think this _is_ a problem, you can't connect branches to leaves (as they are transparent). The only reason this happens on the thin birch is that the `simple_tree` structure is using `birch_branch:0b11` to "illegally" connect to leaves.

image


We can fix the trees by changing the `top` to this state: (I don't know what value this state is)
(line 41 in forest/thin_birch)

image

tillpp:
> Don't merge this jet, there are two issues:

This is ready for merge or rereview, btw
(or should i say re²view)

assets/cubyz/blocks/candy_cane/branch.zig.zon
IntegratedQuantum: Then remove them please. candy cane is not even close to the other tree types in that the top part is made of undecayable materials.

IntegratedQuantum: Also @careeoki what about the mushroom branches

mods/cubyz/rotation/branch.zig
```diff
@@ -33,11 +33,12 @@ const HashMapKey = struct {
 		return std.mem.eql(u8, val1.shellModelId, val2.shellModelId);
 	}
 };
-pub const BranchData = packed struct(u6) {
+pub const BranchData = packed struct(u7) {
 	enabledConnections: u6,
+	placedByHuman: u1,
```
IntegratedQuantum: Please use a `bool`

mods/cubyz/rotation/branch.zig
```diff
@@ -33,11 +33,12 @@ const HashMapKey = struct {
...
 
 	pub inline fn init(blockData: u16) BranchData {
-		return .{.enabledConnections = @truncate(blockData)};
+		return .{.enabledConnections = @truncate(blockData), .placedByHuman = @truncate(blockData >> 6)};
```
IntegratedQuantum: How about a `@bitCast(@truncate(blockData))`?

mods/cubyz/rotation/branch.zig
```diff
@@ -352,7 +353,8 @@ pub fn generateData(
 		const targetVal = ((!neighborBlock.replacable() and (!neighborBlock.viewThrough() or canConnectToNeighbor)) and (canConnectToNeighbor or neighborModel.isNeighborOccluded[neighbor.?.reverse().toInt()]));
 		currentData.setConnection(neighbor.?, targetVal);
 
-		const result: u16 = currentData.enabledConnections;
+		currentData.placedByHuman = 1;
+		const result: u16 = @as(u7, @bitCast(currentData));
```
IntegratedQuantum:
```suggestion
		const result: u7 = @bitCast(currentData);
```

mods/cubyz/rotation/branch.zig
```diff
@@ -375,7 +377,7 @@ pub fn updateData(block: *Block, neighbor: Neighbor, neighborBlock: Block) bool
 		currentData.setConnection(neighbor, false);
 	}
 
-	const result: u16 = currentData.enabledConnections;
+	const result: u16 = @as(u7, @bitCast(currentData));
```
IntegratedQuantum:
```suggestion
	const result: u7 = @bitCast(currentData);
```

src/callbacks/block/server/decay.zig
```diff
@@ -26,7 +38,14 @@ fn getIndexInCheckArray(relativePosition: Vec3i, checkRange: comptime_int) usize
 	const arrayIndexZ = relativePosition[2] + checkRange;
 	return @as(usize, @intCast((arrayIndexX*checkLength + arrayIndexY)*checkLength + arrayIndexZ));
 }
-fn foundWayToLog(world: *Server.ServerWorld, leaf: Block, wx: i32, wy: i32, wz: i32) bool {
+fn isDecayPreventer(self: *@This(), log: Block) bool {
```
IntegratedQuantum:
```suggestion
fn preventsDecay(self: *@This(), log: Block) bool {
```

src/callbacks/block/server/decay.zig
```diff
@@ -7,15 +7,27 @@ const vec = main.vec;
 const Vec3i = vec.Vec3i;
 const ZonElement = main.ZonElement;
 const Server = main.server;
+const Branch = main.rotation.list.@"cubyz:branch";
 
 decayReplacement: blocks.Block,
+prevention: main.ListUnmanaged(main.Tag) = .{},
```
IntegratedQuantum: We don't need resizability, this should be a const slice.

src/callbacks/block/server/decay.zig
```diff
@@ -43,21 +62,30 @@ fn foundWayToLog(world: *Server.ServerWorld, leaf: Block, wx: i32, wy: i32, wz:
 	queue.pushBack(Vec3i{0, 0, 0});
 	checked[getIndexInCheckArray(Vec3i{0, 0, 0}, checkRange)] = true;
 
+	const branchRotation = main.rotation.getByID("cubyz:branch");
+	const sourceIsBranch = leaf.mode() == branchRotation;
```
IntegratedQuantum, line 71: This should be done during loading instead of doing this on every block update.

src/callbacks/block/server/decay.zig
```diff
@@ -77,14 +105,18 @@ pub fn run(self: *@This(), params: main.callbacks.ServerBlockCallback.Params) ma
...
 	} else {
-		std.log.err("Expected {s} to have cubyz:decayable rotation", .{params.block.id()});
+		std.log.err("Expected {s} to have cubyz:decayable or cubyz:branch as rotation", .{params.block.id()});
 	}
 
 	if(Server.world) |world| {
 		if(world.getBlock(wx, wy, wz)) |leaf| {
 			// check if there is any log in the proximity?^
-			if(foundWayToLog(world, leaf, wx, wy, wz))
+			if(foundWayToLog(self, world, leaf, wx, wy, wz))
```
IntegratedQuantum:
```suggestion
			if(self.foundWayToLog(world, leaf, wx, wy, wz))
```

src/callbacks/block/server/decay.zig
IntegratedQuantum: Same for the rotation mode checks below.

IntegratedQuantum (changes requested):

mods/cubyz/rotation/branch.zig
tillpp: I can do `return @bitCast(@as(u7,@truncate(blockData)))`, otherwise the compiler is complaining

mods/cubyz/rotation/branch.zig
tillpp: oh didnt know that was allowed in a packed struct.
fixed

src/callbacks/block/server/decay.zig
tillpp: The reason i did this this way, was because you said it was planned to make them compttime soon anyway?
But i can also change it.
done

src/callbacks/block/server/decay.zig
tillpp: done

assets/cubyz/blocks/candy_cane/branch.zig.zon
tillpp: I didnt know what candy canes even looked like,
done.

All that is left now is, how mushrooms branches should or should not decay. so need to wait for caries answer

assets/cubyz/blocks/glimmergill_branch.zig.zon
tillpp: I dont think it makes sense to treat all mushrooms like mushroom branches.

assets/cubyz/blocks/candy_cane/branch.zig.zon
careeoki: They shouldn't decay, they don't function like regular trees (no corresponding leaf block)

tillpp: OK done.
(i would have thought breaking the stem would break the top part, but i am also fine not doing that)

src/callbacks/block/server/decay.zig
IntegratedQuantum: I meant the bool variable itself (leaf.mode() means loading another cache line, and another comparison from the ==), not just the rotation mode, which yes will be loaded at comptime.

src/callbacks/block/server/decay.zig
```diff
@@ -7,15 +7,34 @@ const vec = main.vec;
 const Vec3i = vec.Vec3i;
 const ZonElement = main.ZonElement;
 const Server = main.server;
+const Branch = main.rotation.list.@"cubyz:branch";
 
 decayReplacement: blocks.Block,
+prevention: []const main.Tag,
+branchRotation: *const main.rotation.RotationMode,
+decayRotation: *const main.rotation.RotationMode,
```
IntegratedQuantum: Should be bools, see other comment

src/callbacks/block/server/decay.zig
```diff
@@ -7,15 +7,34 @@ const vec = main.vec;
...
 		result.decayReplacement = main.blocks.parseBlock(blockname);
 	} else result.decayReplacement = main.blocks.Block.air;
-
+	// prevention
+	result.prevention = &.{};
+	if(zon.getChildOrNull("prevention")) |tagNames| {
+		if(tagNames == .array) {
+			var prevention = main.worldArena.alloc(main.Tag, tagNames.array.items.len);
+			for(tagNames.array.items, 0..) |value, index| {
+				const tagName = value.as(?[]const u8, null) orelse @panic("Invalid tagName.");
```
IntegratedQuantum: Don't panic on user error. Instead print an error and try to continue.

src/callbacks/block/server/decay.zig
```diff
@@ -7,15 +7,34 @@ const vec = main.vec;
...
+	// replacement
 	if(zon.get(?[]const u8, "replacement", null)) |blockname| {
 		result.decayReplacement = main.blocks.parseBlock(blockname);
 	} else result.decayReplacement = main.blocks.Block.air;
-
+	// prevention
+	result.prevention = &.{};
+	if(zon.getChildOrNull("prevention")) |tagNames| {
+		if(tagNames == .array) {
+			var prevention = main.worldArena.alloc(main.Tag, tagNames.array.items.len);
```
IntegratedQuantum:
```suggestion
			const prevention = main.worldArena.alloc(main.Tag, tagNames.array.items.len);
```

IntegratedQuantum (changes requested):

src/callbacks/block/server/decay.zig
tillpp: wait you want me to set sourceIsBranch during init()? but init has no access to the current mode()

src/callbacks/block/server/decay.zig
tillpp: branchRotation is also used to check if the neighbours do have branch information

src/callbacks/block/server/decay.zig
IntegratedQuantum: Oh I see I thought it could, then scratch that (and please also revert the moving of the rotation.getById()).

src/callbacks/block/server/decay.zig
```diff
@@ -25,11 +25,16 @@ pub fn init(zon: ZonElement) ?*@This() {
 	if(zon.getChildOrNull("prevention")) |tagNames| {
 		if(tagNames == .array) {
 			const prevention = main.worldArena.alloc(main.Tag, tagNames.array.items.len);
-			for(tagNames.array.items, 0..) |value, index| {
-				const tagName = value.as(?[]const u8, null) orelse @panic("Invalid tagName.");
+			var index:usize = 0;
```
IntegratedQuantum: Please don't reimplement a ListUnmanaged, if you can use a ListUnmanaged.
Just use initCapacity and appendAssumeCapacity.

tillpp: Ah ok didnt know ListUnmanaged was that simple.
fixed

src/callbacks/block/server/decay.zig
tillpp: done

IntegratedQuantum:
Next up I think item drops would be pretty important, branches definitely should drop

PR #2345, opened 2025-11-23
Tweak grass block palette

careeoki:
Wasn't totally happy with the previous version, I made it too soft.
image


du82:
it looks too green, maybe less green makes the game feel more legit / better colors / graphics, not like the 50000 on the playstore haha

PR #2347, opened 2025-11-24
Migrate branches

careeoki:
Move all the branches to `cubyz:branch/{type}`

PR #2348, opened 2025-11-24
Replace branches with logs in LODs

### LINKED ISSUE #2340: Make Branches Appear as Full Blocks in LODs
### Description

Currently, they don't.

careeoki:
Requires #2347 (i think im doing this right)

Fixes #2340


careeoki:
I'll make another PR after #2347 I just don't want to do this right now

PR #2371, opened 2025-12-03
Winter wonderland rare biome

careeoki:
The winter wonderland is a snowy biome with giant candy canes.
image
The forest subbiome is dense with candy canes, and has no pine trees.

image

While exploring, you might see some snow-snales.
image


ikabod-kee:
My thoughts:
1. Maybe some of these "trees" could have multiple branches like this?
image
2. I think adding more to the mountain parameter can help it from feeling flat. Maybe have a mountain sub-biome for some cool snow/ice spikes.
3. Try using ground patches and/or boulders of permafrost to give it a little texture?
4. I know this is nitpicky, but the candy canes don't have roots. I think it would be really funny to imply that candy canes are a naturally occurring plant in this place.

du82:
we need a fork of cubyz without western holidays and cultural things in it, just a base game

tillpp:
> we need a fork of cubyz without western holidays and cultural things in it, just a base game

What if we add holidays of eastern and other cultures too?

careeoki:
>  1. Maybe some of these "trees" could have multiple branches like this?

I want to keep the traditional candy cane shape.


>  2. I think adding more to the mountain parameter can help it from feeling flat. Maybe have a mountain sub-biome for some cool snow/ice spikes. 3. Try using ground patches and/or boulders of permafrost to give it a little texture? 

Increased mountainness and added an icicles subbiome.
image
I would like to keep it mostly flat, like fresh snowfall. (ideal sledding weather)

> 4. I know this is nitpicky, but the candy canes don't have roots. I think it would be really funny to imply that candy canes are a naturally occurring plant in this place.

They are natural. I imagine they have small fibrous roots, like sugarcane.

assets/cubyz/biomes/rare/winter_wonderland/base.zig.zon
```diff
@@ -0,0 +1,43 @@
+.{
+	.chance = 0.02,
```
IntegratedQuantum: all rare biomes should have a chance of 0.01

IntegratedQuantum:
I love the idea of snow sculptures, so here are some more suggestions:

I'd suggest to give them coal blocks as eyes, also how about snail and a slug versions?
Screenshot at 2025-12-04 18-09-51

Also the torso and neck of the snale are bothering me a bit. I think something like this would be more accurate to the actual model:
Screenshot at 2025-12-04 18-13-12

IntegratedQuantum (changes requested):
In my opinion there are too many canes in the base biome. I'd suggest to reduce their chances a bit.

careeoki:
> I'd suggest to give them coal blocks as eyes, also how about snail and a slug versions? 

Sure, will do

> Also the torso and neck of the snale are bothering me a bit. I think something like this would be more accurate to the actual model:

They're not supposed to be perfect sculptures, more like snowmen which are just balls of snow stacked ontop of eachother. So they are more bottom-heavy than the actual snale.

careeoki:
> Thoughts on rounding the corners of the candy canes with branches like so?

I like keeping natural things unchiseled to create a better distinction of between natural and constructed things. Chiseled things can stick out amongst the regular full-block terrain.

careeoki:
> In my opinion there are too many canes in the base biome. I'd suggest to reduce their chances a bit.

Yeah, true, done.

assets/cubyz/biomes/rare/winter_wonderland/base.zig.zon
careeoki: Done. tuften fields and knitted forest both have a chance of 0.02 by the way

du82:
> > we need a fork of cubyz without western holidays and cultural things in it, just a base game
> 
> What if we add holidays of eastern and other cultures too?

I would rather there just not be any holiday themed things, Minecraft doesn't include them

assets/cubyz/biomes/rare/winter_wonderland/base.zig.zon
IntegratedQuantum: Ah, I guess we should fix them as well then.

IntegratedQuantum:
> Minecraft doesn't include them

What about the present box texture for chests during christmas?

Either way I don't get your problem here. This is a super rare biome and it is just remotely related to christmas if at all. Most people from other cultures will probably just think of it as another curiosity in the world of Cubyz, which is exactly what we want.

PR #2373, opened 2025-12-05
Fix server thread sleeping for too long

MiniBaker007:
The calculation for how long the server thread should sleep was broken by #2009. This fixes it.


IntegratedQuantum (approved):
It is truly amazing how many things I managed to mess up in that PR.

PR #2377, opened 2025-12-06
Thin tree decay

tillpp:
After a talk with @careeoki : https://discord.com/channels/443805812390100992/1445170917339824188/1445172933051027627

grafik

We came to the conclusion that thin birch trees, should break differently then the usual decay algorithm. 
Currently, when you update the upper part of the trunc by placing a block next to it, it will decay since it is too far away from the ground. 
In order to not make the decay checking more expensive by expanding the checking area or making the thin tree branches undecayable, we decided do handle them separately:

Every block of thin tree only checks if it is connected to the block under it.

### DIFF
--- a/mods/cubyz/rotation/branch.zig
+++ b/mods/cubyz/rotation/branch.zig
@@ -36,7 +36,7 @@ const HashMapKey = struct {
 pub const BranchData = packed struct(u8) {
 	enabledConnections: u6,
 	placedByHuman: bool,
-	thinTree:bool,
+	thinTree: bool,
 
 	pub inline fn init(blockData: u16) BranchData {
 		return @bitCast(@as(u8, @truncate(blockData)));


tillpp:
request for label

IntegratedQuantum:
I don't think this is good enough. The trunc is not always straight. Take the palm trees for example:
Screenshot at 2025-12-07 12-16-15

I think the root problem is that sometimes branches are placed in place of logs, and I think maybe they also should behave more like logs.
So how about just making the branches of the main stem undecayable and treat undecayable branches like logs?

tillpp:
The Palm trees should perfectly work fine with the old algorithm.

With thin tree i mean only the trees that are made out of branches and are longer then 5 blocks.
(which there is only one right now: the thin birch tree, (see screenshot) )
@careeoki wanted to have them to be completly decayable

IntegratedQuantum:
Yes, the current palm trees work, it's just intended to be an example that one might make thin stems not straight. It's a reasonable edge case that we should be prepared for in my opinion.

If we want to make them fully decayable then I think the best solution is to increase the search radius. Since branches only follow connected branches it might not be that expensive (though of course it should be measured in some worst-case scenarios) to go all the way to a 32 block radius.

Either way in my opinion the player should not be able to chop down a 20 block high thin tree in a single hit.

tillpp:
> Either way in my opinion the player should not be able to chop down a 20 block high thin tree in a single hit.

I am just following orders,
what do you think @careeoki ? 

scratch the idea and make it all just undecaying instead?

careeoki:
> > Either way in my opinion the player should not be able to chop down a 20 block high thin tree in a single hit.
> 
> I am just following orders, what do you think @careeoki ?

I like being able to chop the thin trees quickly, (fast harvest, but less wood) but Quantum has final say

I would be fine with undecaying trunks instead.

IntegratedQuantum:
Alright I think since mods will probably want to have instant decay, I think we should at least support it in the engine.
To make it happen I think the best solution would be to increase the search range on branches by making it a dynamic parameter of the callback. I think this is better than the proposed solution since it doesn't require additional work when creating the models.

My main problem with instant decay of all branches is that it might make it too easy for the player to obtain wood, but I guess we can consider these balancing problems when they actually become a problem. Might be easy to fix by just making branches worth less planks.

tillpp:
> I think this is better than the proposed solution since it doesn't require additional work when creating the models.

closing this PR then

PR #2387, opened 2025-12-08
Chain block

careeoki:
Inspired by #2383 (but separate because it's going to be blocked for awhile)

Edit: Added subtle reflectivity to all the "cast iron" blocks, and tweak the lantern item textures

Adds a chain block, crafted with 1 iron ingot -> 2 chain.

And yes, it's 3D! Egads!!!!!
image
image

It would probably be the most detailed model in the game. If you're worried about it looking _too_ detailed, here's it in a pile of other unique models to show that it's not that bad.
image

I optimized the model nicely, it only has 22 faces. Essentially it's two big cubes to make the outer shell, with flat planes intersecting
to create the insides and tops of the chain links. For comparison, if I just used cubes for each chain link it would be 60+ faces or something


du82:
I really like this! maybe the rope in the other commit can be thinner or more detailed

OneAvargeCoder193:
how many quads does the chain model use?

careeoki:
> how many quads does the chain model use?

22. I talked about it at the end of the description.

OneAvargeCoder193:
Ah i see it now

ikabod-kee:
I'll be honest, this breaks the established art style of the game for me. 

image
Why do the chains get to be 3D while the hook on the lanterns remain 2D?

careeoki:
> Why do the chains get to be 3D while the hook on the lanterns remain 2D?

The hook on the lantern is a much thinner piece of metal, while the chain is large and thick (like something you'd see at a dock)
image
image

We can use both 3D elements and 2D planes to create a greater variety of shapes.

ikabod-kee:
image

IntegratedQuantum:
one day we will also be able to optimize away some of the extra faces in LOD0.5

One question though: Have you thought about how chains would combine? e.g. crossing multiple chains together?

IntegratedQuantum:
I think it might also be a good idea to reconsider how these interact with lanterns or other blocks.
Screenshot at 2025-12-11 17-52-29

Maybe ropes and chains should behave more separately? Maybe being completely independent of the voxel grid, like the planned ziplines? Then lanterns could an attachment point where you could attach the chain or rope to.

careeoki:
> One question though: Have you thought about how chains would combine? e.g. crossing multiple chains together?

Yeah, I'd like it if you could connect chains (and ropes) together. The model would still work with chains crossing together.


> I think it might also be a good idea to reconsider how these interact with lanterns or other blocks.

Yes, the lanterns are tricky. One solution would be to add an extra state where the lantern is directly attached to the ceiling (without the small hook.) Hanging signs would benefit from this as well 


> Maybe ropes and chains should behave more separately? Maybe being completely independent of the voxel grid

Maybe. I feel it would be reasonable for ropes to work like that, since they're more freeform, but chains need to have the individual chain links line up (and not get abruptly cut off)
(I'm imagining this like ziplines, where they can have arbitrary angles)

I think i would prefer something  like terraria, where you can connect up the ropes/chains as blocks. And then add special cases for things like lanterns.

Bloxxel64:
Perhaps we should consider the idea of a block entity implementation? I feel like these ideas could benefit from that.

IntegratedQuantum:
Alright, I think we'll keep them as full blocks for now then, but I would like to rethink this once we have ziplines implemented.

> but chains need to have the individual chain links line up (and not get abruptly cut off)

That should not be an issue, we can just make the chain longer until it has the right number of links.

PR #1930, opened 2025-10-08
Jungle

Argmaster:
This proposal contains a proposal of jungle improvements.
Additionally it contains monstera plant created @careeoki which is heavily used in new jungle.

image

image

image

PR #2396, opened 2025-12-13
Add vine decay

Argmaster:
Adds makeshift vine decay to avoid having flying vines after #1930 pull request is merged.

Related-to: #1930


IntegratedQuantum (changes requested):
I don't think this is the behavior we want.
Also instead of adding more specific behavior to this one function I'd suggest to add a new callback for vine decay (or more generally hanging decay)

Vines should only decay if the block above it is not a vine of the same type and not a solid block and has no model with downwards faces.

PR #2438, opened 2025-12-27
Huge hills

careeoki:
Do you want even more hills? Not really? Well, too bad!

Huge hills are much larger than large hills, but still not reaching the heights of mountain biomes. They provide some very dynamic terrain.

image
image

**Screenshots below lack some of the details from later commits, but still show off the terrain:**

image
image


Crepestrom:
could you make the colder area a bit more rocky with some boulders to make it a bit different

careeoki:
> could you make the colder area a bit more rocky with some boulders to make it a bit different

Sure, done. Also tweaked the temperate one while I was at it.

IntegratedQuantum (approved):

PR #2439, opened 2025-12-27
Plateau subbiomes

careeoki:
Subbiomes for grasslands and hills biomes, which gives their parent biomes a cool unique look.

image

image
There's also a chance of a plateau having another plateau on top of it!
image
(Can't find a temperate one yet...)


Crepestrom:
could you make the platue have a very thin transition subiome so that its not a sheer cliff but a cliff thats rounded off at the bottom

careeoki:
> could you make the platue have a very thin transition subiome so that its not a sheer cliff but a cliff thats rounded off at the bottom

Tried it, it causes problems:

- using subbiomes, you either get no transition at all or a big lump of the transistion biome in an undesirable spot. (so it's not feasible for it to be thin)
- you also lose out on variable biome sizes doing it this way
- transitionBiomes are exactly what I would need here, but they don't seem to work on subbiomes (afaik)

Personally I don't mind the sheer cliffs, I think they look neat, but I know they can look silly sometimes. Ideally yeah, I would have it slope up towards the sheer cliff.

Crepestrom:
ah could you see how mountains do it
i think they solve that kinda of issue with weird sections though i may be wrong

careeoki:
> ah could you see how mountains do it i think they solve that kinda of issue with weird sections though i may be wrong

The mountains work because they're at a much larger scale, and it doesn't matter for them if they are a bit lumpy.

assets/cubyz/biomes/plateau/_defaults.zig.zon
```diff
@@ -0,0 +1,43 @@
+.{
+	.interpolation = .none,
+	.properties = .{},
+	.minHeightLimit = 7,
+	.minHeight = 50,
+	.maxHeight = 70,
```
IntegratedQuantum: I think they stick out too much. Also you can use the `keepOriginalTerrain` trick to make them more consistently sized.
e.g. the following would set the height to 5-10 blocks above the remaining terrain:
```zig.zon
	.minHeight = 500,
	.maxHeight = 1000,
	.keepOriginalTerrain = 0.99,
```

assets/cubyz/biomes/plateau/_defaults.zig.zon
```diff
@@ -0,0 +1,43 @@
+.{
+	.interpolation = .none,
```
IntegratedQuantum, line 2: Have you considered using the `interpolationWeight` to control the strength of the edge?

IntegratedQuantum (changes requested):

assets/cubyz/biomes/plateau/_defaults.zig.zon
careeoki: Sure, I changed it to use this method.

assets/cubyz/biomes/plateau/_defaults.zig.zon
careeoki: I did not know that existed. I smoothed out the edges more.
image

careeoki:
I had to remove the stacked plateaus because I couldn't get the interpolation to work how I wanted. The stacked plateaus always had hard edges, while the base plateau now has smoother edges. Which looked weird.

IntegratedQuantum:
>  had to remove the stacked plateaus because I couldn't get the interpolation to work how I wanted.

If you give the inner plateau a smooth interpolation then you should get the same transition as between the outer plateau and the surrounding terrain.

careeoki:
> If you give the inner plateau a smooth interpolation then you should get the same transition as between the outer plateau and the surrounding terrain.

Oh, yeah, that works.
image

IntegratedQuantum (approved):

PR #2447, opened 2025-12-30
Set Spawn command (continuation of #2374) 

Wunka:
Changes to the PR #2374 from @Vinywar123 : 

- renamed setSpawn command to spawn.
- only the server knows the spawn and gives it the player with the kill sync.
- when the spawn command is run with no parameters the command will return the spawnpoint.
- renamed the playerSpawnPoint variable to spawnPoint.


Crepestrom:
thanks for picking this up

src/game.zig
```diff
@@ -543,8 +543,8 @@ pub const Player = struct { // MARK: Player
 		inventory.placeBlock(selectedSlot);
 	}
 
-	pub fn kill() void {
-		Player.super.pos = world.?.spawn;
```
IntegratedQuantum, line 548: Please also remove this field from the client-side world struct, I don't think it's used anymore.

src/server/command/spawn.zig
```diff
@@ -0,0 +1,47 @@
+const std = @import("std");
+
+const main = @import("main");
+const User = main.server.User;
+
+pub const description = "Get or set the player spawnPoint";
```
IntegratedQuantum: This text is intended for humans, not programmers.
```suggestion
pub const description = "Get or set the player spawn point";
```

IntegratedQuantum (changes requested):

src/game.zig
Wunka: removed. But what I did notice while doing this. Maybe the command should have gobal / world flag so its sets the world spawnPoint instead of only the player one. ~~I could do that in this PR or~~ I will open a issue for that (as I need to focus on different things for like 3 weeks and the to slice refactors are already enough)

src/server/command/spawn.zig
```diff
@@ -0,0 +1,47 @@
...
+		} else if(y == null) {
+			y = num;
+		} else if(z == null) {
+			z = num;
+		} else {
+			source.sendMessage("#ff0000Too many arguments for command /spawn", .{});
+			return;
+		}
+	}
+	if(x == null or y == null) {
```
IntegratedQuantum: Should only trigger when x is null, and needs a separate case for x != null and y == null

IntegratedQuantum (changes requested):

src/server/command/spawn.zig
Wunka: should be done, but cannot test because of #2485

IntegratedQuantum: try it in a new world

Wunka: with different world it worked. But I don't like the wording of the message (I changed it since my comment) maybe just send the usage?

IntegratedQuantum: Yeah, sending the usage would be good.

Also note that with #1425 we would also get more consistent error messages.

Wunka: now its sending the usage

IntegratedQuantum (approved):

PR #2452, opened 2025-12-31
Ruby rebalance

Crepestrom:
Changes the ruby to have more of a purpose; A early game gem

it has easier requirements and a permanent debuff to durability but it only needs 2 precious to activate its 25% damage buff a cheap requirement when players will have less precious resources.
It also now spawns in the gold/silver layers (under -500) higher than diamond to be more easily gained as a early game ore

image

also its old stats are the same eg; durability

### DIFF
--- a/assets/cubyz/blocks/ruby_ore.zig.zon
+++ b/assets/cubyz/blocks/ruby_ore.zig.zon
@@ -24,7 +24,7 @@
 					.restriction = .{
 						.id = .encased,
 						.tag = .precious,
-						.amount = 3,
+						.amount = 2,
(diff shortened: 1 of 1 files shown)


Argmaster:
Hi, those changes don't align with the direction we planned for ruby before: https://github.com/PixelGuys/Cubyz/issues/1276

I am not immediately rejecting this, but be aware that this may be rejected for said reason.

IntegratedQuantum:
We have different plans for rubies, and adding different stats to make it temporarily useful it is not a good idea in my opinion.

PR #2459, opened 2026-01-01
Create a `checkSupportBlocks` block update callback for torch decay instead of handling it through the rotation mode

### LINKED ISSUE #1103: Torches don't drop in survival when their support block is broken.

IntegratedQuantum:
I'm open for better name suggestions.

I considered merging this into the decay callback, but I think it makes more sense as a separate thing since it only looks at direct connections.

In the future this could be extended to include carpet and signs.
@careeoki should lanterns and signs break when their support block is broken?

this is a first step towards #2330 and already demonstrates that this makes it more flexible, since now the addon author can decide how a block behaves when neighbor blocks are broken, instead of depending on the rotation mode code.

fixes #1103

### DIFF
--- a/src/callbacks/block/server/checkSupportBlocks.zig
+++ b/src/callbacks/block/server/checkSupportBlocks.zig
@@ -31,16 +31,9 @@ pub fn run(_: *@This(), params: main.callbacks.ServerBlockCallback.Params) main.
 	var newBlock: Block = params.block;
 
 	if(params.block.mode() == main.rotation.getByID("cubyz:torch")) {
-		var data: main.rotation.list.@"cubyz:torch".TorchData = @bitCast(@as(u5, @truncate(params.block.data)));
-		if(data.center and !neighborSupportive[Neighbor.dirDown.toInt()]) data.center = false;
-		if(data.negX and !neighborSupportive[Neighbor.dirNegX.toInt()]) data.negX = false;
-		if(data.posX and !neighborSupportive[Neighbor.dirPosX.toInt()]) data.posX = false;
-		if(data.negY and !neighborSupportive[Neighbor.dirNegY.toInt()]) data.negY = false;
-		if(data.posY and !neighborSupportive[Neighbor.dirPosY.toInt()]) data.posY = false;
-		newBlock.data = @as(u5, @bitCast(data));
-		if(newBlock.data == 0) newBlock.typ = 0;
+		main.rotation.list.@"cubyz:torch".updateBlockFromNeighborConnectivity(&newBlock, neighborSupportive);
 	} else {
-		std.log.err("Expected {s} to have cubyz:decayable or cubyz:branch as rotation", .{params.block.id()});
+		std.log.err("Expected {s} to have cubyz:torch as rotation", .{params.block.id()});
 	}
 
 	if(newBlock == params.block) return .ignored;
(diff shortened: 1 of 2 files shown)


src/callbacks/block/server/checkSupportBlocks.zig
```diff
@@ -0,0 +1,69 @@
...
+		var data: main.rotation.list.@"cubyz:torch".TorchData = @bitCast(@as(u5, @truncate(params.block.data)));
+		if(data.center and !neighborSupportive[Neighbor.dirDown.toInt()]) data.center = false;
+		if(data.negX and !neighborSupportive[Neighbor.dirNegX.toInt()]) data.negX = false;
+		if(data.posX and !neighborSupportive[Neighbor.dirPosX.toInt()]) data.posX = false;
+		if(data.negY and !neighborSupportive[Neighbor.dirNegY.toInt()]) data.negY = false;
+		if(data.posY and !neighborSupportive[Neighbor.dirPosY.toInt()]) data.posY = false;
+		newBlock.data = @as(u5, @bitCast(data));
+		if(newBlock.data == 0) newBlock.typ = 0;
+	} else {
+		std.log.err("Expected {s} to have cubyz:decayable or cubyz:branch as rotation", .{params.block.id()});
```
Argmaster: This is not acceptable IMO.
This shouldn't be hardcoded.
This behavior is specific to a rotation mode, so it should call into rotation mode and determine output data there.

assets/cubyz/blocks/bolete.zig.zon
```diff
@@ -10,6 +10,9 @@
 	.absorbedLight = 0x010101,
 	.collide = false,
 	.rotation = "cubyz:torch",
+	.onUpdate = .{
```
Argmaster, line 13: Do you intend to add support requirement to foliage of all sorts?

IntegratedQuantum: Yes, I think that's reasonable. For now I'm just preserving status quo though, by giving all torch-rotated blocks this property.

src/callbacks/block/server/checkSupportBlocks.zig
```diff
@@ -0,0 +1,62 @@
...
+
+	for(Neighbor.iterable) |neighbor| {
+		const neighborBlock: Block = main.server.world.?.getBlock(wx +% neighbor.relX(), wy +% neighbor.relY(), wz +% neighbor.relZ()) orelse .{.typ = 0, .data = 0};
+		const neighborModel = main.blocks.meshes.model(neighborBlock).model();
+		neighborSupportive[neighbor.toInt()] = !neighborBlock.replacable() and neighborModel.neighborFacingQuads[neighbor.reverse().toInt()].len != 0;
+	}
+
+	var newBlock: Block = params.block;
+
+	if(params.block.mode() == main.rotation.getByID("cubyz:torch")) {
```
Argmaster, line 33: I think this is a obvious VTable call material.
Otherwise you will have a wall of if statements for everything that require support checks.
After some thinking, I come to a conclusion that calling into **rotation** vtable, as I wanted at first, may not be a good idea, so I guess I will have to let it slide for now.

Argmaster (approved):

src/callbacks/block/server/checkSupportBlocks.zig
IntegratedQuantum: The other reason I kept it outside the VTable is because mods can't put stuff into the VTable either.

It might be worth investigating the usage of a tagged union (with empty fields), which would allow anyone to add new functions to the interface implicitly using an inline switch with `@hasDecl`.

PR #2462, opened 2026-01-03
Optimize light fetching during mesh updates

IntegratedQuantum:
This mainly involves a vectorization efforts, using a @Vector(8, u16) for all internal calculations.

I had some trouble with `mesh.lightingData[1].getValue(x, y, z) ++ mesh.lightingData[0].getValue(x, y, z);` where the compiler wanted to generated an individual instruction for each insertion, but I managed to find a way for it to produce the right assembly.

Overall this gave a ~33% reduction in block placement lag spikes and a ~10% reduction in mesh generation times (at render distance 24):
reduction1
reduction2

Despite all this, the main bottleneck remains the light fetching/interpolation and clearly more effort is going to be needed to optimize this.

### DIFF
--- a/src/renderer/chunk_meshing.zig
+++ b/src/renderer/chunk_meshing.zig
@@ -1,5 +1,6 @@
 const std = @import("std");
 const Atomic = std.atomic.Value;
+const builtin = @import("builtin");
 
 const main = @import("main");
 const blocks = main.blocks;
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
Even though it's far from what I would have wanted in terms of performance, it is a decent improvement and makes the code simpler, so I guess there is no reason not to merge it.

PR #2463, opened 2026-01-03
Jade ore

careeoki:
I've not done all the assets for this because I'd like to gauge interest first.

Jade appears between 800 to -1000. When incased, it increases durability by +150% (not as much as it sounds since durability is the largest of the three tool parameters.)
It is rarer (less veins) than other gems, but since it can be found on the surface, it would probably be a similar rarity to other gems.
You will need at least iron in your pick to mine it.

image
image

- Gives a greater incentive for mining near the surface, and for exploring on the surface, especially mountains
- Damage (from diamond) or swing speed (from amber) isn't very useful for tools like the shover or axe, since most blocks broken with them have low health, so you don't have a reason to do incasing for these tools. However, you always want durability, so jade would give an incentive to do incasing on other tools.
- Durability is the biggest bottleneck i face with tools, especially while building. So it would be nice to have some more durability.

Why jade?
- It has a cool and unique color
- It's a gemstone that people actually used to make tools irl

(also, yes, its a green gem that can spawn in mountains. i swear that wasn't intentional. (at least there's an actual reason to mine this ore))

### DIFF
--- a/assets/cubyz/blocks/jade_ore.zig.zon
+++ b/assets/cubyz/blocks/jade_ore.zig.zon
@@ -28,8 +28,8 @@
 	},
 	.ore = .{
 		.veins = 4,
-		.size = 3,
-		.minHeight = 500,
+		.size = 4,
+		.minHeight = 350,
 		.density = 0.3,
 	},
 	.drops = .{


du82:
I like it a lot!

IntegratedQuantum:
Alright first up, in the range of -1000 to 0 we already have more than enough ores, 8 different ones in total.
Furthermore, I think we have an opportunity here to make a more unique gemstone, that doesn't just rely on encasing. We have some materials that are made obsolete rather quickly at the moment, and I think we could use jade to create some more interesting synergies.

So I propose the following:
Jade spawns in the sky islands #74 and high mountains (maybe 500+ blocks). This would give the players a tease of what's to come when they venture into the sky, and it could be a nice reward for climbing one of those big mountains.
For the tool, I think jade and copper would fit well together, and copper is currently the most useless metal, directly made obsolete by iron. So instead of giving a flat and boring 150% boost, I'd suggest to instead give maybe a +40% boost (additive) for each copper ingot that was used in the tool. This effect should of course still only be active when encased, though I think we could consider reducing it to 3 precious which would allow you to fit it in places where you cannot fit the other gems.

Finally I am somewhat concerned with the fact that we already have a green ore, so I think in order to distinguish itself it would make sense to lean more into the white spectrum of jade colors. Or maybe we could have a mix of white and green gems in the ore, and a mixture of the two colors for the item?

Crepestrom:
we could make the jade ore check if it has less than 4 .metal neighbors to have make non metal materials decent

Argmaster:
> So I propose the following: Jade spawns in the sky islands #74 and high mountains (maybe 500+ blocks). 

I agree with jade going up, not down. Not sure about the 500+, depends how high the mountains go. We have to ensure that going to the top actually gives you jade fairly reliably.

> For the tool, I think jade and copper would fit well together, and copper is currently the most useless metal, directly made obsolete by iron. 

I would discourage creating a new material just to save the old one. It doesn't solve the problem, it adds a new one.

> I'd suggest to instead give maybe a +40% boost (additive) for each copper ingot that was used in the tool.

Did you do the math, how this will compare to current best pickaxe?
Honestly, I am not sure how much durability it would have to have for me to sacrifice the time a best pickaxe saves me.

---

Now, I will be throwing some ideas from the top of my head like I often do.

**Maybe we should scrap the vertical progression completely?**
If we were to make everything minable by everything players would be driven by convenience, not necessity. So we remove the hard walls and we never punish players for choosing poor combinations of materials. Let them go `wood -> stone -> iron` or even `wood -> stone -> diamond`. So what? None of those materials is actually good on their own. Getting them with stone pickaxe is a pain, so if they really want to? No one digs 1k blocks down by mistake, especially not missing every copper and iron ore on the way.
Instead, we could focus on making the materials localized in 3D space, instead of 2D. So there are biomes which have specific combinations of ores, so player may never encounter some of them and its fine, because they won't need them.

Now the question is, how to make players seek what they missed, for reasons other than curiosity?

An idea that came to my mind is chasing an **optimum**. Blocks are broken slower, when tool they are broken with has stats (damage-like) outside of their breaking optimum. I could deliver some creative logical explanation for this behavior, but I won't, I leave this as a homework for a reader.
The trick is to advertise this correctly to the players.
Instead of operating on numbers, which are hard to understand, we would present players with 3 stats of blocks:
`hardness`, `weight` and `stickiness` (or whatever other metrics you think fit better). Each tool player ever crafts, can have one of the N predefined values for each of the 3 stats.
So, I craft a pickaxe, and it will be described as:

```
Good at extracting hard materials.
Average at extracting heavy materials.
Poor at extracting sticky materials.
```

The spectrum for each could be "Very Poor", "Poor", "Average", "Good", "Exceptional", "God like"

Then, a slate would be described as:

```
Hardness: High
Weight: High
Stickiness: Low
```

With the tiers being eg. "Very Low", "Low", "Average", "High", "Very High"

The rest is just a matter of making a heatmap.
Tools which are out of sync with material they are breaking will be punished by longer breaking time. Making the highest damage pickaxe is no longer a jack of all trades.
Tool crafting would remain numeric, but block breaking math would be quantized. TBD

The pros first:
- Its much more human friendly
  - gives good reference points - if you see that stone has high hardness, you can guess quite precisely every material
  - it has a strong base in the language, making it much more intuitive
- It is much, much easier to balance, because there is a limited grid of interactions. Moreover, you can ensure every basic interaction can be fine tuned to be convenient, because there is limited number of them.
- It reduces the immense complexity of procedural crafting to something on a level equivalent to the rest of the game. This game is a toy, not a calculus lecture.
- It solves the swing time problem, as now swing time is not one of balanced properties.
- *Material will be forced into predefined categories for each property,* so they will have to guarantee their uniqueness through modifiers instead of raw numbers

The cons:
- Material will be forced into predefined categories for each property, *so they will have to guarantee their uniqueness through modifiers instead of raw numbers*
- We lose some of the perfectionist optimization aspect of crafting system, since you no longer can see the raw numbers and the resolution of the system is greatly reduced.

Why not just use numbers?
Because numbers suck. What we have here, is a problem surprisingly many games already had: If there is only one dimension viable to optimize, there will be one best solution available. In our case we cannot give up durability nor swing speed, because the tools immediately lose viability (We do now that no? We have made some attempts!), hence a new optimal tool is born. What we can do, is create multiple axes to optimize and let the hell lose, so players have to comply and find the right tool for the job.

careeoki:
Yeah, jade only being higher up makes sense. If we had non-uniform ore distribution I would argue for it starting at 0 veins at +100 and then maxing out at 2 veins at +500 so you can still occasionally find some lower down, but if you want more you have to go up higher.

I like the copper idea, copper is not really that much worse than iron, but obviously iron still wins every time. So jade could give it some extra life. 
But I also understand what ARG said, that it's not actually solving the root problem. Copper might benefit from some modifier gimmick of it's own (definitely based on it's conductivity in some manner)

> Finally I am somewhat concerned with the fact that we already have a green ore

We have 2 yellow ores, and it hasn't really caused any problems (especially with the new gold ore texture which is much more distinct)

Jade and uranium are pretty distinct in value (as in lightness/darkness) and shape (jade is round, uranium is sharp)

image
image

IntegratedQuantum:
> I would discourage creating a new material just to save the old one. It doesn't solve the problem, it adds a new one.

Well, we kind of have the same pattern with silver/gold encasing, silver and gold would be completely useless without it, but encasing gives them a purpose.
Also note that this PR is creating jade as a durability gem, I'm just proposing to add a use-case for copper on top.

Also I think we should have more behaviors like this, modifiers that interact with different materials and placement make the crafting system more fun in my opinion, and since they are in the tooltip they are also more predictable than stats that follow some black-box function in the background (which is kind of a problem of the other stats, how long does it take noobs to figure out that the tip acts differently?).

> Did you do the math, how this will compare to current best pickaxe?
Honestly, I am not sure how much durability it would have to have for me to sacrifice the time a best pickaxe saves me.

With or without copper, you'd never sacrifice one of the gem slot in your pickaxe for jade.
Keep in mind that this PR is motivated by the other tools.

As for the +40%, that's just a rough idea, on implementation we'd of course do more testing to find a better value.

> Let them go wood -> stone -> iron or even wood -> stone -> diamond. So what?

The main thing I fear here is that the player may end up sitting there for 5 minutes straight to mine this one uranium ore they found in a rare surface structure that happen to spawn near their spawn location. In my opinion there should be a clear way of telling the player, no it's not worth it, you should not bother trying.
Now I wouldn't be opposed to do this in different ways. It could e.g. also be a time-based limit, or a more direct, material based restriction, instead of being damage based.

> hardness, weight and stickiness

Now about your idea, I think the fundamental problem is that on one hand we want a system that is easy for the player to understand (and having only one value, DPS, to optimize for definitely helps in that regard), but on the other hand we want a system with depth and a use for exploring different material combinations (which is where multiple parameters are helpful).

We have seen that situational modifiers like "good at sandstone" are really good at giving the player an incentive to try new tools without adding too much complexity (it's really easy to understand what sandstone is). I think it would be better to expand in that direction, we could easily add something like `good at hard materials` or `good at ore` or `good at stone` or other broader categories, which should still get some of the benefits of your proposed system, but without having these parameters visible at all times, without being limited to 3 parameters, and without the hassle of categorizing all blocks in these three parameters.

> It reduces the immense complexity of procedural crafting to something on a level equivalent to the rest of the game. This game is a toy, not a calculus lecture.

How would it do that? I think discretizing the output could actually make it worse, since now swapping any two materials could possibly push it above the threshold, even if it only changes the values by a small amount, encouraging more random ingredient swapping.
Also how would you propose to average these properties over the given slots?

careeoki:
I think this makes more sense now after #2636 , since you can do encasing earlier in progression

IntegratedQuantum:
Could you rebase this with sky islands? I would like to take a look at how the ore density is in the sky.

IntegratedQuantum:
And yes, I agree it makes sense to add this now, I would also like to see how this changes the dynamic of how people explore.

careeoki:
> Could you rebase this with sky islands? I would like to take a look at how the ore density is in the sky.

Done

careeoki:
Maybe I'm missing something but I can't seem to find any jade at all.

IntegratedQuantum:
yeah, it seems the default of the maxHeight for ores is 0. I'll fix it in a bit.

IntegratedQuantum:
done, please rebase again

IntegratedQuantum:
Also unmark it as WIP, when you think it's ready to merge

careeoki:
> done, please rebase again

thanks, done

> Also unmark it as WIP, when you think it's ready to merge

Yep I will, I just need to make the jade block and cut jade block.
What about the copper + jade idea? Do you still want that? Or should I keep the regular encasing it has now?

IntegratedQuantum:
Just keep it regularly encased for now.

careeoki:
It might be difficult to get jade since sky islands don't have caves. And I really don't want players having to stripmine.

TyrrelMeter:
Well then we'll just have to add caves to sky islands.

careeoki:
it's more like "_Quantum_ will have to add caves to sky islands"

TyrrelMeter:
Good point.

IntegratedQuantum:
I'll add that once there are cave layers, configuring the fractal cave generator per layer would be more reasonable.

IntegratedQuantum:
but I'm also worried about how few there are visible on large mountains. I found a total of just 3 exposed veins on one mountain and one volcano.

careeoki:
I can tweak the ore values.

careeoki:
This is ready for merge. Blocks are done.
I also tweaked the ore values, it should be easier to find exposed veins on mountains
image

du82:
I REALLY like these textures

ikabod-kee:
> This is ready for merge. Blocks are done. I also tweaked the ore values, it should be easier to find exposed veins on mountains image

Make it a swirl. Trust.
image
image

careeoki:
> Make it a swirl. Trust.

the item is already based on a magatama. If you're talking about the cut block, I already tried giving it a similar shape, but it didn't work out how I wanted
image

PR #2464, opened 2026-01-05
The Mantle (concept(?))

careeoki:
The Mantle is something I first mentioned in #1463 , basically being a hell/underworld/fiery death pit area.
image
image

This introduces the following cave layers:
- Mantle layer
- Lower mantle layer
- Deep caves layer (very creative name)

The lower mantle is the lava barrier (or at least as close to that as I can make)
Deep caves is just regular caves again, but with some caves excluded (mainly caves that wouldn't appear there anyways and also small mushroom caves and stone forest because I felt they fit better higher up)


Crepestrom:
Made a extra biome for the mantle if you want to use it

Argmaster:
I think it looks like the magma caves. It has typical procedural content issue "I have already seen this". After seeing the screenshots, I would argue to make it a cold core, since it's just above great emptiness. We could make the hot layer higher, where it is separated from void, but still, reusing assets from lava caves seem lame to me :C

careeoki:
A cold core does sound pretty cool...  but I also want a hot layer as well, I think it being higher up like you said could work. something like this:
image

And yeah, I've really not put much time into the generation. I'm going to start making some unique SBBs to appear here, and try to make it as distinct from the lava caves as I can.

du82:
I like it a lot! especially the new blocks! I think we need a lot more block textures and colors

careeoki:
Can this be moved into a "ready for review" column pretty please?

IntegratedQuantum:
>  would love to have a giant lava pool at the bottom of the mantle, separating it and what's below.

Yeah, that's exactly what I'd want to have. So please move the biome to the 15-20k region and add a small lava-only biome at 20k. For now, since we don't have the stuff below it figured out yet, I think you can just fill everything below 20k with the mantle biomes as well.

Also I like the idea of a cold core, we've already figured out that ice is actually quite deadly, so it makes sense for it to be lower in progression, and it could rise to a more bluish rock. The big question is just what we should put in-between of them. The missing color would be green, so maybe some kind of underground jungle thing would fit in between, and we probably have place for more layers as well.

careeoki:
> The big question is just what we should put in-between of them. The missing color would be green, so maybe some kind of underground jungle thing would fit in between, and we probably have place for more layers as well.

Yeah, that would make sense. 

> So please move the biome to the 15-20k region and add a small lava-only biome at 20k. 

Changed the height and added a basic core biome.
When I try to make a lava pool, I just get a big lumpy mess. Ideally it would be a nice flat plane, like an ocean.

image

assets/cubyz/biomes/cave/mantle/core.zig.zon
```diff
@@ -0,0 +1,7 @@
+.{
+	.chance = 1000,
+	.caves = -0.03,
```
IntegratedQuantum: You should make it have less caves, and if you set the cave radius to 0 then you also won't get any of the normal caves.

assets/cubyz/sbb/cave/mantle/ruins.zig.zon
IntegratedQuantum, line 1: for the ruins, I'd suggest to also make the subsurface structure out of bricks
Screenshot at 2026-01-15 19-25-13

assets/cubyz/sbb/cave/mantle/stalagmite.zig.zon
IntegratedQuantum, line 1: Ideally this something that should be provided by the stalagmite generator, and not something you have to handcraft.
Am I correct to assume that all you wanted to do is make them pointier?

assets/cubyz/blocks/pyrolite/_defaults.zig.zon
IntegratedQuantum, line 1: Maybe we can find a better name for the blocks? I'm not too happy with them being named like the biome.
On the other hand Komatiite has an awful pronounciation.
Maybe we can figure out a better name here?
Many minerals are indeed named after location, but they tend to have a special ending, like -ite
I consulted my procedurally generated mineral name list, and I found an entry for Mantlerite which might be fitting here.

IntegratedQuantum (changes requested):

assets/cubyz/biomes/cave/mantle/core.zig.zon
careeoki: With cave radius (caveRadiusFactor) set to 0 I'm still getting normal caves.
image

assets/cubyz/sbb/cave/mantle/stalagmite.zig.zon
careeoki: Yes, I just wanted them more pointy.

assets/cubyz/sbb/cave/mantle/ruins.zig.zon
careeoki: Done.

assets/cubyz/blocks/pyrolite/_defaults.zig.zon
careeoki: "Komatiite" is also derived from a real-life place name, which I would like to avoid (like good old coloradostone)

I changed it to "pyrolite," (literally fire rock) which I ~stole~ borrowed from [this model to represent the earth's mantle](https://en.wikipedia.org/wiki/Pyrolite#)
...I'm not using it in reference to that model, I just like the "pyro" prefix.

assets/cubyz/biomes/cave/mantle/core.zig.zon
IntegratedQuantum: Ah right, it's because the interpolation are is slightly offset from the block placement.
I guess we can leave it for now. Could you please make an issue explaining that you want a flat lava floor.

assets/cubyz/blocks/pyrolite/_defaults.zig.zon
IntegratedQuantum: Yeah, pyrolite sounds good to me.

assets/cubyz/sbb/cave/mantle/stalagmite.zig.zon
IntegratedQuantum: Alright, then if you don't mind I would like to pospone this PR until we have pointier stalagmites. Could you also make an issue for it please? I will put it in the 0.2.0 milestone

careeoki: I've updated to the new stalagmite simple structure.

careeoki:
> For now, since we don't have the stuff below it figured out yet, I think you can just fill everything below 20k with the mantle biomes as well.

I feel like it would be better if it went back to regular caves, since they have a lot more variety. And I would like to have "buffer" areas between layers of more regular caves.
But there isn't a nice way to exclude biomes in a specific area and have them everywhere else.

IntegratedQuantum:
Hmm, alright, I guess we again block this then until I figure out the cave layer system.

The4codeblocks:
idea:
olivine (green), possibly with an exaggerated saturation

TyrrelMeter:
@careeoki, you should probably update this PR to make use of the new cave generation system.

careeoki:
This is updated to use cave layers and SDFs.

Wunka:
Sadly this already has conflicts... else I would have moved it.

careeoki:
Fixed conflicts, moving to easy to review as it's been partially reviewed before

H41ogen:
it might be good to find a name other than just deep_caves now, so we don't have to add a migration when it inevitably gets renamed/removed

careeoki:
> it might be good to find a name other than just deep_caves now, so we don't have to add a migration when it inevitably gets renamed/removed

I'm open to cooler sounding suggestions but for now ill rename it "post mantle layer"

Argmaster:
Depths
Interior
Abyss

careeoki:
Yeah, but we need something that will still make sense when it is sandwiched between other layers, those are pretty broad and could be used for any generic layer

careeoki:
The running plan is mantle -> buffer layer -> green jungle type area -> buffer layer -> cold core (#1463)

IntegratedQuantum:
> it might be good to find a name other than just deep_caves now, so we don't have to add a migration when it inevitably gets renamed/removed

cave layers don't need migrations, they aren't and won't be stored anywhere.

assets/cubyz/cave_layers/post_mantle.zig.zon
```diff
@@ -0,0 +1,5 @@
+.{
+	.tags = .{.post_mantle_layer},
```
IntegratedQuantum: Instead of retagging all the biomes, you can also just use `.cave_layer` here

assets/cubyz/cave_layers/lower_mantle.zig.zon
```diff
@@ -0,0 +1,7 @@
+.{
+	.tags = .{.lower_mantle_layer},
+	.layerHeight = 200,
```
IntegratedQuantum: I saw some gaps, I would make this a bit larger, maybe 300

IntegratedQuantum (changes requested):
Also missing the pillar, but you can add that in a future PR as well if you like

assets/cubyz/cave_layers/post_mantle.zig.zon
careeoki: I excluded specific biomes from the post mantle layer that i felt fit better only in the regular caves layer

IntegratedQuantum: Then I would suggest to remove the `cave_layer` tag only from those biomes, and instead maybe give them a `shallow_cave_layer` tag and add that to the cave layers above the mantle.

careeoki: Alright, done

assets/cubyz/cave_layers/lower_mantle.zig.zon
careeoki: Done

careeoki:
> Also missing the pillar, but you can add that in a future PR as well if you like

I'm leaving for a camping trip tomorrow so I'll do it in another PR

IntegratedQuantum (approved):

PR #2465, opened 2026-01-06
Magma Pool biome

Crepestrom:
Made for Mantle Concept
image

filled with basalt rock and magma
also contains magma pools


Crepestrom:
#2464 made to be added to this biome

PR #2466, opened 2026-01-06
switch to std.Io.net

Wunka:
Related to #2338


Wunka:
While I didn't think I would do it. here we are. This is from my seeing completely working. 
I will put this into low priority because I think this would also make #859 easier.
If there are any major network problems here, I will only maybe continue it because as said before. I am not that deep into network coding.

Argmaster:
Umm (Windows)

```
Got error on receive: ConcurrencyUnavailable
Got error on receive: ConcurrencyUnavailable
Got error on receive: ConcurrencyUnavailable
Got error on receive: ConcurrencyUnavailable
Got error on receive: ConcurrencyUnavailable
Got error on receive: ConcurrencyUnavailable
Got error on receive: ConcurrencyUnavailable
Got error on receive: ConcurrencyUnavailable
```

(a lot of those)

Wunka:
```zig
.net_receive => |*o| {
	// TODO integrate with overlapped I/O or equivalent to avoid this error
	if (concurrency) return error.ConcurrencyUnavailable;
	batchCompleteBlockingWindows(b, operation_userdata, .{
		.net_receive = netReceiveWindows(t, o.socket_handle, o.message_buffer, o.data_buffer, o.flags),
	});
},

```
https://ziglang.org/documentation/0.16.0/std/#std.Io.Threaded.batchDrainSubmittedWindows
This TODO seems to be the problem. 
This is called by `receiveTimout` with `concurrency = true` but is `false` when calling via `receive` so this will probably be the way to do it. I will make another comment here when I switched that

Wunka:
I fixed the windows problem. 
How?
I created a concurrent task for the receiving of messages (so we don't need to timeout). if a message comes it will be added to a `concurrentQueue` of `IncomingMessages` which then get processed where we before did `receiveTImeout`

Wunka:
As I somehow beaten `getLight` in worse performance so much that its not even in the top 5 anymore, I will close this until either we can use receiveTimeout or there is a good other way to do it.
(profiling on commit 5865b5c7, where then this branch diverges)

image


(profiling on this branch)
image

IntegratedQuantum:
It's because you are busy waiting on the other thread.

Also if this function is only a problem on windows, then maybe we could just check if new data i available and sleep, instead of using a timeout (which we do on windows anyways right now, due to a different issue)

Wunka:
> It's because you are busy waiting on the other thread.

You are right I am busy waiting and can make this better.
> Also if this function is only a problem on windows, then maybe we could just check if new data i available and sleep, instead of using a timeout (which we do on windows anyways right now, due to a different issue)

Sadly I don't see an api point to do this (execpt receiveTimeout, or going again lower). While researching I cam across this issue: https://codeberg.org/ziglang/zig/issues/31098. I think the "baybsitter" thing could work to make our own timeout. but I was not able to implement it (at least in the amount of time I spend trying).

The closing should not mean I give up, just that in my current knowledge I am stuck and also am not willing to spend a significant amount of time making my own solution, when we can also just wait another zig release cycle

IntegratedQuantum:
> Sadly I don't see an api point to do this (execpt receiveTimeout, or going again lower).

Does it maybe behave differently internally when the timeout given is zero?

Wunka:
> > Sadly I don't see an api point to do this (execpt receiveTimeout, or going again lower).
> 
> Does it maybe behave differently internally when the timeout given is zero?

Tested it on windows. still gives the error :(
The best way is probably the babysitter. maybe If I have time (not today) I will try it again.

Wunka:
Okay I got the babysitter to work (sadly not in this pr as it seems you are not allowed to force push to closed prs and also not after that able to reopen them).  Currently it seems like I now have expected performance (which means as it is with `receiveTimeout` normally) but I will have to make a bit more profiling to get actual data.

PR #2468, opened 2026-01-06
Prototype of a world type that shows all structures

IntegratedQuantum:
progress towards #1387

### DIFF
--- a/src/server/terrain/structuremapgen/SbbEnumerationGenerator.zig
+++ b/src/server/terrain/structuremapgen/SbbEnumerationGenerator.zig
@@ -57,9 +57,11 @@ pub fn init(parameters: ZonElement) void {
 		}
 	}
 
-	std.sort.insertion(*const StructureBuildingBlock, localSbbList.items, {}, struct{fn lessThanFn(_: void, lhs: *const StructureBuildingBlock, rhs: *const StructureBuildingBlock) bool {
-		return std.ascii.orderIgnoreCase(lhs.id, rhs.id) == .lt;
-	}}.lessThanFn);
+	std.sort.insertion(*const StructureBuildingBlock, localSbbList.items, {}, struct {
+		fn lessThanFn(_: void, lhs: *const StructureBuildingBlock, rhs: *const StructureBuildingBlock) bool {
+			return std.ascii.orderIgnoreCase(lhs.id, rhs.id) == .lt;
+		}
+	}.lessThanFn);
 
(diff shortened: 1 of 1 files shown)


Argmaster:
I guess it's good enough. I think I already highlighted in the past problems I see with lack of proper UI, but in case I didn't:
- how do I find my structure, when there are hundreds? (There aren't now)
- how do I find and change structure segment I dislike?
- structures are presented out of context, hence it's hard to determine which parts are visible irl, which are not (roots)

I admit I didn't read the code in great detail, I hope to find time for that during the weekend.

IntegratedQuantum:
> how do I find my structure, when there are hundreds?

We could add conditional tp command support when it becomes an issue, for now though you can just fly there.

> how do I find and change structure segment I dislike?

Well, there is another dimension here, we can put all the blueprints and maybe child sbbs involved in that structure above and below it.

IntegratedQuantum:
> structures are presented out of context, hence it's hard to determine which parts are visible irl, which are not (roots)

I now have made a second layer of structures that spawn directly on the flat ground.

IntegratedQuantum:
@Argmaster are you still going to review this?

Argmaster:
Uh it good, sorry, I had a look and I was satisfied

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -53,20 +53,20 @@ pub fn loadModel(parameters: ZonElement) ?*SbbGen {
 }
 
 pub fn generate(self: *SbbGen, _: GenerationMode, x: i32, y: i32, z: i32, chunk: *ServerChunk, _: CaveMapView, _: CaveBiomeMapView, seed: *u64, _: bool) void {
-	placeSbb(self, self.structureRef, Vec3i{x, y, z}, Neighbor.dirUp, self.rotation.getInitialRotation(seed), chunk, seed);
+	placeSbb(self, self.structureRef, Vec3i{x, y, z}, null, self.rotation.getInitialRotation(seed), chunk, seed);
```
Argmaster, line 56: Although, why is that?

IntegratedQuantum: It had some trouble because some of the sbbs (→the unused ones) do not have an upwards origin block.
Also I think it would be reasonable to allow any origin block for the start node, to make it more natural to build hanging or directional structures, and to allow reusing existing intermediate models for that (e.g. a vine cluster could hang on a tree or directly a cave ceiling)

Argmaster: Aight, I think I remember roughly what was happening there, it's fine.

PR #2469, opened 2026-01-07
DepositToAny slice for the destination

Wunka:
Splitted of from #2324 
First of two Payloads that need change to accept slices as destinations for the split of the hotbar to work.

Even if the split of the hotbar is not accepted this could still be necessary. See comment: https://github.com/PixelGuys/Cubyz/pull/2324#pullrequestreview-3625012936

I don't like the allocation solution so if anybody has a better solution it would be very welcome. 

The renaming to `destinations` is to use `dest` in the for loops and to show that this is different to the other payloads `dest` parameter

### DIFF
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -1674,7 +1674,7 @@ pub const Command = struct { // MARK: Command
 		source: InventoryAndSlot,
 		amount: u16,
 
-		fn init(destinations: []const Inventory, source: InventoryAndSlot, amount: u16) DepositToAny {
+		pub fn init(destinations: []const Inventory, source: InventoryAndSlot, amount: u16) DepositToAny {
 			return .{
 				.destinations = main.globalAllocator.dupe(Inventory, destinations),
 				.source = source,
@@ -1702,8 +1702,8 @@ pub const Command = struct { // MARK: Command
 
 			var remainingAmount = self.amount;
 			var selectedEmptySlot: ?u32 = null;
-			var selectedEmptyInv: usize = 0;
-			outer: for(self.destinations, 0..) |dest, destInv| {
+			var selectedEmptyInv: ?Inventory = null;
+			outer: for(self.destinations) |dest| {
 				var emptySlot: ?u32 = null;
 				var hasItem = false;
 				for(dest._items, 0..) |*destStack, destSlot| {
@@ -1711,7 +1711,7 @@ pub const Command = struct { // MARK: Command
 						emptySlot = @intCast(destSlot);
 						if(selectedEmptySlot == null) {
 							selectedEmptySlot = emptySlot;
-							selectedEmptyInv = destInv;
+							selectedEmptyInv = dest;
 						}
 					}
 					if(std.meta.eql(destStack.item, sourceStack.item)) {
@@ -1734,11 +1734,12 @@ pub const Command = struct { // MARK: Command
 						.amount = remainingAmount,
 					}});
 					remainingAmount = 0;
+					break :outer;
 				}
 			}
 			if(remainingAmount > 0 and selectedEmptySlot != null) {
 				ctx.execute(.{.move = .{
-					.dest = .{.inv = self.destinations[selectedEmptyInv], .slot = selectedEmptySlot.?},
+					.dest = .{.inv = selectedEmptyInv.?, .slot = selectedEmptySlot.?},
 					.source = self.source,
 					.amount = remainingAmount,
 				}});
@@ -1757,6 +1758,7 @@ pub const Command = struct { // MARK: Command
 		fn deserialize(reader: *utils.BinaryReader, side: Side, user: ?*main.server.User) !DepositToAny {
 			const destinationsSize = try reader.readVarInt(usize);
 			if(destinationsSize == 0) return error.Invalid;
+			if(destinationsSize*@sizeOf(InventoryId) >= reader.remaining.len) return error.Invalid;
 
 			const destinations = main.globalAllocator.alloc(Inventory, destinationsSize);
 			errdefer main.globalAllocator.free(destinations);


src/Inventory.zig
```diff
@@ -1670,16 +1670,21 @@ pub const Command = struct { // MARK: Command
 	};
 
 	const DepositToAny = struct { // MARK: DepositToAny
-		dest: Inventory,
+		destinations: []Inventory,
+		allocated: bool = false,
```
Argmaster: "Owned" is the term used in other places.

src/Inventory.zig
```diff
@@ -1670,16 +1670,21 @@ pub const Command = struct { // MARK: Command
 	};
 
 	const DepositToAny = struct { // MARK: DepositToAny
-		dest: Inventory,
+		destinations: []Inventory,
```
Argmaster: Big fan of variable names that don't stutter!

src/Inventory.zig
Wunka: changed!

src/Inventory.zig
Wunka: So should I change the name or not? If yes what name would you use? Should I use `dest` and find a different name for the for loops?

Argmaster: No, I like it the way it is.
That was a comment of appreciation lol
dest is an example of stuttering variable name.

Wunka:
I changed the priority system between the different destinations to align with https://github.com/PixelGuys/Cubyz/pull/2324#pullrequestreview-3625012936 
For testing purposes you can just use #2324 as I have merged this branch into there. What I used for testing is a chest and I just shift-clicked moved items between the chest and my hotbar/inventory

src/Inventory.zig
```diff
@@ -1670,16 +1670,21 @@ pub const Command = struct { // MARK: Command
...
+		owned: bool = false,
 		source: InventoryAndSlot,
 		amount: u16,
 
 		fn run(self: DepositToAny, ctx: Context) error{serverFailure}!void {
-			if(self.dest.type == .creative) return;
-			if(self.dest.type == .crafting) return;
-			if(self.dest.type == .workbench) return;
+			defer if(self.owned) main.stackAllocator.free(self.destinations);
```
IntegratedQuantum: command may run multiple times on the client side, freeing here causes a double free. Instead you can free this in finalize.
Also the stackAllocator is a poor choice, only use the stackAllocator for local allocations, this allocation persists for multiple threads and likely will be ran from a different thread as well.

Also the allocator shouldn't be hardcoded here, instead the caller should pass it.

Lastly who even owns the memory in the other case? How do you keep track of something that gets used possible many frames after passing it?

Wunka: > command may run multiple times on the client side, freeing here causes a double free. Instead you can free this in finalize. Also the stackAllocator is a poor choice, only use the stackAllocator for local allocations, this allocation persists for multiple threads and likely will be ran from a different thread as well.

done changed to the `globalAllocator` which is used for other payloads

> Also the allocator shouldn't be hardcoded here, instead the caller should pass it.

so should I change  the finalize to parse over the `globalAllocator`? That looks better to me at least.

> Lastly who even owns the memory in the other case? How do you keep track of something that gets used possible many frames after passing it?

Yeah you are right there... I will look into how I want to solve that. I could maybe just dupe the slice so that its always owned? But I don't want to that in `gui.zig` where `depositToAny` is called... I will sleep over it and look tomorrow at the code again.

IntegratedQuantum: > so should I change the finalize to parse over the globalAllocator? That looks better to me at least.

No, it should be passed on creation, since the finalize call is already really disconnected from the creation.
This would prevent accidentally using mismatched allocators.

Wunka: switched the owned variable with an allocator. Also did the other issue with an dupe so it should now always be owned.

IntegratedQuantum: You could also consider making a separate `.init` function that allocates everything inside. Then I would be fine with not having the allocator stored since it's all used locally.

src/Inventory.zig
```diff
@@ -1670,16 +1670,24 @@ pub const Command = struct { // MARK: Command
...
 
+		fn finalize(self: DepositToAny, _: Side, _: *utils.BinaryReader) !void {
+			self.allocator.free(self.destinations);
+		}
+
 		fn run(self: DepositToAny, ctx: Context) error{serverFailure}!void {
-			if(self.dest.type == .creative) return;
-			if(self.dest.type == .crafting) return;
-			if(self.dest.type == .workbench) return;
+			if(self.destinations.len == 0) return;
```
IntegratedQuantum: When would this even happen?

src/Inventory.zig
```diff
@@ -1670,16 +1670,24 @@ pub const Command = struct { // MARK: Command
...
+			for(self.destinations) |dest| {
+				if(dest.type == .creative) return;
+				if(dest.type == .crafting) return;
+				if(dest.type == .workbench) return;
+			}
 			if(self.source.inv.type == .crafting) {
-				ctx.cmd.tryCraftingTo(ctx.allocator, self.dest, self.source, ctx.side, ctx.user);
+				ctx.cmd.tryCraftingTo(ctx.allocator, self.destinations[0], self.source, ctx.side, ctx.user);
```
IntegratedQuantum, line 1696: I assume you intend to address this function in a future PR?

src/Inventory.zig
```diff
@@ -1688,41 +1696,65 @@ pub const Command = struct { // MARK: Command
 
 			var remainingAmount = self.amount;
 			var selectedEmptySlot: ?u32 = null;
-			for(self.dest._items, 0..) |*destStack, destSlot| {
-				if(destStack.item == .null and selectedEmptySlot == null) {
-					selectedEmptySlot = @intCast(destSlot);
+			var selectedEmptyInv: u8 = 0;
```
IntegratedQuantum: Why limit this to a u8? It should be usize, there is nothing gained by using a u8, other than future bugs from the intCast.

src/Inventory.zig
```diff
@@ -1688,41 +1696,65 @@ pub const Command = struct { // MARK: Command
...
+					.dest = .{.inv = self.destinations[selectedEmptyInv], .slot = selectedEmptySlot.?},
 					.source = self.source,
 					.amount = remainingAmount,
 				}});
 			}
 		}
 
 		fn serialize(self: DepositToAny, writer: *utils.BinaryWriter) void {
-			writer.writeEnum(InventoryId, self.dest.id);
+			writer.writeInt(u8, @intCast(self.destinations.len));
```
IntegratedQuantum: Use a varint and send a usize
Putting an arbitrary limit on the number and crashing when exceeded is not helpful.
If you want to limit this then you should have a good reason for it, and 256 might definitely be in reach for a deposit to nearby chests function.

src/Inventory.zig
```diff
@@ -1688,41 +1696,65 @@ pub const Command = struct { // MARK: Command
...
+				writer.writeEnum(InventoryId, dest.id);
+			}
 			self.source.write(writer);
 			writer.writeInt(u16, self.amount);
 		}
 
 		fn deserialize(reader: *utils.BinaryReader, side: Side, user: ?*main.server.User) !DepositToAny {
-			const destId = try reader.readEnum(InventoryId);
+			const destinationsSize = try reader.readInt(u8);
+			var destinations = main.globalAllocator.alloc(Inventory, destinationsSize);
```
```suggestion
			const destinations = main.globalAllocator.alloc(Inventory, destinationsSize);
```

src/gui/gui.zig
```diff
@@ -665,7 +665,8 @@ pub const inventory = struct { // MARK: inventory
 				nextCraftingAction = nextCraftingAction.addDuration(craftingCooldown);
 				craftingCooldown.nanoseconds -= @divTrunc((craftingCooldown.nanoseconds -% minCraftingCooldown.nanoseconds)*craftingCooldown.nanoseconds, std.time.ns_per_s);
 				if(mainGuiButton.modsOnPress.shift) {
-					itemSlot.inventory.depositToAny(itemSlot.itemSlot, main.game.Player.inventory, itemSlot.inventory.getAmount(itemSlot.itemSlot));
+					var destinations = [_]Inventory{main.game.Player.inventory};
```
IntegratedQuantum:
```suggestion
					const destinations = &.{main.game.Player.inventory};
```

src/gui/gui.zig
```diff
@@ -665,7 +665,8 @@ pub const inventory = struct { // MARK: inventory
...
+					itemSlot.inventory.depositToAny(itemSlot.itemSlot, destinations[0..], itemSlot.inventory.getAmount(itemSlot.itemSlot));
```
IntegratedQuantum: Also maybe just inline it? It's only 4 more characters compared to what was there before.
```suggestion
					itemSlot.inventory.depositToAny(itemSlot.itemSlot, &.{main.game.Player.inventory}, itemSlot.inventory.getAmount(itemSlot.itemSlot));
```

src/gui/gui.zig
```diff
@@ -683,12 +684,14 @@ pub const inventory = struct { // MARK: inventory
 				var iterator = std.mem.reverseIterator(openWindows.items);
 				while(iterator.next()) |window| {
 					if(window.shiftClickableInventory) |inv| {
-						itemSlot.inventory.depositToAny(itemSlot.itemSlot, inv, itemSlot.inventory.getAmount(itemSlot.itemSlot));
+						var destinations = [_]Inventory{inv};
+						itemSlot.inventory.depositToAny(itemSlot.itemSlot, destinations[0..], itemSlot.inventory.getAmount(itemSlot.itemSlot));
```
IntegratedQuantum:
```suggestion
						itemSlot.inventory.depositToAny(itemSlot.itemSlot, &.{inv}, itemSlot.inventory.getAmount(itemSlot.itemSlot));
```

src/gui/gui.zig
```diff
@@ -683,12 +684,14 @@ pub const inventory = struct { // MARK: inventory
...
 						break;
 					}
 				}
 			} else {
-				itemSlot.inventory.depositToAny(itemSlot.itemSlot, main.game.Player.inventory, itemSlot.inventory.getAmount(itemSlot.itemSlot));
+				var destinations = [_]Inventory{main.game.Player.inventory};
+				itemSlot.inventory.depositToAny(itemSlot.itemSlot, destinations[0..], itemSlot.inventory.getAmount(itemSlot.itemSlot));
```
IntegratedQuantum: same here

src/Inventory.zig
```diff
@@ -1688,41 +1696,65 @@ pub const Command = struct { // MARK: Command
...
-						.dest = .{.inv = self.dest, .slot = @intCast(destSlot)},
-						.source = self.source,
-						.amount = amount,
-					}});
-					remainingAmount -= amount;
-					if(remainingAmount == 0) break;
+				if(emptySlot != null and (selectedEmptySlot == null or (hasItem and !selectedEmptyInvHasItem))) {
+					selectedEmptySlot = emptySlot;
+					selectedEmptyInv = @intCast(destInv);
+					selectedEmptyInvHasItem = hasItem;
```
IntegratedQuantum: I think this is different from the behavior suggested in https://github.com/PixelGuys/Cubyz/pull/2324#pullrequestreview-3625012936, here you seem to be first topping off partially filled stacks in the other inventories.

src/Inventory.zig
Wunka: A modified Client can send anything. I just want to put a check in before it later causes bugs.

src/Inventory.zig
Wunka: As it's needed for this function yes I can set this as the next goal

src/Inventory.zig
IntegratedQuantum: If this is about safety of client data, then you should put this check in the deserialization function.

IntegratedQuantum: Client data should generally be sanitized during or immediately after receiving.

src/Inventory.zig
Wunka: all done

src/Inventory.zig
Wunka: This function seems also very easy to allow multiple destinations, so if you want I can address this before this PR gets merged.

IntegratedQuantum: no, it's fine to do it later.

src/Inventory.zig
```diff
@@ -1718,10 +1728,13 @@ pub const Command = struct { // MARK: Command
...
-					selectedEmptySlot = emptySlot;
-					selectedEmptyInv = @intCast(destInv);
-					selectedEmptyInvHasItem = hasItem;
+				if(emptySlot != null and hasItem) {
+					ctx.execute(.{.move = .{
+						.dest = .{.inv = dest, .slot = emptySlot.?},
+						.source = self.source,
+						.amount = remainingAmount,
+					}});
+					remainingAmount = 0;
```
IntegratedQuantum, line 1736: Also should break after this, there is no need to search further.

src/Inventory.zig
```diff
@@ -1743,8 +1756,10 @@ pub const Command = struct { // MARK: Command
 		}
 
 		fn deserialize(reader: *utils.BinaryReader, side: Side, user: ?*main.server.User) !DepositToAny {
-			const destinationsSize = try reader.readInt(u8);
-			var destinations = main.globalAllocator.alloc(Inventory, destinationsSize);
+			const destinationsSize = try reader.readVarInt(usize);
+			if(destinationsSize == 0) return error.Invalid;
```
IntegratedQuantum, line 1760: Also please add a check to ensure that reader.remaining has enough data for the destinationsSize, otherwise this can be used to crash the server due to allocating a possibly user-defined amount of memory.

src/Inventory.zig
```diff
@@ -1670,16 +1670,30 @@ pub const Command = struct { // MARK: Command
 	};
 
 	const DepositToAny = struct { // MARK: DepositToAny
-		dest: Inventory,
+		destinations: []const Inventory,
 		source: InventoryAndSlot,
 		amount: u16,
 
+		fn init(destinations: []const Inventory, source: InventoryAndSlot, amount: u16) DepositToAny {
```
IntegratedQuantum: Should be pub, since someone else may want to access it from the outside.

src/Inventory.zig
```diff
@@ -1688,41 +1702,72 @@ pub const Command = struct { // MARK: Command
 
 			var remainingAmount = self.amount;
 			var selectedEmptySlot: ?u32 = null;
-			for(self.dest._items, 0..) |*destStack, destSlot| {
-				if(destStack.item == .null and selectedEmptySlot == null) {
-					selectedEmptySlot = @intCast(destSlot);
+			var selectedEmptyInv: usize = 0;
```
IntegratedQuantum: Why not store the `Inventory` directly? Do you really need the index?

src/Inventory.zig
Wunka: I actually remember that I wanted to that because shortly before that I saw a video how exactly this kinda bug lead to the easy reading of data from a database but then I forgot it nevertheless... I don't how that much knowledge how the data is organized when doing it per reader writer but my safeguard should at least prevent something. Maybe someone with higher understanding of the memory layout can put a more precise check there.

Wunka:
rest of the requested changes are also done

src/Inventory.zig
IntegratedQuantum: that's good enough, we just need to prevent allocating absurdly large buffers. If the buffers can be a few bytes larger than the actual data, it's not a big deal and checked by the later attempts to read it.

PR #2476, opened 2026-01-09
Increase the timer resolution for a more precise sleep on Windows

MiniBaker007:
On windows:
At the start of the application, the timer resolution is set to the maximum supported value, usualy 0.5 or 1 millisecond. To get consistant timing, we undersleep based on the timer resolution and busy wait the rest. 

I also made it set the game's process priority class to "Above normal" ,which helps sleeping be more consistant. This might also generally improve the game's performance a little, I'm not sure.

From my testing, this achieves a quite consistant framerate, but that can depend on a lot of variables that are difficult to control for. So, if anyone is able to test this, that would be helpful.

Also, I'm not convinced with some of the ways I named things and structured the code, but I decided to get others' opinions rather than to keep overthinking it.

### DIFF
--- a/src/server/server.zig
+++ b/src/server/server.zig
@@ -1,5 +1,4 @@
 const std = @import("std");
-const builtin = @import("builtin");
 const Atomic = std.atomic.Value;
 
 const main = @import("main");


Argmaster:
Hey, sorry but I will give it low prio, because it's technically already solved, even if the solution is crude. I do appreciate the effort tho!

src/main.zig
```diff
@@ -513,7 +514,8 @@ pub fn main() void { // MARK: main()
 	std.log.info("Starting game with version {s}", .{settings.version.version});
 
 	if(builtin.os.tag == .windows) {
```
IntegratedQuantum, line 516: Since you made a new file for this, I think all operating system nuances should be handled inside of it, and the calling code should be OS-agnostic.

I'd suggest to structure the file similarly to the file_monitor.zig

src/utils/windows_scheduler.zig
```diff
@@ -0,0 +1,97 @@
...
+pub fn increaseTimerResolution() void {
+	std.debug.assert(builtin.os.tag == .windows);
+
+	_ = NtQueryTimerResolution(&minResolution, &maxResolution, &initialResolution);
+	_ = NtSetTimerResolution(maxResolution, 1, &currentResolution);
+
+	std.log.info("Set system timer resolution: {d} -> {d}.", .{initialResolution, currentResolution});
+}
+
+pub fn resetTimerResolution() void {
```
IntegratedQuantum, line 21: unused function

src/utils/windows_scheduler.zig
```diff
@@ -0,0 +1,97 @@
...
+	std.log.info("Reset system timer resolution: {d}.", .{initialResolution});
+}
+
+pub fn preciseSleep(sleepDuration: std.Io.Duration) void {
+	std.debug.assert(builtin.os.tag == .windows);
+
+	const start = main.timestamp();
+	const targetTime = start.addDuration(sleepDuration);
+
+	// Update currentResolution because it can change
```
IntegratedQuantum, line 35: Wait what? Can windows just change this behind our back? Otherwise I'd remove this, since this is a value that we should only set once.

src/utils/windows_scheduler.zig
```diff
@@ -0,0 +1,97 @@
...
+	_ = NtQueryTimerResolution(&minResolution, &maxResolution, &currentResolution);
+
+	// Negative value for relative time
+	var delayInterval = -(@divFloor(@as(windows.LARGE_INTEGER, @intCast(sleepDuration.nanoseconds)), 100) - currentResolution);
```
IntegratedQuantum, line 39:
```suggestion
	var delayInterval = currentResolution - @divFloor(@as(windows.LARGE_INTEGER, @intCast(sleepDuration.nanoseconds)), 100);
```

src/utils/windows_scheduler.zig
```diff
@@ -0,0 +1,97 @@
...
+	if(delayInterval < 0) {
+		_ = NtDelayExecution(windows.FALSE, &delayInterval);
```
IntegratedQuantum, line 41: Is this different from what Io.sleep calls internally?

src/utils/windows_scheduler.zig
```diff
@@ -0,0 +1,97 @@
...
+	const end = main.timestamp();
+	if(targetTime.durationTo(end).nanoseconds > currentResolution*100) {
+		const desired = @as(f64, @floatFromInt(sleepDuration.nanoseconds))/1_000_000;
+		const actual = @as(f64, @floatFromInt(start.durationTo(end).nanoseconds))/1_000_000;
+		std.log.warn("Overslept: Desired duration: {d:.2}ms, Actual duration: {d:.2}ms, Timer resolution: {d}", .{desired, actual, currentResolution});
+	}
+
+	while(main.timestamp().durationTo(targetTime).nanoseconds > 0) {
+		std.atomic.spinLoopHint();
+	}
```
IntegratedQuantum, line 53: I wonder if we actually need this busy loop here. Do we need to be this precise? I think it won't be a big deal if the actual time varies by +/- 1 ms.

src/utils/windows_scheduler.zig
```diff
@@ -0,0 +1,97 @@
...
+			.below_normal => "Below Normal",
+			.normal => "Normal",
+			.above_normal => "Above Normal",
+			.high => "High",
+			.realtime => "Realtime",
+		};
+	}
+};
+
+pub fn setPriorityClass(priority: PriorityClass) void {
```
IntegratedQuantum, line 76: I'm not sure if this is a good idea. In general I'd say we should play nicely with the OS, and setting the entire process priority could seriously harm other applications running (e.g. discord) since Cubyz uses a ton of threads. If at all, we should only prioritize the render thread.

src/utils/windows_scheduler.zig
```diff
@@ -0,0 +1,97 @@
...
+	if(SetPriorityClass(windows.GetCurrentProcess(), @intFromEnum(priority)) == 0) {
+		std.log.warn("Failed to set process priority class: {s}'", .{priority.displayName()});
+	} else {
+		std.log.info("Set process priority class: {s}", .{priority.displayName()});
+	}
+}
+
+extern "ntdll" fn NtQueryTimerResolution(MinimumResolution: *windows.ULONG, MaximumResolution: *windows.ULONG, CurrentResolution: *windows.ULONG) callconv(.winapi) windows.NTSTATUS;
```
IntegratedQuantum, line 84: I'd prefer to just import the corresponding headers with a C import. This reduces the chance to make mistakes.

src/utils/windows_scheduler.zig
IntegratedQuantum: I think it would be better to handle the oversleeping case by faking the deltaTime and making the sleep in the next frame shorter to account for the difference.

src/utils/windows_scheduler.zig
MiniBaker007: From what I can tell, the windows implementation of Io.sleep only uses millisecond precision, TBD if that's good enough.

src/utils/windows_scheduler.zig
MiniBaker007: I think doing it only for the render thread sounds like a reasonable compromise. Perhaps it could only be boosted when the FPS is capped, too. This seems pretty standard for video games, from what I can tell.

PR #2480, opened 2026-01-10
Convert ChunkPosition.equals to use guard clauses

Argmaster:


src/chunk.zig
```diff
@@ -166,17 +166,20 @@ pub const ChunkPosition = struct { // MARK: ChunkPosition
...
+
+		if(@TypeOf(other) == ChunkPosition)
 			return self.wx == other.wx and self.wy == other.wy and self.wz == other.wz and self.voxelSize == other.voxelSize;
-		} else if(@TypeOf(other.*) == ServerChunk) {
-			return self.wx == other.super.pos.wx and self.wy == other.super.pos.wy and self.wz == other.super.pos.wz and self.voxelSize == other.super.pos.voxelSize;
-		} else if(@typeInfo(@TypeOf(other)) == .pointer) {
-			return self.wx == other.pos.wx and self.wy == other.pos.wy and self.wz == other.pos.wz and self.voxelSize == other.pos.voxelSize;
-		} else @compileError("Unsupported");
+
+		if(@TypeOf(other.*) == ServerChunk)
```
IntegratedQuantum: I think ifs should either be in a single line, or have parenthesis around the body.

src/chunk.zig
```diff
@@ -166,17 +166,20 @@ pub const ChunkPosition = struct { // MARK: ChunkPosition
...
-			if(other) |notNull| {
-				return self.equals(notNull);
-			}
+			if(other) |notNull| return self.equals(notNull);
 			return false;
-		} else if(@TypeOf(other) == ChunkPosition) {
+		}
+
+		if(@TypeOf(other) == ChunkPosition)
 			return self.wx == other.wx and self.wy == other.wy and self.wz == other.wz and self.voxelSize == other.voxelSize;
```
IntegratedQuantum: How about using `std.meta.eql(self, other)` I think this would also be nicer in the cases below.

Argmaster:
> In general I'm not really happy with the existence of this function. It's only really used by the caching system and maybe we can find a better solution for that use case?

I have already spent some time thinking about it and to no avail. I am currently concerned with different parts of `chunk.zig` and I want that change out of the way.
This PR addresses repetition which is trivial to eliminate, If I come across a better solution down the road, I will surely open a PR.

src/chunk.zig
Argmaster:
```suggestion
			return std.meta.eql(self, other);
```

PR #2482, opened 2026-01-11
Add enum for representing LOD values

Argmaster:
This is an intermediate step PR;

Long term goal is to use Lod enum instead of `ChunkPosition.voxelSize`, `Chunk.width`, `Chunk.voxelSizeShift`, `Chunk.voxelSizeMask` and possibly other values with similar meaning scattered around codebase.

Even long-er term goal is to standardize how we pass around chunk positions, local and global block positions and whatever is passed as random i32/Vec3i coordinates in different contexts. A lot of bit masking is done in various contexts and it is really hard to determine which type of coordinate we are looking at in some random part of the code, without reading long long before and after.

### DIFF
--- a/src/chunk.zig
+++ b/src/chunk.zig
@@ -234,7 +234,7 @@ pub const Lod = enum(u5) {
 	}
 
 	test "Lod.localMask() max" {
-		try std.testing.expectEqual(Lod.max.localMask(), ~@as(i32, 1023));
+		try std.testing.expectEqual(~@as(i32, 1023), Lod.max.localMask());
 	}
 };
 


src/chunk.zig
```diff
@@ -148,6 +148,111 @@ pub fn deinit() void {
...
 
+pub const Lod = enum(u5) {
+	const Self = @This();
+
+	LOD0 = 0,
+	LOD1,
+	LOD2,
+	LOD3,
+	LOD4,
+	LOD5 = 5,
```
IntegratedQuantum: How about taking some inspiration from `std.mem.Alignment` and calling them based on their voxelSize:
```zig
    @"1" = 0,
    @"2" = 1,
    @"4" = 2,
    @"8" = 3,
    @"16" = 4,
    @"32" = 5,
```

src/chunk.zig
```diff
@@ -148,6 +148,111 @@ pub fn deinit() void {
...
+
+	pub inline fn min() Lod {
```
IntegratedQuantum: Constants should be constants, not functions.

src/chunk.zig
```diff
@@ -148,6 +148,111 @@ pub fn deinit() void {
...
+		return @enumFromInt(@typeInfo(Lod).@"enum".fields[0].value);
+	}
+
+	pub inline fn max() Lod {
```
IntegratedQuantum: same here

src/chunk.zig
```diff
@@ -148,6 +148,111 @@ pub fn deinit() void {
...
+		return @intFromEnum(self);
+	}
+
+	pub inline fn voxelSize(self: Lod) u31 {
+		comptime var table: [Lod.max().toInt() + 1]u31 = @splat(0);
+		comptime for(@typeInfo(Lod).@"enum".fields, 0..) |field, i| {
+			const lod = @as(Lod, @enumFromInt(field.value));
+			table[lod.toInt()] = 1 << i;
+		};
+		return table[self.toInt()];
```
IntegratedQuantum: Don't use a lookup table for something that takes one cycle to compute. No need to stress the cache with it.

src/chunk.zig
```diff
@@ -148,6 +148,111 @@ pub fn deinit() void {
...
+		std.debug.assert(Lod.max().voxelSize() == (1 << Lod.max().toInt()));
+	}
+
+	pub inline fn chunkWidth(self: Lod) u31 {
+		comptime var table: [Lod.max().toInt() + 1]u31 = @splat(0);
+		comptime for(@typeInfo(Lod).@"enum".fields) |field| {
+			const lod = @as(Lod, @enumFromInt(field.value));
+			table[lod.toInt()] = lod.voxelSize()*chunkSize;
+		};
+		return table[self.toInt()];
```
IntegratedQuantum: same here

src/chunk.zig
```diff
@@ -148,6 +148,111 @@ pub fn deinit() void {
...
+	}
+	test "Lod.chunkWidth() min" {
```
IntegratedQuantum, line 208: I would prefer to put all the tests at the end, so they don't hinder reading the interface.

src/chunk.zig
IntegratedQuantum: Also note that the compiler will likely have an easier time to optimize this computation when it's known. e.g. with a lookup table it will probably have a harder time to figure out that `*lod.voxelSize()` is that same as `<< lod.toInt()`.
Unless you have a concrete benchmark proving that the lookup table is faster, please don't use one.

src/chunk.zig
```diff
@@ -148,6 +148,111 @@ pub fn deinit() void {
...
+	}
+
+	// Mask for converting global coordinates to Lod resolution coordinates.
+	pub inline fn voxelSizeMask(self: Lod) i32 {
+		comptime var table: [Lod.max().toInt() + 1]i32 = @splat(0);
+		comptime for(@typeInfo(Lod).@"enum".fields) |field| {
+			const lod = @as(Lod, @enumFromInt(field.value));
+			table[lod.toInt()] = ~@as(i32, lod.voxelSize() - 1);
+		};
+		return table[self.toInt()];
```
IntegratedQuantum: even here, applying `& voxelSizeMask()` is a single `bzhi` instruction but with a lookup table the compiler has no chance to make use of that fact. [godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:2,lang:zig,selection:(endColumn:2,endLineNumber:36,positionColumn:2,positionLineNumber:36,selectionStartColumn:2,selectionStartLineNumber:36,startColumn:2,startLineNumber:36),source:'pub+const+Lod+%3D+enum(u5)+%7B%0A%09const+Self+%3D+@This()%3B%0A%0A%09LOD0+%3D+0,%0A%09LOD1,%0A%09LOD2,%0A%09LOD3,%0A%09LOD4,%0A%09LOD5+%3D+5,%0A%7D%3B%0Apub+export+fn+voxelSize(self:+*Lod)+u32+%7B%0A++++return+@as(u31,+1)+%3C%3C+@intFromEnum(self.*)%3B%0A%7D%0Apub+export+fn+voxelSize2(self:+*Lod)+u32+%7B%0A++++comptime+var+table:+%5B6%5Du31+%3D+@splat(0)%3B%0A++++comptime+for(@typeInfo(Lod).@%22enum%22.fields,+0..)+%7Cfield,+i%7C+%7B%0A++++++++const+lod+%3D+@as(Lod,+@enumFromInt(field.value))%3B%0A++++++++table%5B@intFromEnum(lod)%5D+%3D+1+%3C%3C+i%3B%0A++++%7D%3B%0A++++return+table%5B@intFromEnum(self.*)%5D%3B%0A%7D%0Apub+export+fn+voxelSizeMask(self:+*Lod)+i32+%7B%0A++++return+(@as(i32,+1)+%3C%3C+@intFromEnum(self.*))+-+1%3B%0A%7D%0Apub+export+fn+voxelSizeMask2(self:+*Lod)+i32+%7B%0A++++comptime+var+table:+%5B6%5Di32+%3D+@splat(0)%3B%0A++++comptime+for(@typeInfo(Lod).@%22enum%22.fields,+0..)+%7Cfield,+i%7C+%7B%0A++++++++const+lod+%3D+@as(Lod,+@enumFromInt(field.value))%3B%0A++++++++table%5B@intFromEnum(lod)%5D+%3D+(1+%3C%3C+i)+-+1%3B%0A++++%7D%3B%0A++++return+table%5B@intFromEnum(self.*)%5D%3B%0A%7D%0A%0Apub+export+fn+applyMask(self:+*Lod,+wx:+i32)+i32+%7B%0A++++return+wx+%26+voxelSizeMask(self)%3B%0A%7D%0A%0Apub+export+fn+applyMask2(self:+*Lod,+wx:+i32)+i32+%7B%0A++++return+wx+%26+voxelSizeMask2(self)%3B%0A%7D'),l:'5',n:'0',o:'Zig+source+%232',t:'0')),header:(),k:32.526932141232386,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:ztrunk,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:zig,libs:!(),options:'-OReleaseFast',overrides:!(),selection:(endColumn:13,endLineNumber:44,positionColumn:13,positionLineNumber:44,selectionStartColumn:9,selectionStartLineNumber:44,startColumn:9,startLineNumber:44),source:2),l:'5',n:'0',o:'+zig+trunk+(Editor+%232)',t:'0')),header:(),k:34.1397345254343,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'zig+trunk',editorid:2,fontScale:14,fontUsePx:'0',j:2,wrap:'1'),l:'5',n:'0',o:'Output+of+zig+trunk+(Compiler+%232)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

src/chunk.zig
IntegratedQuantum: also in my opinion the tests aren't even particularly good since they just contain a copy of the function's code.

IntegratedQuantum (changes requested):

src/chunk.zig
Argmaster: It was done this way to preserve status quo; I generally agree.

Since we are touching the topic of changing the status quo, I want to warn that this LOD thing is a shitton of lines to change, so I would prefer to change as little as possible at one time. Therefore, whenever possible, I would like to apply changes in follow up PRs rather than changing ones I already submitted.
This one is definitely something to do on this PR, but I already know that in the future I will end up with redundant transformations which would require stacking refactor on a refactor and I don't want to do that.

src/chunk.zig
Argmaster: I will hard code expected values. Tests are necessary.
I wanted to avoid hard coding the expected values since what I really wanted to test was whether I coded the lookup tables right.

src/chunk.zig
Argmaster: Status quo was the value was precomputed and put into field on Chunk object

src/chunk.zig
Argmaster: Is that consistent with what you are advertising to end user?
~~I don't think it is -  I think we are calling it lod0?~~ Edit: Actually, I don't know :)))
There is LOD0.5 in some places? In UI there is in fact LOD1, which is good.

Argmaster: Done as requested

src/chunk.zig
```diff
@@ -148,6 +148,96 @@ pub fn deinit() void {
...
+	pub inline fn voxelSizeShift(self: Lod) u5 {
+		return self.toInt();
+	}
+
+	// Mask for converting global coordinates to Lod resolution coordinates.
+	pub inline fn voxelSizeMask(self: Lod) i32 {
+		return ~@as(i32, self.voxelSize() - 1);
+	}
+
+	// Mask for converting global coordinates to chunk local coordinates.
```
IntegratedQuantum, line 195: From what I understand this a mask that converts to chunk start coordinates (you inverted it)

src/chunk.zig
```diff
@@ -148,6 +148,96 @@ pub fn deinit() void {
...
+
+	test "Lod.voxelSize() max" {
+		std.debug.assert(Lod.max.voxelSize() == 32);
+	}
+
+	test "Lod.chunkWidth() min" {
+		std.debug.assert(Lod.min.chunkWidth() == 32);
+	}
+
+	test "Lod.chunkWidth() max" {
```
IntegratedQuantum, line 212: I would prefer to group small numerical tests of the same function together.

src/chunk.zig
```diff
@@ -148,6 +148,96 @@ pub fn deinit() void {
...
+	@"8" = 3,
+	@"16" = 4,
+	@"32" = 5,
+
+	const min: Lod = @enumFromInt(@typeInfo(Lod).@"enum".fields[0].value);
+	const max: Lod = blk: {
+		const fields = @typeInfo(Lod).@"enum".fields;
+		const maxValue = fields[fields.len - 1].value;
+		break :blk @enumFromInt(maxValue);
+	};
```
IntegratedQuantum, line 164: I think it would be more readable with the direct values `.@"1"` and `.@"32"`
Instead you can add a test that asserts that these are indeed the first and last entries in the enum,

src/chunk.zig
Argmaster: No, they test different edge cases.

src/chunk.zig
Argmaster: Not sure, but I am pretty sure that it should have never existed. I am pretty sure that back then i didn't know what i was doing. This requires more systematical approach and better nomenclature.
I can remove the comment, but the goal is to remove that mask in current form and replace it with intermediate types with better names and conversions between them.

src/chunk.zig
Argmaster: You will (hopefully!) have hundreds, thousands or even more tests regardless of consolidation attempts. 
Consolidation makes results less readable, not more readable.

---

Unless there is a technical reason to change this, I don't think the fact you would write this code differently is a good reason to ask me to change it. Did I miss something technically relevant?

src/chunk.zig
Argmaster: How is that beneficial? Its metaprogramming, we do it so we don't have to fetch max and min manually, not to make it readable. I am pretty sure the fact we won't need to touch it in the future outweigh the readability issue of 3 lines of code.

src/chunk.zig
IntegratedQuantum: > No, they test different edge cases.

I wouldn't call them edge cases, they are testing 2 normal cases that happen all the time, and `@"4"` is just as close to an edge as `@"32"`.

> Consolidation makes results less readable, not more readable.

Here it seems that we disagree. I would say that
```zig
test "Lod.chunkWidth()" {
	std.debug.assert(Lod.min.voxelSize() == 1);
	std.debug.assert(Lod.max.voxelSize() == 32);
}
```
is both locally (my brain has an easier time to see the pattern and filter duplicate information) and globally (less code to navigate through) more readable than
```zig
test "Lod.chunkWidth() min" {
	std.debug.assert(Lod.min.voxelSize() == 1);
}

test "Lod.chunkWidth() max" {
	std.debug.assert(Lod.max.voxelSize() == 32);
}
```

src/chunk.zig
```diff
@@ -148,6 +148,96 @@ pub fn deinit() void {
...
+		return ~@as(i32, self.voxelSize() - 1);
+	}
+
+	// Mask for converting global coordinates to chunk local coordinates.
+	pub inline fn localMask(self: Lod) i32 {
+		return ~@as(i32, self.voxelSize()*chunkSize - 1);
+	}
+
+	test "Lod.voxelSize() min" {
+		std.debug.assert(Lod.min.voxelSize() == 1);
```
IntegratedQuantum: I just noticed that you are using assertions. This is generally harder to debug, since we do not know what value was actually outputed by the function. Please use `std.testing.expectEqual` instead (note that the first argument is the "expected" value, so you have to swap them around).
```suggestion
		try std.testing.expectEqual(1, Lod.min.voxelSize());
```

IntegratedQuantum (changes requested):

src/chunk.zig
Argmaster: > I wouldn't call them edge cases, they are testing 2 normal cases that happen all the time, and `@"4"` is just as close to an edge as `@"32"`.

[An edge case is a problem or situation that occurs only at an extreme (maximum or minimum) operating parameter. ](https://en.wikipedia.org/wiki/Edge_case)
These are edge cases according to definition from Wikipedia. They verify two ends of the spectrum which will fail in different scenarios. First one will fail if we attempt to add a lower LOD and math is not mathing for it, second one will fail for more likely case of adding a higher max LOD. (It might be that language itself prevents those problems to occur, but those test cases are here to confirm that if it doesn't, we can verify our assumptions)

While I cannot be sure which of those is more readable, I can be sure they are verifying different assumptions and their failure will give more information when they fail separate, compared to when they fail consolidated.

Now, this particular case is trivial, but it is the general rule that is important;

Consequence of consolidation is also that, on average, each test will fail more often, so it will require you to read it more often and it will be more text to read every time because the test body is longer.

Some network resources with more nuanced responses to this problem:

https://stackoverflow.com/questions/13130226/unit-testing-do-edge-cases-require-separate-test-methods

https://softwareengineering.stackexchange.com/a/446870

src/chunk.zig
Argmaster: Oh yeah that is a very relevant comment; I guess I had to long of a break from writing tests in zig.

src/chunk.zig
IntegratedQuantum: Well, I'd argue in an exhaustive enum all values are extreme, `@"4"` is extreme, since there exist no entries for `@"5"` or `@"3"`.

But to be honest my main problem here is that this is taking 50 lines of testing code (and as you said, this is code that I only want to read when it fails) in the middle of a rather busy file.

Argmaster: Well, that means the problem is the file itself, not those test cases. Mind that it's nothing unusual for test cases to take same or greater number of lines as the implementation itself.

I would suggest to make attempts on dividing the files instead. Server chunk is a low hanging fruit to evict. 
Chunk position and block position is a material for another file, alongside some clean up in coordinate notation we use. 
Neighbor structure has stopped being chunk only construct long ago, it could be a separate file too. 
At that point there should be a folder with a bunch of files.
Suddenly having double the code volume for each of those is not that scary.

Yes, more code is more maintenance, but less code doesn't always equal less maintenance. Tests verify that what was true about your code back when you wrote it remains true over time. Tests document your code by showing how to use it and what to expect. This is a documentation that never silently goes obsolete. Tests help with refactoring. They are a list of important conditions that have to be true about your code.

---

Technically we could move test cases to mirrored file structure, but I admit I like having those tests nearby. *They are great documentation for verifying if I read the code right.* Additionally this is basically hiding the problem, not solving it. Cubyz source files frequently consolidate a lot of standalone functionality. That functionality would do perfectly fine as separate files.

---

> Well, I'd argue in an exhaustive enum all values are extreme, @"4" is extreme, since there exist no entries for @"5" or @"3".

You are talking about variable names :)
The values are discrete positive integers in range 0 - 5 inclusive. 0 and 5 are the edge cases.

Yes, we explicitly wrote all the values, maybe we need a test that verifies that values are in fact consecutive integers, unless there is a language feature that guarantees this.

---

I feel like a fkn AI chat now how do I always need so many words to explain myself -_-

IntegratedQuantum: Alright, let's leave it for now. I think ideally this would be something that the IDE should do: Fold all blocks of continuous test statements, sadly it doesn't seem like VSCode can do that.

src/chunk.zig
```diff
@@ -148,6 +148,96 @@ pub fn deinit() void {
...
+	test "Lod.voxelSizeMask() max" {
+		try std.testing.expectEqual(~@as(i32, 0b11111), Lod.max.voxelSizeMask());
+	}
+
+	test "Lod.localMask() min" {
+		try std.testing.expectEqual(~@as(i32, 31), Lod.min.localMask());
+	}
+
+	test "Lod.localMask() max" {
+		try std.testing.expectEqual(Lod.max.localMask(), ~@as(i32, 1023));
```
IntegratedQuantum: you forgot to swap it here.

Argmaster:
```suggestion
		try std.testing.expectEqual(~@as(i32, 1023), Lod.max.localMask());
```

Issue #2486, opened 2026-01-12
Break some blocks when their support block breaks


IntegratedQuantum:
- [x] normal flowers/grass
- [x] carpet flowers
- [x] piles
- [ ] lilypads and duckweed (water based plants, require special care)
- [x] vines
- [ ] ivy (should it act like vines, or like carpet flowers that break when breaking their support)
- [ ] signs, should signs break?
- [ ] what about lanterns?

PR #2487, opened 2026-01-12
Break simple flowers when their support block breaks

IntegratedQuantum:
progress towards #2486

this does raise some problems:
- [x] ~~while it does decay in unsupported positions (placed on top of another flower), in can be placed in these positions without a problem.~~ moved to #2489 #2491
- [x] item drop count is incorrect, noticable for the pile rotation mode, but also affects torches and carpets, at least theoretically.

### DIFF
--- a/src/callbacks/block/server/checkSupportBlocks.zig
+++ b/src/callbacks/block/server/checkSupportBlocks.zig
@@ -43,20 +43,23 @@ pub fn run(_: *@This(), params: main.callbacks.ServerBlockCallback.Params) main.
 	if(newBlock == params.block) return .ignored;
 
 	if(main.server.world.?.cmpxchgBlock(wx, wy, wz, params.block, newBlock) == null) {
+		const dropAmount = params.block.mode().itemDropsOnChange(params.block, newBlock);
 		const drops = params.block.blockDrops();
-		for(drops) |drop| {
-			if(drop.chance == 1 or main.random.nextFloat(&main.seed) < drop.chance) {
-				for(drop.items) |stack| {
-					var dir = main.vec.normalize(main.random.nextFloatVectorSigned(3, &main.seed));
-					// Bias upwards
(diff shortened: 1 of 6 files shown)


src/callbacks/block/server/checkSupportBlocks.zig
```diff
@@ -30,10 +30,14 @@ pub fn run(_: *@This(), params: main.callbacks.ServerBlockCallback.Params) main.
 
 	var newBlock: Block = params.block;
 
-	if(params.block.mode() == main.rotation.getByID("cubyz:torch")) {
-		main.rotation.list.@"cubyz:torch".updateBlockFromNeighborConnectivity(&newBlock, neighborSupportive);
-	} else {
-		std.log.err("Expected {s} to have cubyz:torch as rotation", .{params.block.id()});
+	inline for(comptime std.meta.declarations(main.rotation.list)) |rotationMode| {
```
Argmaster, line 33: I think that if we index-ize rotation modes we can do this in O(1) time. 
Add `index` to VTable, 
At build time make a list of rotation mode modules.
At runtime when initing VTable write rotation mode index.
Add getModule() comptime function, you are good to go.

Argmaster: Never mind, compiler won't be happy with indexing comptime table with runtime index XD

Argmaster: Maybe something along those lines this could be used for mods to make their own VTables?
So, we keep the index part, but instead use it to index table containing non-rotation-mode VTables containing our own interfaces.

Argmaster: Am I reinventing inheritance.

IntegratedQuantum: I think maybe we just need an enum instead of a VTable.

PR #2492, opened 2026-01-14
Sulfuric crater biome

codemob-dev:
Some images from various points in development:
image
image
image
image
image
image

### DIFF
--- a/assets/cubyz/biomes/rare/sulfuric_crater/_defaults.zig.zon
+++ b/assets/cubyz/biomes/rare/sulfuric_crater/_defaults.zig.zon
@@ -1,0 +1,13 @@
+.{
+	.properties = .{
+		.hot,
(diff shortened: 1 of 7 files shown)


ikabod-kee:
I feel like this could wait until we have more advanced biome features. The block palette could also use refinement as right now it honestly looks like a garbage waste.

ikabod-kee:
Now that I think about it, a garbage waste biome would go incredibly hard.... I'll have to keep this in mind.

Wbjitscool:
Interesting looking biome hmm

codemob-dev:
Closing this as it didn't turn out all that great

Wbjitscool:
> Closing this as it didn't turn out all that great

ah okay maybe there could be a different biome that looks similar but better not sure?

PR #2506, opened 2026-01-19
Introduce new Sync command `CraftFrom` and make the crafting inventories client side only

### LINKED ISSUE #2497: Inventory crafting should first take items from partial stacks instead of always targettting the first stack

Wunka:
Extracted from: https://github.com/PixelGuys/Cubyz/pull/2500#issue-3826914352

Fixes: #2497

- [ ] fix the check for `deserialize` (As already mentioned in https://github.com/PixelGuys/Cubyz/pull/2469#discussion_r2679048179 I am not that knowledgable on the reader code so if anybody has a better idea here, it would be very welcome)
- [x] find out if any mechanic was removed (what scenario made Drop call tryCrafting?) 

This PR also replaces #2483 as its features are also done here.
`CraftFrom` accepts as source a slice to prepare for #2324  or/and making the user able to craft from items in a chest


Argmaster:
Ok, so I am getting a bit confused by the description. What exactly is getting client sided here? Recipe resolution right? Not whole item transformation?
I admit my knowledge is a little bit outdated due to recent refactors I didn't have time to review, so please forgive my stupid questions.

Wunka:
> Ok, so I am getting a bit confused by the description. What exactly is getting client sided here? Recipe resolution right? Not whole item transformation? I admit my knowledge is a little bit outdated due to recent refactors I didn't have time to review, so please forgive my stupid questions.

For reference see: #2500 . There Quantum introduced the ClientInventory this can either be synced with a Normal inventory on the server or can be completly be handeld on client side. This PR introduces the CraftFrom sync so that the server doesn't have to have access to the crafting inventories (which are displayed in the crafting menu). Which makes it possible to make them also only client side as is the creative menu through #2500

Argmaster:
Aight, since you are clearly keeping up with the topic, could you please answer for sake of documentation: Can the client invoke arbitrary item transformation that was not defined by server side recipe table? Will the server reject such transformation?

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+					if(recipe.sourceItems.len != sourceStacksSize) continue;
+					for(recipe.sourceItems, recipe.sourceAmounts, sourceStacks) |recipeItem, recipeAmount, sourceStack| {
+						if(recipeItem != sourceStack.item.baseItem) continue :outer;
+						if(recipeAmount != sourceStack.amount) continue :outer;
+					}
+					break :blk true;
+				}
+				break :blk false;
+			};
+			if(!found) return error.Invalid;
```
Wunka: This looks if the recipe actually exist, but I have not yet tested if there is a flaw in my logic here.

Argmaster: No worries, I am just trying to understand the intentions

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+			}
+			self.resultStack.toBytes(writer);
+			writer.writeVarInt(usize, self.sourceStacks.len);
+			for(self.sourceStacks) |sourceStack| {
+				sourceStack.toBytes(writer);
+			}
+		}
+
+		fn deserialize(reader: *BinaryReader, side: Side, user: ?*main.server.User) !CraftFrom {
+			const destinationsSize = try reader.readVarInt(usize);
```
Argmaster: Oh yes, varInt. (This comment mainly to addressed @IntegratedQuantum) How often is this message send so this varInt makes sense here? It's saving 6? 7? Bytes. Out of how large impact over eg. Hour of average gameplay? Anyway, my skepticism aside, shouldn't we make the enums varInts too? Since we care about those few bytes, I am almost sure we are wasting double that when all the entries are summed up.

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+				ctx.execute(.{.create = .{
+					.dest = .{.inv = selectedEmptyInv.?, .slot = selectedEmptySlot.?},
+					.amount = remainingAmount,
+					.item = self.resultStack.item,
+				}});
+			}
+		}
+
+		fn serialize(self: CraftFrom, writer: *BinaryWriter) void {
+			writer.writeVarInt(usize, self.destinations.len);
```
Argmaster: Think we should consider more realistic limits than usize. Range of InventoryIds is u32 if I recall correctly. Even then it's not possible for all of the lists to have that length. This is also potential DOS vector, since apparently we started worrying about server crashing.

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+			main.globalAllocator.free(self.sources);
+			main.globalAllocator.free(self.sourceStacks);
+		}
+
+		fn run(self: CraftFrom, ctx: Context) error{serverFailure}!void {
+			for(self.destinations) |dest| if(dest.type != .normal) return;
+			for(self.sources) |source| if(source.type != .normal) return;
+
+			// Can we even craft it?
+			outer: for(self.sourceStacks) |requiredStack| {
```
Argmaster: I suspect we have to wait with improvements for next PR?
This cries for a helper function. Abstraction level in here is too low and the function is too long overall.
This is clearly a separate function that would be better off with lower indentation level and a proper name, rather than "outer" label.

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+			for(sources) |*source| {
+				const invId = try reader.readEnum(InventoryId);
+				source.* = Inventory.getInventory(invId, side, user) orelse return error.InventoryNotFound;
+			}
+
+			const resultStack = try ItemStack.fromBytes(reader);
+			const sourceStacksSize = try reader.readVarInt(usize);
+			if(sourceStacksSize == 0) return error.Invalid;
+			std.log.info("sourceStacksSize: {d}, sourceStacksSize*@sizeOf(ItemStack): {d}, remaining: {d}\n", .{sourceStacksSize, sourceStacksSize*@sizeOf(ItemStack), reader.remaining.len});
+			// if(sourceStacksSize*@sizeOf(ItemStack) > reader.remaining.len) return error.Invalid; // this check sadly fails
```
Argmaster: Umm >=?

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+				writer.writeEnum(InventoryId, source.id);
+			}
+			self.resultStack.toBytes(writer);
+			writer.writeVarInt(usize, self.sourceStacks.len);
+			for(self.sourceStacks) |sourceStack| {
+				sourceStack.toBytes(writer);
+			}
+		}
+
+		fn deserialize(reader: *BinaryReader, side: Side, user: ?*main.server.User) !CraftFrom {
```
Argmaster: ~~I think we could unit test this fairly easily. We would have to be naught and add `mock` entry to Side enums that result in getInventory being noop and reserve a sentinel InventoryId, but we could get serialization edge case coverage from that. Alternatively we can just implement declarative serialization C:~~ nvm we can't easily make a dymmy CraftFrom object.

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+					break :outer;
+				}
+			}
+			if(remainingAmount > 0 and selectedEmptySlot != null) {
+				ctx.execute(.{.create = .{
+					.dest = .{.inv = selectedEmptyInv.?, .slot = selectedEmptySlot.?},
+					.amount = remainingAmount,
+					.item = self.resultStack.item,
+				}});
+			}
```
Wunka: Just wanted to mention it here. I do want probably make this a helper function as the only difference to example the same code from `DepositToAny` is that this uses a `ctx.execute(.{create` while `DepositToAny` uses `ctx.execute(.{.move`

src/sync.zig
Argmaster: Btw this is `destinationCount` (possibly with plural destinations), size has to strong size-in-bytes connotations in this context.

src/sync.zig
Wunka: When writting the code first I changed to > because this is the last read so if its maybe = then this would still be ok.

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+			writer.writeVarInt(usize, self.sourceStacks.len);
+			for(self.sourceStacks) |sourceStack| {
+				sourceStack.toBytes(writer);
+			}
+		}
+
+		fn deserialize(reader: *BinaryReader, side: Side, user: ?*main.server.User) !CraftFrom {
+			const destinationsSize = try reader.readVarInt(usize);
+			if(destinationsSize == 0) return error.Invalid;
+			if(destinationsSize*@sizeOf(InventoryId) >= reader.remaining.len) return error.Invalid;
```
Argmaster: Do we have to check this ahead of time? We can just for loop for destinationCount times and try reading. Reader should report error when it runs out of buffer data I think? We then catch it and possibly transform to Invalid or whatever.

src/sync.zig
Argmaster: I think we should cap it around 32k / 64k assuming we want to support crafting from nearby chests. Otherwise I can't see this exceeding few inventories. For the nearby chests I think a +-16 block radius is a fair assumption.

src/sync.zig
Argmaster: Ye without a proper abstraction level adjustment I am not reviewing that, but I know a person who will.

src/sync.zig
Argmaster: Anyway my suggestion is to read until it fails and hard cap max count we can process, as described In nearby comment.

src/sync.zig
```diff
@@ -988,16 +888,221 @@ pub const Command = struct { // MARK: Command
...
+			// if(sourceStacksSize*@sizeOf(ItemStack) > reader.remaining.len) return error.Invalid; // this check sadly fails
+
+			const sourceStacks = main.globalAllocator.alloc(ItemStack, sourceStacksSize);
+			errdefer main.globalAllocator.free(sourceStacks);
+
+			for(sourceStacks) |*sourceStack| {
+				sourceStack.* = try ItemStack.fromBytes(reader);
+			}
+
+			// check if recipe is valid
```
Argmaster: Very pretty comment. Now make it a function name.

src/sync.zig
Wunka: Comes from https://github.com/PixelGuys/Cubyz/pull/2469/changes#r2679023133

Argmaster: Well, then I guess it's bad I missed that one. Suggestion still remains valid IMO, unless proven otherwise.

Argmaster: Tho I admit that IMO a best solution would be an automagical serialization for structs, but Quantum disagrees.

src/sync.zig
IntegratedQuantum: Well, let me point you to the [contributing guidelines](https://github.com/PixelGuys/Cubyz/blob/master/docs/CONTRIBUTING.md#a-note-on-performance-optimizations), using varint is non-pessimization. Of course in this case it makes no difference, but if we save a few bytes in each packet, then suddenly we are looking at different dimensions.
And there is no reason not to use varint, it's just 3 characters more to write.

You are also correct about the enums here, since they (and all non-exhaustive enums) are indices, it would make sense to serialize them as varints as well, and this might actually make a meaningful enough difference for item drop/entitiy/projectile ids. So maybe we should think about automatically doing this in the background.

src/sync.zig
IntegratedQuantum: Adding hardcoded buffer size limits is just annoying to work with, you need that extra intCast at the source, and you need to argue about what limit to choose, and then document why you chose it.
I think we should address the root cause instead by having a limit on the network buffers.

src/sync.zig
Wunka: moved

src/sync.zig
Wunka: I did create a helper with #2515

Wunka:
- I have updated this to use the `put_items_into`.
- I also looked what scenario made `Drop` call `tryCraftingTo` and I didn't find any scenario, so the remove seems to not remove any mechanic
- I see for now no conclusion to the `deserialize` problem, so I didn't change anything there.
- In my opinion the run function from CraftFrom has now engough abstraction but if there are different opinions I can move more code to helper functions (There is also this comment: https://github.com/PixelGuys/Cubyz/pull/2515#discussion_r2722647921 so they may be changes to this order in the future in a follow up PR)

Wunka:
while rebasing to the new format changes, it seems I choose a wrong option and I saw that to late... but I think I have now recovered all changes which through that disappeared.

src/Inventory.zig
```diff
@@ -475,6 +467,23 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+		std.debug.assert(craftingInv.type == .crafting);
+
+		if (slot != craftingInv.super._items.len - 1) return;
+		const destinationsCanHold = blk: {
+			for (destinations) |dest| {
+				if (dest.canHold(craftingInv.getStack(slot))) break :blk true;
+			}
+			break :blk false;
+		};
+		if (!destinationsCanHold) return;
```
IntegratedQuantum: Do we really need this? It's already checked in the inventory command anyways

src/gui/windows/inventory_crafting.zig
```diff
@@ -92,9 +92,18 @@ fn findAvailableRecipes(list: *VerticalList) bool {
 		}
 		// All ingredients found: Add it to the list.
 		if (recipe.cachedInventory == null) {
-			recipe.cachedInventory = ClientInventory.init(main.globalAllocator, recipe.sourceItems.len + 1, .crafting, .serverShared, .{.recipe = recipe}, .{});
+			recipe.cachedInventory = ClientInventory.init(main.globalAllocator, recipe.sourceItems.len + 1, .normal, .crafting, .other, .{});
```
IntegratedQuantum: The caching was mainly to prevent redownloading these from the server every time you open the window.
Now this isn't needed and I'd prefer to remove caching and always reopen it.

src/sync.zig
```diff
@@ -248,6 +249,7 @@ pub const Command = struct { // MARK: Command
 		updateBlock: UpdateBlock,
 		addHealth: AddHealth,
 		chatCommand: ChatCommand,
+		craftFrom: CraftFrom,
```
IntegratedQuantum: Please sort it into the block of pure inventory actions

src/sync.zig
```diff
@@ -779,6 +781,22 @@ pub const Command = struct { // MARK: Command
 		};
 	}
 
+	fn validRecipe(sourceStacks: []ItemStack, resultStack: ItemStack) bool {
```
IntegratedQuantum: Since this is only used by the crafting command, please put it inside of its namespace.

src/sync.zig
```diff
@@ -1054,16 +970,162 @@ pub const Command = struct { // MARK: Command
 		}
 	};
 
+	const CraftFrom = struct { // MARK: CraftFrom
+		destinations: []const Inventory,
+		sources: []const Inventory,
+		resultStack: ItemStack,
+		sourceStacks: []const ItemStack,
```
IntegratedQuantum: Should contain the whole `Recipe`, and not reiterate its entries.

src/sync.zig
IntegratedQuantum: Or I think it would be better to just add a deserialize (and serialize) method to the `Recipe` struct that deserializes and validates the recipe.

IntegratedQuantum (changes requested):

src/Inventory.zig
Wunka: can you elaborate where this is already done?

src/gui/windows/inventory_crafting.zig
Wunka: done

src/sync.zig
Wunka: done. Now `CraftFrom` just has a pointer to a Recipe from the recipes list.

src/Inventory.zig
IntegratedQuantum: I guess it's missing then.
It needs to be done in the Inventory command, otherwise it may break if e.g. the player collects and item before the server receives the packet.

Wunka: I moved it now into `CraftFrom`. Is this so right or do I need to return an error?

IntegratedQuantum: No, the error is only for kicking people, we shouldn't kick people on valid race conditions (like crafting while collecting items)

src/sync.zig
```diff
@@ -1054,16 +954,149 @@ pub const Command = struct { // MARK: Command
 		}
 	};
 
+	const CraftFrom = struct { // MARK: CraftFrom
```
IntegratedQuantum: The structs should have the same order as in the enum above

src/sync.zig
```diff
@@ -1054,16 +954,149 @@ pub const Command = struct { // MARK: Command
...
+
+		pub fn init(destinations: []const Inventory.ClientInventory, sources: []const Inventory.ClientInventory, resultStack: ItemStack, sourceStacks: []const ItemStack) CraftFrom {
+			const destinationsCopy = main.globalAllocator.alloc(Inventory, destinations.len);
+			for (destinationsCopy, destinations) |*d, s| d.* = s.super;
+			const sourcesCopy = main.globalAllocator.alloc(Inventory, sources.len);
+			for (sourcesCopy, sources) |*d, s| d.* = s.super;
+			return .{
+				.destinations = destinationsCopy,
+				.sources = sourcesCopy,
+				.recipe = main.items.Recipe.getValidRecipeThroughStacks(sourceStacks, resultStack) catch unreachable,
```
IntegratedQuantum: the recipe should be passed in, the inventory should know what recipe it is. This search is pointless.

src/sync.zig
```diff
@@ -1054,16 +954,149 @@ pub const Command = struct { // MARK: Command
...
+			main.globalAllocator.free(self.sources);
+		}
+
+		fn run(self: CraftFrom, ctx: Context) error{serverFailure}!void {
+			for (self.destinations) |dest| if (dest.type != .normal) return;
+			for (self.sources) |source| if (source.type != .normal) return;
+
+			const destinationsCanHold = blk: {
+				for (self.destinations) |dest| {
+					if (dest.canHold(.{.item = .{.baseItem = self.recipe.resultItem}, .amount = self.recipe.resultAmount})) break :blk true;
```
IntegratedQuantum: This doesn't account for the fact that we could split the stack onto multiple partially filled slots from separate inventories.

src/sync.zig
```diff
@@ -1054,16 +954,149 @@ pub const Command = struct { // MARK: Command
...
+				// There might be duplicate entries:
+				for (self.recipe.sourceItems, self.recipe.sourceAmounts) |otherItem, otherAmount| {
+					if (std.meta.eql(requiredItem, otherItem))
+						amount += otherAmount;
+				}
+				for (self.sources) |source| {
+					for (source._items) |otherStack| {
+						if (otherStack.item != .null and std.meta.eql(requiredItem, otherStack.item.baseItem)) {
+							amount -|= otherStack.amount;
+							if (amount == 0) break :outer;
```
IntegratedQuantum: I think this should continue not break.

src/sync.zig
```diff
@@ -1054,16 +954,149 @@ pub const Command = struct { // MARK: Command
...
+						}
+					}
+				}
+				// Not enough ingredients
+				if (amount != 0)
+					return;
```
IntegratedQuantum:
```suggestion
				if (amount != 0) return;
```

src/sync.zig
```diff
@@ -1054,16 +954,149 @@ pub const Command = struct { // MARK: Command
...
+				}
+				if (remainingAmount > 0 and fullSlot != null) {
+					ctx.execute(.{.delete = .{
+						.source = .{.inv = fullInv.?, .slot = fullSlot.?},
+						.amount = @min(remainingAmount, requiredItem.stackSize()),
+					}});
+				}
+			}
+
+			put_items_into.do(ctx, self.destinations, self.recipe.resultAmount, .{.create = .{.baseItem = self.recipe.resultItem}});
```
IntegratedQuantum: I think we should have this return the leftover items, so we can assert that there are none.

src/sync.zig
```diff
@@ -1054,16 +954,149 @@ pub const Command = struct { // MARK: Command
...
+			outer: for (self.recipe.sourceItems, self.recipe.sourceAmounts) |requiredItem, requiredAmount| {
+				var fullSlot: ?u32 = null;
+				var fullInv: ?Inventory = null;
+				var remainingAmount: usize = requiredAmount;
+				for (self.sources) |source| {
+					for (source._items, 0..) |*otherStack, i| {
+						if (otherStack.item != .null and std.meta.eql(requiredItem, otherStack.item.baseItem)) {
+							if (otherStack.amount == otherStack.item.stackSize()) {
+								if (fullSlot == null) {
+									fullSlot = @intCast(i);
```
IntegratedQuantum: Unlike for adding items, removing them should prioritize the last slot in each inventory.

src/gui/windows/inventory_crafting.zig
```diff
@@ -91,10 +94,15 @@ fn findAvailableRecipes(list: *VerticalList) bool {
 			continue :outer; // Ingredient not found.
 		}
 		// All ingredients found: Add it to the list.
-		if (recipe.cachedInventory == null) {
-			recipe.cachedInventory = ClientInventory.init(main.globalAllocator, recipe.sourceItems.len + 1, .crafting, .serverShared, .{.recipe = recipe}, .{});
+		const inv = ClientInventory.init(main.globalAllocator, recipe.sourceItems.len + 1, .normal, .crafting, .other, .{});
```
IntegratedQuantum: Here the recipe should be added to the type information `.{.crafting = recipe}` (see other comment)

src/items.zig
```diff
@@ -1061,7 +1061,58 @@ pub const Recipe = struct { // MARK: Recipe
 	sourceAmounts: []u16,
 	resultItem: BaseItemIndex,
 	resultAmount: u16,
-	cachedInventory: ?Inventory.ClientInventory = null,
+
+	pub fn toBytes(self: *const Recipe, writer: *BinaryWriter) void {
+		writer.writeEnum(BaseItemIndex, self.resultItem);
+		writer.writeInt(u16, self.resultAmount);
```
IntegratedQuantum: writeVarInt

src/items.zig
```diff
@@ -1061,7 +1061,58 @@ pub const Recipe = struct { // MARK: Recipe
...
+		writer.writeVarInt(usize, self.sourceItems.len);
+		for (self.sourceItems, self.sourceAmounts) |item, amount| {
+			writer.writeEnum(BaseItemIndex, item);
+			writer.writeInt(u16, amount);
```
IntegratedQuantum: writeVarInt

src/items.zig
```diff
@@ -1061,7 +1061,58 @@ pub const Recipe = struct { // MARK: Recipe
...
+		const sourceAmounts = main.stackAllocator.alloc(u16, sourceCount);
+		defer main.stackAllocator.free(sourceAmounts);
+		for (sourceItems, sourceAmounts) |*item, *amount| {
+			item.* = try reader.readEnum(BaseItemIndex);
+			amount.* = try reader.readInt(u16);
+		}
+		return getValidRecipe(.{.sourceItems = sourceItems, .sourceAmounts = sourceAmounts, .resultItem = resultItem, .resultAmount = resultAmount});
+	}
+
+	pub fn getValidRecipeThroughStacks(sourceStacks: []const ItemStack, resultStack: ItemStack) error{Invalid}!*Recipe {
```
IntegratedQuantum: Can be removed (see other comment)

src/items.zig
```diff
@@ -1061,7 +1061,58 @@ pub const Recipe = struct { // MARK: Recipe
...
+				if (recipeAmount != selfAmount) continue :outer;
+			}
+			return recipe;
+		}
+		return error.Invalid;
+	}
+
+	pub fn fromBytes(reader: *BinaryReader) !*Recipe {
+		const resultItem = try reader.readEnum(BaseItemIndex);
+		const resultAmount = try reader.readInt(u16);
```
IntegratedQuantum: varint

src/items.zig
```diff
@@ -1061,7 +1061,58 @@ pub const Recipe = struct { // MARK: Recipe
...
+		const sourceCount = try reader.readVarInt(usize);
+
+		const sourceItems = main.stackAllocator.alloc(BaseItemIndex, sourceCount);
+		defer main.stackAllocator.free(sourceItems);
+		const sourceAmounts = main.stackAllocator.alloc(u16, sourceCount);
+		defer main.stackAllocator.free(sourceAmounts);
+		for (sourceItems, sourceAmounts) |*item, *amount| {
+			item.* = try reader.readEnum(BaseItemIndex);
+			amount.* = try reader.readInt(u16);
```
IntegratedQuantum: varint

IntegratedQuantum (changes requested):

Wunka:
all done.

src/sync.zig
```diff
@@ -958,22 +899,14 @@ pub const Command = struct { // MARK: Command
...
-						writer.writeInt(u16, val.sourceAmounts[i]);
-						writer.writeWithDelimiter(val.sourceItems[i].id(), 0);
-					}
-				},
 				.blockInventory => |val| {
 					writer.writeVec(Vec3i, val);
 				},
 				.other => {},
-				.alreadyFreed => unreachable,
+				.alreadyFreed, .recipe => unreachable,
```
IntegratedQuantum: Please don't just use the source again, this is shared without the server and adds another edge case to the logic that deals with it.
Please instead put this into the client-only `Type` by making it a union. (And we'll probably need to do a similar thing when it comes to removing the workbench inventory type)

src/sync.zig
```diff
@@ -779,6 +781,15 @@ pub const Command = struct { // MARK: Command
 		};
 	}
 
+	fn destinationsCanHold(destinations: []const Inventory, itemStack: ItemStack) bool {
```
IntegratedQuantum: What we really want to do here, is test if we can perform the below `put_items_into` action. So maybe it would be best to namespace it in there instead of the global scope.

src/sync.zig
```diff
@@ -779,6 +781,15 @@ pub const Command = struct { // MARK: Command
...
+		var remainingAmount = itemStack.amount;
+		for (destinations) |dest| {
+			remainingAmount -|= dest.getPossibleHoldingAmount(itemStack);
+			if (remainingAmount == 0) return true;
+		}
+		return false;
+	}
+
 	const put_items_into = struct {
```
IntegratedQuantum: I also noticed this earlier: You forgot to add a `MARK` section here, which makes it hard to find.

src/sync.zig
```diff
@@ -1471,11 +1348,143 @@ pub const Command = struct { // MARK: Command
...
+				if (amount != 0) return;
+			}
+
+			// Craft it
+			outer: for (self.recipe.sourceItems, self.recipe.sourceAmounts) |requiredItem, requiredAmount| {
+				var fullSlot: ?u32 = null;
+				var fullInv: ?Inventory = null;
+				var remainingAmount: usize = requiredAmount;
+				for (self.sources) |source| {
+					for (0..source._items.len) |r| {
```
IntegratedQuantum: `i` is only allowed to be single-letter because it has a clear convention. No such convention exists for `r`.
```suggestion
					for (0..source._items.len) |reverseIndex| {
```

src/Inventory.zig
```diff
@@ -601,18 +603,22 @@ pub fn getAmount(self: Inventory, slot: usize) u16 {
 	return self._items[slot].amount;
 }
 
-pub fn canHold(self: Inventory, sourceStack: ItemStack) bool {
-	if (sourceStack.amount == 0) return true;
+pub fn getPossibleHoldingAmount(self: Inventory, sourceStack: ItemStack) u16 {
```
IntegratedQuantum: This name is different from what the function is doing.
Getting the possible holding amount (apart from the fact that this name is rather odd, like what is a "holding amount" supposed to be?) is different from what you are doing here, which is return the `min(possible amount, sourceStack.amount)`

IntegratedQuantum: Maybe it would be best solved by changing the original function return type and keeping the `canHold` name:
`union { yes: void, remainingAmount: u16 }`

src/items.zig
```diff
@@ -1061,7 +1061,46 @@ pub const Recipe = struct { // MARK: Recipe
...
+			for (recipe.sourceItems, recipe.sourceAmounts, self.sourceItems, self.sourceAmounts) |recipeItem, recipeAmount, selfItem, selfAmount| {
+				if (recipeItem != selfItem) continue :outer;
+				if (recipeAmount != selfAmount) continue :outer;
+			}
+			return recipe;
+		}
+		return error.Invalid;
+	}
+
+	pub fn fromBytes(reader: *BinaryReader) !*Recipe {
```
IntegratedQuantum, line 1089: Also just noticed: serialization and deserialization functions should always be declared right next to each other.

IntegratedQuantum (changes requested):

src/Inventory.zig
Wunka: I like that. destinationsCanHold also now has that return type for when that is needed. Because they use the same return Type I have now put destinationsCanHold to canHold, but I also like your idea to put it into the namespace... but then canHold should also be there?

src/sync.zig
Wunka: I myself don't use them currently so I forgot it. Added!

src/sync.zig
Wunka: done

src/Inventory.zig
```diff
@@ -601,18 +599,35 @@ pub fn getAmount(self: Inventory, slot: usize) u16 {
...
-	return false;
+	return .{.remainingAmount = sourceStack.amount - remainingAmount};
+}
+
+pub fn destinationsCanHold(destinations: []const Inventory, itemStack: ItemStack) canHoldReturn {
+	var remainingAmount = itemStack.amount;
+	for (destinations) |dest| {
+		remainingAmount -|= switch (dest.canHold(itemStack)) {
+			.yes => return .yes,
+			.remainingAmount => |amount| amount,
```
IntegratedQuantum: This looks wrong. It's subtracting the remainingAmount from the previous remaining amount.

src/Inventory.zig
```diff
@@ -601,18 +599,35 @@ pub fn getAmount(self: Inventory, slot: usize) u16 {
 	return self._items[slot].amount;
 }
 
-pub fn canHold(self: Inventory, sourceStack: ItemStack) bool {
-	if (sourceStack.amount == 0) return true;
+const canHoldReturn = union(enum) {
```
```suggestion
const CanHoldReturn = union(enum) {
```

src/sync.zig
```diff
@@ -1471,11 +1339,143 @@ pub const Command = struct { // MARK: Command
...
+
+			if (Inventory.destinationsCanHold(self.destinations, .{.item = .{.baseItem = self.recipe.resultItem}, .amount = self.recipe.resultAmount}) != .yes) return;
+
+			// Can we even craft it?
+			outer: for (self.recipe.sourceItems) |requiredItem| {
+				var amount: usize = 0;
+				// There might be duplicate entries:
+				for (self.recipe.sourceItems, self.recipe.sourceAmounts) |otherItem, otherAmount| {
+					if (std.meta.eql(requiredItem, otherItem))
+						amount += otherAmount;
```
IntegratedQuantum:
```suggestion
					if (requiredItem == otherItem) amount += otherAmount;
```

src/sync.zig
```diff
@@ -1471,11 +1339,143 @@ pub const Command = struct { // MARK: Command
...
+							if (remainingAmount == 0) continue :outer;
+						}
+					}
+				}
+				if (remainingAmount > 0 and fullSlot != null) {
+					ctx.execute(.{.delete = .{
+						.source = .{.inv = fullInv.?, .slot = fullSlot.?},
+						.amount = @min(remainingAmount, requiredItem.stackSize()),
+					}});
+				}
```
IntegratedQuantum: I think it might sense to put this into a helper as well. While we may not need it outside of crafting it would make this code easier to read and have a more consistent level of abstractions:
```zig
self.recipe.sourceItems, self.recipe.sourceAmounts) |requiredItem, requiredAmount| {
	remove_items_from.do(...);
}
put_items_into.do(...);
```

src/Inventory.zig
```diff
@@ -601,18 +599,35 @@ pub fn getAmount(self: Inventory, slot: usize) u16 {
...
 			remainingAmount -= amount;
-			if (remainingAmount == 0) return true;
+			if (remainingAmount == 0) return .yes;
 		}
 	}
-	return false;
+	return .{.remainingAmount = sourceStack.amount - remainingAmount};
+}
+
+pub fn destinationsCanHold(destinations: []const Inventory, itemStack: ItemStack) canHoldReturn {
```
IntegratedQuantum: I'm also still not happy with this name, and seeing all the duplicated serialization logic below, I would suggest the following:
How about we add a struct for a group of inventories, it would then have member functions for `canHold`, `putItemsInto`, `removeItemsFrom`, `hasItems` as well as all the serialization logic.

IntegratedQuantum (changes requested):

src/Inventory.zig
Wunka: everything other than this is done. For this I put the general structure in but didn't use it anywhere yet. To get your first opinion of it, if that is what you imagined. The namespaces are probably then not needed anymore or do you want the member functions to call the namespace `do` ?

IntegratedQuantum: Yeah, that's a good start. Though I think it should be in Inventory.zig since it all exclusively relates to inventories and there is no restriction that requires it to be in sync.zig

IntegratedQuantum: And yeah, the namespaces aren't needed

Wunka: moved to Inventory.zig and made `CraftFrom` use it. I do have to say I had to make a lot `pub` because of the namespaces which use a few sync structs / functions. 
The other sync commands are for now not migrated.

IntegratedQuantum: > The other sync commands are for now not migrated.

Since the PR is also getting uncomfortably long, maybe you could introduce this type in a separate PR first, and implement migrate these other functions there directly.

Wunka: created with: #2524

Wunka:
this is now updated to use #2524. 
I also want to mention that I had an idea. The last read where I get the problems for the check is about the recipe. But we know all recipes, so should I maybe just add a variable to `recipe.zig` for `maxRecipeSourceCount` wich while loading the recipes tracks the max amount then I can just compare it to the user given source amount. I think it probably needs a better solution in the future (at least for the inventory fromBytes) but that would at least solve for now the problem that I have no check there.

src/Inventory.zig
```diff
@@ -795,4 +793,38 @@ pub const Inventories = struct { // MARK: Inventories
...
+
+	pub fn removeItems(self: Inventories, ctx: sync.Command.Context, itemAmount: u16, baseItem: main.items.BaseItemIndex) void {
+		var fullSlot: ?u32 = null;
+		var fullInv: ?Inventory = null;
+		var remainingAmount: usize = itemAmount;
+		for (self.inventories) |source| {
+			for (0..source._items.len) |reverseIndex| {
+				const i: usize = source._items.len - reverseIndex - 1;
+				const otherStack: *ItemStack = &source._items[i];
+				if (otherStack.item != .null and baseItem == otherStack.item.baseItem) {
```
```suggestion
				if (otherStack.item == .baseItem and baseItem == otherStack.item.baseItem) {
```
Otherwise this may cause problems when there are tools in the inventory.

src/sync.zig
```diff
@@ -1353,11 +1219,63 @@ pub const Command = struct { // MARK: Command
...
+
+		fn run(self: CraftFrom, ctx: Context) error{serverFailure}!void {
+			for (self.destinations.inventories) |dest| if (dest.type != .normal) return;
+			for (self.sources.inventories) |source| if (source.type != .normal) return;
+
+			if (self.destinations.canHold(.{.item = .{.baseItem = self.recipe.resultItem}, .amount = self.recipe.resultAmount}) != .yes) return;
+
+			// Can we even craft it?
+			for (self.recipe.sourceItems, self.recipe.sourceAmounts) |requiredItem, requiredAmount| {
+				self.sources.removeItems(ctx, requiredAmount, requiredItem);
```
IntegratedQuantum, line 1266: Please first check if we have the needed ingredients, before removing anything (you accidentally removed that code)

IntegratedQuantum:
Hmm, I'm not sure. I think for now it would be better to have one clear pattern: Check if the reader has enough data left before allocating.

IntegratedQuantum (changes requested):
also add the check to fromBytes please

src/sync.zig
Wunka: readded. It seems I removed it when introduced the remove_items namespace...

Wunka:
So now we go back to the problem I described at the beginning of the PR. The check for recipe `fromBytes` fails. So if you would run this currently the game crashes when crafting something.

src/items.zig
```diff
@@ -1061,7 +1061,48 @@ pub const Recipe = struct { // MARK: Recipe
...
+			writer.writeVarInt(u16, amount);
+		}
+	}
+
+	pub fn fromBytes(reader: *BinaryReader) !*Recipe {
+		const resultItem = try reader.readEnum(BaseItemIndex);
+		const resultAmount = try reader.readVarInt(u16);
+		const sourceCount = try reader.readVarInt(usize);
+		if (sourceCount == 0) return error.Invalid;
+		if (sourceCount*@sizeOf(Recipe) > reader.remaining.len) return error.Invalid; // this check fails
```
IntegratedQuantum: Should be `@sizeOf(u16) + @sizeOf(BaseItemIndex)`, but even that doesn't account for the compression of the varint.

So instead, I'd suggest to just use a pair of Lists, that are maybe preallocated to `@min(256, sourceCount)`.

src/sync.zig
```diff
@@ -1353,11 +1219,82 @@ pub const Command = struct { // MARK: Command
...
+
+			if (self.destinations.canHold(.{.item = .{.baseItem = self.recipe.resultItem}, .amount = self.recipe.resultAmount}) != .yes) return;
+
+			// Can we even craft it?
+			outer: for (self.recipe.sourceItems) |requiredItem| {
+				var amount: usize = 0;
+				// There might be duplicate entries:
+				for (self.recipe.sourceItems, self.recipe.sourceAmounts) |otherItem, otherAmount| {
+					if (std.meta.eql(requiredItem, otherItem))
+						amount += otherAmount;
```
IntegratedQuantum: https://github.com/PixelGuys/Cubyz/pull/2506#discussion_r2725682507

IntegratedQuantum (changes requested):

src/items.zig
Wunka: If I understood correctly what you wanted. Its now there. (but please no merge It seems I removed a check so now if you click a sourceitem the game crashes)

src/items.zig
```diff
@@ -1061,7 +1061,49 @@ pub const Recipe = struct { // MARK: Recipe
...
+		defer sourceItems.deinit();
+		var sourceAmounts: main.List(u16) = .initCapacity(main.stackAllocator, @min(256, sourceCount));
+		defer sourceAmounts.deinit();
+
+		while (reader.remaining.len > 0) {
+			sourceItems.append(try reader.readEnum(BaseItemIndex));
+			sourceAmounts.append(try reader.readVarInt(u16));
+		}
+
+		return getValidRecipe(.{.sourceItems = sourceItems.toOwnedSlice(), .sourceAmounts = sourceAmounts.toOwnedSlice(), .resultItem = resultItem, .resultAmount = resultAmount});
```
IntegratedQuantum: We don't need to realloc here.
```suggestion
		return getValidRecipe(.{.sourceItems = sourceItems.items, .sourceAmounts = sourceAmounts.items, .resultItem = resultItem, .resultAmount = resultAmount});
```

IntegratedQuantum (changes requested):

Wunka:
check is revived.

src/gui/gui.zig
```diff
@@ -650,8 +650,8 @@ pub const inventory = struct { // MARK: inventory
 		};
 		const mainGuiButton = main.KeyBoard.key("mainGuiButton");
 		const secondaryGuiButton = main.KeyBoard.key("secondaryGuiButton");
-
-		if (itemSlot.inventory.type == .crafting and itemSlot.mode == .takeOnly and mainGuiButton.pressed and (recipeItem != .null or itemSlot.pressed)) {
+		if (itemSlot.inventory.type == .crafting and mainGuiButton.pressed and (recipeItem != .null or itemSlot.pressed)) {
+			if (itemSlot.mode != .takeOnly) return;
```
IntegratedQuantum: Could you explain this change?

Wunka: I actual thought this was also a case where I needed to add the check again. But yeah... it also works without this change so I reverted that.

IntegratedQuantum:
I'm also still wondering about the other change from that "check is revived" part, why did it need to change?

Wunka:
> I'm also still wondering about the other change from that "check is revived" part, why did it need to change?

When you click the left "source" inventories of the recipe menu you are in a recipe inventory but its not `takeOnly`. so we trigger a `depositOrSwap` with a non sharedInventory which triggers the assert. Before in `depositOrSwap` `tryCrafting` would be called.

src/gui/gui.zig
```diff
@@ -650,8 +650,8 @@ pub const inventory = struct { // MARK: inventory
 		};
```
IntegratedQuantum, line 650: Hmm, I see. Then your solution is only temporary, it fixes the issue for immutable slots that are a part of crafting inventories, but we will have other immutable slots in the future.
Ideally we'd have a check right here at the beginning of the function that returns if the slot is immutable.

Wunka: added the check there. apply should also have that check too but with the split Components this could of course be better made. I can still add it if you want.

src/items.zig
```diff
@@ -1061,7 +1061,49 @@ pub const Recipe = struct { // MARK: Recipe
...
+			writer.writeEnum(BaseItemIndex, item);
+			writer.writeVarInt(u16, amount);
+		}
+	}
+
+	pub fn fromBytes(reader: *BinaryReader) !*Recipe {
+		const resultItem = try reader.readEnum(BaseItemIndex);
+		const resultAmount = try reader.readVarInt(u16);
+		const sourceCount = try reader.readVarInt(usize);
+		if (sourceCount == 0) return error.Invalid;
```
IntegratedQuantum: I wouldn't dismiss this so quickly. The rest of the engine allows it, and there is no reason not to allow addon creators to define recipes like this. This might in fact make more sense once tool durability and special crafting stations are added to the mix.

src/items.zig
```diff
@@ -1061,7 +1061,49 @@ pub const Recipe = struct { // MARK: Recipe
...
+
+		var sourceItems: main.List(BaseItemIndex) = .initCapacity(main.stackAllocator, @min(256, sourceCount));
+		defer sourceItems.deinit();
+		var sourceAmounts: main.List(u16) = .initCapacity(main.stackAllocator, @min(256, sourceCount));
+		defer sourceAmounts.deinit();
+
+		while (reader.remaining.len > 0) {
```
IntegratedQuantum: Should add most add `sourceCount` items. Otherwise we may cause problems if there is other stuff afterwards that expects to read things from the same reader.

src/sync.zig
```diff
@@ -1353,11 +1219,81 @@ pub const Command = struct { // MARK: Command
...
+			// Can we even craft it?
+			outer: for (self.recipe.sourceItems) |requiredItem| {
+				var amount: usize = 0;
+				// There might be duplicate entries:
+				for (self.recipe.sourceItems, self.recipe.sourceAmounts) |otherItem, otherAmount| {
+					if (requiredItem == otherItem) amount += otherAmount;
+				}
+				for (self.sources.inventories) |source| {
+					for (source._items) |otherStack| {
+						if (otherStack.item != .null and std.meta.eql(requiredItem, otherStack.item.baseItem)) {
```
IntegratedQuantum: That case is automatically handle by std.meta.eql.
```suggestion
						if (std.meta.eql(requiredItem, otherStack.item.baseItem)) {
```

IntegratedQuantum (changes requested):
a few more things

src/gui/gui.zig
IntegratedQuantum: should be fine

src/sync.zig
IntegratedQuantum: Oh wait, nevermind it uses .baseItem, it should have been
```zig
if (otherStack.item == .baseItem and requiredItem == otherStack.item.baseItem) {
```

Wunka: your message and the error in my console came at the same time. but all good. that you even have to point that out is my fault. requests are done.

IntegratedQuantum (approved):
Thank you for working on this, I think it's a great improvement both in terms of readability and feature consistency.
If you are up to it, the next step in that direction would be #2529, it would be amazing if we could get rid of all the edge cases that came with the server-shared inventory type.

Wunka:
> If you are up to it, the next step in that direction would be #2529

maybe. I will have to read into the tool logic a little bit before to understand whats exactly going on there. 
If I take it up I will again start first with the proposal for the structure of the Sync Operation like I did for this.

PR #2511, opened 2026-01-20
Update hover and press window functions to return a Result.

### LINKED ISSUE #2162: Can't drop items through crosshair

IntegratedQuantum:
This allows dealing with ordering and item dropping in the correct way

fixes #2162 among other small inconsistencies.

### DIFF
--- a/src/gui/components/HorizontalList.zig
+++ b/src/gui/components/HorizontalList.zig
@@ -77,6 +77,7 @@ pub fn updateSelected(self: *HorizontalList) void {
 }
 
 pub fn updateHovered(self: *HorizontalList, mousePosition: Vec2f) main.callbacks.Result {
+	// reverse order of rendering, the last-rendered element is the first one that we should try to interact with
 	var i: usize = self.children.items.len;
 	while (i != 0) {
 		i -= 1;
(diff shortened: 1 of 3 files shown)


Argmaster:
I think checkboxes don't work, i tried toggling bloom and vsync and they don't change in UI

IntegratedQuantum:
good find, it was actually a mistake in the VerticalList's scrolling behavior and affected all components after scrolling.

src/gui/components/HorizontalList.zig
```diff
@@ -96,16 +96,14 @@ pub fn render(self: *HorizontalList, mousePosition: Vec2f) void {
 	draw.restoreTranslation(oldTranslation);
 }
 
-pub fn mainButtonPressed(self: *HorizontalList, mousePosition: Vec2f) void {
-	var selectedChild: ?*GuiComponent = null;
-	for (self.children.items) |*child| {
+pub fn mainButtonPressed(self: *HorizontalList, mousePosition: Vec2f) main.callbacks.Result {
+	var iterator = std.mem.reverseIterator(self.children.items);
```
Argmaster, line 102: Why is it reverse?

IntegratedQuantum: because the element rendered last is on top, and we want to prefer it when clicking.

IntegratedQuantum: Should i add a comment for it?

Argmaster: Yeah, but its not the only place where it would need a comment.
I think we could settle on comments for now and create an issue to add some more abstract construct (or maybe just a helper function?) that would allow us to consolidate that knowledge into one place.

IntegratedQuantum: How would you even want to abstract that?

PR #2512, opened 2026-01-20
Metal bars

careeoki:
image

Building blocks for each metal, crafted 1 ingot -> 2 bars.

### DIFF
--- a/assets/cubyz/recipes/special_recipes.zig.zon
+++ b/assets/cubyz/recipes/special_recipes.zig.zon
@@ -49,11 +49,11 @@
 	},
 	.{
 		.inputs = .{"1 cubyz:iron_ingot"},
-		.output = "2 cubyz:chain/iron",
+		.output = "4 cubyz:chain/iron",
 	},
 	.{
 		.inputs = .{"{mod}:{type}_ingot"},
-		.output = "2 {mod}:bars/{type}",
+		.output = "4 {mod}:bars/{type}",
 	},
 	.{
 		.inputs = .{"4 cubyz:silver_ingot"},


Crepestrom:
these look very nice
could we get metal tiles and/or metal bricks like the stone varients?

Crepestrom:
also do the uranium fences glow like the uranium block

Crepestrom:
oh wait uranium blocks dont glow but uranium ore does

Crepestrom:
#2513 i made a small pull request for uranium blocks giving off light

careeoki:
> could we get metal tiles and/or metal bricks like the stone varients?

Yeah, i would like something like that, I just need to think if how they would look.

IntegratedQuantum:
For the copper bars, have you considered adding a 2 pixel thick horizontal bar in the center instead of leaving a space?
That space it kind of bothering me, and I think filling it could match the style of the other bars, which all do have a horizontal bar in the center.

careeoki:
Yeah, the uneven gap is kinda silly looking.
I think you mean like this?
image
image

IntegratedQuantum:
Yeah, but I would put them one further up, such that there is always one row of pixels at the top and bottom where the texture ends.

careeoki:
Alright, sure
image

assets/cubyz/blocks/bars/uranium.zig.zon
```diff
@@ -0,0 +1,7 @@
+.{
+	.blockHealth = 25,
```
IntegratedQuantum, line 3: I think these should also have the same resistance as their corresponding materials

assets/cubyz/recipes/special_recipes.zig.zon
```diff
@@ -51,6 +51,10 @@
 		.inputs = .{"1 cubyz:iron_ingot"},
 		.output = "2 cubyz:chain/iron",
 	},
+	.{
+		.inputs = .{"{mod}:{type}_ingot"},
+		.output = "2 {mod}:bars/{type}",
```
IntegratedQuantum: I think we can be a bit more generous here (and also for the chain), I think 4 for each ingot would be nicer.

assets/cubyz/blocks/bars/uranium.zig.zon
careeoki: Done. A lot of other metal/gem blocks lack correct resistances, so I should do a separate PR fixing that.

assets/cubyz/recipes/special_recipes.zig.zon
careeoki: Sure, done for both

assets/cubyz/blocks/bars/uranium.zig.zon
IntegratedQuantum: yes, please

PR #2513, opened 2026-01-20
Uranium block light

Crepestrom:
Makes uranium blocks glow like uranium ore


careeoki:
This was an intentional decision so that if you built something with uranium blocks, it wouldn't be constantly glowing at night, which usually is not something you want.

For a logical explanation: raw uranium ore glows, processed uranium metal does not.

Regardless you also need an emission texture.

Crepestrom:
Ah k

PR #2494, opened 2026-01-16
depositOrDrop now accepts slice

Wunka:
With #2483 together these two are the last to slice refactors I need to do for #2324 
So if these refactors shall be continued for other payloads there should maybe be an issue. 

As with the `tryCrafting` refactor the priority code is almost just copied over. 

And also as with the `tryCrafting` refactor I will keep #2324 up to Date with this PR.

PR #2515, opened 2026-01-21
PutItemsInto Helper

Wunka:
This is extracted from: https://github.com/PixelGuys/Cubyz/pull/2506#discussion_r2707904989

Initially this was only supposed to be a helper function, but as I wanted to improve the abstraction it grew into a struct to avoid namespace pollution.

This will also be helpful if #2506 is not accepted, as I have at least one other slice refactor that currently copies this code (#2494), and we may extend additional functions to slices or make more operations use the same priority system.

### DIFF
--- a/src/sync.zig
+++ b/src/sync.zig
@@ -779,6 +779,67 @@ pub const Command = struct { // MARK: Command
 		};
 	}
 
+	const PutItemsInto = struct {
+		const Source = union(enum) {
+			move: InventoryAndSlot,
+			create: Item,
+		};
+
+		fn getBaseOperation(dest: InventoryAndSlot, amount: u16, source: Source) BaseOperation {
+			return switch(source) {
+				.move => |slot| .{.move = .{
+					.dest = dest,
+					.amount = amount,
+					.source = slot,
+				}},
+				.create => |item| .{.create = .{
+					.dest = dest,
+					.amount = amount,
+					.item = item,
+				}},
+			};
+		}
+
+		pub fn do(ctx: Context, destinations: []const Inventory, itemAmount: u16, source: Source) void {
+			const item = switch(source) {
+				.move => |slot| slot.ref().item,
+				.create => |item| item,
+			};
+			var remainingAmount = itemAmount;
+			var selectedEmptySlot: ?u32 = null;
+			var selectedEmptyInv: ?Inventory = null;
+			outer: for(destinations) |dest| {
+				var emptySlot: ?u32 = null;
+				var hasItem = false;
+				for(dest._items, 0..) |*destStack, destSlot| {
+					if(destStack.item == .null and emptySlot == null) {
+						emptySlot = @intCast(destSlot);
+						if(selectedEmptySlot == null) {
+							selectedEmptySlot = emptySlot;
+							selectedEmptyInv = dest;
+						}
+					}
+					if(std.meta.eql(destStack.item, item)) {
+						hasItem = true;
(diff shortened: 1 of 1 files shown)


src/sync.zig
```diff
@@ -779,6 +779,67 @@ pub const Command = struct { // MARK: Command
 		};
 	}
 
+	const PutItemsInto = struct {
```
IntegratedQuantum: structs as namespaces should formatted in snake_case, as mentioned in [the contributing guidelines](https://github.com/PixelGuys/Cubyz/blob/master/docs/CONTRIBUTING.md#follow-the-style-guide)

src/sync.zig
```diff
@@ -779,6 +779,67 @@ pub const Command = struct { // MARK: Command
...
+		const Source = union(enum) {
+			move: InventoryAndSlot,
+			create: Item,
+		};
+
+		fn getBaseOperation(dest: InventoryAndSlot, amount: u16, source: Source) BaseOperation {
```
IntegratedQuantum: Should be a member function of `Source` in my opinion.

src/sync.zig
```diff
@@ -779,6 +779,67 @@ pub const Command = struct { // MARK: Command
 		};
 	}
 
+	const PutItemsInto = struct {
+		const Source = union(enum) {
```
IntegratedQuantum: Maybe it should be named differently, maybe provider makes more sense?

src/sync.zig
```diff
@@ -779,6 +779,67 @@ pub const Command = struct { // MARK: Command
...
+				}},
+				.create => |item| .{.create = .{
+					.dest = dest,
+					.amount = amount,
+					.item = item,
+				}},
+			};
+		}
+
+		pub fn do(ctx: Context, destinations: []const Inventory, itemAmount: u16, source: Source) void {
```
IntegratedQuantum: Should have a separate mode based on how to behave when it doesn't fit (for crafting it should not do anything in that case, since we cannot craft partial recipes)

IntegratedQuantum (changes requested):
Please also check if you can apply this for other similar actions (→ deposit or drop, shift-click crafting, item collecting) as well.

src/sync.zig
Wunka: For the tryCrafting example this is a little bit harder as tryCrafting first deletes the source Items from the recipe. So the check needs to be (like its currently) before the deletions happen or I return an error or something else to tell that it didn't fit. (or I delete after the creation?)

IntegratedQuantum: Hmm, I guess it should be done in a follow-up PR then.
Either way the transfer check should be done before deletions, I'd say:
Check if items are available → try transfer → delete items if transfer succeeds.

Wunka: for DepositOrDrop this is would also be good so we know if there is anything to drop

Wunka:
Ok so for the other ones: 

- `tryCraftingto`: supported as I tested this already with #2506 
- `DepositOrDrop`: with the amount check we don't need the return of `put_items_into`, so added. 
- `Collect`: this is a helper like tryCraftingTo and uses an entire sync command (fillFromCreative) so I would need to add a little bit but should be possible

src/sync.zig
Wunka: didn't see that, changed

Wunka:
Correction: `Collect` needs the return so it knows if the item actually has been collected.

IntegratedQuantum (approved):
Please apply it to the other functions in future PRs then.

Issue #1558, opened 2025-05-30
Require sickles for collecting leaves, grass plants, etc


careeoki:
Requires #856 

This makes sickles more important and also reduces inventory clutter when you don't want to collect plants.

Blocks that would require a sickle to collect:
- All leaves
- All leaf piles
- All grass vegetation
- Fern
- Tussock
- Duckweed
- Ivy

DellieDelta:
The only one I'd leave (leaf?) out of this is leaf piles. To me they feel more like flowers the way you collect them

careeoki:
> The only one I'd leave (leaf?) out of this is leaf piles. To me they feel more like flowers the way you collect them

They appear in pretty large patches and can clutter your inventory pretty quick. I think it's better if you have to intentionally use your sickle to collect them instead of picking them up all the time.

IntegratedQuantum:
Wouldn't it make sense to have all plants require the sickle?
At least conceptually I think it makes sense to use a sickle to collect flowers and mushrooms, instead of punching them with your fist.
Plus a sickle is trivial to craft.

DellieDelta:
Fair enough, plants = sickle without exceptions makes sense!

careeoki:
I imagine it more as plucking the plant with your hands.

I think the problem would be that flowers are more "special" and if you can accidentally punch them away you would lose them completely, which could be upsetting.

IntegratedQuantum:
I think we should probably also not even make them selectable if you don't have a chisel.

careeoki:
Yeah, in that case, only harvesting with sickles would make sense.

DellieDelta:
I'll first work on ignoring the hitbox of foliage when not holding a sickle. Otherwise it might be jarring that foliage suddenly doesn't drop when punched by hand. Shipping them together also makes testing easierr

DellieDelta:
Also another thing. When sickles are the only way to collect foliage, what should be done if the replaceable foliage is broken by placing a block into it or if its supporting block is broken? 

Currently it drops the item, which I like because you get all the blocks, but if the goal is reducing inventory clutter it might be better to not drop the items in those casess

IntegratedQuantum:
We have a similar issue for decay, e.g. if you break the support block, or if you break the top block of a vine (with or without a sickle, the game cannot distinguish).

[...]

PR #2522, opened 2026-01-25
Texture pile grass plants (+ lots of grass)

careeoki:
As originally described in #1409 , I really want multiple grass heights. We still can't have 2-tall grass, but we can do 2 grass stages with texture piles.
I also generally want a ton more grass, it makes the world much more lush and organic feeling, and reduces the "flatness" a lot of our biomes have.
I did only* one biome per grass type to showcase, so other biomes would have to be changed in followup PRs.
image
image

image

image
Also did tussock as well

annoyances with so much grass:
- without #1558 you get a ton of grass in your inventory all the time
- #2267 is extra annoying


H41ogen:
random offset/rotation (maybe even non-90 degree rotations?) would be nice, probably not in this pr but in another.

careeoki:
> random offset/rotation (maybe even non-90 degree rotations?) would be nice, probably not in this pr but in another.

That's outside my abilities, but yes, I want random offsets and texture variants for grass

careeoki:
I'll redo this sometime after #3298 because i dont want to bother with the conflicts

PR #2524, opened 2026-01-25
Inventories wrapper

Wunka:
While working on #2506, it became apparent that handling multiple inventories was becoming repetitive and that more functions required their own `[]const Inventory` implementations.

This PR introduces a wrapper for handling multiple Inventories and migrates the existing Sync Commands that operate on them to use it.

It also changes the behavior of `canHold`. Before it just returned if the Inventory can hold the `ItemStack` but now it either returns `yes` for it can hold or the `remainingAmount` it can not hold.

`removeItems` is currently unused, but will be used by #2506.

### DIFF
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -671,10 +671,6 @@ pub const Inventories = struct { // MARK: Inventories
 		alloctor.free(self.inventories);
 	}
 
-	pub fn getInventories(self: Inventories) []const Inventory {
-		return self.inventories;
-	}
-
 	pub fn canHold(self: Inventories, itemStack: ItemStack) Inventory.CanHoldReturn {
 		var remainingAmount = itemStack.amount;
 		for (self.inventories) |dest| {
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -722,11 +722,10 @@ pub const Inventories = struct { // MARK: Inventories
 	pub fn canHold(self: Inventories, itemStack: ItemStack) Inventory.CanHoldReturn {
 		var remainingAmount = itemStack.amount;
 		for (self.inventories) |dest| {
-			remainingAmount -|= switch (dest.canHold(itemStack)) {
+			remainingAmount = switch (dest.canHold(.{.item = itemStack.item, .amount = remainingAmount})) {
 				.yes => return .yes,
-				.remainingAmount => |amount| (itemStack.amount - amount),
+				.remainingAmount => |amount| amount,
 			};
-			if (remainingAmount == 0) return .yes;
 		}
 		return .{.remainingAmount = remainingAmount};
 	}
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -773,38 +773,4 @@ pub const Inventories = struct { // MARK: Inventories
 		return remainingAmount;
 	}
 
-	pub fn removeItems(self: Inventories, ctx: sync.Command.Context, itemAmount: u16, baseItem: main.items.BaseItemIndex) void {
-		var fullSlot: ?u32 = null;
-		var fullInv: ?Inventory = null;
-		var remainingAmount: usize = itemAmount;
-		for (self.inventories) |source| {
-			for (0..source._items.len) |reverseIndex| {
-				const i: usize = source._items.len - reverseIndex - 1;
-				const otherStack: *ItemStack = &source._items[i];
-				if (otherStack.item != .null and baseItem == otherStack.item.baseItem) {
-					if (otherStack.amount == otherStack.item.stackSize()) {
-						if (fullSlot == null) {
-							fullSlot = @intCast(i);
-							fullInv = source;
-						}
-						continue;
-					}
-					const amount = @min(remainingAmount, otherStack.amount);
-					ctx.execute(.{.delete = .{
-						.source = .{.inv = source, .slot = @intCast(i)},
-						.amount = amount,
-					}});
-					remainingAmount -= amount;
-					if (remainingAmount == 0) return;
-				}
-			}
-		}
-		if (remainingAmount > 0 and fullSlot != null) {
-			ctx.execute(.{.delete = .{
-				.source = .{.inv = fullInv.?, .slot = fullSlot.?},
-				.amount = @min(remainingAmount, baseItem.stackSize()),
-			}});
-		}
-	}
-
 };
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -772,5 +772,4 @@ pub const Inventories = struct { // MARK: Inventories
 		}
 		return remainingAmount;
 	}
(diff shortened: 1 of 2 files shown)


src/Inventory.zig
```diff
@@ -644,3 +649,165 @@ pub fn fromBytes(self: Inventory, reader: *BinaryReader) void {
...
+		return .{
+			.inventories = copy,
+		};
+	}
+
+	pub fn deinit(self: Inventories, alloctor: NeverFailingAllocator) void {
+		alloctor.free(self.inventories);
+	}
+
+	pub fn getInventories(self: Inventories) []const Inventory {
```
IntegratedQuantum: this function is unnecessary

src/Inventory.zig
```diff
@@ -644,3 +649,165 @@ pub fn fromBytes(self: Inventory, reader: *BinaryReader) void {
...
+		return self.inventories;
+	}
+
+	pub fn canHold(self: Inventories, itemStack: ItemStack) Inventory.CanHoldReturn {
+		var remainingAmount = itemStack.amount;
+		for (self.inventories) |dest| {
+			remainingAmount -|= switch (dest.canHold(itemStack)) {
+				.yes => return .yes,
+				.remainingAmount => |amount| (itemStack.amount - amount),
```
IntegratedQuantum: Why make it so complicated, just set remainingAmount to the remaining amount, no subtraction needed.

src/Inventory.zig
```diff
@@ -644,3 +649,165 @@ pub fn fromBytes(self: Inventory, reader: *BinaryReader) void {
...
+			};
+			if (remainingAmount == 0) return .yes;
```
IntegratedQuantum: Cannot happen

src/sync.zig
```diff
@@ -260,11 +260,11 @@ pub const Command = struct { // MARK: Command
 		addEnergy = 6,
 	};
 
-	const InventoryAndSlot = struct {
+	pub const InventoryAndSlot = struct {
```
IntegratedQuantum: This should probably also be moved to Inventory.zig.

src/Inventory.zig
```diff
@@ -644,3 +649,165 @@ pub fn fromBytes(self: Inventory, reader: *BinaryReader) void {
...
+			}
+		}
+		if (remainingAmount > 0 and selectedEmptySlot != null) {
+			ctx.execute(provider.getBaseOperation(.{.inv = selectedEmptyInv.?, .slot = selectedEmptySlot.?}, remainingAmount));
+			remainingAmount = 0;
+		}
+		return remainingAmount;
+	}
+
+	pub fn removeItems(self: Inventories, ctx: sync.Command.Context, itemAmount: u16, baseItem: main.items.BaseItemIndex) void {
```
IntegratedQuantum: Since it isn't used yet, please save it for the crafting PR.

src/Inventory.zig
```diff
@@ -644,3 +649,165 @@ pub fn fromBytes(self: Inventory, reader: *BinaryReader) void {
...
+	}
+
+	pub fn toBytes(self: Inventories, writer: *BinaryWriter) void {
+		writer.writeVarInt(usize, self.inventories.len);
+		for (self.inventories) |inv| {
+			writer.writeEnum(InventoryId, inv.id);
+		}
+	}
+
+	pub fn fromBytes(allocator: NeverFailingAllocator, reader: *BinaryReader, side: sync.Side, user: ?*main.server.User) !Inventories {
```
IntegratedQuantum: Since it is a kind of init, it belongs to the other init functions.
The corresponding toBytes should then be directly below deinit, to keep them close together.

src/Inventory.zig
```diff
@@ -644,3 +649,165 @@ pub fn fromBytes(self: Inventory, reader: *BinaryReader) void {
...
+pub const Inventories = struct { // MARK: Inventories
+	inventories: []const Inventory,
+
+	pub fn init(alloctor: NeverFailingAllocator, inventories: []const Inventory) Inventories {
+		return .{
+			.inventories = alloctor.dupe(Inventory, inventories),
+		};
+	}
+
+	pub fn initWithClientInventories(alloctor: NeverFailingAllocator, clientInventories: []const Inventory.ClientInventory) Inventories {
```
```suggestion
	pub fn initFromClientInventories(alloctor: NeverFailingAllocator, clientInventories: []const Inventory.ClientInventory) Inventories {
```

IntegratedQuantum (changes requested):

src/Inventory.zig
Wunka: the `itemstack` amount is not updated, so `remainingAmount` tracks for all destinations while `dest.canHold` tries to put in always the same item with the same amount.
If I change it to `=` would it then not be the same issue like it was before?

Wunka: I took inspiration from the original `canHold` which also instead of changing `itemStack.amount` tracks the `remaingAmount` over all slots

IntegratedQuantum: Then I think it should pass an item stack with the remaining amount each time.

Wunka: changed. The other things are also done

src/sync.zig
```diff
@@ -1379,92 +1288,62 @@ pub const Command = struct { // MARK: Command
...
-			if (destinationsSize*@sizeOf(InventoryId) >= reader.remaining.len) return error.Invalid;
-
-			const destinations = main.globalAllocator.alloc(Inventory, destinationsSize);
-			errdefer main.globalAllocator.free(destinations);
-
-			for (destinations) |*dest| {
-				const invId = try reader.readEnum(InventoryId);
-				dest.* = Inventory.getInventory(invId, side, user) orelse return error.InventoryNotFound;
-			}
+			const destinations: Inventory.Inventories = try .fromBytes(main.globalAllocator, reader, side, user);
```
IntegratedQuantum, line 1296: Needs an errdefer deinit

src/sync.zig
```diff
@@ -1379,92 +1288,62 @@ pub const Command = struct { // MARK: Command
...
-			errdefer main.globalAllocator.free(destinations);
-
-			for (destinations) |*dest| {
-				const invId = try reader.readEnum(InventoryId);
-				dest.* = Inventory.getInventory(invId, side, user) orelse return error.InventoryNotFound;
-			}
-
 			return .{
-				.destinations = destinations[0..],
+				.destinations = try .fromBytes(main.globalAllocator, reader, side, user),
```
IntegratedQuantum: also needs an errdefer deinit

IntegratedQuantum (changes requested):

src/sync.zig
Wunka: both added

IntegratedQuantum (approved):

PR #2530, opened 2026-01-30
Beginnings of a permission layer

Wunka:
This is a first step in the direction of #1961 

Usage: 
Users / Groups can be either whitelisted or blacklisted from a permissionPath.
For example a user can be granted the permission: "command" but be blacklisted from "command/spawn" 
This would mean the user can run any command except spawn. 

Users can be added to groups. 
User permissions precede the group permissions, so if a user is blacklisted from the spawn command and joines a group who is whitelisted for the spawn command, the user would still have no permission to use spawn.

For handling all this 2 Commands are introduced: 
- `/perm` to add permssionPaths to the whitelist / blacklist of the user or a certain group
- `/group` to create / delete / join or leave groups

Tasks for a complete permission layer (can be follow-up issues):

- [x] Add user blacklists/whitelists
- [x] Add groups
- [x] Add tests
- [x] Save permissions in zon
- [ ] Send the client the complete permission tree (to enable #2414)
- [ ] Add argument-based command permissions (after #1425)
- [x] Add commands to modify permissions and groups
- [x] Update permission/group commands: sanitize arguments and improve error messages
- [x] add assertions for functions that modify permissions / groups
- [ ] Decide on the permission hierarchy (e.g. all commands under `/command`)
- [ ] Add user parameter to commands to enable changing permissions of someone else


IntegratedQuantum:
Could you give an example of how this would look?

Also one relevant case would be to allow for localized permissions to e.g. forbid editing in a region (#81)
Would this fit into your system?

Wunka:
Example for `/spawn`: (we could also add the prefix `command` to show that we are talking about command permissions)
server hoster has complete access so: `user.addPermision("spawn.*")`  
other users can only get their spawnpoint: `user.addPermision("spawn.get")`

you can chain the dots as long as we don't reach the limit with recursions (for checking and adding permissions)
so making something like `user.addPermission("blocks.protecotor.[some way to identify the explicit block]")` would be possible

IntegratedQuantum:
Yeah, that seems reasonable (though with how much this is like a file system, I'd prefer if it used slashes).
For the protector block it would probably not address the block directly but rather mentioning a string that can be shared for multiple protector blocks (otherwise it would be a nightmare to add another player to your region).

And of course we should pick a reasonable hierarchy, there should be a subcategory for commands, and inside of commands there should be a subcategory of get commands.

Also one problem I see in your code is that if you enable a parent node then all the subtree information is lost. So if I give you permissions to run all commands, then take this permission away at a later point, you cannot even run /help anymore.
On the other hand it could also be reasonable to be able to disable child nodes from an enabled parent (maybe I want you to be able to run all commands except from world edit commands, or all getter commands, except from getting the seed).

Wunka:
okay so to adresss a few of your things I think I need somekind of global saved permission tree. So I think I will change the system so that as an example commands register themself like `registerPermission("command/get/spawn")` that way we have a complete set. 

> if you enable a parent node then all the subtree information is lost. So if I give you permissions to run all commands, then take this permission away at a later point, you cannot even run /help anymore.

I think I will introduce an optional `fn checkPermission(msg) bool` to `_commands.zig` which in default checks just the command name (like it is now) and for help: overwrite it so that it just checks the permission of the command you want to run help on.

> On the other hand it could also be reasonable to be able to disable child nodes from an enabled parent (maybe I want you to be able to run all commands except from world edit commands, or all getter commands, except from getting the seed).

This is already supported? There is just not a good way to enable that. But through the global saved permission tree you could do things like: `user.addPermission("commands/get/*):` `user.removePermission("commands/get/seed");`

Also in that regard what do you think about * ? should it mabe just be usable at the top so `user.addPermission("*")` and for commands you don't need it. Just: `user.addPermission("commands/get")` to get permission to all get commands

IntegratedQuantum:
> okay so to adresss a few of your things I think I need somekind of global saved permission tree.

I don't see how a global permission tree makes this easier, if at all I think it adds more edge cases, since the tree can change between updates (or just during normal gameplay).

Wunka:
> This is already supported? There is just not a good way to enable that. But through the global saved permission tree you could do things like: user.addPermission("commands/get/*): user.removePermission("commands/get/seed");

A lot can probably be done without the global tree but this not for what I see. 
(execept I introduce being able to switch between whitelist and blacklist)

Wunka:
The tree could maybe be also only used to get a picture of the current tree. So "*" instead of giving .all could just loop over all subPermissions in the global tree and add them all to the user.

IntegratedQuantum:
> (execept I introduce being able to switch between whitelist and blacklist)

Yeah, that's what I would propose, just add an enum to each node which has 3 states allowed, forbidden and neutral.
Then during iteration you have to go the entire way from top to bottom of the tree and switch states.

Another approach would be to kill recursion (and I'd always be in favor of that), and have just a single per player white and a single per player black list with the full permission path, then to check permission you'd basically just do:
```zig
var iterator = std.mem.findRightMostIteratorWhateverIDontKnow(permissionPath, '/');
while(iterator.next()) |endIndex| {
    if(blacklist.has(permissionPath[0..endIndex])) return false;
    if(whitelist.has(permissionPath[0..endIndex])) return true;
}
return false;
```

Wunka:
> Another approach would be to kill recursion (and I'd always be in favor of that), and have just a single per player white and a single per player black list with the full permission path, then to check permission you'd basically just do:

Do I understand that right that your `findRightMostIteratorWhateverIDontKnow` would make out of example "command/get/spawn" -> {"command/get/spawn", "command/get", "command"} ? 
that way all whitelist/blacklist would only work on the deepest specialization? 
If yes I think I will do that.

IntegratedQuantum:
Yeah, exactly.
And the list itself could just be a `StringHashMapUnmanaged(void)`

Wunka:
Changed to the black/whitelist system. (much better)
While on that I also added groups. 

For the whitelist / blacklist system as long as one group has whitelist access you are good to go. 
But the userlists have still higher priority so if a user is blacklisted from a command the game doesn't care even if he is part of some root group which has access to everything.

And for commands in general: I don't want to spend as much time into things like seperate permissions for the seperate subcommands as I think this would be much easier with #1425

IntegratedQuantum:
The other important thing would be sharing with the client. This is relevant for anything that has client-side prediction (e.g. gamemode permissions, protector block, #2414, ...)

For that I'd suggest to precompute the full set of permission from the user+group permission and send that to the client on each change. I think that would be easier and more efficient.

Also another important reminder: Add testing! The permission system should have unit tests for all of its functionality.

Wunka:
added tests and updated the description of the PR to reflect the changes

Wunka:
Oh what I also wanted to mention is that I am thinking of adding a kinda root permission. currently there is no way for someone to have access to literally everything (without specifing every top level permission like "command"). So maybe if someone has access to "root" or some other keyword they just skip every check.

Or we start everything with "/" and then you can get access to that

IntegratedQuantum:
Starting everything with a slash seems to be a reasonable option.

Wunka:
The permissions / groups are now saved into zon files. (player permission in player zon and groups into a new file `groups.zig.zon`) 
Before this can be passed up as a bare bones system. there is one problem I don't know how to adrress. 
See this scenario:

1. group created with name "group1"
2. "player1" joines "group1"
3.  "player1" leaves the server
4. "group1" is deleted (currently only online players are updated and offlines ones are when they join the server)
5. another person creates his own group, but it has the same name "group1"
6. "player1" joines the server again, but is not not removed from "group1" as it has the same name

So either we update every player zon file when a group is deleted 
or 
groups save the players instead (I think this solves the problem better, but I would like some input)

Wunka:
So I fixed my mentioned issue now by the groups also having a list of members.
I also added commands to manage permissions and groups.

I think I want to tackle the client side full permission tree as a follow up issue and would like this first to be fully reviewed. 
(The commands are for now pretty barebones with almost no good error messages but they work. I want to look at them again in a few days and work them complety out)

Because of the commands there is also now the requirement that groups don't start with "/".

src/server/command/permission/group.zig
```diff
@@ -0,0 +1,65 @@
...
+	var split = std.mem.splitScalar(u8, args, ' ');
+	if (split.next()) |arg| {
+		if (std.ascii.eqlIgnoreCase(arg, "create")) {
+			op = .create;
+		} else if (std.ascii.eqlIgnoreCase(arg, "delete")) {
+			op = .delete;
+		} else if (std.ascii.eqlIgnoreCase(arg, "join")) {
+			op = .join;
+		} else if (std.ascii.eqlIgnoreCase(arg, "leave")) {
+			op = .leave;
```
IntegratedQuantum: std.meta.stringToEnum()

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+	map.put(allocator.allocator, duped, {}) catch unreachable;
+}
+
+fn fillMap(allocator: NeverFailingAllocator, map: *std.StringHashMapUnmanaged(void), zon: ZonElement) void {
+	if (zon != .array) return;
+
+	for (zon.array.items) |item| {
+		switch (item) {
+			.string => |string| fillMapHelper(allocator, map, string),
+			.stringOwned => |string| fillMapHelper(allocator, map, string),
```
IntegratedQuantum: No need for a poorly named helper:
```suggestion
			.string, .stringOwned => |string| {
				...
			},
```

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+fn fillMapHelper(allocator: NeverFailingAllocator, map: *std.StringHashMapUnmanaged(void), permissionPath: []const u8) void {
+	if (map.contains(permissionPath)) return;
+	const duped = allocator.dupe(u8, permissionPath);
+	map.put(allocator.allocator, duped, {}) catch unreachable;
+}
+
+fn fillMap(allocator: NeverFailingAllocator, map: *std.StringHashMapUnmanaged(void), zon: ZonElement) void {
+	if (zon != .array) return;
+
+	for (zon.array.items) |item| {
```
IntegratedQuantum:
```suggestion
	for (zon.toSlice()) |item| {
```

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const ZonElement = main.ZonElement;
+
+fn fillMapHelper(allocator: NeverFailingAllocator, map: *std.StringHashMapUnmanaged(void), permissionPath: []const u8) void {
+	if (map.contains(permissionPath)) return;
+	const duped = allocator.dupe(u8, permissionPath);
+	map.put(allocator.allocator, duped, {}) catch unreachable;
+}
+
+fn fillMap(allocator: NeverFailingAllocator, map: *std.StringHashMapUnmanaged(void), zon: ZonElement) void {
```
IntegratedQuantum: I'd prefer symmetrical naming, mapFromZon, mapToZon

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+		switch (item) {
+			.string => |string| fillMapHelper(allocator, map, string),
+			.stringOwned => |string| fillMapHelper(allocator, map, string),
+			else => {},
+		}
+	}
+}
+
+fn mapToZon(allocator: NeverFailingAllocator, map: *std.StringHashMapUnmanaged(void)) ZonElement {
+	var zon: ZonElement = .initArray(allocator);
```
IntegratedQuantum:
```suggestion
	const zon: ZonElement = .initArray(allocator);
```

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+		var it = self.permissionWhiteList.keyIterator();
+		while (it.next()) |key| {
+			self.allocator.free(key.*);
+		}
+		self.permissionWhiteList.deinit(self.allocator.allocator);
+		it = self.permissionBlackList.keyIterator();
+		while (it.next()) |key| {
+			self.allocator.free(key.*);
+		}
+		self.permissionBlackList.deinit(self.allocator.allocator);
```
IntegratedQuantum: I think things like this are the perfect use case for local arena allocators.

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+			return true;
+		}
+		return false;
+	}
+
+	pub fn hasPermission(self: *Permissions, permissionPath: []const u8) PermissionResult {
+		var it = std.mem.splitBackwardsScalar(u8, permissionPath, '/');
+		var current = permissionPath;
+
+		while (it.next()) |path| {
```
IntegratedQuantum: Since this iterator is not really suited, I'd suggest
```
while(std.mem.lastIndexOfScalar(u8, current, '/)) |nextPos| {
    ...
    current = current[0..nextPos];
}
```

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+		return mapToZon(allocator, self.list(listType));
+	}
+
+	pub fn addPermission(self: *Permissions, listType: ListType, permissionPath: []const u8) void {
+		self.list(listType).put(self.allocator.allocator, self.allocator.dupe(u8, permissionPath), {}) catch unreachable;
+	}
+
+	pub fn removePermission(self: *Permissions, listType: ListType, permissionPath: []const u8) bool {
+		const _key = self.list(listType).getKeyPtr(permissionPath);
+		if (_key) |key| {
```
IntegratedQuantum:
```suggestion
		const key = self.list(listType).getKeyPtr(permissionPath) orelse return false;
```

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+			.white => &self.permissionWhiteList,
+			.black => &self.permissionBlackList,
+		};
+	}
+
+	pub fn fillList(self: *Permissions, listType: ListType, zon: ZonElement) void {
+		fillMap(self.allocator, self.list(listType), zon);
+	}
+
+	pub fn listToZon(self: *Permissions, allocator: NeverFailingAllocator, listType: ListType) ZonElement {
```
IntegratedQuantum: Instead of leaving the serialization of both lists to the caller, I'd prefer a function that stores it all, you can do this by taking the ZonElement instead of returning it.

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,389 @@
...
+			current = permissionPath[0..len];
+		}
+		return if (self.permissionWhiteList.contains("/")) .yes else .neutral;
+	}
+};
+
+pub const PermissionGroup = struct {
+	allocator: NeverFailingAllocator,
+	permissions: Permissions,
+	members: std.StringHashMapUnmanaged(void) = .{},
```
IntegratedQuantum: You do a lot of bookkeeping here, but I cannot see any place where you actually use these for anything.

src/server/permission.zig
```diff
@@ -0,0 +1,389 @@
...
+	}
+	groups.deinit();
+}
+
+pub fn createGroup(name: []const u8, allocator: NeverFailingAllocator) error{AlreadyExists}!void {
+	if (groups.contains(name)) return error.AlreadyExists;
+	groups.put(allocator.dupe(u8, name), .init(allocator)) catch unreachable;
+}
+
+pub fn deleteGroup(name: []const u8) bool {
```
IntegratedQuantum: Many of these C-style functions should be member functions (of either the PermissionGroup or the User or if you want them in this file, then I'd suggest a wrapper struct around all the stuff that's stored in the player) with only a single global `getGroup(name)` function to access them by name.

IntegratedQuantum (changes requested):
> So either we update every player zon file when a group is deleted
> or
> groups save the players instead (I think this solves the problem better, but I would like some input)

groups saving which players use them might cause other problems.

What we can do instead is give groups a unique ID. Every time you delete a group id, you don't reuse it, and when you load a player with a deleted id, then you can just detect this and remove it.

Also a thought on multithreading: I want to multithread player actions in the future, so we need to support reading permissions from multiple threads. In order to achieve this without adding locks everywhere, I'd suggest to add assertions to every modifying function that assert that we are on the server thread (see `sync.threadContext`).

> Because of the commands there is also now the requirement that groups don't start with "/"

Why is that a problem?

Wunka:
> Why is that a problem?

poorly made commands... thats the problem. will fix when I get back to this

src/server/permission.zig
Wunka: Everything that has not directly something to do with the groups list is now somewhere else. 
everything else is also for now I think done. 

One problem I have is that the asserts to cause the tests to fail...

IntegratedQuantum: In that case I'd suggest to just disable the assert when in testing (do this inside the assert function)

Wunka: added. Now all test pass again

src/server/command/_list.zig
```diff
@@ -21,3 +21,6 @@ pub const rotate = @import("worldedit/rotate.zig");
 pub const set = @import("worldedit/set.zig");
 pub const mask = @import("worldedit/mask.zig");
 pub const replace = @import("worldedit/replace.zig");
+
+pub const perm = @import("permission/perm.zig");
```
IntegratedQuantum: Please sort them alphabetically, permission comes before worledit

src/server/permission.zig
IntegratedQuantum, line 1: Files should be named in snake_case, unless they have top-level fields.
Also I would suggest to just call it `permission`

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+const User = server.User;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const NeverFailingArenaAllocator = main.heap.NeverFailingArenaAllocator;
+const ZonElement = main.ZonElement;
+const sync = main.sync;
+
+fn mapFromZon(allocator: NeverFailingAllocator, map: *std.StringHashMapUnmanaged(void), zon: ZonElement) void {
+	if (zon != .array) return;
+
+	for (zon.toSlice()) |item| {
```
IntegratedQuantum: The check is unnecessary, toSlice does the check internally.
```suggestion
	for (zon.toSlice()) |item| {
```

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+
+const main = @import("main");
+const server = main.server;
+const User = server.User;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const NeverFailingArenaAllocator = main.heap.NeverFailingArenaAllocator;
+const ZonElement = main.ZonElement;
+const sync = main.sync;
+
+fn mapFromZon(allocator: NeverFailingAllocator, map: *std.StringHashMapUnmanaged(void), zon: ZonElement) void {
```
IntegratedQuantum: Just a thought, but we have a specific hashmap, and two member functions, how about making a wrapper struct for it?

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+	const zon: ZonElement = .initArray(allocator);
+
+	var it = map.keyIterator();
+	while (it.next()) |key| {
+		zon.append(key.*);
+	}
+	return zon;
+}
+
+pub const Permissions = struct {
```
IntegratedQuantum: Please add `MARK` comments for all the big types and the start of the test section

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+	pub const ListType = enum {
+		white,
+		black,
+	};
+
+	arenaAllocator: NeverFailingArenaAllocator,
+	permissionWhiteList: std.StringHashMapUnmanaged(void) = .{},
+	permissionBlackList: std.StringHashMapUnmanaged(void) = .{},
+
+	pub fn deinit(self: *Permissions) void {
```
IntegratedQuantum, line 55: Generally everything that has a deinit function should also have an init function, in this case to create the arena.

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+
+	fn list(self: *Permissions, listType: ListType) *std.StringHashMapUnmanaged(void) {
+		return switch (listType) {
+			.white => &self.permissionWhiteList,
+			.black => &self.permissionBlackList,
+		};
+	}
+
+	pub fn fromZon(self: *Permissions, zon: ZonElement) void {
+		mapFromZon(self.arenaAllocator.allocator(), self.list(.white), zon.getChild("permissionWhiteList"));
```
IntegratedQuantum: whitelist and blacklist is one word
```suggestion
		mapFromZon(self.arenaAllocator.allocator(), self.list(.white), zon.getChild("permissionWhitelist"));
```

IntegratedQuantum: Please correct this everywhere

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+		white,
+		black,
+	};
+
+	arenaAllocator: NeverFailingArenaAllocator,
+	permissionWhiteList: std.StringHashMapUnmanaged(void) = .{},
+	permissionBlackList: std.StringHashMapUnmanaged(void) = .{},
+
+	pub fn deinit(self: *Permissions) void {
+		self.arenaAllocator.deinit();
```
IntegratedQuantum: Also add the assertion to deinit please, so we can be sure that it's not accidentally freed on the wrong thread.

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	pub fn toZon(self: *Permissions, allocator: NeverFailingAllocator, zon: *ZonElement) void {
+		zon.put("permissionWhiteList", mapToZon(allocator, self.list(.white)));
+		zon.put("permissionBlackList", mapToZon(allocator, self.list(.black)));
+	}
+
+	pub fn addPermission(self: *Permissions, listType: ListType, permissionPath: []const u8) void {
+		sync.threadContext.assertCorrectContext(.server);
+		self.list(listType).put(self.arenaAllocator.allocator().allocator, self.arenaAllocator.allocator().dupe(u8, permissionPath), {}) catch unreachable;
```
IntegratedQuantum: Should check if it's already in the list before allocating a new entry. I'd suggest to use getOrPut

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	pub fn removePermission(self: *Permissions, listType: ListType, permissionPath: []const u8) bool {
+		sync.threadContext.assertCorrectContext(.server);
+		const key = self.list(listType).getKeyPtr(permissionPath) orelse return false;
+		const slice = key.*;
+		_ = self.list(listType).remove(permissionPath);
+		self.arenaAllocator.allocator().free(slice);
```
IntegratedQuantum: arena can't free. Do nothing instead. It's not a problem to leak a few bytes here, no one cares. To cause a denial of service, you'd need a player to add and remove permissions for many hours, and if their goal is to take down the server, they can also just add permissions without removing them for the same effect.

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+		return if (self.permissionWhiteList.contains("/")) .yes else .neutral;
+	}
+};
+
+pub const PermissionGroup = struct {
+	permissions: Permissions,
+	id: u32,
+
+	pub fn init(allocator: NeverFailingAllocator) PermissionGroup {
+		currentId += 1;
```
IntegratedQuantum: If you access a non-atomic global variable, please add the assertion that you are on the server thread.

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+		return .{
+			.permissions = .{.arenaAllocator = .init(allocator)},
```
```suggestion
			.permissions = .init(allocator),
```

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+		self.permissions.deinit();
+	}
+
+	pub fn hasPermission(self: *PermissionGroup, permissionPath: []const u8) Permissions.PermissionResult {
+		return self.permissions.hasPermission(permissionPath);
+	}
+};
+
+var groups: std.StringHashMap(PermissionGroup) = undefined;
+var currentId: u32 = 0;
```
IntegratedQuantum: What's the id used for anyways?

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+
+pub fn init(allocator: NeverFailingAllocator) void {
+	groups = .init(allocator.allocator);
+}
+
+pub fn deinit() void {
+	var it = groups.iterator();
+	while (it.next()) |entry| {
+		groups.allocator.free(entry.key_ptr.*);
```
IntegratedQuantum: Better store a HashMapUnmanaged (again, the managed one is deprecated) and a separate allocator.
Also that one should be used for the allocation of these keys as well.

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+	pub fn deinit(self: *PermissionGroup) void {
+		self.permissions.deinit();
+	}
+
+	pub fn hasPermission(self: *PermissionGroup, permissionPath: []const u8) Permissions.PermissionResult {
+		return self.permissions.hasPermission(permissionPath);
+	}
+};
+
+var groups: std.StringHashMap(PermissionGroup) = undefined;
```
IntegratedQuantum: I think you forgot to reset the groups. I'd suggest to tie the init/deinit call to the world init/deinit to prevent any problems with persisting resources.

src/server/permission.zig
```diff
@@ -0,0 +1,306 @@
...
+pub fn deinit() void {
+	var it = groups.iterator();
+	while (it.next()) |entry| {
+		groups.allocator.free(entry.key_ptr.*);
+		entry.value_ptr.deinit();
+	}
+	groups.deinit();
+}
+
+pub fn groupsToZon(allocator: NeverFailingAllocator) ZonElement {
```
IntegratedQuantum: Could unified with init (see comment above)
or if not, should assert that `groups` is empty.

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,306 @@
...
+	defer deinit();
+
+	try createGroup("test", main.heap.testingAllocator);
+	const group = groups.getPtr("test").?;
+	group.permissions.addPermission(.white, "/command/test");
+
+	try std.testing.expectEqual(true, group.permissions.removePermission(.white, "/command/test"));
+}
+
+test "inValidGroupPermission" {
```
IntegratedQuantum:
```suggestion
test "invalidGroupPermission" {
```

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,306 @@
...
+	init(main.heap.testingAllocator);
+	defer deinit();
+
+	try createGroup("test", main.heap.testingAllocator);
+	try std.testing.expectError(error.GroupNotFound, getGroup("root"));
+}
+
+test "inValidGroupPermissionEmptyGroups" {
```
IntegratedQuantum:
```suggestion
test "invalidGroupPermissionEmptyGroups" {
```

src/server/permissionLayer.zig
```diff
@@ -0,0 +1,306 @@
...
+	init(main.heap.testingAllocator);
+	defer deinit();
+
+	try std.testing.expectError(error.GroupNotFound, getGroup("root"));
+}
+
+test "inValidGroupCreation" {
```
IntegratedQuantum:
```suggestion
test "invalidGroupCreation" {
```

IntegratedQuantum (changes requested):
Also now that I think about more, you are freeing resources everywhere, so in order to access them from other threads you need to either make use of garbage collection or also assert that they are on the server thread.

src/server/permission.zig
Wunka: I unified init and groupsFromZon. 
I let groupsToZon stay if we want to save it sometimes in between joining and leaving a world.

src/server/permission.zig
Wunka: this is for the ids of the groups so they are identifiable even if a group A was deleted and then a new group A is created. That is the replacement for saving the members as you suggested.

src/server/permission.zig
Wunka: Done. groups also now use an arenaAllocator.
Also so I don't need to comment everything the rest is also done.

Wunka: Is there something I forgot here? I can and will wait for the next review cycle but also don't want this PR to go stale if both parties think they wait for the other one.

IntegratedQuantum: Yeah, sorry, I just haven't found the time to review this big PR again. 600 lines really is too much. Could you maybe split it further?

Wunka: Groups are now removed. Now its only +221. Sadly all tests relied on them so they are also now gone

Wunka: Edit: 298 now as I added as many tests as possible back

IntegratedQuantum: Thank you, that should be more managable to review

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+const sync = main.sync;
+
+const PermissionMap = struct { // MARK: PermissionMap
+	map: std.StringHashMapUnmanaged(void) = .{},
+
+	pub fn fromZon(self: *PermissionMap, allocator: NeverFailingAllocator, zon: ZonElement) void {
+		for (zon.toSlice()) |item| {
+			switch (item) {
+				.string, .stringOwned => |string| {
+					if (self.map.contains(string)) return;
```
IntegratedQuantum: Always use getOrPut instead of contains + put.

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+const User = server.User;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const NeverFailingArenaAllocator = main.heap.NeverFailingArenaAllocator;
+const ZonElement = main.ZonElement;
+const sync = main.sync;
+
+const PermissionMap = struct { // MARK: PermissionMap
+	map: std.StringHashMapUnmanaged(void) = .{},
+
+	pub fn fromZon(self: *PermissionMap, allocator: NeverFailingAllocator, zon: ZonElement) void {
```
IntegratedQuantum: Allocator parameters that are assumed to be arenas should be called `arena`.

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+		for (zon.toSlice()) |item| {
+			switch (item) {
+				.string, .stringOwned => |string| {
```
IntegratedQuantum: Instead of switching on the internals, it would be better to use `const string = item.as(?[]const u8, null) orelse continue;`

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+
+		var it = self.map.keyIterator();
+		while (it.next()) |key| {
+			zon.append(key.*);
+		}
+		return zon;
+	}
+
+	pub fn put(self: *PermissionMap, allocator: NeverFailingAllocator, key: []const u8) void {
+		_ = self.map.getOrPut(allocator.allocator, key) catch unreachable;
```
IntegratedQuantum: Please only dupe the key if it isn't already present. (also the dupe should be done internally)

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+		self.list(.black).fromZon(self.arena.allocator(), zon.getChild("permissionBlacklist"));
+	}
+
+	pub fn toZon(self: *Permissions, allocator: NeverFailingAllocator, zon: *ZonElement) void {
+		zon.put("permissionWhitelist", self.list(.white).toZon(allocator));
+		zon.put("permissionBlacklist", self.list(.black).toZon(allocator));
+	}
+
+	pub fn addPermission(self: *Permissions, listType: ListType, permissionPath: []const u8) void {
+		sync.threadContext.assertCorrectContext(.server);
```
IntegratedQuantum, line 86: Please also add the assertion to the fromZon and toZon functions, as well as the hasPermission function. If either of these are called from a wrong thread, we have a possible crash.

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+// ––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––
+// MARK: Testing
+// ––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––
+
+test "WhitePermission" {
+	var permissions: Permissions = .init(main.heap.testingAllocator);
+	defer permissions.deinit();
+
+	permissions.addPermission(.white, "/command/test");
+	try std.testing.expectEqual(.yes, permissions.hasPermission("/command/test"));
```
IntegratedQuantum, line 118: Please also check the parent permission `/command`

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+
+test "Blacklist" {
+	var permissions: Permissions = .init(main.heap.testingAllocator);
+	defer permissions.deinit();
+
+	permissions.addPermission(.white, "/command");
+	permissions.addPermission(.black, "/command/test");
+
+	try std.testing.expectEqual(.no, permissions.hasPermission("/command/test"));
+	try std.testing.expectEqual(.yes, permissions.hasPermission("/command"));
```
IntegratedQuantum, line 131: Please also check the parent permission `/`

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+
+	pub fn addPermission(self: *Permissions, listType: ListType, permissionPath: []const u8) void {
+		sync.threadContext.assertCorrectContext(.server);
+		self.list(listType).put(self.arena.allocator(), self.arena.allocator().dupe(u8, permissionPath));
+	}
+
+	pub fn removePermission(self: *Permissions, listType: ListType, permissionPath: []const u8) bool {
+		sync.threadContext.assertCorrectContext(.server);
+		_ = self.list(listType).map.remove(permissionPath);
+		return true;
```
IntegratedQuantum: What's the point of the return value? Did you mean to return the result of remove?

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+	permissions.addPermission(.white, "/server/command/testing/test");
+
+	try std.testing.expectEqual(.yes, permissions.hasPermission("/server/command/testing/test"));
+	try std.testing.expectEqual(.neutral, permissions.hasPermission("/server/command/testing"));
+	try std.testing.expectEqual(.neutral, permissions.hasPermission("/server/command"));
+	try std.testing.expectEqual(.neutral, permissions.hasPermission("/server"));
+	try std.testing.expectEqual(.neutral, permissions.hasPermission("/server/command/testing/test2"));
+}
+
+test "RootPermission" {
```
IntegratedQuantum, line 149: Please also add a test that checks whether the blacklist works if the root permission is set.

src/server/permission.zig
```diff
@@ -0,0 +1,185 @@
...
+	var permissions: Permissions = .init(main.heap.testingAllocator);
+	defer permissions.deinit();
+
+	permissions.addPermission(.white, "/");
+
+	try std.testing.expectEqual(.yes, permissions.hasPermission("/command/test"));
+}
+
+test "ddRemovePermission" {
```
```suggestion
test "addRemovePermission" {
```
Please also add a test for attempting to remove a permission that doesn't exist

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,84 @@
...
+		if (std.ascii.eqlIgnoreCase(arg, "remove")) {
+			const helper = Helper.parseHelper(source, &split) catch |err| switch (err) {
+				error.InvalidAmount => {
+					source.sendMessage("#ff0000Not the right amount of arguments for /perm remove", .{});
+					return;
+				},
+				error.InvalidArg => return,
+			};
+			if (helper.permissionPath) |permissionPath| {
+				_ = source.permissions.removePermission(helper.listType, permissionPath);
```
IntegratedQuantum: Please print a message to inform the user if the permission wasn't present.

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,84 @@
...
+				return;
+			} else {
+				source.sendMessage("#ff0000Not the right amount of arguments for /perm remove", .{});
+				return;
+			}
+		}
+	}
+	split.reset();
```
IntegratedQuantum: Instead of resetting, I'd suggest to `peek` in the above case.

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,84 @@
...
+				},
+				error.InvalidArg => return,
+			};
+			if (helper.permissionPath) |permissionPath| {
+				_ = source.permissions.removePermission(helper.listType, permissionPath);
+				return;
+			} else {
+				source.sendMessage("#ff0000Not the right amount of arguments for /perm remove", .{});
+				return;
+			}
```
IntegratedQuantum, line 39: This case should be catched by the helper, to avoid duplicating the logic here.

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,84 @@
...
+
+const main = @import("main");
+const User = main.server.User;
+const permission = main.server.permission;
+const ListType = permission.Permissions.ListType;
+
+pub const description = "Performs changes on the permissions of the player or a groupp.";
+pub const usage =
+	\\/perm <whitelist/blacklist> <permissionPath>
+	\\/perm remove <whitelist/blacklist> <permissionPath>
```
IntegratedQuantum, line 11: Following the pattern used by other commands, I'd suggest three options: `add`, `remove` and ` ` (empty) which just queries the state.

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,84 @@
...
+	if (args.len == 0) {
+		source.sendMessage("#ff0000Too few arguments for command /perm. Expected at least two arguments.", .{});
+		return;
+	}
+	var split = std.mem.splitScalar(u8, args, ' ');
+	if (split.next()) |arg| {
+		if (std.ascii.eqlIgnoreCase(arg, "remove")) {
+			const helper = Helper.parseHelper(source, &split) catch |err| switch (err) {
+				error.InvalidAmount => {
+					source.sendMessage("#ff0000Not the right amount of arguments for /perm remove", .{});
```
IntegratedQuantum: Why can thios not be printed inside the helper function?

IntegratedQuantum (changes requested):

src/server/command/permission/perm.zig
Wunka: done, for now it only prints if the user has permission when using ` ` if your plan was to print the whole tree then I would wait with that until the follow up issue about sending the hole tree to the user for prediction

src/server/permission.zig
Wunka: done as the rest. As I needed to check before to not dupe the key the `getOrPut` is gone and replaced with a `contains`, `put`

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,80 @@
...
+pub const description = "Performs changes on the permissions of the player or shows the if has permission for a specific permission path";
+pub const usage =
+	\\/perm add <whitelist/blacklist> <permissionPath>
+	\\/perm remove <whitelist/blacklist> <permissionPath>
+	\\/perm <permissionPath>
+;
+
+pub fn execute(args: []const u8, source: *User) void {
+	if (args.len == 0) {
+		source.sendMessage("#ff0000Too few arguments for command /perm. Expected at least one arguments.", .{});
```
```suggestion
		source.sendMessage("#ff0000Too few arguments for command /perm. Expected at least one argument.", .{});
```

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,80 @@
...
+	if (split.next()) |arg| {
+		if (std.ascii.eqlIgnoreCase(arg, "remove")) {
+			const helper = Helper.parseHelper(source, &split) catch return;
+			if (!source.permissions.removePermission(helper.listType, helper.permissionPath)) {
+				source.sendMessage("#ff0000Permission path {s} is not present inside users permission {s}list", .{helper.permissionPath, @tagName(helper.listType)});
+			}
+		} else if (std.ascii.eqlIgnoreCase(arg, "add")) {
+			const helper = Helper.parseHelper(source, &split) catch return;
+			source.permissions.addPermission(helper.listType, helper.permissionPath);
+		} else if (arg[0] == '/') {
```
IntegratedQuantum, line 30: You probably should print an error if it's not a `/`.

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,80 @@
...
+			}
+			if (source.hasPermission(arg)) {
+				source.sendMessage("#00ff00User has permission for path: {s}", .{arg});
+			} else {
+				source.sendMessage("#ff0000User has no permission for path: {s}", .{arg});
+			}
+		}
+		return;
+	}
+	source.sendMessage("#ff0000Not the right amount of arguments for /perm remove", .{});
```
```suggestion
	source.sendMessage("#ff0000Not the right amount of arguments for /perm", .{});
```

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,80 @@
...
+			if (std.ascii.eqlIgnoreCase(arg, "whitelist")) {
+				listType = .white;
+			} else if (std.ascii.eqlIgnoreCase(arg, "blacklist")) {
+				listType = .black;
+			} else {
+				source.sendMessage("#ff0000Expected either whitelist or blacklist, found \"{s}\"", .{arg});
+				return error.InvalidArgs;
+			}
+		} else {
+			source.sendMessage("#ff0000Not the right amount of arguments for /perm remove", .{});
```
```suggestion
			source.sendMessage("#ff0000Not the right amount of arguments for /perm", .{});
```

src/server/command/permission/perm.zig
IntegratedQuantum: Also is this even reachable? Isn't this already covered by the above `args.len != 0` check?

src/server/command/permission/perm.zig
```diff
@@ -65,20 +56,25 @@ const Helper = struct {
...
 
 		var permissionPath = split.next();
 
 		if (permissionPath != null and permissionPath.?[0] != '/') {
 			source.sendMessage("#ff0000Permission paths always begin with a \"/\", got: {s}", .{permissionPath.?});
-			return error.InvalidArg;
+			return error.InvalidArgs;
+		}
+
+		if (split.next() != null or permissionPath == null) {
```
IntegratedQuantum: I'd suggest to separate these into one path for "Not enough arguments" and one path for "too many arguments"

src/server/command/permission/perm.zig
```diff
@@ -65,20 +56,25 @@ const Helper = struct {
 				listType = .black;
 			} else {
 				source.sendMessage("#ff0000Expected either whitelist or blacklist, found \"{s}\"", .{arg});
-				return error.InvalidArg;
+				return error.InvalidArgs;
 			}
-		} else return error.InvalidAmount;
+		} else {
+			source.sendMessage("#ff0000Not the right amount of arguments for /perm remove", .{});
```
IntegratedQuantum: Should say too few arguments.
Also I'd prefer if you used orelse instead of `if () || {} else {}`

src/server/permission.zig
```diff
@@ -34,8 +28,9 @@ const PermissionMap = struct { // MARK: PermissionMap
 		return zon;
 	}
 
-	pub fn put(self: *PermissionMap, allocator: NeverFailingAllocator, key: []const u8) void {
-		_ = self.map.getOrPut(allocator.allocator, key) catch unreachable;
+	pub fn put(self: *PermissionMap, arena: NeverFailingAllocator, key: []const u8) void {
+		if (self.map.contains(key)) return;
```
IntegratedQuantum: Again, please use getOrPut, instead of contains+put (note you can dupe the string after insertion by changing the key pointer)

IntegratedQuantum (changes requested):
Also one thing from the previous review is not complete

src/server/permission.zig
Wunka: ah yes that makes sense. done

src/server/command/permission/perm.zig
Wunka: ... I  looked at the everything except the helper if I removed the "remove"... Rest is also done (this time really)

src/server/command/permission/perm.zig
```diff
@@ -70,8 +69,13 @@ const Helper = struct {
 			return error.InvalidArgs;
 		}
 
-		if (split.next() != null or permissionPath == null) {
-			source.sendMessage("#ff0000Not the right amount of arguments for /perm", .{});
+		if (permissionPath == null) {
```
IntegratedQuantum: Could be done with orelse, to avoid the `.?` and != null checks elsewhere.

IntegratedQuantum (changes requested):

src/server/command/permission/perm.zig
Wunka: true. Changed

src/server/command/permission/perm.zig
```diff
@@ -0,0 +1,82 @@
...
+		if (std.ascii.eqlIgnoreCase(arg, "whitelist")) {
+			listType = .white;
+		} else if (std.ascii.eqlIgnoreCase(arg, "blacklist")) {
+			listType = .black;
+		} else {
+			source.sendMessage("#ff0000Expected either whitelist or blacklist, found \"{s}\"", .{arg});
+			return error.InvalidArgs;
+		}
+
+		var permissionPath = split.next() orelse {
```
IntegratedQuantum: And should be const
```suggestion
		const permissionPath = split.next() orelse {
```

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #2531, opened 2026-01-30
Change block interaction override to crouch keybind instead of shift

hazelmeowing:
"Block interaction override" being the key you hold to place a block while looking at something that would normally do something else, currently it is always `mods.shift` which can be problematic. For example I use shift for sprint and toggle sprint, meaning whenever I want to use this I toggle whether I am running, which is bothersome (additionally, some people might not have `mods.shift` bound for anything else, which would make this very unnatural). I don't think it is necessary to add a special keybind for this, and changing it to `"crouch"` will be familiar for people coming from Minecraft.

I did notice however there is `dropFromHand` which also checks `mods.shift` to decide whether to drop a whole stack or not, currently I have not changed this as I checked in Minecraft and for some reason this is always shift unlike block interaction overriding, I do feel this is an inconsistency and I obviously don't think Cubyz should seek to be inconsistent just because Minecraft is, but I am unsure whether it would be better to add a new keybind for this or make it `"crouch"`

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -529,14 +529,14 @@ pub const Player = struct { // MARK: Player
 		}
 	}
 
-	pub fn placeBlock(mods: main.Window.Key.Modifiers) void {
+	pub fn placeBlock() void {
 		if (main.renderer.MeshSelection.selectedBlockPos) |blockPos| {
-			if (!mods.shift) {
+			if (!KeyBoard.key("crouch").pressed) {
 				if (main.renderer.mesh_storage.triggerOnInteractBlockFromRenderThread(blockPos[0], blockPos[1], blockPos[2]) == .handled) return;
 			}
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
We used to have a key for this (which got removed in the refactoring of #2144)

I do agree that we should have this remappable, but I do not think that this is the right approach. Tying it to crouching, a completely unrelated activity.

Also I do like the modifiers in the sense that now there is a clear and fixed list of key modifiers you can check, instead of having to query a random key.
So maybe we could just introduce a new entry in the modifers struct that is remappable.

BoySanic:
@IntegratedQuantum Did you mean to close this or leave it open?

IntegratedQuantum:
I kept it open in case they wanted to work on the suggested alternative approach or have other suggestions, I usually don't just close PRs like this.

hazelmeowing:
I will close it I guess because I don't really feel inspired to do anything more complicated with keybinds, it was just something I found bothersome in game

H41ogen:
> So maybe we could just introduce a new entry in the modifers struct that is remappable.

or just a new keybind called "modifier" or smth that's shift by default

PR #2533, opened 2026-01-31
Add advanced controls setting creativeReach

hazelmeowing:
Adds creativeReach setting which modifies the players block selection range if they are in creative mode. Ranges from 6 to 240 moving by increments of 6. I put it in advanced controls because it seems pretty similar to block place/break speed/delay to me.

~~Also I accidentally added `Merge branch 'PixelGuys:master' into creativereach_setting` to this branch and as I'm inexperienced with git I'm worried about messing things up if I try to remove it, but it doesn't seem to actually change anything? So I think it is fine..~~ Edit: I think I fixed it


IntegratedQuantum:
With #2420 the engine could at most support a reach of 32 blocks.

BoySanic:
Did you have more to review here @IntegratedQuantum? It wasn't closed from your comment and you didn't directly suggest changes, so I'm wondering what the next step here is.

IntegratedQuantum:
The next step would be to limit the reach to ensure that it stays within 32 blocks - player model size or something like that.

PR #2544, opened 2026-02-07
Make blueprints store and paste block entity data

### LINKED ISSUE #1743: [WE] Add support for capturing block entity data in blueprints
### Description

Support for capturing block entity data in a blueprint is necessary to support undo / redo of edits containing block entities.
This must be supported at game engine level to be viable.

Related to: #1739 
Related to: #1446

OneAvargeCoder193:
closes #1743

### DIFF
--- a/src/blueprint.zig
+++ b/src/blueprint.zig
@@ -35,15 +35,11 @@ pub const BlueprintCompression = enum(u16) {
 
 pub const Blueprint = struct {
 	blocks: Array3D(Block),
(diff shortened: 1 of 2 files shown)


IntegratedQuantum:
As I already said on discord: The current binary format of block entities has some problems when it comes to allowing multiple block entities for a block, so to avoid having to migrate blueprints (which we cannot really do), I would like to have this sorted out first.
I'll try to work on this soon.

PR #2546, opened 2026-02-09
Add bobbing

lukaswrz:
This PR adds bobbing so that walking doesn't feel "on-rails".


RanPix:
could you please upload a video of how that looks?

lukaswrz:
Sure: https://youtu.be/MIPdgzWBX_c

RanPix:
yes.

ikabod-kee:
I'm going to try this out!

ikabod-kee (approved):
I don't know how you did it, but you managed to get it feeling right on the very first try. No complaints from me. Well done!!

(Now we just have to make our up & down bobbing meet the same quality as this!)

BoySanic:
Can we add an option for enabling/disabling bobbing?

lukaswrz:
> Can we add an option for enabling/disabling bobbing?

Should that disable the vertical movement (i.e. when jumping/falling) as well (which is what Minecraft does, for example), or should it just disable the bobbing that is being added by this PR?

ikabod-kee:
https://github.com/user-attachments/assets/c54f5d8c-b956-480b-8049-740cfb884a9c

It seems the animation doesn't reset when stopping and starting. Also a slow fade-in like you suggested in the Discord would make sense!

ikabod-kee:
https://github.com/user-attachments/assets/d603b989-20c3-4e5b-9605-e8bf8b4438a3

Also when jumping around, the item always seems to go to the right position, even when jumping when it's positioned to the left.

ikabod-kee:
Wait I just posted the same video twice 😭 my bad.

https://github.com/user-attachments/assets/d64724c1-b9ce-48ce-a99e-5eea143010ec

IntegratedQuantum:
One thing that is weird is how fast the bobbing frequency slows down when you move slower. It feels like one of these toys that have wave their arm as you move them over the ground.

In real life if you walk slower, then your step length also reduces, ranging from maybe 1/4 of a meter to several meters when sprinting. Here your step length seems to be constant.

lukaswrz:
> It seems the animation doesn't reset when stopping and starting

This is intentional. I think it would look weirder if the item being held just kept going in the same direction every time you pressed W. That said, I don't think it matters all that much, since this isn't something that you would be doing during actual gameplay.

lukaswrz:
@ikabod-kee Feel free to test it again, I've made some changes that should make it look better.

ikabod-kee:
Alright! I'll get to it soon

ikabod-kee (changes requested):
Okay, this is definitely a lot better now! I will make one suggestion right quick though:
Could you make it so the hand slows down its animation when jumping instead of immediately stopping?

ikabod-kee:
Also, when they are airborne, could you make it so the hand (very) slowly goes back to the original position? (with some easing, of course)

image

ikabod-kee:
Another complaint: The fade-in for going from stopped to running is too long.
Sorry to be a bother about these adjustments!

lukaswrz:
> The fade-in for going from stopped to running is too long

Can you tell me the value that you'd prefer? You can test a few values by adjusting the `fadeInSpeed`.

>  Sorry to be a bother about these adjustments!

No worries, all good.

ikabod-kee:
> Can you tell me the value that you'd prefer? You can test a few values by adjusting the fadeInSpeed. The max is 1.

Sure, I can tell you! Where can I find this setting?

lukaswrz:
It's a constant defined in the code. I don't think it makes sense to introduce a GUI setting for this, instead, I just added one to turn bobbing on or off entirely, because I think that's what most players would be looking for.

ikabod-kee:
Where is it in the code? I'm not a programmer XD

lukaswrz:
Line 580 in src/itemdrop.zig.

ikabod-kee:
I decided on 5.0 as the "fadeInSpeed" value, because I like how snappy it feels. You immediately know when you are walking!
As for the "fadeOutSpeedOnGround," I couldn't notice a difference when changing it.

ikabod-kee:
I actually think having it be an immediate start/stop when pressing the walk keys would make it feel extremely responsive! The only issue I really have with the bobbing otherwise would how it acts when airborne.

ikabod-kee (changes requested):
You managed to fix most of my complaints with this recent commit! It's pretty much ready to merge, but I ask that you try to make the animation reset when stopped so that the bobbing goes *right* every time you begin walking. I believe this will help a bit in its "feel."

Basically "trust me bro."

ikabod-kee:
Oh, and if you can, could you make it so going into fly mode doesn't immediately jerk it like this?

https://github.com/user-attachments/assets/7385ed8b-3309-475c-94ea-3b9ccb5113a4

ikabod-kee:
It'd also be nice to have the player only bob when a movement key is held, so it doesn't look like they're walking on ice when they're sliding (because you know, they're sliding.)

You don't have to do this one as I feel tying it to movement keys may leave a slight period in between switching directions that can stop the animation. Unless you want to. It's just a "would be nice to have" thing.

https://github.com/user-attachments/assets/eff673f8-409c-4d57-bdc2-0eff49c0ab5f

lukaswrz:
Going into fly mode is smoother now, and sliding on ice should look better as well.

ikabod-kee (approved):
PERFECT!!

ikabod-kee:
I love the little detail you added where the item gets lowered when crouch-walking.

IntegratedQuantum:
I've noticed that running down feels kind of weird with the way the animation resets in the air:

https://github.com/user-attachments/assets/78505e7d-5946-41b7-9493-5be7a251fd50

lukaswrz:
I've made some adjustments to the animation reset speed (e.g., when running down hillsides). It's fairly subjective, but I'm pretty happy with how it looks now.

IntegratedQuantum:
Sorry for the late reply. It feels better, but it still feels a bit off when running down. Sometimes it feels like the animation stops too rapidly, and it the direction of the animation also feels kind of random. Would it be possible to always alternate the direction every time the player makes contact with the ground while walking down?

H41ogen:
Is this abandoned...? :(

Crepestrom:
i am willing to take it then

lukaswrz:
> Is this abandoned...? :(

No, I think it's a bit early to assume that.

Crepestrom:
could you make the item droop down a bit when the player hits the ground?

IntegratedQuantum:
Just to be clear, I'm still waiting on a fix for https://github.com/PixelGuys/Cubyz/pull/2546#issuecomment-3941167953

lukaswrz:
> Just to be clear, I'm still waiting on a fix for [#2546 (comment)](https://github.com/PixelGuys/Cubyz/pull/2546#issuecomment-3941167953)

Yeah, sorry for the delay.

> Sometimes it feels like the animation stops too rapidly, and it the direction of the animation also feels kind of random. Would it be possible to always alternate the direction every time the player makes contact with the ground while walking down?

The animation will now continue/finish rather than instantly reset when airborne, which also has the effect of alternating the direction.

lukaswrz:
> could you make the item droop down a bit when the player hits the ground?

Sure, but I'm not certain if it's a good idea to add that in this PR. That could easily be added later, entirely separate from the current bobbing code.

IntegratedQuantum:
Yes, this is much better. @ikabod-kee do you have anything more to add?

ikabod-kee:
Nope! I tested it out and it's great!

src/itemdrop.zig
```diff
@@ -543,27 +543,130 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
 	}
 };
 
-// Going to handle item animations and other things like - bobbing, interpolation, movement reactions
+const BobManager = struct {
+	phase: f32 = 0,
+	scale: f32 = 0,
+	sneakOffset: f32 = 0,
```
IntegratedQuantum: do you intend to use this for anything besides the local player? If not there is no reason to make this an object.

src/itemdrop.zig
```diff
@@ -543,27 +543,130 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
...
+	const phaseResetSpeedFlying = 3;
+	const phaseResetSpeedStanding = 2;
+
+	const scaleThreshold = 0.1;
+	const scaleFadeSpeedWalking = 4;
+	const scaleFadeSpeedFlying = 3;
+	const scaleFadeSpeedStanding = 2;
+
+	const sneakScaleMul = 0.2;
+	const sneakFadeSpeed = 5;
```
IntegratedQuantum: I'm a bit overwhelmed by all these constants, like what even does `5` fade speed even mean?
I think it would help if you could add some units, and try to express it in more tangible units. e.g. fade should probably be expressed as a half-life time, not per-second (what does that even mean? 5 fades per second???).

src/itemdrop.zig
```diff
@@ -543,27 +543,130 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
...
+
+	fn fadeScale(self: *@This(), dt: f32, newScale: f32, fadeSpeed: f32) void {
+		self.scale = std.math.lerp(self.scale, newScale, @min(dt*fadeSpeed, 1));
```
IntegratedQuantum: Should use exponential relationship to be fps-independent and avoid the odd min.

src/itemdrop.zig
```diff
@@ -543,27 +543,130 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
 	}
 };
 
-// Going to handle item animations and other things like - bobbing, interpolation, movement reactions
+const BobManager = struct {
+	phase: f32 = 0,
+	scale: f32 = 0,
```
IntegratedQuantum: I think a better word would be amplitude.

IntegratedQuantum (changes requested):

src/itemdrop.zig
```diff
@@ -543,27 +543,131 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
...
+
+		const inputSpeed: f32 = @floatCast(game.Player.inputSpeed);
+
+		const horizontalVel: Vec3f = .{
+			@floatCast(game.Player.super.vel[0]),
+			@floatCast(game.Player.super.vel[1]),
+			0,
+		};
+
+		const actualSpeed = vec.length(horizontalVel);
```
IntegratedQuantum, line 611: Shouldn't that be named horizontalSpeed?

src/itemdrop.zig
```diff
@@ -543,27 +543,131 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
...
+
+		const newScale = @min(inputSpeed/inputSpeedMax, actualSpeed/actualSpeedMax, 1);
```
IntegratedQuantum: I've noticed that on ice the animation is delayed because it takes half a second before the actualSpeed reaches a high enough value. Is there a reason why we can't just only use the input speed?

IntegratedQuantum (changes requested):

src/itemdrop.zig
lukaswrz: I named it `actualSpeed` to make it clear that it's analogous to `inputSpeed`, but I'd be fine with naming it `horizontalSpeed` as well.

src/itemdrop.zig
lukaswrz: If we only use the input speed, the animation will play even when e.g. the player is walking against a wall, which I think doesn't really make sense. The animation would also keep the same speed when the player is being slowed down by e.g. walking in water.

H41ogen: how about only check whether the actualSpeed is above the threshols if there's no input speed?

IntegratedQuantum: Maybe the right solution would then be to ignore the input velocity in a direction, if the actual velocity in that direction is exactly zero? That should prevent the wall issues without impacting anything else.

ikabod-kee:
Any updates on this? Would love to have this in the game.

Wbjitscool:
yeah same

ikabod-kee:
Will someone be taking over this PR at some point? It might get stale.

PR #2549, opened 2026-02-13
Bitter columns

TyrrelMeter:
A cold hilly biome with some pillars. Not to much of a description right now because just look at the pictures. Name is a work in progress.

Cubyz 2_21_2026 8_36_38 AM
Cubyz 2_21_2026 8_36_19 AM
Cubyz 2_21_2026 8_36_03 AM
Cubyz 2_21_2026 8_35_58 AM
Cubyz 2_21_2026 8_35_52 AM


TyrrelMeter:
As much as I'd like to, unless they've changed something putting ores on structures breaks things.

careeoki:
I think there is too many different blocks in these small structures. It's not that it looks bad, and it would be perfectly fine on a player build, but for natural structures it's too much. 
If you wanted to tear down one of these, 6 slots of inventory space would be taken up by all the different blocks, which is very irritating.
I think glacite -> gravel -> slate would be more reasonable.

There should also be much more size and height variation for these.

PR #2557, opened 2026-02-14
Abandoned Cavern

TyrrelMeter:
Explore caverns with structures already built into them that you did not place. How did they get here, and who made them? You haven't been here before.
Cubyz 2_14_2026 6_08_09 PM
Cubyz 2_14_2026 6_08_00 PM


careeoki:
"Abandoned" implies something was living here, but clearly it was just passing through and pillaring/bridging around.
Something like "vitals" might be more fitting.

H41ogen:
perhaps "explored cavern"

Argmaster:
These are *the ugliest bridges* you can make and you want to generate them in the world by default 💀

tillpp:
> These are the ugliest bridges you can make and you want to generate them in the world by default 💀

Makes the Player feel good about themself :skull:

Wbjitscool:
I know now that we have pillars and stuff we can make better looking bridges and etc and make them look ruined like a real abandoned cavern with broken bridges with iron chains or something and moss and etc

Wbjitscool:
something more like this maybe better?
image

TyrrelMeter:
Wow people are actually looking at this again. I was going to close it because I didn't feel like updating it.

PR #2559, opened 2026-02-15
Glimmer Worms attempt 2

Crepestrom:
Re-Formatted #2255


IntegratedQuantum:
In the future please don't open a new PR for an existing one. Instead I'd suggest to learn how to properly use git locally, it is worth it.

IntegratedQuantum (approved):
@TyrrelMeter since you also used glimmergill in some of your phantasmal biomes, it might be worth to consider adding these there as well.

PR #2560, opened 2026-02-15
Update the main menu layout

### LINKED ISSUE #2040: Display version number on title screen

I think it would be helpful if we put a label on the title screen (top left maybe?) with contents:

1. `"Cubyz version {}", .{.{settings.version.version}}`

OneAvargeCoder193:
New Layout:
image

This is based of Ikabod's drawings from last year:
image

(The logo is temporary for now, at least until carrie figures something out, it is just an image made of a large scale unscii font with a slight darkening around the borders)
This also fixes #2040


OneAvargeCoder193:
if we want to do the darker buttons I would need the assets for the darker buttons, as well as their pins

OneAvargeCoder193:
specifically i would need the idle, pressed, and hovered nine slice textures, and the pin texture that i can just paste on the sides

ikabod-kee:
While you're at it, could put the version number somewhere to fix #2040

ikabod-kee:
Would you be able to remove the automatic shadow that's applied to the logo?

OneAvargeCoder193:
done

src/gui/components/Button.zig
```diff
@@ -134,12 +167,15 @@ pub fn mainButtonReleased(self: *Button, mousePosition: Vec2f) void {
 }
 
 pub fn render(self: *Button, mousePosition: Vec2f) void {
+	const pressed = if (self.style == .mainMenu) pressedMainMenuTextures else pressedTextures;
+	const hovered = if (self.style == .mainMenu) hoveredMainMenuTextures else hoveredTextures;
+	const normal = if (self.style == .mainMenu) normalMainMenuTextures else normalTextures;
```
IntegratedQuantum: Please put them into a helper struct to avoid this mess.
Also I think this should be grouped together with the below logic into a helper function in the style enum.
```zig
const textures = style.getTextures(self.pressed, GuiComponent.contains(self.pos, self.size, mousePosition) and self.hovered)
```

Also @ikabod-kee @careeoki what is the plan here? Do we keep two different types of buttons?

src/gui/components/Button.zig
```diff
@@ -163,6 +199,11 @@ pub fn render(self: *Button, mousePosition: Vec2f) void {
 		graphics.draw.boundSubImage(self.pos + Vec2f{0, cornerSize[1]}, Vec2f{cornerSize[0], self.size[1] - 2*cornerSize[1]}, .{0, lowerTexture[1]}, .{cornerSizeUV[0], upperTexture[1] - lowerTexture[1]});
 		graphics.draw.boundSubImage(self.pos + Vec2f{self.size[0] - cornerSize[0], cornerSize[1]}, Vec2f{cornerSize[0], self.size[1] - 2*cornerSize[1]}, .{upperTexture[0], lowerTexture[1]}, .{cornerSizeUV[0], upperTexture[1] - lowerTexture[1]});
 	}
+	if (self.style == .mainMenu) {
+		nailTexture.bindTo(0);
+		graphics.draw.boundImage(self.pos + Vec2f{6, self.size[1]/2 - 4}, Vec2f{8, 8});
+		graphics.draw.boundImage(self.pos + Vec2f{self.size[0] - 12, self.size[1]/2 - 4}, Vec2f{8, 8});
```
IntegratedQuantum, line 204: Since you add new textures to the sides of the button, it might make sense to adjust the text wrapping so it doesn't overlap.

Image

src/gui/windows/main.zig
```diff
@@ -1,30 +1,50 @@
...
 	main.Window.c.glfwSetWindowShouldClose(main.Window.window, main.Window.c.GLFW_TRUE);
 }
 pub fn onOpen() void {
-	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
-	list.add(Button.initText(.{0, 0}, 128, "Singleplayer", gui.openWindowCallback("save_selection")));
-	list.add(Button.initText(.{0, 0}, 128, "Multiplayer", gui.openWindowCallback("multiplayer")));
-	list.add(Button.initText(.{0, 0}, 128, "Settings", gui.openWindowCallback("settings")));
-	list.add(Button.initText(.{0, 0}, 128, "Touch Grass", .init(exitGame)));
+	const list = VerticalList.init(.{padding, 16 + padding}, 300, 2);
+	list.add(Icon.init(.{0, 0}, .{384, 96}, logo, true));
```
IntegratedQuantum: There is a reason why we don't (and shouldn't) have full-screen windows that fill the full height of the window:
Image
Also be aware that languages may produce text wrapping on the button, which would also put stuff outside the window.

I'd suggest to use separate windows for logo, buttons and version number. These could then also be freely moved by the player if we add an equivalent of the "Reorder Hud" button.

src/gui/components/Button.zig
IntegratedQuantum: Also Ikabod's drawing positioned the nails pixel perfect, yours does not. Could you fix that?

IntegratedQuantum (changes requested):
There is also a weird edge case that allows you to move stuff behind the main menu:
Image

src/gui/windows/main.zig
OneAvargeCoder193: i dont believe that the window takes up the full size of the screen, it only uses enough for the buttons, for the outline it just does some things that let it render outside the window's bounds

OneAvargeCoder193:
> There is also a weird edge case that allows you to move stuff behind the main menu

is there a way to force a window to be behind everything else?

src/gui/components/Button.zig
ikabod-kee: Brother I know naught of what you speak.

ikabod-kee:
If you haven't already, could you make the logo pixel perfect somehow? I know it *looks* pixel perfect, but it's not XD

src/gui/components/Button.zig
OneAvargeCoder193: there is issues with slight pixel offsets everywhere in the gui, by the math it should be pixel perfect, however it is not. a similar issue can be seen in scroll bars. i think the problem with it not being pixel perfect is somewhere deeper in the gui

src/gui/components/Button.zig
OneAvargeCoder193: > Brother I know naught of what you speak.

i think quantum is asking if we kep a seperate button style for the main menu and for the rest of the gui, or if we have them be the same

ikabod-kee: I think having the button style be separate is fine. It's just a darker version of the current button.

ikabod-kee: Unless you guys want to change that. Be my guess! Anything for visual clarity!

src/gui/components/Button.zig
OneAvargeCoder193: also in terms of text wrapping, isn't this an argument for the buttons, not something calculated by the buttons?

OneAvargeCoder193:
> If you haven't already, could you make the logo pixel perfect somehow? I know it _looks_ pixel perfect, but it's not XD

this should be done now

src/gui/windows/main.zig
IntegratedQuantum: Well, as you can see in the image, the window is too big, whether it's full size or almost full size doesn't really matter.

src/gui/components/Button.zig
IntegratedQuantum: It is calculated directly using the width of the button `width - 3*border`

src/gui/components/Button.zig
IntegratedQuantum: Well, but is it really desirable to have two different button textures?

careeoki: I would rather have 1 button style. The darker wood colour has better contrast against the white text than the old buttons anyway.
It does create other problems though, I would want to redo the UI colour palette to match the new buttons better. Which means changing a bunch of textures. Which is probably outside this PR's scope

careeoki:
[I did a PR with a new wordmark](https://github.com/OneAvargeCoder193/Cubyz/pull/8):

image

And yes, I know not everyone is a fan of the spiral b. However, there's more people who prefer the spiral b than those who prefer the normal b.
And as the one who actually _made_ it, I prefer the spiral b. It creates visual interest and gives a unique characteristic to otherwise a pretty basic font.

src/gui/windows/main.zig
OneAvargeCoder193: i seperated the title and the buttons into their own windows, however i dont know how to fix the problem of pre-existing gui_layouts, since i moved the position of the main window to be at 0.5, 0.75

careeoki:
[another PR to swap to the latest 1 pixel outline version](https://github.com/OneAvargeCoder193/Cubyz/pull/9)

BoySanic:
Since master now includes the version in the titlebar, do we need the version in the topleft?

ikabod-kee:
image

@OneAvargeCoder193 Could you go off of this image instead? The layout only changed a slight amount and the title and buttons were given a drop shadow.

Oh, and keep the cool outline around the screen :)

ikabod-kee:
cubyz3
Here is the logo

Argmaster:
@ikabod-kee well since you requested use of different assets, were do we get those button assets from? ~~Or did you just mean the layout?~~ clearly not

ikabod-kee:
image
Bbutton. I'll have hovered and pressed versions soon

careeoki:
You should probably split that into individual textures like this: https://discord.com/channels/443805812390100992/695680181075050586/1477020648164294797

ikabod-kee:
Alright, I'll do that.

ikabod-kee:
bg
button
buttonhover
buttonpress

PR #2563, opened 2026-02-15
Landmarks

Crepestrom:
Adds landmarks
One time large structures that look cool and are eye catching

image
* not every structure added

A few are made for most biomes but more can be easily added on later

### DIFF
--- a/assets/cubyz/biomes/cave/stalagmite_cave.zig.zon
+++ b/assets/cubyz/biomes/cave/stalagmite_cave.zig.zon
@@ -37,7 +37,7 @@
 		},
 		.{
 			.id = "cubyz:sbb",
-			.structure = "cubyz:landmarks/cave/marble/marble_landmarks",
+			.structure = "cubyz:landmarks/cave/cave_landmarks",
 			.chance = 0.01,
 		},
 	},


ikabod-kee:
No screenshots??

Crepestrom:
ah i didn't make any baobab landmarks

Crepestrom:
also the formatter isnt really helping

ikabod-kee (changes requested):
image
This structure seems to be completely broken. It also spawns way too much and has zero variation

image
This one too.

image
It feels rather unnatural with its disc shape. Reminds me of Minecraft modded structures.

Okay I'm not going to go over every structure you added, but it seems to me like there's too much in this PR and it's all broken or not given enough care. These landmarks are too common to be called "landmarks." I feel like a landmark should wait until heightmaps or larger structure sizes are added. Also for future reference, limit the amount of things you add in one PR, as this is far too much to review.

On their own though? These builds are pretty nice-looking.

BoySanic:
I think this would benefit a lot from #2129 which should be very close to merging

BoySanic:
Idk what happened to the CI but that failed too. Hmm. I'll put this in for review anyway just in case it's a non-issue

PR #2565, opened 2026-02-16
Root Bridges

TyrrelMeter:
Adds bridges of varying length to the root. They serve a similar purpose to torches and workbenches in the root.
Cubyz 0 2 0-dev 4_3_2026 6_41_41 PM
Cubyz 0 2 0-dev 4_3_2026 6_37_06 PM
Cubyz 0 2 0-dev 4_3_2026 6_36_47 PM


BoySanic:
@ikabod-kee @careeoki Tagging since you're artsy folk :)

ikabod-kee:
I like it! I'd make it just a teensy bit rarer though for good measure.

careeoki:
I get the idea with using random different blocks, but it looks ugly. I would just stick to slate and planks, and you could do cobblestone as well

TyrrelMeter:
I'll work on that, though I think that I won't use cobble because who in their right mind uses that? You may as well just use slate if your just making a one wide bridge. It isn't meant for aesthetics.

careeoki:
Cobble naturally generates in boulders, so you might pick up some while caving. A bridge is the perfect time to use it up.

IntegratedQuantum (changes requested):
In my opinion the bridges are too obviously randomly generated:

I noticed that it sometimes generates sections of only one or a few blocks of a given type. I'd suggest to prefix each section with at least 4 blocks that always appear.
I particularly noticed that often there is just a single cobblestone, when a player is that deep, they probably have either none or more than one in their inventory.

I don't like how often the block type changes, frequently enough you see it change and then change back to the original type, this is not how players build normally. When they are out of material they are out, and don't find more of it later. This is especially true for wood.
I'd suggest to make the chance to change color much smaller.

The placement of torches is a mess. Especially when building an incomplete bridge, I think most players would place the torch at the end. This should be easy to do if you add a custom end cap that places a torch on the previous block instead of using null.

IntegratedQuantum:
also needs a rebase due to #2597 which renamed planks

TyrrelMeter:
I've updated the bridges to not have the block switches and only generate with a torch at the end.

Issue #2555, opened 2026-02-14
Typo in account creation/login


BoySanic:
### Steps to Reproduce and Observed Behavior

A pop up appears in red text that says "You clipboard has been cleared". It should probably say "Your"

IntegratedQuantum:
I can't believe I managed to mess that up

IntegratedQuantum:
Of course I also managed to typo the `fixees` in the commit message

PR #2571, opened 2026-02-19
fix: typo on account cleared clipboard message

tacf:
Pretty straight forward fix correction.

fix for #2555


IntegratedQuantum:
Sorry for this, but I had already planned to make this a part of the seed phrase rename which I now did in 4ab9d98b58d84bd9d640d244a70ee104662a7863 alongside another small clipboard fix.

PR #2578, opened 2026-02-21
Add hemispheres to more biomes

careeoki:
Ice caves: Many smaller hemispheres to create some treacherous cliffs
Lava caves: large hemispheres, though smaller than crystal caves
Mushroom caves: bigger hemispheres than regular caves


Crepestrom:
ah can you only control the size and not the width and hight seperatly?

IntegratedQuantum:
thank you

PR #2579, opened 2026-02-22
Make silver useful (and copper)

H41ogen:
Gives silver better durability and swing speed so that it can compete with gold, along with reducing the durability of gold. Gives copper slightly better swing speed than iron so that it has a use sometimes late-game if the player is willing to sacrifice durability. Also, gold spawns lower now, so the player gets an opportunity to use silver's `.precious` property before finding gold.

### DIFF
--- a/assets/cubyz/blocks/gold_ore.zig.zon
+++ b/assets/cubyz/blocks/gold_ore.zig.zon
@@ -5,7 +5,7 @@
 	.ore = .{
 		.veins = 2,
 		.size = 12,
-		.height = -1400,
+		.height = -2000,
 		.density = 0.25,
 	},
 	.drops = .{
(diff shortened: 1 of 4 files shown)


IntegratedQuantum:
Hmm, in my opinion it already is pretty competitive to gold, in the sense that if you don't have any gold it's not too bad to downgrade to silver.
I do agree on making gold spawn lower though.
But maybe, by making it deeper (and I'd propose to maybe make it spawn a lot deeper, a while below uranium), we could also give it more of a right to be more powerful than silver.

I'm also opposed to the copper change: Iron is meant to be a direct upgrade. We shouldn't make them usable interchangeable or we may hurt the sense of progression the player has. There will other uses for copper in the future though.

H41ogen:
it has been done

PR #2581, opened 2026-02-22
Update README.md

H41ogen:
tool crafting doesnt let you craft anything you want anymore, so the readme should reflect that

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Cubyz is a 3D voxel sandbox game (inspired by Minecraft).
 Cubyz has a bunch of interesting/unique features such as:
 - Level of Detail (→ This enables far view distances.)
 - 3D Chunks (→ There is no height or depth limit.)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,7 @@ Cubyz is a 3D voxel sandbox game (inspired by Minecraft).
 Cubyz has a bunch of interesting/unique features such as:
 - Level of Detail (→ This enables far view distances.)
 - 3D Chunks (→ There is no height or depth limit.)
-- Procedural Crafting (→ You can craft anything you want, and the game will figure out what kind of tool you tried to make.)
+- Procedural Crafting (→ There are near-infinite possibilites for tool crafting.)
```
IntegratedQuantum: I don't think we should advertise this restriction, as it's practically infinite (and with addons it really is countably infinite)
```suggestion
- Procedural Crafting (→ There are infinite possibilites for tool crafting.)
```

PR #2582, opened 2026-02-22
Workable hemaspheres

Crepestrom:
For #2577 
Makes Hemaspheres more powerfull as sdfs as they can now be defined with a tallness mult
I have set the crystal caves to have a tallness mult of 3

image

this should allow for tall pillars of empty space

### DIFF
--- a/src/server/terrain/sdf_models/hemisphere.zig
+++ b/src/server/terrain/sdf_models/hemisphere.zig
@@ -12,13 +12,13 @@ pub const id = "cubyz:hemisphere";
 
 minRadius: f32,
 maxRadius: f32,
+tallnessMult: f32,
 
 pub fn init(zon: ZonElement) ?*@This() {
 	const result = main.worldArena.create(@This());
 	result.minRadius = zon.get(f32, "minRadius", 16);
 	result.maxRadius = zon.get(f32, "maxRadius", result.minRadius);
-	result.heightMult  = zon.get(f32, "heightMult", 1);
-	result.widthMult  = zon.get(f32, "widthMult", 1);
+	result.tallnessMult = zon.get(f32, "tallnessMult", 1);
 	return result;
 }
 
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
looks like you put the PR onto the wrong branch.

Also it is not that easy, you squished the SDF, which does not produce a valid SDF anymore, an SDF gives you at every point the distance to the perimeter of the shape, by squishing it you break that constraint, which means that combining shapes won't work as well, and this also may introduce artifacts even when it's just a single shape.

Crepestrom:
ok fixed the branch issue
also forgot to divide instead of multiply

Crepestrom:
Alright ive actually properly made this this time 
sorry about that earlier i just had to sleep

src/server/terrain/sdf_models/hemisphere.zig
```diff
@@ -38,7 +40,10 @@ pub fn generate(self: *@This(), output: main.utils.Array3D(f32), interpolationSm
 		while (y < maxInt[1]) : (y += voxelSize) {
 			var z = minInt[2] & ~(voxelSize - 1);
 			while (z < maxInt[2]) : (z += voxelSize) {
-				const distanceSquare: f32 = @floatFromInt((x - relPos[0])*(x - relPos[0]) + (y - relPos[1])*(y - relPos[1]) + (z - relPos[2])*(z - relPos[2]));
+				const xDifference: f32 = @floatFromInt(x - relPos[0]);
+				const yDifference: f32 = @floatFromInt(y - relPos[1]);
+				const zDifference: f32 = @floatFromInt(z - relPos[2]);
+				const distanceSquare: f32 = (xDifference*xDifference + yDifference*yDifference + zDifference*zDifference/(self.tallnessMult*self.tallnessMult));
```
IntegratedQuantum, line 46: This produces the correct boundary, but this still results in an incorrect SDF, which breaks the merging behavior, here look at this:
This is without smooth merging (`caveSmoothness` parameter set to 0):
Image
This is with smooth merging (`caveSmoothness` parameter set to 4):
Image

As you can see: Almost no difference, but with only a 1 block gap they should have merged. This is not happening though because the SDF value above the thing increases by 4 unit (tallnessMult is set to 0.25), but should be set to 1.

The opposite is true for a high tallnessMult, here is 16, and it's suddenly very eager to merge with anything (to the point where it's hard to believe that these are even the same scene:

Image

Image

Beyond that this also influences how noise interacts with the SDF, with correct SDF, adding noise of strength 8 will at most displace the wall by 8 blocks, but with your PR it may displace it by 2 block (tallnessMult 0.25) or 32 blocks (tallnessMult 4), and if it reaches the end of the SDF computation area (the 16 block `perimeter`), it will just cut it off flat.

Crepestrom: ah do you know which variable i should look at to fix that then

Crepestrom: also thanks for the very detailed response

IntegratedQuantum: Well, I don't really know how to fix this, it seems like a rather math-heavy problem to figure out the SDF for a spheroid. Maybe try to look how other people solved a spheroid/ellipsoid or 2D ellipse.

Crepestrom:
welp i cant solve this

PR #2583, opened 2026-02-22
Fix typo in the login screen

H41ogen:


H41ogen:
is it not a waste to do the compilation/format check on a pr like this? is there a way to make it not do that?

IntegratedQuantum:
> is it not a waste to do the compilation/format check on a pr like this? is there a way to make it not do that?

Yes, but I don't see how we can easily detect this automatically. And even for these one-line string literal changes, there is always a possibility to mess things up, like e.g. including a tab in the string.

PR #2584, opened 2026-02-22
Add shapes to the rest of the caves

careeoki:
Notable changes:
- the dropoff has very large caverns
- phantasmal has one large central hemisphere
- crystal forest has large hemispheres
- 

Most of the other ones are pretty simple.

### DIFF
--- a/assets/cubyz/biomes/cave/mushroom/_defaults.zig.zon
+++ b/assets/cubyz/biomes/cave/mushroom/_defaults.zig.zon
@@ -1,6 +1,5 @@
 .{
 	.isCave = true,
-	.caveNoiseStrength = 10,
 	.fogDensity = 2,
 	.music = "cubyz:heart-of-the-beast",
 	.structures = .{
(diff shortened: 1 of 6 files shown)


assets/cubyz/biomes/cave/mushroom/_defaults.zig.zon
```diff
@@ -1,5 +1,6 @@
 .{
 	.isCave = true,
+	.caveNoiseStrength = 10,
```
IntegratedQuantum: before you copy paste that value everywhere, note that I only added this value to the base caves to show that this parameter exists. If you think 10 works better, then we could also make it the default (which is currently 8)

careeoki: Oh, I wasn't really sure how much of a difference that was making so I just carried it over. 8 is fine.
I removed it from some biomes where I see fit since it's that small of a difference

PR #2585, opened 2026-02-22
Saplings and Seeds

tillpp:
https://github.com/user-attachments/assets/e57848ed-bbaa-4ab5-a976-f27147196fd9

Be able to to add saplings, and other structures growable structures.
```zig
.{
	..
	.onTick = .{
		.type = .sapling,
		.structures = .{
			.{
				.id = "cubyz:sbb",
				.structure = "cubyz:tree/oak/white", // what structure should be grown?
				.placeMode = .degradable,
				.chance = 0.03, // chance, that tick will cause this to grow
			},
			// other alternative structures, this sapling can grow into
		},		
	},
	..
}

```


IntegratedQuantum:
Why do some of them decay instantly?

tillpp:
> Why do some of them decay instantly?

I think it's because default tree generation creates too many leafs, and they then instantly decay because their are too far away from a branch/log

IntegratedQuantum:
Are you sure that's the case?
These things could also happen if it does the updates during your placement.

tillpp:
> Are you sure that's the case? These things could also happen if it does the updates during your placement.

oh you mean, sometimes leafs are generated before stems are generated?

IntegratedQuantum:
yes

tillpp:
I think this is a general problem with the Blueprint.paste() function, so I think It's better to fix that in a separate PR.

src/server/terrain/simple_structures/Boulder.zig
```diff
@@ -33,7 +33,7 @@ pub fn loadModel(parameters: ZonElement) ?*Boulder {
 	return self;
 }
 
-pub fn generate(self: *Boulder, _: GenerationMode, x: i32, y: i32, z: i32, chunk: *main.chunk.ServerChunk, caveMap: CaveMapView, _: CaveBiomeMapView, seed: *u64, _: bool) void {
+pub fn generate(self: *Boulder, _: GenerationMode, x: i32, y: i32, z: i32, chunk: *main.chunk.ServerChunk, caveMap: CaveMapView, _: CaveBiomeMapView, seed: *u64, _: bool, _: bool) void {
```
IntegratedQuantum: I think only SBBs should be allowed, everything else (especially given that it isn't implemented correctly) is a hazard.

src/callbacks/block/server/sapling.zig
```diff
@@ -0,0 +1,64 @@
...
+	const vegetationModels = self.vegetationModels.?;
+
+	const wx = params.chunk.super.pos.wx + params.blockPos.x;
+	const wy = params.chunk.super.pos.wy + params.blockPos.y;
+	const wz = params.chunk.super.pos.wz + params.blockPos.z;
+	const ch = params.chunk;
+
+	// the originals have already deinited long time ago at this point :(
+	const caveMap = main.server.terrain.CaveMap.CaveMapView.init(main.stackAllocator, ch.super.pos, ch.super.width, 32);
+	defer caveMap.deinit(main.stackAllocator);
```
IntegratedQuantum: Don't use generation structures for runtime placement. In the future we will use these to determine that a tree can't spawn in a cave. But since then maybe the player already removed the ceiling to make space for the tree.

src/callbacks/block/server/sapling.zig
```diff
@@ -0,0 +1,64 @@
...
+	const biomeMap = main.server.terrain.CaveBiomeMap.CaveBiomeMapView.init(main.stackAllocator, ch.super.pos, ch.super.width, 32);
+	defer biomeMap.deinit();
+
+	// copied from SimpleStructureGen.generate.
+	var seed = main.random.initSeed3D(main.seed, .{wx, wy, wz});
+	var randomValue = main.random.nextFloat(&seed);
+	for (vegetationModels) |*model| { // TODO: Could probably use an alias table here.
```
IntegratedQuantum: Why even allow more then one? Chances can also be done from inside the SBB.

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -52,11 +52,11 @@ pub fn loadModel(parameters: ZonElement) ?*SbbGen {
...
 }
 
-pub fn generate(self: *SbbGen, _: GenerationMode, x: i32, y: i32, z: i32, chunk: *ServerChunk, _: CaveMapView, _: CaveBiomeMapView, seed: *u64, _: bool) void {
-	placeSbb(self, self.structureRef, Vec3i{x, y, z}, null, self.rotation.getInitialRotation(seed), chunk, seed);
+pub fn generate(self: *SbbGen, _: GenerationMode, x: i32, y: i32, z: i32, chunk: *ServerChunk, _: CaveMapView, _: CaveBiomeMapView, seed: *u64, _: bool, inGeneration: bool) void {
+	placeSbb(self, self.structureRef, Vec3i{x, y, z}, Neighbor.dirUp, self.rotation.getInitialRotation(seed), chunk, seed, inGeneration);
 }
 
-fn placeSbb(self: *SbbGen, structure: *const sbb.StructureBuildingBlock, placementPosition: Vec3i, placementDirection: ?Neighbor, rotation: sbb.Rotation, chunk: *ServerChunk, seed: *u64) void {
+fn placeSbb(self: *SbbGen, structure: *const sbb.StructureBuildingBlock, placementPosition: Vec3i, placementDirection: ?Neighbor, rotation: sbb.Rotation, chunk: *ServerChunk, seed: *u64, inGeneration: bool) void {
```
IntegratedQuantum: The new parameter should be `comptime` to avoid any unnecessary runtime cost.

src/callbacks/block/server/sapling.zig
```diff
@@ -0,0 +1,64 @@
...
+vegetationModels: ?[]SimpleStructureModel,
+
+pub fn init(zon: main.ZonElement) ?*@This() {
+	const result = main.worldArena.create(@This());
+	result.structures = zon.getChild("structures").clone(main.worldArena);
+	result.vegetationModels = null;
+	return result;
+}
+
+fn initAfterBiomesHaveBeenInited(self: *@This()) void {
```
IntegratedQuantum: Please just call init after all biomes have been inited.

Also why does it even need to happen after all biomes were inited? Where is the dependency on the biome?

IntegratedQuantum (changes requested):
There are some flaws with your approach. Instead of unnecessarily trying to bend your code around the terrain generation system, I'd suggest to directly reference the underlying SBB.

src/callbacks/block/server/sapling.zig
tillpp: This was a **Draft** PR, not really meant for review yet  😅 (That's why i haven't really renamed the names yet)

IntegratedQuantum: Ah right, I guess I was too quick to assume after I read your last comment.

tillpp:
One problem i have, is that It feels too laggy if you have too many trees growing at once,
what do you think?

Maybe the way we paste Blueprint has to be optimized.

(Btw does someone know why the check failed)

Wunka:
> (Btw does someone know why the check failed)

You get a recursion of imports through `
pub const SimpleStructures = @import("simple_structures/_list.zig");` in `biomes.zig`
At least that is what I think because the test passes If I remove that and instead do 
`const sbbGen = @import("../../../server/terrain/simple_structures/SbbGen.zig");` inside `sapling.zig` but that is pretty cursed so I hope for you that you find some other way

IntegratedQuantum:
> One problem i have, is that It feels too laggy if you have too many trees growing at once,
what do you think?

Well, if you have a performance problem: Measure it.

If the problem is mesh light sampling then #2405 might help improve it a bit. Other than that, there is a good chance that it's just the light updates.

tillpp:
I think the lagg spikes are soo big, that i dont really need to meassure something:
(each line is a tree that grows)
I think optimizing Blueprint pasting is something we need to work on

grafik

also thanks @Wunka  for the tip

IntegratedQuantum:
> I think the lagg spikes are soo big, that i dont really need to meassure something:

The measuring is to figure out what exactly is responsible for this. And I doubt that it's the blueprint pasting.

tillpp:
> The measuring is to figure out what exactly is responsible for this. And I doubt that it's the blueprint pasting.

I doubt that random.nextFloat() that is called once is going to be the bottleneck, rather the hundreds of Blueprint.paste calls are going to be the cause. Aside from Blueprint.paste and recursive placeSbb calls that call Blueprint.paste, there isn't really something that could cause this otherwise.

messuring  (996 samples) the **Blueprint.paste** that get called during the saplings growth gives me:
**92.46 ± 77.85µs** (for each Blueprint.paste command, that is called)
 
Messuring the **root placeSbb function call** in sapling.zig (55 samples) that recursivly calls placeSbb and then calls all of those blueprint.paste gives:
**6.89 ± 6.85 ms**

and yes, i did deactivate leaf and branch decay to make sure they aren't the cause.

IntegratedQuantum:
Huh, that's actually quite bad, given that we can generate thousands of them without it even showing up in the profiler.
I guess it's probably related to the lock when placing a block, and all the other random stuff that happens in there.

Either way 6 ms server-side lag does not explain the 60 ms client side lag you saw.

tillpp:
> Either way 6 ms server-side lag does not explain the 60 ms client side lag you saw.

probably just multiple trees that grew up at the same time. ( i was at a place with a lot of saplings around me)
Or do you have an idea where the rest could come from?
light updates?

IntegratedQuantum:
I'd bet on light updates or remeshing

tillpp:
nearly forgot that this PR existed

Wbjitscool:
I hope this pr can get merged one day once some of the kinks get straightened out

IntegratedQuantum:
Please rebase and check if the performance issue persists.

tillpp:
Seems fixed, but it might be worth to test this again on a worse CPU, 

grafik
grafik

tillpp:
But it does feel like as if the server would be overwhelmed.
some things don't work, like picking up items after a while for example

assets/cubyz/blocks/leaves/_defaults.zig.zon
```diff
@@ -4,11 +4,13 @@
 		.prevention = .{.log, .branch},
 		.drops = .{
 			.{.chance = 0.01, .items = .{"cubyz:apple"}},
+			.{.chance = 0.001, .items = .{"cubyz:sapling/oak"}},
 		},
 	},
 	.drops = .{
 		.{.items = .{.auto}, .allowedToolTags = .{.cuttable}},
 		.{.chance = 0.01, .items = .{"cubyz:apple"}, .forbiddenToolTags = .{.cuttable}},
+		.{.chance = 0.001, .items = .{"cubyz:sapling/oak"}, .forbiddenToolTags = .{.cuttable}},
```
IntegratedQuantum, line 13: How many leaf blocks are there in a tree?

src/callbacks/block/server/sapling.zig
```diff
@@ -0,0 +1,41 @@
+const std = @import("std");
+
+const main = @import("main");
+const Vec3i = main.vec.Vec3i;
+const sbbGen = @import("../../../server/terrain/simple_structures/SbbGen.zig");
```
IntegratedQuantum: Don't import files like this, access it through main.server.terrain..., if it isn't public, then maybe it should be
Also don't alias stuff under a different name.

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -68,12 +68,20 @@ fn placeSbb(self: *SbbGen, structure: *const sbb.StructureBuildingBlock, placeme
...
-	rotated.blueprint.pasteInGeneration(pastePosition, chunk, self.placeMode);
+	if (duringWorldGeneration) {
+		rotated.blueprint.pasteInGeneration(pastePosition, chunk, self.placeMode);
+	} else {
+		var pos = Vec3i{0, 0, 0};
+		pos[0] = chunk.super.pos.wx + pastePosition[0];
+		pos[1] = chunk.super.pos.wy + pastePosition[1];
+		pos[2] = chunk.super.pos.wz + pastePosition[2];
 
+		rotated.blueprint.paste(pos, .{.noUpdate = true});
```
IntegratedQuantum: I'd argue it should never do block updates during pasting, this has caused many other difficulties when pasting blueprints

src/blueprint.zig
```diff
@@ -147,6 +147,7 @@ pub const Blueprint = struct {
 
 	pub const PasteFlags = struct {
 		preserveVoid: bool = false,
+		noUpdate: bool = false,
```
IntegratedQuantum: unused variable

src/server/world.zig
```diff
@@ -1270,6 +1270,9 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		for (userList) |user| {
 			main.network.protocols.blockUpdate.send(user.conn, &.{.{.pos = .{wx, wy, wz}, .newBlock = newBlock, .blockEntityData = &.{}}});
 		}
+		if (noUpdate)
+			return null;
+
 		// onBreak event
```
IntegratedQuantum, line 1295: The onBreak event should still be triggered, otherwise the player can neutralize negative effects by planting a tree.

src/server/world.zig
```diff
@@ -1270,6 +1270,9 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		for (userList) |user| {
 			main.network.protocols.blockUpdate.send(user.conn, &.{.{.pos = .{wx, wy, wz}, .newBlock = newBlock, .blockEntityData = &.{}}});
 		}
+		if (noUpdate)
```
IntegratedQuantum: This check does not appear to work, when I tried it, there was a lot of decay still, and cactus branches weren't connected sometimes.

src/callbacks/block/server/sapling.zig
```diff
@@ -0,0 +1,41 @@
...
+const SimpleStructureModel = main.server.terrain.biomes.SimpleStructureModel;
+
+structures: main.ZonElement,
+sbb: ?*sbbGen.SbbGen,
+generationMode: SimpleStructureModel.GenerationMode,
```
IntegratedQuantum: unused

src/callbacks/block/server/sapling.zig
```diff
@@ -0,0 +1,41 @@
+const std = @import("std");
+
+const main = @import("main");
+const Vec3i = main.vec.Vec3i;
+const sbbGen = @import("../../../server/terrain/simple_structures/SbbGen.zig");
+const SimpleStructureModel = main.server.terrain.biomes.SimpleStructureModel;
+
+structures: main.ZonElement,
```
IntegratedQuantum: Can you place load this without storing this forever in memory? Maybe callback loading should be deferred until after all other assets are loaded.

IntegratedQuantum (changes requested):

assets/cubyz/blocks/leaves/_defaults.zig.zon
tillpp: probably less then 1000

IntegratedQuantum: It would be nice to have those numbers, I think for all saplings the number of saplings per tree should be consistent.

tillpp: feel free to count them

tillpp: also the amount of leaves varies because of the generation

careeoki:
How close are we to having saplings with growth stages that have different models?

Argmaster:
Nowhere close

tillpp:
requires: https://github.com/PixelGuys/Cubyz/issues/3043

tillpp:
I think my PR broke the formatter:
grafik
(it still keeps printing)

but the game compiles

src/server/world.zig
tillpp: > cactus branches weren't connected sometimes.

That happens if you spawn multiple cactus right next to each other.

tillpp:
good  news: no problem on client side
bad news: server gets overwhelmed if you spam saplings

tillpp:
@careeoki 
> How close are we to having saplings with growth stages that have different models?

it depends.
In  this PR one block and summon a SBB structure.
of course the resulting SBB structure can itself summon a SBB structure.
You can use this to chain SBB together to simulate something growing.
(like sugar cane or something)

but more complex stuff isn't possible with this yet. (like checking if the sugar cane isn't too big already (well technically it is possible, but it wouldn't be nice to implement), or removing a bigger SBB with a smaller SBB afterwards)

Crepestrom:
do you have any clue why the server is dieing?

IntegratedQuantum:
> server gets overwhelmed if you spam saplings

even in release?

tillpp:
> even in release?

its better in release.
But my computer is also unrepresentative of the average gamer. (humble flex)
So I hope someone with a bad PC can test this by doing:

1. select a medium sized region with the selection wand (but z height = 1)
2. /set cubyz:sapling/oak
3. while seeing the saplings grow, try to drop item
4. can you see the items drop and move?

IntegratedQuantum:
There was an issue in chunk saving, it always saved it immediately instead of waiting 5 seconds in case other changes happened.
With bccdbdad290894d01cb1eb76265d20cd6424cb9b it should be more balanced now.

tillpp:
i did merge master and my computer still struggles
nothing grows, no items drop when pressing Q. but when rejoining i can see the dropped items and the trees growing. So maybe it is the network who is overwhelmed: 
grafik

IntegratedQuantum:
yes, this is #3250, it should at least be a little better though.

IntegratedQuantum:
When I was testing this, I also noticed that the roots are replacing all ground blocks, instead of checking for decayable.

tillpp:
> decayable

I assume you meant degradeable?

tillpp:
>  it should at least be a little better though.

yeah a little bit: 

grafik

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -68,12 +68,20 @@ fn placeSbb(self: *SbbGen, structure: *const sbb.StructureBuildingBlock, placeme
...
-	rotated.blueprint.pasteInGeneration(pastePosition, chunk, self.placeMode);
+	if (duringWorldGeneration) {
+		rotated.blueprint.pasteInGeneration(pastePosition, chunk, self.placeMode);
+	} else {
+		var pos = Vec3i{0, 0, 0};
+		pos[0] = chunk.super.pos.wx + pastePosition[0];
+		pos[1] = chunk.super.pos.wy + pastePosition[1];
+		pos[2] = chunk.super.pos.wz + pastePosition[2];
 
+		rotated.blueprint.paste(pos, .{.replaceUndegradable = false});
```
IntegratedQuantum, line 79: Should be configurable via self.placeMode, like when placed in generation.

src/server/terrain/simple_structures/SbbGen.zig
```diff
@@ -16,7 +16,7 @@ const NeverFailingAllocator = main.heap.NeverFailingAllocator;
 pub const id = "cubyz:sbb";
 pub const generationMode = .floor;
 
-const SbbGen = @This();
+pub const SbbGen = @This();
```
IntegratedQuantum, line 19: It makes no sense to make this public

src/server/world.zig
```diff
@@ -1231,7 +1231,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 	}
 
 	/// Returns the actual block on failure
-	pub fn cmpxchgBlock(self: *ServerWorld, wx: i32, wy: i32, wz: i32, oldBlock: ?Block, _newBlock: Block) ?Block {
+	pub fn cmpxchgBlock(self: *ServerWorld, wx: i32, wy: i32, wz: i32, oldBlock: ?Block, _newBlock: Block, comptime noUpdate: bool) ?Block {
```
IntegratedQuantum, line 1234: Please use an enum to improve reasability at the callsite

IntegratedQuantum (changes requested):

Argmaster:
I think I will look into taking over this PR soon, after I'm done with #3446

ikabod-kee:
What if- to prevent the trees from popping into existence- we have them grow when not observed? Quantum trees.

IntegratedQuantum:
Too difficult. The server cannot know what you are observing (let alone what you will observe when you receive the block updates) due to network delay.

PR #2587, opened 2026-02-23
Permission Groups

Wunka:
Follow Up PR for #2530 

This adds the  rest from the `permissionLayer` PR which was split of. For Documentation see #2530


src/server/command/_list.zig
```diff
@@ -10,6 +10,7 @@ pub const time = @import("time.zig");
 pub const tp = @import("tp.zig");
 
 pub const perm = @import("permission/perm.zig");
+pub const group = @import("permission/group.zig");
```
IntegratedQuantum: Should be ordered alphabetically

src/server/command/permission/group.zig
```diff
@@ -0,0 +1,62 @@
+const std = @import("std");
+
+const main = @import("main");
+const User = main.server.User;
+const permission = main.server.permission;
+
+pub const description = "lets you create, delete, join and leave groups";
+pub const usage = "/group <create/delete/join/leave> <groupName>";
+
+const operations = enum {
```
IntegratedQuantum:
```suggestion
const Operation = enum {
```

src/server/command/permission/perm.zig
IntegratedQuantum, line 1: I would suggest to add this to the group command instead, maybe `/group <whitelist/blacklist> <add/remove> <groupName> <path>`

src/server/permission.zig
```diff
@@ -106,6 +106,102 @@ pub const Permissions = struct { // MARK: Permissions
 	}
 };
 
+pub const PermissionGroup = struct { // MARK: PermissionGroup
+	permissions: Permissions,
+	id: u32,
```
IntegratedQuantum, line 118: What's the purpose of the id, when groups are addressable via their name?

src/server/server.zig
```diff
@@ -175,6 +176,11 @@ pub const User = struct { // MARK: User
 		}
 
 		self.permissions.deinit();
+		var permissionGroupsIterator = self.permissionGroups.keyIterator();
+		while (permissionGroupsIterator.next()) |group| {
+			main.globalAllocator.free(group.*);
```
IntegratedQuantum: Why is the group freed here? This looks like an invalid free.

src/server/permission.zig
```diff
@@ -106,6 +106,102 @@ pub const Permissions = struct { // MARK: Permissions
...
+	sync.threadContext.assertCorrectContext(.server);
+	const result = groups.getOrPut(groupsArena.allocator().allocator, name) catch unreachable;
+	if (result.found_existing) return error.AlreadyExists;
+
+	result.key_ptr.* = groupsArena.allocator().dupe(u8, name);
+	result.value_ptr.* = .init(groupsArena.allocator());
+}
+
+pub fn getGroup(name: []const u8) error{GroupNotFound}!*PermissionGroup {
+	return groups.getPtr(name) orelse return error.GroupNotFound;
```
IntegratedQuantum: hashmaps don't have stable pointers

src/server/world.zig
```diff
@@ -639,6 +641,23 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		try files.cubyzDir().writeZon(path, worldData);
 	}
 
+	pub fn loadPermissionGroups(self: *ServerWorld) !void {
+		const path = std.fmt.allocPrint(main.stackAllocator.allocator, "saves/{s}/groups.zig.zon", .{self.path}) catch unreachable;
```
IntegratedQuantum: before we put everything into a single file, I would like to at least see an estimation of how big the file could be on a large server (for comparison the survival playtest despite having a peak of only around 10 concurrent players has seen over 500 distinct accounts).

IntegratedQuantum (changes requested):

src/server/permission.zig
Wunka: Names are good until we remove a group and a player who was in that group, doesn't join the server until a new group is created with the same name.
https://github.com/PixelGuys/Cubyz/pull/2530#pullrequestreview-3740146794

IntegratedQuantum: Then please add a comment for it.

src/server/world.zig
Wunka: Its for me the question what these groups will be used for. If we say only for command permissions I don't see the file being big. But if we say that for #81 there will actually be a group for every block then the file could grow quite big. 

But separating them opens other issues. Should we load all groups at the start of a world? Or should we only load the groups in use? (would prevent commands on them).

src/server/permission.zig
Wunka: I added the assert but thats probaly not enough right? I am sorry I am not that knowledgeable to know what would be the best course of action here.

src/server/command/permission/perm.zig
Wunka: I did `/group <whitelist/blacklist> <groupName> <add/remove/[blank]>  <path>` 
I switch `groupName` and `add/remove/[blank]` for easy looking up if `[blank]` is used. but If you really want the other way around I can look into that

src/server/server.zig
Wunka: Good catch. Fixed!

src/server/world.zig
IntegratedQuantum: For #81 you should be allowed to assign the same group to multiple blocks. But it will still likely end up at 1 group per player on average.

src/server/permission.zig
Wunka: I finally brought myself back to this pr. I have changed the hashmap from `StringHashMapUnmanaged(PermissionGroup)` to `StringHashMapUnmanaged(*PermissionGroup)`.
This should from my understanding fix the problem with the stable pointers

src/server/world.zig
Wunka: So you want seperate files correct? What is then the answer to to the my follow up question about when to load the zon then?

IntegratedQuantum: Since that makes management easier, I think for now we can just load them all into memory.

Wunka: after finnally looking back at this. Where should I then save the `currentId`? in the `world.zig.zon`?

Wunka: I have fixed the problem with a `metadata.zon` file inisde the group folder. 
The groups are now split into seperate files per group.

I do want to update the group commands a bit, but I will do that in a follow up PR (@<playerIndex> syntax, list subcommand to list all groups a player is member of etc)

src/server/command/permission/group.zig
```diff
@@ -0,0 +1,136 @@
+const std = @import("std");
+
+const main = @import("main");
+const User = main.server.User;
+const permission = main.server.permission;
+
+pub const description = "lets you create, delete, join, leave groups and modify there permission paths";
```
Argmaster: Descriptions should start with upper case letter, eg.: https://github.com/PixelGuys/Cubyz/blob/e0dac4995dca8150423949b04f036274bd7dbee2/src/server/command/permission/perm.zig#L9

src/server/command/permission/group.zig
```diff
@@ -0,0 +1,136 @@
...
+			permission.createGroup(groupName) catch {
+				source.sendMessage("#ff0000Group {s}#ff0000 already exists.", .{groupName});
+			};
+		},
+		.delete => {
+			if (!permission.deleteGroup(groupName)) {
+				source.sendMessage("#ff0000Group {s}#ff0000 could not be removed as it already doesn't exist.", .{groupName});
+			}
+		},
+		.join => {
```
Argmaster: Usefulness of joining a group is rather limited, administration would mostly rely on adding/removing others to a group... Is it planned in future or just missed?

Wunka: I can add that. I created these options when there was no `@<playerIndex>` option

src/server/command/permission/group.zig
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if (pathOp) |_op| {
+		switch (_op) {
+			.add => group.permissions.addPermission(listType, path),
+			.remove => {
+				if (!group.permissions.removePermission(listType, path)) {
+					source.sendMessage("#ff0000Permission path {s} is not present inside group {s} permission {s}list", .{path, groupName, @tagName(listType)});
+				}
+			},
+		}
```
Argmaster: Imo all code paths should end with return, to prevent mistakes in future, where behavior is by mistake added to multiple code paths. This makes sense because all code paths are mutually exclusive. Such change would also flatten the structure a little bit by moving else contents directly into function body. Definitely not critical tho.

src/server/command/permission/group.zig
```diff
@@ -0,0 +1,136 @@
...
+		return;
+	}
+
+	if (split.next() != null) {
+		source.sendMessage("#ff0000Too many arguments for command /group. Usage: " ++ usage, .{});
+		return;
+	}
+	if (pathOp) |_op| {
+		switch (_op) {
+			.add => group.permissions.addPermission(listType, path),
```
Argmaster: Please add feedback to all positive paths.
"Permission path added to group X" or alike.

src/server/permission.zig
```diff
@@ -106,6 +106,123 @@ pub const Permissions = struct { // MARK: Permissions
...
+		return self;
+	}
+
+	pub fn deinit(self: *PermissionGroup, allocator: NeverFailingAllocator) void {
+		sync.threadContext.assertCorrectContext(.server);
+		self.permissions.deinit();
+		allocator.destroy(self);
+	}
+
+	pub fn hasPermission(self: *PermissionGroup, permissionPath: []const u8) Permissions.PermissionResult {
```
Argmaster: I think this is excessively descriptive. You have permission.zig PermissionGroup, addPermission and permissionPath. We get it, permissions, we don't need to repeat it 4 times unless there is ambiguity. If there is, we should be consistent, so you should `savePermissionGroup` instead of `saveGroups`.
But there isn't. It can be `Group`, `hasPermission`, `path` and it should be fine.

src/server/permission.zig
```diff
@@ -106,6 +106,123 @@ pub const Permissions = struct { // MARK: Permissions
...
+	var metadatZon: ZonElement = .initObject(main.stackAllocator);
+	defer metadatZon.deinit(main.stackAllocator);
+	metadatZon.put("currentId", currentId);
+	try main.files.cubyzDir().writeZon(metadatPath, metadatZon);
+
+	var it = groups.iterator();
+	while (it.next()) |group| {
+		const path = std.fmt.allocPrint(allocator.allocator, "{s}/{d}.zon", .{groupsPath, group.value_ptr.*.id}) catch unreachable;
+		defer allocator.free(path);
+		var groupZon: ZonElement = .initObject(allocator);
```
Argmaster: Move into PermissionGroup next to fromZon as toZon

src/server/server.zig
```diff
@@ -487,11 +493,55 @@ pub const User = struct { // MARK: User
...
 
+	pub fn addToGroup(self: *User, groupName: []const u8) error{GroupNotFound}!void {
+		sync.threadContext.assertCorrectContext(.server);
+		const group = try permission.getGroup(groupName);
+		self.permissionGroups.put(main.globalAllocator.allocator, main.globalAllocator.dupe(u8, groupName), group) catch unreachable;
+	}
+
+	pub fn removeFromGroup(self: *User, groupName: []const u8) bool {
+		sync.threadContext.assertCorrectContext(.server);
+		const slice = (self.permissionGroups.getKeyPtr(groupName) orelse return false).*;
```
Argmaster: Variable names should carry the logical meaning of their content, not the type they contain. Type system already describe contained type, you can retrieve it at any time using language server. Comptime type parameters are an exception, as zig doesn't have contracts/protocols/interfaces, but that's not what we have here.

After this lengthy explanation, please rename to... Idk, groupNamePtr? Not sure what it contains.

src/server/world.zig
```diff
@@ -645,6 +647,40 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		const metaDataZon: ZonElement = dir.readToZon(main.stackAllocator, "metadata.zon") catch .initObject(main.stackAllocator);
+		defer metaDataZon.deinit(main.stackAllocator);
+
+		permission.init(main.globalAllocator, metaDataZon.get(u32, "currentId", 0));
+
+		var groupDir = try dir.openIterableDir("groups");
+		defer groupDir.close();
+		var iterator = groupDir.iterate();
+		while (try iterator.next()) |file| {
+			if (file.kind == .file and std.mem.endsWith(u8, file.name, ".zon")) {
```
Argmaster: Would be better to make two guard clauses out of that condition.

src/server/world.zig
```diff
@@ -651,6 +653,40 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		permission.init(main.globalAllocator, metaDataZon.get(u32, "currentId", 0));
+
+		var groupDir = try dir.openIterableDir("groups");
+		defer groupDir.close();
+		var iterator = groupDir.iterate();
+		while (try iterator.next(main.io)) |file| {
+			if (file.kind == .file and std.mem.endsWith(u8, file.name, ".zon")) {
+				const zon = try groupDir.readToZon(main.stackAllocator, file.name);
+				defer zon.deinit(main.stackAllocator);
+				if (std.mem.eql(u8, file.name, "metadata.zon")) continue;
```
IntegratedQuantum: This is stored outside the groups subdirectory, so no need to check for it here right?

src/server/world.zig
```diff
@@ -651,6 +653,40 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		try files.cubyzDir().writeZon(path, worldData);
 	}
 
+	pub fn loadPermissionGroups(_: *ServerWorld, dir: main.files.Dir) !void {
```
IntegratedQuantum: I think this function should also be in permission.zig.

src/server/permission.zig
```diff
@@ -106,6 +106,123 @@ pub const Permissions = struct { // MARK: Permissions
...
+	// Example scenario:
+	// - User1 joins Group1
+	// - Group1 is deleted while User1 is offline (so their data isn’t updated)
+	// - A new Group1 is created
+	// - When User1 reconnects, they are incorrectly treated as a member of the new Group1
+	id: u32,
+
+	pub fn init(allocator: NeverFailingAllocator) *PermissionGroup {
+		sync.threadContext.assertCorrectContext(.server);
+		currentId += 1;
```
IntegratedQuantum, line 124: Should immediately save the metadata file.

src/server/permission.zig
```diff
@@ -184,6 +301,29 @@ test "RemoveNonExistentPermission" {
...
 
+test "invalidGroupPermission" {
+	init(main.heap.testingAllocator, 0);
+	defer deinit();
+
+	try createGroup("test");
+	try std.testing.expectError(error.GroupNotFound, getGroup("root"));
+}
+
+test "invalidGroupPermissionEmptyGroups" {
```
IntegratedQuantum: Also test the success case at least once.

src/server/world.zig
```diff
@@ -1058,6 +1096,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		try self.saveWorldConfig();
 
 		try self.saveAllPlayers();
+		try self.savePermissionGroups();
```
IntegratedQuantum: Should be done if and only if the data changed.

IntegratedQuantum (changes requested):
In order to keep the line changes down, could you please split this into 3 PRs:
- one only adds the core groups system and saving/loading in the world
- one adds the changes to the `User`
- one adds the `group` command

src/server/world.zig
Wunka: no. its inside the groups folder.

Wunka: It seems I saved it inside and tried to load it outside.... I don't know how i didn't see that. But I do want it to b inside the folder

Wunka:
Ok I did a few changes, here is a summary:

- for each group and the metadata zon exists now a `bool` which says if we need to save it. That means on saving only changed things are overwritten (means also we can save not only on closing of the world but also in between, but currently not done)
- I noticed with the change to one file per group I never addressed deleting the file, so I introduced a `queue` for saving these removed groups (inspired by the `userdeinitList`) to make this work.
- added more tests for success cases as asked by quantum
- after making the most changes, I removed in commit https://github.com/PixelGuys/Cubyz/pull/2587/changes/6d272bb7d1d7a7aff1e710855287eb9f88b98346 and https://github.com/PixelGuys/Cubyz/pull/2587/changes/e82e54ac0bd9186274fd7c64ce2c96f59fd41036 the command and changes to the user, so that will be done in follow up PRs

src/server/world.zig
```diff
@@ -1061,6 +1073,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
```
IntegratedQuantum: Is force save called frequently enough? I think it may not be.

Is there any reason against immediately saving on change?

IntegratedQuantum (commented):
Also there are some unaddressed comments from the last review

src/server/world.zig
Wunka: I removed it completly

src/server/world.zig
Wunka: I think its just my thinking of not having these "different" logics mixed up. I have added to world update that the groups are saved. But if you really want it I can just save where I currently have `updated = true`

IntegratedQuantum: Yes, please save it directly. There is no need to defer saving it. (We only defer saving when it is relevant for performance, but here it is one small file and it's unlikely that the same group gets modified multiple times an a short timespan)

Wunka: I rewrote the save logic. So I don't have to pass the name of a group each time something is applied to them I have given the group a reference to it but its not owned by the group

src/server/permission.zig
```diff
@@ -106,11 +107,184 @@ pub const Permissions = struct { // MARK: Permissions
...
+		return self;
+	}
+
+	pub fn toZon(self: *Group, allocator: NeverFailingAllocator, zon: *ZonElement) void {
+		sync.threadContext.assertCorrectContext(.server);
+		self.permissions.toZon(allocator, zon);
+	}
+
+	pub fn save(self: *Group, allocator: NeverFailingAllocator) void {
+		if (builtin.is_test) return;
```
IntegratedQuantum, line 162: also needs the assertion

src/server/permission.zig
```diff
@@ -106,11 +107,184 @@ pub const Permissions = struct { // MARK: Permissions
...
+		self.permissions.fromZon(zon);
+		return self;
+	}
+
+	pub fn toZon(self: *Group, allocator: NeverFailingAllocator, zon: *ZonElement) void {
+		sync.threadContext.assertCorrectContext(.server);
+		self.permissions.toZon(allocator, zon);
+	}
+
+	pub fn save(self: *Group, allocator: NeverFailingAllocator) void {
```
IntegratedQuantum: I think all the internal functions should be private.

src/server/permission.zig
```diff
@@ -106,11 +107,184 @@ pub const Permissions = struct { // MARK: Permissions
...
+		var groupZon: ZonElement = .initObject(allocator);
+		defer groupZon.deinit(allocator);
+		groupZon.put("name", self.name);
+		self.toZon(allocator, &groupZon);
+		main.files.cubyzDir().writeZon(path, groupZon) catch |err| {
+			std.log.err("Couldn't save permission group: {s} {t}", .{self.name, err});
+		};
+	}
+
+	pub fn deinit(self: *Group, allocator: NeverFailingAllocator) void {
```
IntegratedQuantum: deinit should be next to init, and also private

src/server/permission.zig
```diff
@@ -106,11 +107,184 @@ pub const Permissions = struct { // MARK: Permissions
...
+			std.log.err("Couldn't parse group file {s}: {s} Skipping.", .{file.name, @errorName(err)});
+			continue;
+		};
+		addGroupFromZon(id, zon);
+	}
+}
+
+fn saveMetaData(allocator: NeverFailingAllocator) !void {
+	if (builtin.is_test) return;
+	const metadatPath = std.fmt.allocPrint(allocator.allocator, "{s}/metadata.zon", .{main.server.world.?.path}) catch unreachable;
```
IntegratedQuantum: Should be placed in the "groups" subfolder

src/server/permission.zig
```diff
@@ -106,11 +107,184 @@ pub const Permissions = struct { // MARK: Permissions
...
+	const result = groups.getOrPut(groupsArena.allocator().allocator, name) catch unreachable;
+	if (result.found_existing) return error.AlreadyExists;
+
+	result.key_ptr.* = groupsArena.allocator().dupe(u8, name);
+	result.value_ptr.* = .init(groupsArena.allocator(), result.key_ptr.*);
+}
+
+pub fn getGroup(name: []const u8) error{GroupNotFound}!*Group {
+	sync.threadContext.assertCorrectContext(.server);
+	return (groups.getPtr(name) orelse return error.GroupNotFound).*;
```
IntegratedQuantum: Why not use `.get()`?
```suggestion
	return groups.get(name) orelse return error.GroupNotFound;
```

src/server/permission.zig
```diff
@@ -106,11 +107,184 @@ pub const Permissions = struct { // MARK: Permissions
...
+}
+
+pub fn deleteGroup(allocator: NeverFailingAllocator, name: []const u8) bool {
+	sync.threadContext.assertCorrectContext(.server);
+	const group = groups.fetchRemove(name) orelse return false;
+
+	const path = std.fmt.allocPrint(allocator.allocator, "{s}/{d}.zon", .{main.server.world.?.path, group.value.id}) catch unreachable;
+	defer allocator.free(path);
+	if (main.files.cubyzDir().hasFile(path)) {
```
IntegratedQuantum: Is this really needed?
Doing this causes two syscalls instead of one, and it also hides the error when the file does not exist (which probably should be an error).

src/server/permission.zig
```diff
@@ -106,11 +107,184 @@ pub const Permissions = struct { // MARK: Permissions
...
+pub fn getGroup(name: []const u8) error{GroupNotFound}!*Group {
+	sync.threadContext.assertCorrectContext(.server);
+	return (groups.getPtr(name) orelse return error.GroupNotFound).*;
+}
+
+pub fn deleteGroup(allocator: NeverFailingAllocator, name: []const u8) bool {
+	sync.threadContext.assertCorrectContext(.server);
+	const group = groups.fetchRemove(name) orelse return false;
+
+	const path = std.fmt.allocPrint(allocator.allocator, "{s}/{d}.zon", .{main.server.world.?.path, group.value.id}) catch unreachable;
```
IntegratedQuantum: Should be stored in the "groups" subfolder

src/server/permission.zig
```diff
@@ -184,7 +358,58 @@ test "RemoveNonExistentPermission" {
...
+}
+
+test "invalidGroupCreation" {
+	init(main.heap.testingAllocator, 0);
+	defer deinit();
+
+	try createGroup("test");
+	try std.testing.expectError(error.AlreadyExists, createGroup("test"));
+}
+
```
IntegratedQuantum, line 410: Please also add a roundtrip test for toZon->deinit->init->fromZon

IntegratedQuantum (changes requested):

Wunka:
All done

src/server/permission.zig
```diff
@@ -106,11 +107,183 @@ pub const Permissions = struct { // MARK: Permissions
...
+		return self;
+	}
+
+	fn deinit(self: *Group, allocator: NeverFailingAllocator) void {
+		sync.threadContext.assertCorrectContext(.server);
+		self.permissions.deinit();
+		allocator.destroy(self);
+	}
+
+	pub fn fromZon(allocator: NeverFailingAllocator, zon: ZonElement, id: u32, name: []const u8) *Group {
```
```suggestion
	fn fromZon(allocator: NeverFailingAllocator, zon: ZonElement, id: u32, name: []const u8) *Group {
```

src/server/permission.zig
```diff
@@ -106,11 +107,183 @@ pub const Permissions = struct { // MARK: Permissions
...
+		self.* = .{
+			.permissions = .init(allocator),
+			.id = id,
+			.name = name,
+		};
+		self.permissions.fromZon(zon);
+		return self;
+	}
+
+	pub fn toZon(self: *Group, allocator: NeverFailingAllocator, zon: *ZonElement) void {
```
```suggestion
	fn toZon(self: *Group, allocator: NeverFailingAllocator, zon: *ZonElement) void {
```

src/server/permission.zig
```diff
@@ -106,11 +107,183 @@ pub const Permissions = struct { // MARK: Permissions
...
+		sync.threadContext.assertCorrectContext(.server);
+		self.permissions.toZon(allocator, zon);
+	}
+
+	pub fn save(self: *Group, allocator: NeverFailingAllocator) void {
```
```suggestion
	fn save(self: *Group, allocator: NeverFailingAllocator) void {
```

IntegratedQuantum (changes requested):

Wunka:
all applied

IntegratedQuantum (approved):

PR #2595, opened 2026-02-24
Make a few subbiomes more common

H41ogen:
Makes forest clearings and plateaus more common because right now they're way too rare

### DIFF
--- a/assets/cubyz/biomes/forest/mixed/oak_birch_clearing.zig.zon
+++ b/assets/cubyz/biomes/forest/mixed/oak_birch_clearing.zig.zon
@@ -75,7 +75,7 @@
 	.parentBiomes = .{
 		.{
 			.id = "cubyz:forest/mixed/oak_birch",
-			.chance = 6,
+			.chance = 2,
 		},
 	},
 }
--- a/assets/cubyz/biomes/forest/mixed/oak_pine_clearing.zig.zon
+++ b/assets/cubyz/biomes/forest/mixed/oak_pine_clearing.zig.zon
@@ -93,7 +93,7 @@
 	.parentBiomes = .{
 		.{
 			.id = "cubyz:forest/mixed/oak_pine",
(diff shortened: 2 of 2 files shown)


assets/cubyz/biomes/plateau/temperate.zig.zon
```diff
@@ -39,19 +39,19 @@
 	.parentBiomes = .{
 		.{
 			.id = "cubyz:grassland",
-			.chance = 0.5,
+			.chance = 1.5,
```
IntegratedQuantum, line 42: The plateaus are deliberately rarer to leave some of the terrain nice and flat. You may enjoy plateaus, but it's important to have enough flat areas that are easier to traverse.

assets/cubyz/biomes/forest/mixed/oak_birch_clearing.zig.zon
```diff
@@ -74,7 +75,7 @@
 	.parentBiomes = .{
 		.{
 			.id = "cubyz:forest/mixed/oak_birch",
-			.chance = 0.25,
+			.chance = 6,
```
IntegratedQuantum: Could you also get a comparison screenshot of how the old/new densities look? This is quite a drastic difference.

H41ogen:
It seems I never pushed the changes where I set it back down to 2

H41ogen:
It's hard to compare with a screenshot since clearings are still pretty rare, and you'll pretty much always either find 1 or none at all. `0.25` was just way over-the-top rare.

careeoki:
If you want more plateaus you can always make a separate uncommon biome that generates a bunch of them.

H41ogen:
maybe. I just feel like they're too pretty (and not crazy enough) to be as rare as they are

H41ogen:
I did some thinking, and I think I'm going to close this PR and make a new one with @careeoki's idea

PR #2596, opened 2026-02-24
Branch rebalance

### LINKED ISSUE #2573: Branches are kind of bad for what they cost, this is also potentially confusing to newcomers
Should we just use the same stats as the corresponding planks?

H41ogen:
Fixes #2573

Branches now give 35% durable bonus, and planks give 25%. Branches are still in their natural state, unprocessed, so by my logic that means more strong. Open to feedback :3

### DIFF
--- a/assets/cubyz/blocks/baobab_planks.zig.zon
+++ b/assets/cubyz/blocks/baobab_planks.zig.zon
@@ -13,7 +13,7 @@
 			.modifiers = .{
 				.{
 					.id = "durable",
-					.strength = 0.4,
+					.strength = 0.25,
 				},
 			},
 		},
--- a/assets/cubyz/blocks/birch_planks.zig.zon
+++ b/assets/cubyz/blocks/birch_planks.zig.zon
@@ -14,7 +14,7 @@
 			.modifiers = .{
 				.{
 					.id = "durable",
-					.strength = 0.4,
+					.strength = 0.25,
 				},
 			},
 		},
--- a/assets/cubyz/blocks/mahogany_planks.zig.zon
+++ b/assets/cubyz/blocks/mahogany_planks.zig.zon
@@ -14,7 +14,7 @@
 			.modifiers = .{
 				.{
 					.id = "durable",
-					.strength = 0.4,
+					.strength = 0.25,
 				},
 			},
 		},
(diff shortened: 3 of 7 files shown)


BoySanic:
Would this close #2573?

H41ogen:
yes, sorry lol

BoySanic:
Pls edit your initial description then so it says 'Fixes #2573' so that if this gets merged, it'll close the issue

H41ogen:
i was doing that...

assets/cubyz/blocks/baobab_planks.zig.zon
```diff
@@ -13,7 +13,7 @@
 			.modifiers = .{
 				.{
 					.id = "durable",
-					.strength = 0.4,
+					.strength = 0.25,
```
IntegratedQuantum, line 16: Please explain why you are nerfing normal planks so much.

assets/cubyz/blocks/birch_branch.zig.zon
```diff
@@ -24,17 +24,14 @@
...
 			.textureRoughness = 0.7,
 			.colors = .{
 				0xffb8aa9e, 0xffd0c2b5, 0xffeedfd0, 0xfffff6ed, 0xffeedfd0,
 			},
 			.modifiers = .{
 				.{
-					.id = "fragile",
-					.strength = 0.10,
+					.id = "durable",
+					.strength = 0.35,
```
IntegratedQuantum, line 34: I would prefer to see a difference only in the other parameters, as proposed in https://github.com/PixelGuys/Cubyz/issues/2573#issuecomment-3938113113

assets/cubyz/blocks/baobab_planks.zig.zon
H41ogen: Just because it makes more sense (to me) for the tool handle material to be branches. If you want I can try what was in #2573 instead

assets/cubyz/blocks/birch_branch.zig.zon
H41ogen: changing only the other parameters wouldn't do much, since the handle material doesn't have much of an effect on the tool stats (unless I'm remembering wrong)

IntegratedQuantum: All parameters still have a small, but noticable impact even in the handle. We could accompy it with a small, maybe 5% difference in the durable parameter if that isn't noticable enough.

H41ogen:
I'm going to start over, since there are merge conflicts and what I'm going to try next is completely different.

PR #2597, opened 2026-02-24
Migrate fences & planks to new naming scheme

H41ogen:


H41ogen:
> Would you mind also doing branches next?

branches have the same problem as logs iirc. if I knew how to work with sbbs I would lol

IntegratedQuantum:
I think that migrations were working for structure models, it's only a problem for logs because the rotation mode changed, and because of the simple tree model.

H41ogen:
got it. I will try and see if any issues arise

PR #2604, opened 2026-02-25
Add relative notation to commands

### LINKED ISSUE #2574: relative notation in /tp
### Description

similar to mc
/tp ~a b c → add a to x, assign b to y, assign c to z
/tp ^a ^b ^c → add a,b,c rotated by facing direction to x,y,z

H41ogen:
Closes #2574

uses ~ to specify that a coordinate in the `/tp` command is relative

also now moves the coordinate parsing logic to a common location (_command.zig)

Example usage:

```
/tp ~ ~ ~10 # 10 blocks up
/tp 0 0 ~ # go to origin, keep you z axis
/tp 1938 7691 100 # go to specific coords
```

### DIFF
--- a/src/server/command/spawn.zig
+++ b/src/server/command/spawn.zig
@@ -13,7 +13,7 @@ pub const usage =
 
 pub fn execute(args: []const u8, source: *User) void {
 	var split = std.mem.splitScalar(u8, args, ' ');
-	if (!std.mem.eql(u8, split.peek().?, "")) {
+	if (split.peek().?.len > 0) {
 		const pos = command.parseCoordinates(&split, source) catch return;
 		if (split.next()) |_| {
 			source.sendMessage("#ff0000Too many arguments for command /spawn", .{});


Wunka:
Did you look into `/particles` for that? you could maybe just use the same system until all command argument parsing is merged together with #1425 
https://github.com/PixelGuys/Cubyz/blob/48ec7a2d76197d0cc2b821693569f62576a61ea3/src/server/command/particles.zig#L40C1-L73C2

H41ogen:
I wasn't aware /particles was even a command

H41ogen:
No more `/tp x y` to tp while keeping your z coordinate, instead now you'd do `/tp x y ~`

I'll probably implement `_command.parseCoordinates` for the other commands in a different PR.

OneAvargeCoder193:
ideally i would like to use the system ********* uses where you just put a ~ before the number, as this lets you specify specific axes to be relative

H41ogen:
this is the same system.

src/server/command/_command.zig
```diff
@@ -46,3 +46,23 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+			source.sendMessage("#ff0000Expected number or \"~\", found \"{s}\"", .{arg});
+			return error.InvalidNumber;
+		}
+		break :ret 0;
+	};
+
+	return if (hasTilde) playerPos + num else num;
+}
+
+pub fn parseCoordinates(x: *f64, y: *f64, z: *f64, split: *std.mem.SplitIterator(u8, .scalar), source: *User) !void {
```
Wunka: why are you using pointers here? 
You could just return a Vec3d.
If you need them sepereate you can just destruct the vector

src/server/command/particles.zig
```diff
@@ -43,9 +45,11 @@ fn parseArguments(source: *User, args: []const u8) anyerror!void {
 	var split = std.mem.splitScalar(u8, std.mem.trimRight(u8, args[0..zonIndex], " "), ' ');
 	const particleId = split.next() orelse return error.TooFewArguments;
 
-	const x = try parsePosition(split.next() orelse return error.TooFewArguments, source.player.pos[0], source);
-	const y = try parsePosition(split.next() orelse return error.TooFewArguments, source.player.pos[1], source);
-	const z = try parsePosition(split.next() orelse return error.TooFewArguments, source.player.pos[2], source);
+	var x: f64 = undefined;
+	var y: f64 = undefined;
+	var z: f64 = undefined;
+	try command.parseCoordinates(&x, &y, &z, &split, source);
```
Wunka: With the the vector return this could be 
```suggestion
	var x: f64 = undefined;
	var y: f64 = undefined;
	var z: f64 = undefined;
	x, y, z = try command.parseCoordinates(&split, source);
```
or

```suggestion
	const x, const y, const z = try command.parseCoordinates(&split, source);
```

H41ogen: woah, i had no idea this was a thing in zig. very cool

Wunka: also I just noticed that /particles at the end just puts them into a vector again so here you can just do 
```suggestion
	const pos = try command.parseCoordinates(&split, source);
```

src/server/command/tp.zig
```diff
@@ -63,35 +65,20 @@ pub fn execute(args: []const u8, source: *User) void {
...
-	if (z == null) {
-		z = source.player.pos[2];
-	}
-	x = std.math.clamp(x.?, -1e9, 1e9); // TODO: Remove after #310 is implemented
-	y = std.math.clamp(y.?, -1e9, 1e9);
-	z = std.math.clamp(z.?, -1e9, 1e9);
-	main.network.protocols.genericUpdate.sendTPCoordinates(source.conn, .{x.?, y.?, z.?});
+	x = std.math.clamp(x, -1e9, 1e9); // TODO: Remove after #310 is implemented
+	y = std.math.clamp(y, -1e9, 1e9);
+	z = std.math.clamp(z, -1e9, 1e9);
```
Wunka: From the documentation clamp seems to work on vectors, so I think you also don't need to destruct the vector here

Wunka: Also, if I am understanding this right, is the clamp not always requiered when working with coordiantes? Could this maybe be a bug inside /particles that its not done there? We should maybe think about doing the clamp inside the `parseCoordiantes `function

H41ogen: I don't know; the clamp was added because teleporting really far out would cause crashes.

Wunka: or better in parseAxis as its not a coordiantes but in general an axis problem

Wunka: > I don't know; the clamp was added because teleporting really far out would cause crashes.

Then I say we just wait with that. Quantum will probably comment here if its better or not.

H41ogen: I'm working on testing if /particle can cause crashes right now, since that'll be what matters

H41ogen: indeed it (any value over 2^31) does, so i will move the clamp

H41ogen:
I swear literally every time I forget to run the formatter it fails

src/server/command/_command.zig
```diff
@@ -46,3 +46,25 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+	};
+
+	return std.math.clamp(if (hasTilde) playerPos + num else num, -1e9, 1e9); // TODO: Remove clamp after #310 is implemented
+}
+
+pub fn parseCoordinates(split: *std.mem.SplitIterator(u8, .scalar), source: *User) !main.vec.Vec3d {
+	return .{
+		try parseAxis(split.next() orelse return error.TooFewArguments, source.player.pos[0], source),
+		try parseAxis(split.next() orelse return error.TooFewArguments, source.player.pos[1], source),
+		try parseAxis(split.next() orelse return error.TooFewArguments, source.player.pos[2], source),
```
Wunka: Also, just a heads up. That you `sendMessage` on one error but not the other is kinda wierd and not desired.
Better is when all `sendMessage` for errors are handeld inside the helper function.

H41ogen: I left the error for the too few arguments up to the caller because the resulting error message depends on the command being used. but I agree it is weird.

src/server/command/_command.zig
```diff
@@ -46,3 +46,25 @@ pub fn execute(msg: []const u8, source: *User) void {
...
 	}
 }
+
+fn parseAxis(arg: []const u8, playerPos: f64, source: *User) !f64 {
+	const hasTilde = if (arg.len == 0) false else arg[0] == '~';
+	const numberSlice = if (hasTilde) arg[1..] else arg;
+	const num: f64 = std.fmt.parseFloat(f64, numberSlice) catch ret: {
+		if (!hasTilde) {
+			source.sendMessage("#ff0000Expected number or \"~\", found \"{s}\"", .{arg});
+			return error.InvalidNumber;
```
Wunka: I have to sadly take back my approve... 
I think `arg > 1` is better here.
Else you would on `~a` just see the `a` as a `0` and not return an error.

Wunka: Edit: Better is how it was before `arg > 1 or arg == 0`

Wunka: no... wait. `!hasTilde or arg > 1` ? I think this should cover all cases

H41ogen: i split it into 2 separate ifs for the error message to be more clear

Argmaster:
Hey, please add examples of use to the pr description.
I am not sure if relative+offset (eg. `~34`) notation was already included, but it should be.

src/server/command/_command.zig
```diff
@@ -46,3 +46,29 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+		}
+		if (arg.len > 1) {
```
IntegratedQuantum: I think it would less confusing if this case was handled before parsing as an `if (hasTilde and numberSlice.len == 0) return playerPos`.

src/server/command/spawn.zig
```diff
@@ -3,49 +3,24 @@ const std = @import("std");
...
-			return;
-		};
-		if (x == null) {
-			x = num;
-		} else if (y == null) {
-			y = num;
-		} else if (z == null) {
-			z = num;
-		} else {
+	if (!std.mem.eql(u8, split.peek().?, "")) {
```
IntegratedQuantum:
```suggestion
	if (split.peek().len != 0) {
```

IntegratedQuantum (changes requested):

src/server/command/spawn.zig
H41ogen: omg i cant believe i did that

PR #2616, opened 2026-02-26
Check if the biome exists before teleporting

### LINKED ISSUE #2606: Attempting to teleport to a biome that doesn't exist throws an error

Ideally, this should just display the message that the biome is invalid, no?

Image

BoySanic:
This prevents a pop-up error when a  user teleports to a non-existent biome. Seemed excessive to give them a pop-up for what should just be user error.

Fixes #2606

### DIFF
--- a/src/server/command/tp.zig
+++ b/src/server/command/tp.zig
@@ -8,60 +8,59 @@ pub const usage = "/tp <x> <y>\n/tp <x> <y> <z>\n/tp <biome>";
 
 pub fn execute(args: []const u8, source: *User) void {
 	if (std.mem.containsAtLeast(u8, args, 1, ":")) {
-		if (main.server.terrain.biomes.getByIdOptional(args)) |biome| {
-			if (biome.isCave) {
-				source.sendMessage("#ff0000Teleport to biome is only available for surface biomes.", .{});
-				return;
-			}
-			const radius = 16384;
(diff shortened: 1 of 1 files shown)


src/server/command/tp.zig
```diff
@@ -8,11 +8,11 @@ pub const usage = "/tp <x> <y>\n/tp <x> <y> <z>\n/tp <biome>";
 
 pub fn execute(args: []const u8, source: *User) void {
 	if (std.mem.containsAtLeast(u8, args, 1, ":")) {
-		const biome = main.server.terrain.biomes.getById(args);
```
IntegratedQuantum, line 11: I think it would be better to have a version that returns an optional.
A similar effort has also been started for the blocks getById, ideally all these functions should return optional and error handling should be done when needed at the callsite.

BoySanic: Done. This should be a little closer to what you wanted, though the PR is larger.

IntegratedQuantum: Please use `const biome = main.server.terrain.biomes.getByIdOptional(args) orelse {...error path...}` to avoid having excessive line changes, and to avoid this random else at the bottom of the block that no one can tell what it even belongs to.

IntegratedQuantum (changes requested):

PR #2617, opened 2026-02-26
fix: add guard to prevent crouching when ui open

### LINKED ISSUE #977: Player can crouch while the inventory is open

tacf:
Simple fix to prevent the constant crouching when typing in chat. Since crouch is global this seemed an "easy" way to do it.

fixes #977


Wilker-uwu:
i thought that was an intended feature

tacf:
I couldn't find much evidence of anyone complaining about it so i'm not suprised if that's the case. Feel free to close the issue not to pollute the project work.

Wunka:
Please add "Fixes #977" into the PR description, so the issue is closed automatically

Issue #2566, opened 2026-02-17
Player whitelist


IntegratedQuantum:
Requires a window to get a public key of the player.

For now a command would be good enough to add people to the whitelist.

Wunka:
what is exactly ment with "window to get a public key of the player"? Something like this?: 

<img width="694" height="610" alt="Image" src="https://github.com/user-attachments/assets/51d624f2-c0a9-4f46-a8e6-b252e9f1e85c" />

IntegratedQuantum:
Yeah, though I think ideally we'd only show one key to avoid confusion (and also show it as text).
After first login the server can then figure out whether it wants to use a different key or not.
Since it only applies for the first login it wouldn't be a big deal to use a wrong key, even if the algorithm got broken.

H41ogen:
I feel like this should be part of 0.2.0

PR #2618, opened 2026-02-26
adds the players public key to the social settings.

Wunka:
This is needed for #2566

### DIFF
--- a/src/gui/windows/social.zig
+++ b/src/gui/windows/social.zig
@@ -7,6 +7,8 @@ const gui = @import("../gui.zig");
 const GuiWindow = gui.GuiWindow;
 const Button = @import("../components/Button.zig");
 const CheckBox = @import("../components/CheckBox.zig");
+const HorizontalList = @import("../components/HorizontalList.zig");
+const Label = @import("../components/Label.zig");
 const VerticalList = @import("../components/VerticalList.zig");
 
 pub var window: GuiWindow = GuiWindow{
@@ -15,6 +17,7 @@ pub var window: GuiWindow = GuiWindow{
 };
 
 const padding: f32 = 8;
+var publicKeysZon: main.ZonElement = undefined;
 
 fn toggleStreamerMode(value: bool) void {
 	main.settings.streamerMode = value;
@@ -31,9 +34,25 @@ fn logout() void {
 	gui.openWindow("authentication/login");
 }
 
+fn copy(zon: *main.ZonElement) void {
+	main.Window.setClipboardString(zon.as(?[]const u8, null) orelse return);
+}
+
 pub fn onOpen() void {
 	const list = VerticalList.init(.{padding, 16 + padding}, 400, 16);
 	list.add(CheckBox.init(.{0, 0}, 316, "Streamer Mode (hides sensitive data)", main.settings.streamerMode, &toggleStreamerMode));
+	publicKeysZon = main.network.authentication.KeyCollection.getPublicKeys(main.globalAllocator);
+	if (publicKeysZon == .object) {
+		list.add(Label.init(.{0, 0}, 200, "Your public keys:", .center));
+		var it = publicKeysZon.object.iterator();
+		while (it.next()) |entry| {
+			const row = HorizontalList.init();
+			const keyLabel = Label.init(.{0, 0}, 128, entry.key_ptr.*, .left);
+			row.add(keyLabel);
+			row.add(Button.initText(.{0, 0}, 70, "Copy", .initWithPtr(copy, entry.value_ptr)));
+			list.add(row);
+		}
+	}
 	if (main.game.world == null) {
 		list.add(Button.initText(.{0, 0}, 128, "Change Name", gui.openWindowCallback("change_name")));
 		list.add(Button.initText(.{0, 0}, 128, "Logout", .init(logout)));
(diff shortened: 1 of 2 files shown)


src/gui/windows/social.zig
```diff
@@ -31,9 +34,25 @@ fn logout() void {
...
 }
 
+fn copy() void {
+	main.Window.setClipboardString(key orelse return);
+}
+
 pub fn onOpen() void {
 	const list = VerticalList.init(.{padding, 16 + padding}, 400, 16);
 	list.add(CheckBox.init(.{0, 0}, 316, "Streamer Mode (hides sensitive data)", main.settings.streamerMode, &toggleStreamerMode));
+	const publicKeysZon = main.network.authentication.KeyCollection.getPublicKeys(main.globalAllocator);
```
IntegratedQuantum: Please create a function that only returns the one key we want. Creating and then parsing a zon makes no sense and just makes this harder to read.

IntegratedQuantum (changes requested):

src/gui/windows/social.zig
Wunka: yeah... sorry about that. Now there is a getPublicKey function.

src/gui/windows/social.zig
```diff
@@ -31,9 +34,19 @@ fn logout() void {
...
 }
 
+fn copy() void {
+	main.Window.setClipboardString(key);
+}
+
 pub fn onOpen() void {
 	const list = VerticalList.init(.{padding, 16 + padding}, 400, 16);
 	list.add(CheckBox.init(.{0, 0}, 316, "Streamer Mode (hides sensitive data)", main.settings.streamerMode, &toggleStreamerMode));
+	key = main.network.authentication.KeyCollection.getPublicKey(main.globalAllocator, .ed25519);
```
IntegratedQuantum: Please use the key type from the launch config.

src/network/authentication.zig
```diff
@@ -95,6 +95,15 @@ pub const KeyCollection = struct { // Provides multiple methods to allow server
 		return result;
 	}
 
+	pub fn getPublicKey(allocator: NeverFailingAllocator, comptime keyType: KeyTypeEnum) []const u8 {
+		const bytes = if (@hasDecl(@TypeOf(@field(Storage, @tagName(keyType)).public_key), "toBytes"))
+			@field(Storage, @tagName(keyType)).public_key.toBytes()
+		else
+			@field(Storage, @tagName(keyType)).public_key.toUncompressedSec1();
+		var base64: [std.base64.standard.Encoder.calcSize(bytes.len)]u8 = undefined;
+		return allocator.dupe(u8, std.base64.standard.Encoder.encode(&base64, &bytes));
```
IntegratedQuantum: Please also prefix this with the key type, as is done for the player data that's saved, so the server knows what key type this is.

src/gui/windows/social.zig
```diff
@@ -31,9 +34,19 @@ fn logout() void {
...
+	const row = HorizontalList.init();
+	list.add(Label.init(.{0, 0}, 128, "Your public key", .left));
+	row.add(Label.init(.{0, 0}, 200, key, .left));
```
IntegratedQuantum: Sorry, for changing my opinion again, but I think this is too much junk on the window. I think a single button that just says `Copy public key` would be better.
Please then also reduce the scope of the key to the copy function.

IntegratedQuantum (changes requested):

src/network/authentication.zig
Wunka: all done. its now [keyType]:[publicKey]

src/network/authentication.zig
```diff
@@ -95,6 +95,21 @@ pub const KeyCollection = struct { // Provides multiple methods to allow server
...
+				const bytes = if (@hasDecl(@TypeOf(@field(Storage, decl.name).public_key), "toBytes"))
+					@field(Storage, decl.name).public_key.toBytes()
+				else
+					@field(Storage, decl.name).public_key.toUncompressedSec1();
+				var base64: [std.base64.standard.Encoder.calcSize(bytes.len)]u8 = undefined;
+				const key = std.base64.standard.Encoder.encode(&base64, &bytes);
+				return std.mem.concat(allocator.allocator, u8, &.{@tagName(keyType), ":", key}) catch unreachable;
+			}
+		}
+		@panic("KeyCollection.Storage does not have the same elements as KeyTypeEnum");
```
Wunka: Here I didn't really know what to do as this should never happen.

IntegratedQuantum: Using a switch with inline else is the way to go, then you get a compile error

Wunka: I am blind... If I just looked 3 lines lower into the sign function where the exact solution is.... it is now a switch

IntegratedQuantum (changes requested):
Please make the buttons wider so it fits in one line
Image

Wunka:
changed all to 150

src/gui/windows/social.zig
```diff
@@ -31,12 +34,19 @@ fn logout() void {
 	gui.openWindow("authentication/login");
 }
 
+fn copy() void {
+	const key = main.network.authentication.KeyCollection.getPublicKey(main.globalAllocator, settings.launchConfig.preferredAuthenticationAlgorithm);
```
IntegratedQuantum: local allocations should use the stackAllocator

IntegratedQuantum (changes requested):

src/gui/windows/social.zig
Wunka: changed!

IntegratedQuantum (approved):

PR #2619, opened 2026-02-26
Override glfw's keynames to distinguish between keys that print the same character

### LINKED ISSUE #2605: Keybind UI: Number row and number pad are not distinguished

Image

on a full-sized keyboard, the label for the many characters in the number row and the number pad have the exact same label despite the game being able to read them as separate keys just fine.

BoySanic:
Fixes #2605

X11, Wayland, etc provide the printable character, not some other name that describes the key. So Keypad 1 and 1 come up as the same name but different scancodes. 

To fix this, I reordered `getName` so our table of names gets used first, and if we don't have the key defined ourselves, then it'll ask GLFW.

Updating GLFW would not have fixed this. The function does the same thing in the latest commit from what I can tell.
(And we're on the latest release of GLFW)

### DIFF
--- a/src/gui/windows/controls.zig
+++ b/src/gui/windows/controls.zig
@@ -96,19 +96,23 @@ fn unbindKey(keyPtr: usize) void {
 }
 
 pub fn onOpen() void {
+	const controlsListWidth: u32 = 256;
+	const keybindButtonWidth: u32 = 160;
+	const unbindButtonWidth: u32 = 64;
+
 	const list = VerticalList.init(.{padding, 16 + padding}, 364, 8);
-	list.add(Button.initText(.{0, 0}, 192, if (editingKeyboard) "Gamepad" else "Keyboard", .init(toggleKeyboard)));
-	list.add(ContinuousSlider.init(.{0, 0}, 256, 0, 5, if (editingKeyboard) main.settings.mouseSensitivity else main.settings.controllerSensitivity, &updateSensitivity, &sensitivityFormatter));
-	list.add(CheckBox.init(.{0, 0}, 256, "Invert mouse Y", main.settings.invertMouseY, &invertMouseYCallback));
-	list.add(CheckBox.init(.{0, 0}, 256, "Toggle sprint", main.KeyBoard.key("sprint").isToggling == .yes, &sprintIsToggleCallback));
+	list.add(Button.initText(.{0, 0}, keybindButtonWidth, if (editingKeyboard) "Gamepad" else "Keyboard", .init(toggleKeyboard)));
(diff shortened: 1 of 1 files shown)


src/graphics/Window.zig
```diff
@@ -439,7 +437,23 @@ pub const Key = struct { // MARK: Key
...
+				c.GLFW_KEY_KP_3 => "Keypad 3",
+				c.GLFW_KEY_KP_4 => "Keypad 4",
+				c.GLFW_KEY_KP_5 => "Keypad 5",
+				c.GLFW_KEY_KP_6 => "Keypad 6",
+				c.GLFW_KEY_KP_7 => "Keypad 7",
+				c.GLFW_KEY_KP_8 => "Keypad 8",
+				c.GLFW_KEY_KP_9 => "Keypad 9",
+				c.GLFW_KEY_KP_DECIMAL => "Keypad .",
+				c.GLFW_KEY_KP_DIVIDE => "Keypad /",
+				c.GLFW_KEY_KP_MULTIPLY => "Keypad *",
```
IntegratedQuantum: `*` does not show up, because it's a formatting character. Also on my keyboard it's a `×` and the `DECIMAL` is a `,`. So I'd suggest to keep the name as text for all the symbols, matching the GLFW string

src/graphics/Window.zig
```diff
@@ -452,6 +466,16 @@ pub const Key = struct { // MARK: Key
 				c.GLFW_KEY_UNKNOWN => "(Unbound)",
 				else => "Unknown Key",
 			};
+			if (std.mem.eql(u8, key_name, "Unknown Key")) {
```
IntegratedQuantum: Just move this logic into the else case of the switch

IntegratedQuantum (changes requested):

src/graphics/Window.zig
BoySanic: So, I changed this to your suggestion. A side effect of doing this is that "Keypad Multiply" is too long for a single line, so it wants to wrap onto a 2nd line.

I'm thinking we either need to expand the button or shorten the name, because 2 lines for a single control when the rest take 1 seems odd.

I also like "Keypad" more than "KP" looking at the controls menu. But that's subjective.

For now I'll look at expanding the button width.

BoySanic: image
192 width works, but it's possible smaller values will work too. It was just the next multiple of 64 so I tried that.

IntegratedQuantum: Yeah I'd prefer it smaller. Maybe 160 could work

BoySanic: 160, indeed, worked.

I also moved the widths to constants so they're not magic numbers. (Though, it's not hard to figure out what they are. Just tedious to change them in a dozen places.)

PR #2620, opened 2026-02-26
More islands

### LINKED ISSUE #2044: Secondary child biome generates outside of intended biome

When a secondary child biome is nested inside of a primary child biome, the secondary child biome spawns outside of the primary child biome, even if the chance for both child biomes is set to 0.
The secondary child will frequently spawn directly inside of the parent of the primary child and will sometimes spawn outside of that parent biome entirely, spilling into neighboring biomes.

[oasis_test.zip](https://github.com/user-attachments/files/22981195/oasis_test.zip)

Attached is a simple addon containing 2 biomes - oasis and oasis_water. Oasis is a child of desert, while oasis_water is a child of oasis. Both biomes have a chance of 0, but oasis_water patches spawn in biomes near deserts, detached from the desert or the oasis biome it is meant to spawn inside.

careeoki:
Now each ocean type gets its own island!

**Temperate**: The original cubyz islands, but now with some sandy beaches
image
**Warm**: Sandy islands with palm trees.
image
**Cold**: Grassy islands with young pine trees.
image

Misc changes to islands:
- Island base biomes now use the song "Sincerely", which previously was only used in island shelves for some reason
- Reduced the radius of island shelves so there would be less weird flat areas below the water
- Increased the height of island base biomes by 1 block to minimize underwater islands a small slightly
- Removed some redundant ground structures which placed slate on top of slate

Included is also a fix to sub-subbiomes appearing without their parent biomes, fixes #2044


TyrrelMeter:
For the cold island maybe make it generate with more trees. (I am just looking at pictures right now, I'll check it out on Cubyz later).

careeoki:
I think it's a fine number of trees for a little island. And the more trees you have, the more often you get trees submerged in the ocean

IntegratedQuantum:
> And the more trees you have, the more often you get trees submerged in the ocean

That's what the shelf is for, to have a base level of other stuff around the core island. If the shelf height and radius are chosen right, then it should prevent underwater trees and grass.

TyrrelMeter:
> I think it's a fine number of trees for a little island. And the more trees you have, the more often you get trees submerged in the ocean

I think that it could at least use something to make it less flat. More trees would help, but rocks and such also work.

careeoki:
> That's what the shelf is for, to have a base level of other stuff around the core island. If the shelf height and radius are chosen right, then it should prevent underwater trees and grass.


It doesn't seem to work in practice though. With the current islands on master I see underwater trees very often.

IntegratedQuantum (changes requested):
Since you are migrating the base island anyways: I think the islands, since they are subbiomes of the oceans, should be in the ocean folder, e.g. `cubyz:ocean/cold/island/base`.

careeoki:
> Since you are migrating the base island anyways: I think the islands, since they are subbiomes of the oceans, should be in the ocean folder, e.g. `cubyz:ocean/cold/island/base`.

Yeah, makes sense, done.

TyrrelMeter:
Have you considered some sort of volcanic island?

careeoki:
> Have you considered some sort of volcanic island?

That's outside the scope of this PR.

TyrrelMeter:
> That's outside the scope of this PR.

Ok. Maybe I'll go ahead and make a volcanic island at some point.

IntegratedQuantum:
I think there is a bug in the subbiome placement, I'll take a look at this.

IntegratedQuantum:
#2856 should fix the issues with trees underwater

careeoki:
I'm not sure how to get it to work right, what value did you use when testing it in #2856 ?

IntegratedQuantum:
The default of 32 should be fine here.

careeoki:
Oh, well, if I leave it at default I get this:
image
I can increase the radius of the shelf, but that's also weird.
image

IntegratedQuantum:
Can you send the test.png for that failed attempt?

careeoki:
From the first image?

IntegratedQuantum:
yes

careeoki:
Here's from the first image. Does test.png show where you are when you load into a world? because I /tp'd away from spawn to get to this island.
test

IntegratedQuantum:
No, it's just whatever it was last generating. But from the looks of it, there is shelf all around the island.

IntegratedQuantum:
Alright, I think the problem is that there are diagonal ocean neighbors. To prevent them you have to set the `.parentEdgeDistance = 45,` (32·√2) on the island biome and make the difference in radius between the island and the shelf is at least `45`, so only 5 larger than what it currently is.

IntegratedQuantum:
I think this situation will also get better with #213, since there are no diagonal neighbors in a hexagonal grid.

careeoki:
Thanks for figuring that out.

careeoki:
Updated this, and since things are less jank now, I went and upgraded the islands a bit. They are a bit larger, less flat, and have small amounts of plants on them.'
image
image
image

By doing this I probably created more problems.

careeoki:
something really weird is going on here- this is a temperate ocean, there's a tiny splotch of warm ocean, and this shelf-less island is on it

image

this one is similar, but is on the edge of a temperate ocean and a warm ocean. again no shelf biome
image

here's my test.png but i cant really tell where to look

test

IntegratedQuantum:
I do also see them on my side, and in the biome map (when increasing the ocean chance), you can also see them appear, and they only appear next to biome borders:
test

IntegratedQuantum:
Alright, there was a bug in the code that rejected subbiomes that were placed outside their parent biome: It still generated sub-subbiomes despite the subbiome being rejected. I pushed a fix for this directly to your PR.

IntegratedQuantum (changes requested):
I think islands would also be prime candidates for amber, so please also add some surface amber to the islands (and maybe also their shelves)

careeoki:
> Alright, there was a bug in the code that rejected subbiomes that were placed outside their parent biome: It still generated sub-subbiomes despite the subbiome being rejected

Sweet, it's not my fault.

> I think islands would also be prime candidates for amber

Oh yeah, true. Done. I didn't want amber on grass so the temperate and cold island bases only have them on the shelves.
It's a bit more common here than on beaches, so there should be at least one patch of amber on an island.

IntegratedQuantum:
> I didn't want amber on grass so the temperate and cold island bases only have them on the shelves.

I'd suggest to double the chance then, I still see plenty of temperate/cold islands without any amber.

careeoki:
> I'd suggest to double the chance then, I still see plenty of temperate/cold islands without any amber.

Done, there should be at least one patch on an island. Also made it slightly more common on warm islands, but not as much since there's more surface area

IntegratedQuantum (approved):

PR #2621, opened 2026-02-27
Allow the player to choose biome search radius in /tp command

H41ogen:
The limit is set at 100 000 000 blocks, since the limit is somewhere between 200 and 300 million, and it's already way further than you should ever have to search.

### DIFF
--- a/src/server/command/tp.zig
+++ b/src/server/command/tp.zig
@@ -4,23 +4,39 @@ const main = @import("main");
 const User = main.server.User;
 
 pub const description = "Teleport to location.";
-pub const usage = "/tp <x> <y>\n/tp <x> <y> <z>\n/tp <biome>";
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
Giving this decision to the client could cause some serious lag on the server. There is a reason why I chose stricter limits.
This should be done after the permission system is ready.

H41ogen:
Got it. I'll leave this open in the meantime. I'm going to bed, lol

Wunka:
> Giving this decision to the client could cause some serious lag on the server. There is a reason why I chose stricter limits. This should be done after the permission system is ready.

Would you say it is now ready enough? 
I think we just need to specify what the `permissionPath` would look like.

IntegratedQuantum:
> Would you say it is now ready enough?

I think it is ready enough when we can specify that by default a player should not be allowed to change it.

PR #2626, opened 2026-02-27
Don't hardcode the path to bash

lukaswrz:
Considering `/bin/bash` isn't guaranteed to exist on some systems, it makes sense to use `/usr/bin/env bash` so that the path is dynamically looked up via `$PATH`.

### DIFF
--- a/debug_linux.sh
+++ b/debug_linux.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 cd "$(dirname "$0")"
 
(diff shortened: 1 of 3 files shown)


lukaswrz:
In general, `/usr/bin/env` is present on more systems than `/bin/bash`. NixOS (and, AFAIK, Guix System) would be an example of a system that doesn't have `/bin/bash`. I'm not sure if Cubyz runs on FreeBSD, but the story would be similar there.

IntegratedQuantum:
freebsd is not currently supported, I'm not aware of anyone who even uses this as a desktop.

PR #2629, opened 2026-02-27
Shaft cave

careeoki:
A simple cave biome that creates large vertical shafts. They aren't as common as most cave biomes, but still... watch your step.

image


IntegratedQuantum:
Out of curiousity: How would it look with rectangular caves?

careeoki:
You can spot the edges pretty easily (with cave noise at 15,) though i guess there's the benefit of more x/y width variation
image

PR #2630, opened 2026-02-27
Ravines

careeoki:
Ravine-shaped cave biome.
Since this uses the rectagular cuboid SDF, they are always axis aligned. And there has to be 2 biomes for x and y directions (I don't want to create square-shaped ravines.)
Both of which are annoying.
image


IntegratedQuantum:
let's just add randomly rotated boxes then.

careeoki:
Yes please

PR #2631, opened 2026-02-27
crying_cave

TyrrelMeter:
A cave with strange beams of glass, almost as though rain is falling.

Cubyz 0 2 0-dev 4_3_2026 5_22_42 PM
Cubyz 0 2 0-dev 4_3_2026 5_22_31 PM
Cubyz 0 2 0-dev 4_3_2026 5_23_01 PM

### DIFF
--- a/assets/cubyz/biomes/cave/rare/crying_cave.zig.zon
+++ b/assets/cubyz/biomes/cave/rare/crying_cave.zig.zon
@@ -2,7 +2,7 @@
 	.isCave = true,
 	.fogDensity = 8,
 	.fogColor = 0x3b5c8c,
-	.chance = 1,
+	.chance = 0.01,
 	.maxHeight = -1000,
 	.minHeight = -7500,
 	.music = "cubyz:cubyz",


H41ogen:
What if they were attached to the ceiling?

TyrrelMeter:
> What if they were attached to the ceiling?

I think that while that would be neat it would limit the size of the cave do to structure size limit if you want them to go down more than 20 blocks, which I do.

Wunka:
@careeoki @ikabod-kee another cave biome, now we can't say that it should be in a cave

TyrrelMeter:
> I think it would be more interesting without the glowcrystals at the end of the raindrops. Otherwise, it feels quite similar to the starlight cave or other glowcrystal based biomes we already have.

That actually gives me a good idea for the biome. Expect an overhaul for it later today.

TyrrelMeter:
@careeoki I updated the biome, if you want to take a look.

TyrrelMeter:
@careeoki I've updated the biome.

careeoki:
> @careeoki I've updated the biome.

Doesn't look like you pushed your changes.

TyrrelMeter:
yes I have

Issue #2166, opened 2025-10-29
make leaf piles more random


H41ogen:
### Description

the leaf piles look really odd because all leaf piles in a patch are exactly the same 
- - -
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/927f94df-7ad0-4c99-b935-cbe6eba218c6" />

PR #2632, opened 2026-02-28
Random block states for flower_patch

H41ogen:
Required for #2166 
- - -
Basically, adds this syntax:
image
which performs a similar function to this:
image
but with each block being randomized individually.
- - -
I'm sure my code will need adjustments; I'm having to relearn Zig/Cubyz so it's not my best work. It might also be useful to be able to weight the different blockstates, but I'm not sure if it'd be worth the added complexity in applications that don't need it. Right now, a similar effect can be done by just repeating certain blockstates in the `.blockStates` parameter.
- - -
Here's an example of the first code snippet in action:
Screenshot_20260228_095155

### DIFF
--- a/assets/cubyz/biomes/autumn/dead_forest.zig.zon
+++ b/assets/cubyz/biomes/autumn/dead_forest.zig.zon
@@ -113,7 +113,7 @@
 		},
 		.{
 			.id = "cubyz:flower_patch",
-			.block = "cubyz:dead_leaf_pile:0",
+			.blocks = .{"cubyz:dead_leaf_pile:0"},
 			.chance = 0.01,
 			.width = 10,
 			.variation = 6,
@@ -122,7 +122,7 @@
 		},
 		.{
 			.id = "cubyz:flower_patch",
-			.block = "cubyz:dead_leaf_pile:1",
+			.blocks = .{"cubyz:dead_leaf_pile:1"},
 			.chance = 0.015,
 			.width = 10,
 			.variation = 6,
@@ -131,7 +131,7 @@
 		},
 		.{
 			.id = "cubyz:flower_patch",
-			.block = "cubyz:dead_leaf_pile:2",
+			.blocks = .{"cubyz:dead_leaf_pile:2"},
 			.chance = 0.02,
 			.width = 8,
 			.variation = 6,
@@ -140,7 +140,7 @@
 		},
 		.{
 			.id = "cubyz:flower_patch",
-			.block = "cubyz:dead_leaf_pile:3",
+			.blocks = .{"cubyz:dead_leaf_pile:3"},
 			.chance = 0.015,
 			.width = 6,
 			.variation = 6,
@@ -149,7 +149,7 @@
 		},
 		.{
 			.id = "cubyz:flower_patch",
-			.block = "cubyz:bolete",
+			.blocks = .{"cubyz:bolete"},
 			.chance = 0.035,
 			.width = 8,
 			.variation = 4,
@@ -158,7 +158,7 @@
 		},
 		.{
 			.id = "cubyz:flower_patch",
-			.block = "cubyz:toadstool",
+			.blocks = .{"cubyz:toadstool"},
 			.chance = 0.03,
 			.width = 8,
 			.variation = 4,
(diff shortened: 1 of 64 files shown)


src/server/terrain/simple_structures/FlowerPatch.zig
```diff
@@ -19,15 +19,39 @@ pub const generationMode = .floor;
...
+						}) catch unreachable;
+						defer main.stackAllocator.free(combinedBlockString);
+						block.* = main.blocks.parseBlock(combinedBlockString);
+					}
+					break :blk output;
+				}
+			}
+			const output = main.worldArena.alloc(main.blocks.Block, 1);
+			output[0] = main.blocks.parseBlock(blockString);
+			break :blk output;
```
H41ogen, line 36: This can't be the best way to do this.

nitinankad:
nit: Commit message can be cleaner

src/server/terrain/simple_structures/FlowerPatch.zig
```diff
@@ -19,15 +19,39 @@ pub const generationMode = .floor;
...
-		.block = main.blocks.parseBlock(parameters.get([]const u8, "block", "")),
+		.blocks = blk: {
+			if (parameters.getChildOrNull("blockStates")) |blockStatesZon| {
+				if (std.mem.containsAtLeast(u8, blockString, 2, ":")) {
+					std.log.err("Block state already specified in block field ({s}). Ignoring blockStates field ({s}).", .{
+						blockString,
+						parameters.get([]const u8, "blockStates", ""),
+					});
+				} else {
+					const output = main.worldArena.alloc(main.blocks.Block, blockStatesZon.toSlice().len);
```
nitinankad: question: What if blockStateZon is of length 0 / empty? I'm unfamiliar with this language and repo so not sure what would happen

H41ogen: it crashes! I will look into fixing that

src/server/terrain/simple_structures/FlowerPatch.zig
```diff
@@ -19,15 +19,39 @@ pub const generationMode = .floor;
...
 density: f32,
 
 pub fn loadModel(parameters: ZonElement) ?*FlowerPatch {
 	const self = main.worldArena.create(FlowerPatch);
+	const blockString = parameters.get([]const u8, "block", "");
 	self.* = .{
-		.block = main.blocks.parseBlock(parameters.get([]const u8, "block", "")),
+		.blocks = blk: {
+			if (parameters.getChildOrNull("blockStates")) |blockStatesZon| {
+				if (std.mem.containsAtLeast(u8, blockString, 2, ":")) {
```
nitinankad: What does the 2 mean in this case? Is it checking to see if there's two `:`?

H41ogen: yes; the only case in which that would happen is if the block state was set in the `.block` field

src/server/terrain/simple_structures/FlowerPatch.zig
```diff
@@ -19,15 +19,39 @@ pub const generationMode = .floor;
...
+							state.as(u16, 0),
+						}) catch unreachable;
+						defer main.stackAllocator.free(combinedBlockString);
+						block.* = main.blocks.parseBlock(combinedBlockString);
+					}
+					break :blk output;
+				}
+			}
+			const output = main.worldArena.alloc(main.blocks.Block, 1);
+			output[0] = main.blocks.parseBlock(blockString);
```
nitinankad: There might be a better way to do this rather than mutating the 1st element in the array but not 100% sure

H41ogen: I'm sure there is; that's why I left my own comment about it lol

nitinankad: Ah missed that lol

IntegratedQuantum:
@ikabod-kee @careeoki Please state your opinion on this.

ikabod-kee:
This is awesome and something that me and carrie have both wanted for a while! Awesome to finally have a PR for it.

PR #2639, opened 2026-02-28
Huge caves

H41ogen:
Adds enormous caves that spawn z<-20k
- - -
Screenshot_20260301_115033


Wbjitscool:
interesting

IntegratedQuantum (changes requested):
z < 15k is currently reserved for #2464

H41ogen:
should I move them higher?

TyrrelMeter:
> should I move them higher?

Yes you should

TyrrelMeter:
Though you could also technically move them lower I think, but check with someone else on that.

H41ogen:
changing this to a draft until the cave layers are more complete

PR #2640, opened 2026-03-01
Add language support (pt. 1)

H41ogen:
Things done in this PR vs things that need to be done:

- [x] item translation code
- [x] tag translation code
- [x] tool translation code
- [ ] UI translation code
- [ ] all items translated
- [ ] all tags translated
- [x] all tools translated
- [ ] all UI translated
- [x] translations in `addon:english.zig.zon` should merge into `cubyz:english.zig.zon` (and this should work for any other language too)
- [ ] decide how to sort translations
- [x] support for non-SVO word order
- [ ] some way to switch languages


assets/cubyz/languages/en_us.zig.zon
```diff
@@ -0,0 +1,14 @@
+.{
+	.assets = .{
+		.biomes = .{}, // for autocomplete
+		.blocks = .{},
+		.items = .{},
+		.languages = .{
+			.@"en-us" = "English (United States)",
+		},
+		.particles = .{}, // for autocomplete
```
OneAvargeCoder193: particles, biomes, and sbbs wont need a name translated to a language because in all cases they are referenced as a base id.

H41ogen: I included these for the future when we have autocomplete in commands, but I can remove them for now if it makes the most sense

OneAvargeCoder193: commands would use the id specifier anyways

assets/cubyz/languages/en_us.zig.zon
```diff
@@ -0,0 +1,15 @@
+.{
+	.assets = .{
+		.biomes = .{}, // for autocomplete
+		.blocks = .{},
+		.items = .{},
+		.languages = .{
+			.@"cubyz:en_us" = "English (United States)",
```
OneAvargeCoder193: i also think that languages should just show their own translated name in the language select menu, rather than showing a translated name for your specific language. this would remove needing to have each language reference every other language

src/lang.zig
```diff
@@ -0,0 +1,92 @@
...
+	item,
+	label,
+	language,
+	modifier,
+	tag,
+	tool,
+	world_preset,
+};
+
+var languages: []ZonMapEntry = &.{};
```
Wunka: can you explain to me why you use `ZonMapEntry `and not `StringHashMapUnmanaged `here?

src/gui/lang.zig
```diff
@@ -0,0 +1,92 @@
...
+				}
+			}
+			unreachable;
+		},
+		.modifier => standardTranslate("ui", "modifiers", string),
+		.tag => standardTranslate("assets", "tags", string),
+		.tool => standardTranslate("assets", "tools", string),
+		.world_preset => standardTranslate("assets", "world_presets", string),
+	};
+}
```
Wunka, line 69: I am not much knowledgable about what optimizations zig does, but is it really good to always query the translation? maybe you should have a wrapper like 
```zig
const Translation = struct {
	inner: ?[]const u8 = null,
	languageId: []const u8 = undefined,

	pub fn translate(...) []const u8 {
		if(self.inner) |localisation| {
			if(std.mem.eql(u8, self.languageId, settings.language) return localisation;
		}
		// query localisation
	}
}
```
I do know this can be a later thing, just wanted to mention it

H41ogen: If I'm understanding correctly, this would be pretty burdensome to the code anywhere stuff needs to be translated. Honestly I don't know enough about zig and low level stuff to know about optimizations.

src/lang.zig
H41ogen: Because that's what assets.zig works with, I assumed it would be the best option here. I don't know what else ZonMapEntry would be used for.

src/gui/lang.zig
Wunka: Yeah same for me. I am just thinking that querying from a zonMap everytime you want to display any text is something. But maybe you can just set this aside and at the end you can benchmark how much performance this costs (I don't think a lot but you never know)

H41ogen: It's probably not worth worrying about unless it causes problems. Maybe you can do all the item translations so we can benchmark it :P

tillpp: > Yeah same for me. I am just thinking that querying from a zonMap everytime you want to display any text is a something. But maybe you can just this aside and at the end you can benchmark how much performance this costs (I don't think a lot but you never know)

If the translation is only queried during the initialization of GuiWindows/items, then I think the performance impact should be small

H41ogen: > If the translation is only queried during the initialization of GuiWindows/items

It should be, so hopefully it's fine

src/lang.zig
Wunka: Maybe I am blind but I only see assets.zig work with `StringHashMapUnmanaged`? But even if. You have an array of the entries. Why not just save the map which contains these entries? The only time I would say this should be done is if you filter the map

H41ogen: assets.zig:

image

This is why I'm using it, it was the path of least resistance at the time. I can look into changing to a `StringHashMapUnmanaged` after I finish with what I'm doing.

Wunka: assets.zig: `pub const ZonHashMap = std.StringHashMapUnmanaged(ZonElement);`

Wunka: Thats why I thought its wierd that you use a slice if the data is already in a good format

src/gui/lang.zig
```diff
@@ -0,0 +1,84 @@
...
+					const zon = entry.value_ptr.*;
+					const translated = zon.get(?[]const u8, "language", null);
+					break :blk translated orelse blk2: {
+						std.log.err("Couldn't find name for language {s}", .{string});
+						break :blk2 string;
+					};
+				}
+			}
+			unreachable;
+		},
```
Wunka, line 61:
```suggestion
		.language => blk: {
			const zon = languagesMap.get(string) oresle unreachable;
			break :blk zon.get(?[]const u8, "language", null) orelse blk2: {
				std.log.err("Couldn't find name for language {s}", .{string});
				break :blk2 string;
			};
		},
```

src/lang.zig
```diff
@@ -0,0 +1,84 @@
...
+
+fn load(languageId: []const u8) !void {
+	var iterator = languagesMap.iterator();
+	while (iterator.next()) |entry| {
+		if (std.mem.eql(u8, entry.key_ptr.*, languageId)) {
+			languageZon = entry.value_ptr.*;
+			return;
+		}
+	}
+	return error.LanguageNotFound;
```
Wunka:
```suggestion
	languageZon = languagesMap.get(languageId) orelse return error.LanguageNotFound
```

Wunka (changes requested):

IntegratedQuantum:
Before you go ahead too far with this: Each point mentioned in the list at the top should be its own PR, to simplify review and to reduce the amount of rewriting you have to do if something needs to be redesigned along the way.

H41ogen:
would it be best then to just create a PR with what I've already done?

IntegratedQuantum:
yeah, probably

H41ogen:
ready for review !!!!

src/gui/windows/workbench.zig
```diff
@@ -49,7 +49,7 @@ var needsUpdate: bool = false;
 fn toggleTool() void {
 	currentToolType += 1;
 	currentToolType %= toolTypes.items.len;
-	toolButton.child.label.updateText(toolTypes.items[currentToolType].id());
```
IntegratedQuantum: This doesn't go deep enough. If the player changes the language, then it should automatically update all windows.

IntegratedQuantum (changes requested):

src/gui/windows/workbench.zig
H41ogen: Unless I'm missing something, i feel like adding code to do this for every window would extremely overkill for something that probably happens once.

H41ogen: It might even be reasonable to only load the language on startup, and ask the user to restart the game after changing their language.

IntegratedQuantum: how do other games approach this?

H41ogen: Most games don't have a window system like cubyz, for example minecraft having a dedicated screen for switching languages. Most other games only have to worry about refreshing the current page after the language is changed.

H41ogen: Additionally, maybe we shouldnt bother with this until the language actually can be changed.

H41ogen: Factorio requires you to restart the game for the new language to take effect.

IntegratedQuantum: Ok, then I guess we can go with a restart.

src/tool/modifiers/restrictions/encased.zig
```diff
@@ -31,5 +31,10 @@ pub fn loadFromZon(allocator: NeverFailingAllocator, zon: ZonElement) *const Enc
 }
 
 pub fn printTooltip(self: *const Encased, outString: *main.List(u8)) void {
-	outString.print("encased in {} .{s}", .{self.amount, self.tag.getName()});
+	outString.print("{s} {} {s} {s}", .{
+		main.lang.translate(.restriction, "encased in"),
+		self.amount,
+		main.lang.translate(.tag, self.tag.getName()),
+		main.lang.translate(.restriction, "items"),
```
IntegratedQuantum: This needs runtime format strings, as the grammar may be different depending on the language.

IntegratedQuantum (changes requested):
Also, please stop making this PR bigger, items can be translated in the next PR.

src/tool/modifiers/restrictions/encased.zig
H41ogen: How do you do that without hard-coding every possible format?

H41ogen:
> Also, please stop making this PR bigger, items can be translated in the next PR.

sorry, I just assumed it wouldn't really have an effect on reviewing

IntegratedQuantum:
the effect it has is that it attracts more conflicts

src/tool/modifiers/restrictions/encased.zig
IntegratedQuantum: With runtime format strings (→not implemented in Zig)

H41ogen: So how am i supposed to use runtime format strings if they don't exist

H41ogen: I might be able to figure it out regardless

H41ogen:
I implemented a system for runtime formatting. Let me know what you think

src/lang.zig
```diff
@@ -78,3 +78,23 @@ pub fn translate(category: Category, string: []const u8) []const u8 {
 		.world_preset => translateHelper("assets", "world_presets", string),
 	};
 }
+
+pub fn format(allocator: main.heap.NeverFailingAllocator, category: Category, string: []const u8, args: []const []const u8) []const u8 {
```
IntegratedQuantum: To avoid all the extra string handling, I'd suggest to extend this with a tagged union like this:
```zig
const FormatArg = union(enum) {
	string: []const u8,
	int: i128,
	tag: Tag,
	item: []const u8,
	...
	pub fn string(_string: []const u8) FormatArg {
		return .{.string = _string};
	}
	pub fn tag(_tag: Tag) FormatArg {
		return .{.tag = _tag};
	}
	...
}
```

Then at the callsite it would look like this:
```zig
main.lang.format(main.stackAllocator, .restriction, "encased", &.{.int(self.amount), .tag(self.tag)});
```

And inside the thing you have a switch-case that directly prints the argument into the list, e.g.
```zig
.tag => |tag| {
	outputList.appendSlice(translate(.tag, tag.getName()));
},
.int => |int| {
	outputList.print("{}", .{int});
},
```

H41ogen:
I don't know how happy I am with the float formatting, having a set amount of digits after the decimal might not be the best.

Wunka:
> I don't know how happy I am with the float formatting, having a set amount of digits after the decimal might not be the best.

You could add a comptime known int to the float so the caller of format can specify how many digits they want.

H41ogen:
Where would the int be stored?

H41ogen:
Never mind, that question was phrased badly and I have an idea.

H41ogen:
I'm hitting a wall with struct fields not being able to be comptime-known no matter what (for some reason). Here's an example of something I tried:

image

IntegratedQuantum:
you have to pass it as a runtime parameter and then enumerate all option using a switch case with inline else
(and to reduce code generation, keep a more limited set of options)

assets/cubyz/languages/en_us.zig.zon
```diff
@@ -0,0 +1,424 @@
+.{
+	.language = "English (United States)",
+	.assets = .{
+		.blocks = .{}, // i think only for autocomplete
+		.items = .{
+			.@"cubyz:amber_block" = "Amber Block",
```
IntegratedQuantum: again, please remove the items for now to avoid conflicts

src/tool/modifiers/restrictions/always.zig
```diff
@@ -14,5 +14,5 @@ pub fn loadFromZon(_: NeverFailingAllocator, _: ZonElement) *const anyopaque {
 }
 
 pub fn printTooltip(_: *const anyopaque, outString: *main.List(u8)) void {
-	outString.appendSlice("always");
+	outString.appendSlice(main.lang.translate(.restriction, "always"));
```
IntegratedQuantum: should be `cubyz:always`, we should specify full ids whereever possible to encourage mods to do the same.

src/assets.zig
```diff
@@ -82,7 +85,8 @@ pub const Assets = struct {
 			.structureBuildingBlocks = self.structureBuildingBlocks.clone(allocator.allocator) catch unreachable,
 			.blueprints = self.blueprints.clone(allocator.allocator) catch unreachable,
 			.particles = self.particles.clone(allocator.allocator) catch unreachable,
-			.worldPresets = .{}, // Not accessible inside the world
+			.languages = .{}, // Not needed inside the world
+			.worldPresets = .{},
```
IntegratedQuantum: please leave the old comment, or make it clear that the comment is meant to apply to both

src/items.zig
```diff
@@ -815,13 +817,16 @@ pub const Tool = struct { // MARK: Tool
...
-			\\Damage: {d:.2}
-			\\Durability: {}/{}
+			\\{d:.2} {s}
+			\\{s}: {d:.2}
+			\\{s}: {}/{}
 		, .{
-			self.type.id(),
+			main.lang.translate(.tool, self.type.id()),
 			self.swingSpeed,
+			main.lang.translate(.other, "swings/s"),
```
IntegratedQuantum: we should agree one some kind of consistent wording of these things. I think we should avoid special symbols or case-sensitivity.

src/gui/lang.zig
IntegratedQuantum, line 1: I think maybe we should put this in gui, since it is only relevant for user-displayed strings.

src/lang.zig
```diff
@@ -0,0 +1,150 @@
...
+
+const ZonMapEntry = std.StringHashMapUnmanaged(ZonElement).Entry;
+
+const Category = enum {
+	block,
+	item,
+	label,
+	language,
+	modifier,
+	other,
```
IntegratedQuantum: I think other is too vague. If we are not specific we'll end up with a lot of `.other` stuff.

src/lang.zig
```diff
@@ -0,0 +1,150 @@
...
+pub var numbersFirst = true;
+
+const languagesMap: *const std.StringHashMapUnmanaged(ZonElement) = main.assets.languages();
+
+var languageZon: ZonElement = undefined;
+
+pub fn init() void {
+	load(main.settings.language) catch {
+		std.log.err("Couldn't find language {s}. Switching to english...", .{main.settings.language});
+		setLanguage("cubyz:en_us") catch unreachable;
```
IntegratedQuantum: should we really follow this standard? I kind of dislike that we have to put a random country in there.
Also I think in most applications these days I see full names instead of these codes.
So I would prefer to just call it `cubyz:english`.

src/lang.zig
```diff
@@ -0,0 +1,150 @@
...
+	languageZon = languagesMap.get(languageId) orelse return error.LanguageNotFound;
+	numbersFirst = languageZon.get(bool, "numbersFirst", true);
+}
+
+fn translateHelper(sectionName: []const u8, catrgoryName: []const u8, string: []const u8) []const u8 {
+	const zon = languageZon.getChild(sectionName).getChild(catrgoryName);
+	const translated = zon.get(?[]const u8, string, null);
+	return translated orelse blk: {
+		// uncomment when english is complete
+		// std.log.err("Couldn't find translation for '{s}'. Searched at '{s}/{s}/{s}/{s}'", .{
```
IntegratedQuantum: Should be a warning, we don't want players to be interrupted when their addon doesn't have names in their language.

src/lang.zig
```diff
@@ -0,0 +1,150 @@
...
+pub fn init() void {
+	load(main.settings.language) catch {
+		std.log.err("Couldn't find language {s}. Switching to english...", .{main.settings.language});
+		setLanguage("cubyz:en_us") catch unreachable;
+	};
+}
+
+pub fn setLanguage(newLanguageId: []const u8) !void {
+	main.settings.language = main.globalAllocator.dupe(u8, newLanguageId);
+	main.settings.save();
```
IntegratedQuantum: setting the setting should be done in the menu that sets it.

src/lang.zig
```diff
@@ -0,0 +1,150 @@
...
+	pub fn fromTag(_tag: main.Tag) FormatArg {
+		return .{.tag = _tag};
+	}
+};
+
+pub fn format(allocator: main.heap.NeverFailingAllocator, category: Category, string: []const u8, args: []const FormatArg) []const u8 {
+	const fmt = translate(category, string);
+	var iterator = std.mem.splitAny(u8, fmt, "{}");
+
+	var outputList = main.List(u8).init(allocator);
```
IntegratedQuantum: since we already use a list internally, how about making a second variant of the function that allows printing to a list directly instead of going through an intermediate allocation.

IntegratedQuantum (changes requested):
I think there is one big flaw with this apporach: addons.
How are addon creators supposed to add english translations without editing the cubyz english file?

H41ogen:
> I think there is one big flaw with this apporach: addons. How are addon creators supposed to add english translations without editing the cubyz english file?

> translations in addon:cubyz/en_us should add to cubyz:en_us (and this should work for any other language too)

I just wasn't going to do it in this PR since you wanted me to just submit what I already had. I could do it now though.

src/lang.zig
H41ogen: There are cases where the country is important to specify. Brazilian portuguese is very different from portugal portuguese, for example. If a language doesn't want to specify a country, they could just leave that part out, for example `de`. I do agree though that it might make sense to write out the language name fully instead of abbreviating it.

IntegratedQuantum: Yes, I agree, but I don't want a country name in the default language, and for english the differences are really minor anyways.

H41ogen: sounds good. I will replace with `cubyz:english`.

IntegratedQuantum:
> I just wasn't going to do it in this PR since you wanted me to just submit what I already had. I could do it now though.

Well, at least you should propose an approach.

src/lang.zig
H41ogen: Should I remove the function for now, or try to create the menu for changing language?

IntegratedQuantum: Remove it, the menu can be added in a future PR.

src/lang.zig
H41ogen: I'm not sure I understand.

IntegratedQuantum:
```diff
-	const translated = main.lang.format(main.stackAllocator, .restriction, "encased", &.{.fromInt(self.amount), .fromTag(self.tag)});
-	defer main.stackAllocator.free(translated);
-	outString.appendSlice(translated);
+	main.lang.formatToList(outString, .restriction, "encased", &.{.fromInt(self.amount), .fromTag(self.tag)});
```

src/tool/modifiers/restrictions/always.zig
H41ogen: Is there any translation ID that you think makes sense to not have `cubyz:` before it?

IntegratedQuantum: Hmm, I don't know. Might as well put it everywhere for consistency.

src/lang.zig
H41ogen: So a buffer, essentially?

H41ogen: ohhh is outString a `List(u8)`?

H41ogen: never mind I see now

src/tool/modifiers/restrictions/always.zig
H41ogen: Doing this for tags would require a decent refactor, since a lot of the time the argument passed into `lang.translate()` is along the lines of `tag.getName()`, so I'll leave them out for now.

src/gui/lang.zig
H41ogen: Done.

H41ogen:
`lang.load()` will now search the other language zons for language IDs in the format `addon:namespace/language`. It might make sense to do this consolidation on init, though. I'll try that.

H41ogen:
That wasn't too bad.

src/tool/modifiers/restrictions/always.zig
IntegratedQuantum: Right, tags are by design addon-independent, so we should leave them as is.

src/gui/windows/main.zig
```diff
@@ -21,10 +21,10 @@ fn exitGame() void {
...
 pub fn onOpen() void {
 	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
-	list.add(Button.initText(.{0, 0}, 128, "Singleplayer", gui.openWindowCallback("save_selection")));
-	list.add(Button.initText(.{0, 0}, 128, "Multiplayer", gui.openWindowCallback("multiplayer")));
-	list.add(Button.initText(.{0, 0}, 128, "Settings", gui.openWindowCallback("settings")));
-	list.add(Button.initText(.{0, 0}, 128, "Touch Grass", .init(exitGame)));
+	list.add(Button.initText(.{0, 0}, 128, main.lang.translate(.label, "cubyz:main/singleplayer"), gui.openWindowCallback("save_selection")));
+	list.add(Button.initText(.{0, 0}, 128, main.lang.translate(.label, "cubyz:main/multiplayer"), gui.openWindowCallback("multiplayer")));
+	list.add(Button.initText(.{0, 0}, 128, main.lang.translate(.label, "cubyz:main/settings"), gui.openWindowCallback("settings")));
+	list.add(Button.initText(.{0, 0}, 128, main.lang.translate(.label, "cubyz:main/touch_grass"), .init(exitGame)));
```
IntegratedQuantum: internally we should use the functional name: exit

src/assets.zig
```diff
@@ -100,13 +104,31 @@ pub const Assets = struct {
...
+			addon.readAllZon(allocator, "languages", false, &self.languages, null);
 			addon.readAllZon(allocator, "world_presets", true, &self.worldPresets, null);
 		}
+
+		self.consolidateLanguages();
+	}
+	fn consolidateLanguages(self: *Assets) void {
+		var iterator = self.languages.iterator();
+		while (iterator.next()) |entry| {
+			if (std.mem.countScalar(u8, entry.key_ptr.*, '/') == 1) {
```
IntegratedQuantum: I think all languages with the same name should be merged. The addon name should be ignored entirely.

IntegratedQuantum: I think ideally merging should be done during loading, and not as an afterthought.

src/items.zig
```diff
@@ -815,13 +817,16 @@ pub const Tool = struct { // MARK: Tool
 		self.tooltip.clearRetainingCapacity();
 		self.tooltip.print(
 			\\{s}
-			\\{d:.2} swings/s
-			\\Damage: {d:.2}
-			\\Durability: {}/{}
+			\\{d:.2} {s}
+			\\{s}: {d:.2}
+			\\{s}: {}/{}
```
IntegratedQuantum: I think these 3 should also use runtime formatting.

src/assets.zig
IntegratedQuantum: Also there is one more problem with this entire system:
Addons can be loaded at runtime during world loading. Thus their translations are missing.

IntegratedQuantum (changes requested):

src/assets.zig
H41ogen: > I think all languages with the same name should be merged. The addon name should be ignored entirely.

which one should have priority in case of conflicts?

H41ogen: > Also there is one more problem with this entire system: Addons can be loaded at runtime during world loading. Thus their translations are missing.

why in the world would someone have world-specific languages?

edit: never mind

H41ogen: > I think ideally merging should be done during loading, and not as an afterthought.

what do you mean?

Argmaster: > > I think all languages with the same name should be merged. The addon name should be ignored entirely.
> 
> which one should have priority in case of conflicts?

We could include priority field to give people tools for fixing stuff.
I don't think there is any specific order we could establish that would work in every case.

src/gui/lang.zig
```diff
@@ -0,0 +1,143 @@
+const std = @import("std");
+
+const main = @import("main");
+const ZonElement = main.ZonElement;
+
+const ZonMapEntry = std.StringHashMapUnmanaged(ZonElement).Entry;
+
+const Category = enum {
+	block,
+	item,
```
Argmaster, line 7: Just a reminder, we were supposed to support to modding. Not something to do now tho.

src/gui/lang.zig
```diff
@@ -0,0 +1,143 @@
...
+pub fn init() void {
+	languageZon = languagesMap.get(main.settings.language) orelse blk: {
+		std.log.err("Couldn't find language {s}. Switching to English...", .{main.settings.language});
+		main.settings.language = "cubyz:english";
+		main.settings.save();
+		break :blk languagesMap.get(main.settings.language).?;
+	};
+}
+
+fn translateHelper(sectionName: []const u8, catrgoryName: []const u8, string: []const u8) []const u8 {
```
Argmaster: What the hell is a helper?

src/assets.zig
H41ogen: I actually think that it makes sense for there to be "sub-languages" like there are currently, since if an addon has translations for a language the player doesn't have the rest of the language for, the player shouldn't be able to switch to that language, since most of it would be an untranslated mess

src/gui/lang.zig
H41ogen: a helper function.

src/assets.zig
Argmaster: > > I think ideally merging should be done during loading, and not as an afterthought.
> 
> what do you mean?

Probably that you should merge every language hash map you load into a single hash map.
Make an empty language map before you start and every time you load a language immediately merge it into that initial hash map.
That immediately excludes priority as an option.

H41ogen: I'm confused on how that helps, I must not be understanding what you mean

src/gui/lang.zig
Argmaster: Yeah well that was a suggestion for you to think about a better name for it.
Name should explain what function does and this one clearly extracts category given category of translations.

src/gui/windows/main.zig
tillpp: I think it would be cool to have a quirky saying for exit for each language in the future

src/assets.zig
Argmaster: > I actually think that it makes sense for there to be "sub-languages" like there are currently, since if an addon has translations for a language the player doesn't have the rest of the language for, the player shouldn't be able to switch to that language, since most of it would be an untranslated mess

Hm. Maybe we should support per addon language selection? Otherwise players may never be able to use anything other than English because add-ons simply won't have translations. 

We cannot disallow players switching to a incomplete language tho, as this can easily crash the game. We should have a default language and merge into it whatever is available, but that again would require priorities and preferences so to not result in a huge mess. We could for example force modders to include English translation (by refusing to load add-ons without English translation) and default to it...

H41ogen: What I think I'm going to try is 

1. update the language map + currently loaded language upon loading the world
2. ignore namespace when trying to merge languages
3. when language is switched, merge new language into current language zon instead of setting it (this "solves" the problem of some addon creators not including languages other than english by missing translations just staying in english)

H41ogen: I also think it could make sense to not unload world-specific language data until the game is closed for simplicity, and also because the likelihood of a wold you open having a world-specific full standalone language addon, you closing said world, and then you being annoyed that the language is still in your language list is zero.

H41ogen:
One advantage this system has is that someone could create e.g. british english, and all they'd have to do is call the zon file `english.zig.zon`, have `isSupplement = true`, and they'd only need to include translations that are different from american to british. same goes for if there's a mexican spanish addon and someone wants spain spanish, for example.

src/assets.zig
```diff
@@ -100,13 +104,39 @@ pub const Assets = struct {
...
 			addon.readAllZon(allocator, "particles", true, &self.particles, null);
+			addon.readAllZon(main.globalArena, "languages", false, &self.languages, null);
 			addon.readAllZon(allocator, "world_presets", true, &self.worldPresets, null);
 		}
 	}
+	fn consolidateLanguages(self: *Assets) void {
+		var iterator = self.languages.iterator();
+		while (iterator.next()) |entry| {
+			if (entry.value_ptr.get(bool, "isSupplement", false)) {
+				defer self.languages.removeByPtr(entry.key_ptr);
```
Wunka: how are you allowed to remove something here? A modification would invalidate the iterator

H41ogen: I'm not exactly sure how it works in this case, I'll have to read more about iterators I guess.

H41ogen: You're correct. I'll work on this when I get home. For some reason I was under the impression that this worked fine.

H41ogen: Actually, the current implementation works like I originally thought, since when entries are removed, they are just tombstoned instead of making the list smaller.

tillpp:
i knew this day would come:
https://github.com/PixelGuys/Cubyz/pull/382/changes/76e007c50813901a20880bbe60dcd7262578bcd1

grafik
grafik

src/assets.zig
Argmaster: Is it guaranteed by API and docs or is it an example of Hyrum's Law (google it)?

src/assets.zig
Argmaster: > I also think it could make sense to not unload world-specific language data until the game is closed for simplicity, and also because the likelihood of a wold you open having a world-specific full standalone language addon, you closing said world, and then you being annoyed that the language is still in your language list is zero.

Eh that is very problematic, because it can hide bugs between reloading world. We would really really want that #1483 it would simplify the problem a lot.

src/assets.zig
H41ogen: It's a case of I tried it with 5 supplement languages and it worked. Zig doesn't have enough users for Hyrum's Law to work for this lol.

Argmaster: The point of mentioning Hyrum's Law here is to discourage relying on implementation details unless they are explicitly guaranteed by the documentation. The observed behavior of hash map implementation is irrelevant here. If you found this behavior documented somewhere, please link it here. Otherwise tombstone insertion is not part of standard hash map contract and should not be relied on.

As an example of implementation detail you can rely on I can bring up order of entries in Python hash maps which always preserve order of insertions and this is explicitly stated in documentation since Python 3.7.

H41ogen: https://ziglang.org/documentation/master/std/#std.HashMap:~:text=any%20modification%20invalidates%20live%20iterators

https://ziglang.org/documentation/master/std/#std.array_hash_map.ArrayHashMap:~:text=Modifying%20the%20hash%20map%20while%20iterating%20is%20allowed

It seems this behavior is only guaranteed for ArrayHashMaps.

Wunka: From the source code of the hashmap it seems to me like deletions are through tombstoning also save. But maybe @H41ogen  you can just try to find an implementation that tries to not do it and then we don't even have to talk about it?

H41ogen: I'm looking into switching to an `ArrayHashMap`, but if it doesn't make sense then yes I plan to rewrite the implementation.

IntegratedQuantum: ArrayHashMap needs to iterate the entire array for each access, this doesn't scale well.

src/assets.zig
H41ogen: Best practice would be for worlds to not contain full languages in their assets(they should only contain supplements as part of other add-ons), but I dont know if it makes sense to enforce that.

src/assets.zig
H41ogen: Fixed.

src/assets.zig
IntegratedQuantum: > which one should have priority in case of conflicts?

Doesn't matter. If two addons define the same translation for e.g. base game assets, then neither has the right to come before the other.
That's why in case of conflicts I'd suggest to print an error or warning and move on.

It is important that we support this, as separate addons may redefine the same languages.

> >  I think ideally merging should be done during loading, and not as an afterthought.

> what do you mean?

I mean that e.g. every time you read an addon file named `languages/english` it should be merged into the existing english language during loading.
This helps because then you won't have any duplicate entries of `xxx:english` in the hashmap, because by design only one entry with key `english` exists.

Your current approach to isExtension also seems to be integrated into this in a possibly confusing way. It seems to extend the first language with matching name it finds, but if there are e.g. two `xxx:german` languages installed, then which one does it extend?
All this mess can be avoided if you only have one `german` language, that is the combinations of all `german` language files of all addons.

> Best practice would be for worlds to not contain full languages in their assets

I think the fact that you are even trying to distinguish full language declarations with the same name is a design flaw. If a player installs two `german` language addons, then I'm sure all the player cares about is that they get german translations, not which addon the translations come from (and if one of the addons has bad translations, then that is a bug in the addon, not something the player should need to fix by defining a priority).

As for language extensions: They should always have a separate name, e.g. the file should be called `british_english` instead of silently breaking your `english` translations the moment you have it in your files.
I think extension should instead explicitly define a fallback language e.g. `.fallbacks = .{"english"}`, then when selecting british_english, the language system would merge british_english and all the fallback languages together recursively. In case of multiple or conflicting fallbacks an error should be printed.

PR #2641, opened 2026-03-01
Multiple player models | avatar command

tillpp:
Let's you pick which character you want to play as:
``/avatar cubyz:snale`` and you are a snale
``/avatar cubyz:snail`` and you are a snail
``/avatar cubyz:pig`` and you are a pig
``/avatar ...``
And with AddOns you can add even more.

I also see this is a stepping stone towards Mobs:
entities know carry their entityType. I.e, not every Entity is a player model anymore.

image


tillpp:
I think this PR relates to the last point on https://github.com/PixelGuys/Cubyz/issues/88 "models", which would make this PR technically **High priority**

ikabod-kee:
I think it'd be nice if you included Carrie's Cubert model (when it's done, of course)

Bloxxel64:
isnt it??

src/entity.zig
```diff
@@ -25,6 +25,62 @@ pub const EntityNetworkData = struct {
 	rot: Vec3f,
 };
 
+pub const ClientEntityType = struct {
```
IntegratedQuantum, line 28: This is not an entity type, this is and entity model. These are different things. Entity types will have different models (e.g. maybe there are 3 variants with different textures and maybe slightly different models for one common entity type)

assets/cubyz/models/entity/pig.obj
```diff
@@ -0,0 +1,312 @@
+# Blender v3.0.1 OBJ File: 'pig.blend'
```
IntegratedQuantum: Please remove the pig, I don't think we want pigs at this point. (and honestly it's terrible in its current state)
Instead I'd suggest to add carrie's moffalo model + texture.

assets/cubyz/models/entity/missing.obj
```diff
@@ -0,0 +1,40 @@
...
+vt 0.125000 0.500000
+vt 0.375000 0.500000
+vt 0.125000 0.750000
+vn 0.0000 1.0000 0.0000
+vn 0.0000 0.0000 1.0000
+vn -1.0000 0.0000 0.0000
+vn 0.0000 -1.0000 0.0000
+vn 1.0000 0.0000 0.0000
+vn 0.0000 0.0000 -1.0000
+usemtl Material
```
IntegratedQuantum, line 33: Please remove the material stuff, we can't parse that anyways plus it's referencing files that don't exist, which may make it harder to open it again in an editor.

assets/cubyz/entity/snail.zig.zon
IntegratedQuantum, line 1: again these are not entities, these describe entity models.

assets/cubyz/entity/missing.zig.zon
```diff
@@ -0,0 +1 @@
+.{.model = "cubyz:entity/missing", .texture = "missing.png", .height = 1}
```
IntegratedQuantum, line 1: Please split it onto multiple lines.

IntegratedQuantum (changes requested):

src/entity.zig
tillpp: yeah already renamed it into EntityModel in my most recent PR draft.
and made it that EntityModel,texture and type are seperate

At this point i rewritten so much of the code, that i ask myself if there is even a point doing this inbetween PR

assets/cubyz/models/entity/pig.obj
tillpp: > onestly it's terrible 

:( ouch, i modelled this myself. 
but also ment as programmer art, i can use to build and test other things around it (like pathfinding system).
I can remove it, but i dont have access to @careeoki s moffalo model + texture

careeoki:
Cubert is ready: [pull](https://github.com/tillpp/Cubzio/pull/1)

assets/cubyz/entity/snail.zig.zon
careeoki, line 1: So the whole reason that snale exists and why this snail was replaced is that the snail is smaller than one block, while the player's hitbox is 2 blocks tall. And then all of the problems that come from that, especially in multiplayer.
I know we've talked about this before, but I don't remember ever coming to a conclusion. Obviously players can do whatever they want when they can create their own models, but do we want to provide tiny player models like this by default?

tillpp: I think that some of the models of the /avatar command can forbidden with the permission system, only allowed for admins.

And speaking of physics, if there are issues with small hitboxes, they have to be fixed, regardless if the player is able to have them as a model, because we are probably going to have one block big animals

careeoki: The problems weren't physics, just "fairness" 
if you're small you can hide easier, are harder to hit, etcetera

assets/cubyz/models/entity/pig.obj
careeoki: Well, it's not as bad as the real cubyz pig:
image

I can make a PR with the moffalo assets, though I'm not totally happy with how that model looks anymore.

assets/cubyz/entity/snail.zig.zon
tillpp: ah ok, but for creative mode or server owner i think it's ok?

assets/cubyz/models/entity/pig.obj
tillpp: > I can make a PR with the moffalo assets, though I'm not totally happy with how that model looks anymore.

that would be wonderful, so i have something to test pathfinding and other mob related stuff with.
And we can replace it in the future. (I think you moffalos look good btw)

careeoki: Alright, done: [pull](https://github.com/tillpp/Cubzio/pull/2)

masterminer176:
Just asking for clarification/curiosity, where exactly does a player's avatar get saved? Is it per world? And furthermore how does this tie into addons?
I figured I'd ask here instead of discord to keep discussions focused.

tillpp:
> Just asking for clarification/curiosity, where exactly does a player's avatar get saved? Is it per world? 

yes per world. Maybe in the distant future when custom character exist, they will be per client. But for now at least they are per world.

> And furthermore how does this tie into addons?

Addons can create their own entityModels, and you can then select them. (if you got the right permission)

> I figured I'd ask here instead of discord to keep discussions focused.

This PR is actually a bit out of date.
I am working on the continuation https://github.com/PixelGuys/Cubyz/pull/2652, where i have basicly rewritten most of the code of this PR. (So there isnt really a point in this one, if the code is completly rewritten in the next PR)
If you have more question, ~~i recommend asking them there.~~. Actually, here is fine too

masterminer176:
> yes per world. Maybe in the distant future when custom character exist, they will be per client. But for now at least they are per world.

Perhaps this is a bit selfish of me to say but I do think I prefer it that way. Makes it easier to have a "Thematically" fitting character for your world(s) without having to switch out each time. Also I imagine that makes it easier to work with World-Based vs Global addons (For example, you set your client avatar to one that only exists in a specific world/server. Solveable with a fallback, sure, but an edge case nonetheless)

tillpp:
REPLACED BY: https://github.com/PixelGuys/Cubyz/pull/2865

PR #2645, opened 2026-03-01
Make the music slider be linear

### LINKED ISSUE #2644: Music volume slider doesn't feel linear

The music volume when the slider is at around 80% feels like 50% of the volume, and when the slider is at 50% it is almost inaudible. Ideally a 50% volume slider should sound like 50%

OneAvargeCoder193:
A simple change that makes it so that halfway across the music slider actually sounds like 50% volume, rather than having the slider be the amount of decibels, which is non-linear.
fixes #2644

### DIFF
--- a/src/gui/windows/sound.zig
+++ b/src/gui/windows/sound.zig
@@ -18,23 +18,12 @@ pub var window = GuiWindow{
 };
 
 fn musicCallback(newValue: f32) void {
-	settings.musicVolume = deziBelToLinear(newValue);
+	settings.musicVolume = newValue;
 	settings.save();
 }
 
-fn deziBelToLinear(x: f32) f32 {
-	if (x < -59.95) return 0;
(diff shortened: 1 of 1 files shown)


Crepestrom:
human hearing is also non linear

OneAvargeCoder193:
the problem is that it doesnt feel linear, when the slider is at 50% it should be half as audible, but it ends up being barely audible for me

tillpp:
I agree with @OneAvargeCoder193 , the music slider feels very odd, either way too silent or way too loud, you really need to aim at the right pixel on screen to make it sound right.

IntegratedQuantum:
Well maybe the range is too big then. But at the end of the day we should allow someone to blast their speakers to maximum and then configure Cubyz to run at a reasonably quite volume using the slider. This is not possible with a linear slider.

PR #2646, opened 2026-03-02
More modifier types

Crepestrom:
Adds two new tool modifier conditions

ConnectedTo and ConductedWith

they function similariliy but here is how they work;
 - ConnectedTo recursivly checks orthogonally adjacent slots in the workbench for a tag and counts up how many of that tag is seen
 - ConductedWith recusivly checks orthoganal slots with a Conductivetag but also checks for a different SourceTag and counts up the ammount of SourceTags are adjacent to connected ConductiveTags

Test Addon
[strange.zip](https://github.com/user-attachments/files/25676365/strange.zip)

I also made sure to eliminate double counting of the same slot for both methods

### DIFF
--- a/src/tool/modifiers/restrictions/conductedWith.zig
+++ b/src/tool/modifiers/restrictions/conductedWith.zig
@@ -79,7 +79,7 @@ pub fn loadFromZon(allocator: NeverFailingAllocator, zon: ZonElement) *const Enc
 	const result = allocator.create(Encased);
 	result.* = .{
 		.sourceTag = main.Tag.find(zon.get([]const u8, "sourceTag", "not specified")),
-		.conductorTag = main.Tag.find(zon.get([]const u8, "conductorTag", "not specified")),
+		.conductorTag = main.Tag.find(zon.get([]const u8, "conductorTag", .sourceTag)),
 		.amount = zon.get(usize, "amount", 8),
 	};
 	return result;


Crepestrom:
no
the modifiers are different even if similar
one requiring one tag while the other needing two tags
we can just keep 2 simple modifiers instead of one complex one that can do both
also when it prints out to the player its more intuitive which one you are using
but if we combine them it can make it harder to understand
thanks for trying to make it simplier

IntegratedQuantum:
Oh right, sorry I just glanced over it and misunderstood what connectedTo is actually doing. As far as I understand it checks if the item is directly connected to a specific tag and then counts the size of the area of connected materials.
If that's the case, then it is actually equivalent to just your conductedWith with source and conductor being the same.
In that case I'd say it's even more desirable to combine them, since the player then has to only figure out the meaning of one restriction.

Crepestrom:
Ok i removed the connectedTo

src/tool/modifiers/restrictions/conductedWith.zig
```diff
@@ -0,0 +1,90 @@
...
+	return @as(usize, @intCast((arrayIndexX*checkLength + arrayIndexY)));
+}
+
+pub fn satisfied(self: *const Encased, tool: *const Tool, x: i32, y: i32) bool {
+	var count: usize = 0;
+	count = 0;
+	// init array to mark already searched blocks.
+	const checkRange = 5;
+	const checkLength = checkRange*2 + 1;
+	var checked: [checkLength*checkLength]bool = undefined;
```
IntegratedQuantum: Why not use a 2D array?
Also if you hardcode the size then please use the `Tool.craftingGridSize`, or ideally make a new constant for the side length.

src/tool/modifiers/restrictions/_list.zig
```diff
@@ -3,3 +3,4 @@ pub const @"and" = @import("and.zig");
 pub const encased = @import("encased.zig");
 pub const not = @import("not.zig");
 pub const @"or" = @import("or.zig");
+pub const conductedWith = @import("conductedWith.zig");
```
IntegratedQuantum, line 6: I'm a bit unhappy with the name. I think that `connected_to` would better describe both use cases, either you are connected to it, or you are connected to it through a conductor.

src/tool/modifiers/restrictions/conductedWith.zig
```diff
@@ -0,0 +1,90 @@
...
+	for (0..checkLength*checkLength) |i| {
+		checked[i] = false;
+	}
```
IntegratedQuantum, line 34: I would suggest to use a grid that is the same size as the tool grid, this reduces the size and avoids any need for different offsets.
```suggestion
	const checkRange = main.items.Tool.craftingGridSideLength;
	var checked: [checkLength][checkLength]bool = @splat(@splat(false));
```

src/tool/modifiers/restrictions/conductedWith.zig
```diff
@@ -0,0 +1,90 @@
...
+
+	// queue for breadth-first search
+	var queue = main.utils.CircularBufferQueue(Vec2i).init(main.stackAllocator, 32);
+	defer queue.deinit();
+
+	queue.pushBack(Vec2i{1, 0});
+	queue.pushBack(Vec2i{-1, 0});
+	queue.pushBack(Vec2i{0, 1});
+	queue.pushBack(Vec2i{0, -1});
```
IntegratedQuantum, line 43: see comment above
```suggestion
	queue.pushBack(Vec2i{x + 1, y});
	queue.pushBack(Vec2i{x - 1, y});
	queue.pushBack(Vec2i{x, y + 1});
	queue.pushBack(Vec2i{x, y - 1});
```

src/tool/modifiers/restrictions/conductedWith.zig
```diff
@@ -0,0 +1,90 @@
...
+
+			// mark as checked
+			if (checked[getIndexInCheckArray(value, checkRange)]) continue;
+			count += 1;
+			checked[getIndexInCheckArray(value, checkRange)] = true;
+			if ((tool.getItemAt(x + value[0], y + value[1]) orelse continue).hasTag(self.conductorTag)) {
+
+				// mark as checked
+				if (checked[getIndexInCheckArray(value, checkRange)]) continue;
+				checked[getIndexInCheckArray(value, checkRange)] = true;
```
IntegratedQuantum, line 57: see comment above
```suggestion
				if (checked[@intCast(value[0])][@intCast(value[1])]) continue;
				checked[@intCast(value[0])][@intCast(value[1])] = true;
```

src/tool/modifiers/restrictions/conductedWith.zig
```diff
@@ -0,0 +1,90 @@
...
+	checked[getIndexInCheckArray(Vec2i{0, 0}, checkRange)] = true;
+
+	while (queue.popFront()) |value| {
+		if ((tool.getItemAt(x + value[0], y + value[1]) orelse continue).hasTag(self.sourceTag)) {
+
+			// mark as checked
+			if (checked[getIndexInCheckArray(value, checkRange)]) continue;
+			count += 1;
+			checked[getIndexInCheckArray(value, checkRange)] = true;
+			if ((tool.getItemAt(x + value[0], y + value[1]) orelse continue).hasTag(self.conductorTag)) {
```
IntegratedQuantum, line 53: x + is no longer needed (see comment above)

src/tool/modifiers/restrictions/conductedWith.zig
```diff
@@ -0,0 +1,90 @@
...
+			queue.pushBack(Vec2i{0 + value[0], -1 + value[1]});
+		}
+	}
+	return count >= self.amount;
+}
+
+pub fn loadFromZon(allocator: NeverFailingAllocator, zon: ZonElement) *const Encased {
+	const result = allocator.create(Encased);
+	result.* = .{
+		.sourceTag = main.Tag.find(zon.get([]const u8, "sourceTag", "not specified")),
```
IntegratedQuantum, line 81: Please use `?[]const u8` and do proper error handling on null.

src/tool/modifiers/restrictions/conductedWith.zig
```diff
@@ -0,0 +1,90 @@
...
+		.conductorTag = main.Tag.find(zon.get([]const u8, "conductorTag", "not specified")),
```
IntegratedQuantum: Should default to the source tag if not specified.

Crepestrom:
being moved to a new pr

PR #2648, opened 2026-03-02
Swap the order of iron and silver in progression

### LINKED ISSUE #2636: Swap the order of iron and silver in progression
This makes copper more useful in the early game (it's too easy to get iron), and makes silver accessible earlier, its preciousness is useful even before finding diamonds, mostly for mining stone, but also to encase amber which can be found on beaches.

H41ogen:
Closes #2636 

I spent a while on this before realizing I was overthinking it and went with simply swapping the level at which the two generate, as well as their block health/resistance values.

### DIFF
--- a/assets/cubyz/blocks/chain/iron.zig.zon
+++ b/assets/cubyz/blocks/chain/iron.zig.zon
@@ -1,7 +1,7 @@
 .{
 	.tags = .{.mineable},
-	.blockHealth = 15,
-	.blockResistance = 1,
+	.blockHealth = 20,
+	.blockResistance = 2,
 	.drops = .{
 		.{.items = .{.auto}},
 	},


careeoki:
You'll also have to swap the resistance/health of the other iron and silver things (blocks, bars)

H41ogen:
ah yeah, I don't know how i forgot that

TyrrelMeter:
I think it might also be a good idea to make iron rarer(at least at the point where it first generates) to encourage players to go deeper for even larger and more common veins.

H41ogen:
i agree, but we don't have a system to do that yet

assets/cubyz/blocks/iron_ore.zig.zon
```diff
@@ -1,11 +1,11 @@
...
 	.tags = .{.mineable},
-	.blockHealth = 35,
-	.blockResistance = 5,
+	.blockHealth = 60,
+	.blockResistance = 7,
 	.ore = .{
 		.veins = 4,
 		.size = 20,
-		.height = -250,
+		.height = -500,
```
IntegratedQuantum: I'd prefer to put it a little lower even. Maybe -600

H41ogen:
> Just to make sure: Did you test if you can break iron with a silver+copper pickaxe?

image

Iron ore's resistance is 7.0

IntegratedQuantum (changes requested):
there is also the iron chain

H41ogen:
done

PR #2649, opened 2026-03-03
Add word deletion in text inputs with control modifier

H41ogen:
This was a very simple change, let me know if there was a reason it wasn't implemented.


IntegratedQuantum:
> let me know if there was a reason it wasn't implemented.

Yes, I never knew this shortcut exists

PR #2650, opened 2026-03-03
Even more modifiers

Crepestrom:
Adds
-onDiagonal
diagonal-visual
-onOrthogonal
Orthoganal-visual
-Encased Rework
Encased-visual
-At range
atRange-visual
(all at range 2)

All can use a range so that it is even more customizable
like counting for each .metal within a 5x5 area instead of a 3x3

Test Addon:
[strange.zip](https://github.com/user-attachments/files/25708342/strange.zip)


Crepestrom:
Finished Formatting

PR #2652, opened 2026-03-03
EntityComponentSystem | Summon command

tillpp:
continuation of https://github.com/PixelGuys/Cubyz/pull/2641
 **This is just a Draft, don't review yet**!
 **Once i am done, i am going to split this PR into many smaller reviewable PRs**

credit for the cubert and moffalo model: @careeoki 

TODO List:
- [ ] src/server/EntitySystem.zig should use a freed List, to recycle old entity IDs. remove inUse
- [ ] make Entities be stored when chunk unloads, if they are NPC (player entities already get stored in the Player file)
- [x] Have a EntityModel pallete
- [ ] SparseSet to  store entities
- [x] Give everyone permission to /avatar
- [ ] view entityModel in player Inventory
- [ ] change entityModel
- [ ] upload custom texture

Long term things:
- [ ] display amour around entityModels
- [ ] display amour around Pathfinding
- [ ] display amour around enemyBehaviour

**requested to be removed by quantum**
- [ ] ~~Items should use the EntitySystem instead of ItemDrops~~

**moved to issue https://github.com/PixelGuys/Cubyz/issues/2670 in order to be finished before 0.2.0:**
- [ ] ~~Only give Information about Entities in a range around that player.~~
- [ ] ~~Keep some entity information hidden, depending on who and where the audience is.~~
- [ ] ~~Only send Entitites that moved.~~

**finished:**
- [x] Split Components from system
- [x] Let servers/world define themself what entityModel players are allowed, pick one at random when joining
- [x] Custom protocol for component addition/remove and edit
- [x] component Data in storing and transmit should be BInary
- [x] Bug: components aren't stored in Player files, if you exit the world.

This PR adds a ``/summon <entityType> ?<name>`` command that let's you spawn an Entity
summon_1


IntegratedQuantum:
> Items should use the EntitySystem instead of ItemDrops

No, items should remain in a separate system. This is more performant, as they only have the data they really need.

PR #2653, opened 2026-03-03
.Hot and .Cold block tags

Crepestrom:
adds .hot and .cold tags to the respective blocks

### DIFF
--- a/assets/cubyz/blocks/frost.zig.zon
+++ b/assets/cubyz/blocks/frost.zig.zon
@@ -1,5 +1,5 @@
 .{
-	.tags = .{.mineable},
+	.tags = .{.mineable, .cold},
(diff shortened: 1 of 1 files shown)


TyrrelMeter:
What would these tags be for?

IntegratedQuantum:
yeah, I would also like to know the reasons for this.

Crepestrom:
ah for certain material modifiers
it would be nice to have these in the game as some items that like cold materials or hate hot materials could be interesting

TyrrelMeter:
So something like if encased by .cold or .cold encased by precious or some other modifier it gives a buff or debuff of some sort?

IntegratedQuantum (approved):

PR #2654, opened 2026-03-03
.Natural block tag

Crepestrom:
added to mushrooms, logs, flowers, and grass


delete
Wunka, line 1: Why does this file exist? Please delete delete

Crepestrom: my bad i was using the github website and that was the only way for me to make a branch
i know how to do it properly now

Issue #2615, opened 2026-02-26
Player needs a formatting-stripped identifier


BoySanic:
#2610 brought up stripping formatting for logs, but there are many uses for being able to target a player by the formatting-stripped name.

Maybe to execute a command as a player (#2446), or for #413, or for #2610.

I propose this should be a field in the User struct in server.zig. Use the formatted name for displaying it in-game, but use the formatting-stripped version of the name for commands and logging.

IntegratedQuantum:
As people can have the same name, a name, whether stripped or unstripped is not sufficient for usage in commands. I'd suggest that commands use the player index instead, and have autocomplete figure out which index to use.

BoySanic:
That's a good point, especially since we allow many characters that aren't able to be typed by keyboard.

Wunka:
Question for this. How would a command user then know what the right index is? Should names in game / chat be prefixed with something like [index]?

H41ogen:
maybe it could show their index next to their name in the tablist (which we don't have yet)

Wunka:
> maybe it could show their index next to their name in the tablist

Problem I see there is what if you see two people with the name "GenericName" and one is doing something which requires you to ban him. 
How would you know through the tablist which one it is? 
That's why I think the index would be needed everywhere, even if through autocomplete you don't have to always type it out

IntegratedQuantum:
We could have a toggle that renders it in-game behind their name

Wunka:
I will look that I implement this toggle and add the option of the index to the /perm command as soon as possible, as currently the permission layer is only a hinderness without being able to change others permissions

PR #2660, opened 2026-03-04
Display User with id when setting enabled

Wunka:
Related to #2615

This PR adds a setting to the social settings to enable showing the player id when the name is displayed.
And of course the actual showing of the id.

For the ingame showing I am uncertain if I can just use the entity id. In the User struct there is a todo to use the entity id. So is it the same or not? If not should the displaying of the id above the player be implemented later or do I need to do shenanigans to get it?

There is also the question how exactly now we want it formated
Currently:
image


src/server/server.zig
```diff
@@ -484,6 +484,14 @@ pub const User = struct { // MARK: User
 			.no, .neutral => false,
 		};
 	}
+
+	pub fn format(user: User, writer: *std.Io.Writer) std.Io.Writer.Error!void {
+		if (main.settings.showIdWithName) {
+			try writer.print("{s} [{d}]", .{user.name, user.id});
```
IntegratedQuantum: I think we should use the same character here that is used when entering it in commands
```suggestion
			try writer.print("{s}@{d}", .{user.name, user.id});
```

IntegratedQuantum (changes requested):
There is a problem here: All of this is done server side, but in my opinion this should be a clientside option or keybind.

For now, since chat is just text (which will at some point need to change), that means the only thing we can change is the player nametag.

Wunka:
I deleted all printing of the id when it goes to all users. But let it stay for logs (~~maybe add the setting to the `launchconfig` for headless servers?~~ people can just edit the settings inside cubyzDir) and with #2673 the entity id is also the player id. So I made it that when enabled the names also have the id

src/client/Entity.zig
```diff
@@ -71,3 +71,11 @@ pub fn update(self: *@This(), time: i16, lastTime: i16) void {
 	self.rot[1] = @floatCast(self.interpolatedValues.outPos[4]);
 	self.rot[2] = @floatCast(self.interpolatedValues.outPos[5]);
 }
+
+pub fn format(self: *const @This(), writer: *std.Io.Writer) std.Io.Writer.Error!void {
+	if (main.settings.showIdWithName) {
+		try writer.print("{s}@{d}", .{self.name, self.id});
```
IntegratedQuantum: This is not the player id, this is  the entity id. These are different things, one is static, the other one is dynamic and may change on rejoin.

IntegratedQuantum (changes requested):

src/client/Entity.zig
Wunka: they are not different anymore from my testing and understanding. inside `connectInternal` https://github.com/PixelGuys/Cubyz/blob/1ee3d978ce5af60b383dfe6615d920a01d908333/src/server/server.zig#L743-L753 the user Id and the user name is taken and send inside a zon to each player. this id is then taken for the entity id

tillpp: Ah, you want to display the temporary entityID, which is then enough to kick and bann people?

i.e. /kick entitiyID instead of /kick playerID ?

or did i misunderstood you?=

IntegratedQuantum: user.playerIndex is the thing we need here.

Wunka: ... that is an early reveal... all my PRs about this topic handeld everything with the player id.... I will make a PR to fix that. Then we currently don't have a way for a client to know what the playerIndex is

tillpp: Idea:
you could transmit it with the entity data over protocol 8 (entity) for now
(client/entit_manager.zig) getting parsed with addEntity
grafik

and when i have my ECS PRs done, i could move this information into it's own EntitiyComponent: "playerInformation" for example, that stores playerIDs and other information that is only specific to players

(Before you implement it this way, i would like to see quantums approval, before you implement it unnessarily)

IntegratedQuantum: Yeah, sorry for that, I didn't realize you used the wrong thing in your other PRs.

Wunka: What about the suggestion from Till? I would like to do it that way (would open an issue about separating it into its own component)

IntegratedQuantum: Yeah, for that seems like the best option, and apparently it's going to take a bit longer until components are ready and usable.

Wunka: added `playerIndex` and also... I didn't see that `playerIndex` is a `usize` and not `u32`... so that Is also changed in the `getPlayerByIndex` function

src/client/Entity.zig
```diff
@@ -30,13 +30,15 @@ rot: Vec3f = undefined,
...
 name: []const u8,
+playerIndex: usize, // TODO extract into own component #2760
 
 pub fn init(self: *@This(), zon: ZonElement, allocator: NeverFailingAllocator) void {
 	self.* = @This(){
 		.id = zon.get(u32, "id", std.math.maxInt(u32)),
 		.width = zon.get(f64, "width", 1),
 		.height = zon.get(f64, "height", 1),
 		.name = allocator.dupe(u8, zon.get([]const u8, "name", "")),
+		.playerIndex = zon.get(usize, "playerIndex", 1),
```
IntegratedQuantum: I'd suggest to use `std.math.maxInt(usize)` as the default, to make it clear to the player that something is wrong.

src/server/server.zig
```diff
@@ -485,6 +485,14 @@ pub const User = struct { // MARK: User
 			.no, .neutral => false,
 		};
 	}
+
+	pub fn format(user: User, writer: *std.Io.Writer) std.Io.Writer.Error!void {
+		if (main.settings.showIdWithName) {
+			try writer.print("{s}@{d}", .{user.name, user.playerIndex});
```
IntegratedQuantum: I think for the server logs, it should always print the id instead of hiding it behind a setting.

IntegratedQuantum (changes requested):

src/server/server.zig
Wunka: both done.

IntegratedQuantum (approved):

PR #2661, opened 2026-03-04
Add parsePlayerId for command usage

### LINKED ISSUE #2615: Player needs a formatting-stripped identifier

Wunka:
With #2660 I think this will close #2615

This PR adds a helper function to `_command.zig` for parsing `playerIds` and also tests this by implementing it for `/perm`
This also solves the problem that currently the permission layer is only a hindrance because you weren't in a good way able to change other players permissions.

### DIFF
--- a/src/server/command/_command.zig
+++ b/src/server/command/_command.zig
@@ -77,8 +77,12 @@ pub fn parseCoordinates(split: *std.mem.SplitIterator(u8, .scalar), source: *Use
 }
 
 pub fn parsePlayerId(playerId: []const u8, source: *User) !*User {
-	const id = std.fmt.parseInt(u32, playerId, 10) catch {
-		source.sendMessage("#ff0000Player ids must be integers, found \"{s}\"", .{playerId});
+	if (!std.ascii.startsWithIgnoreCase(playerId, "@")) {
+		source.sendMessage("#ff0000Player id specifiers always start with @, found \"{s}\"", .{playerId});
+		return error.InvalidArg;
+	}
+	const id = std.fmt.parseInt(u32, playerId[1..], 10) catch {
+		source.sendMessage("#ff0000Player ids must be integers, found \"{s}\"", .{playerId[1..]});
 		return error.InvalidArg;
 	};
 	return main.server.getUserById(id) catch {
--- a/src/server/command/_command.zig
+++ b/src/server/command/_command.zig
@@ -75,3 +75,14 @@ pub fn parseCoordinates(split: *std.mem.SplitIterator(u8, .scalar), source: *Use
 		break :blk output;
 	};
 }
+
+pub fn parsePlayerId(playerId: []const u8, source: *User) !*User {
+	const id = std.fmt.parseInt(u32, playerId, 10) catch {
+		source.sendMessage("#ff0000Player ids must be integers, found \"{s}\"", .{playerId});
+		return error.InvalidArg;
+	};
+	return main.server.getUserById(id) catch {
+		source.sendMessage("#ff0000Player with id {d} not found", .{id});
+		return error.InvalidArg;
+	};
+}
--- a/src/server/command/_command.zig
+++ b/src/server/command/_command.zig
@@ -76,7 +76,7 @@ pub fn parseCoordinates(split: *std.mem.SplitIterator(u8, .scalar), source: *Use
 	};
 }
 
-pub fn parsePlayerIdAndIncreaseRefCount(playerId: []const u8, source: *User) !*User {
+fn parsePlayerIdAndIncreaseRefCount(playerId: []const u8, source: *User) !*User {
 	if (!std.ascii.startsWithIgnoreCase(playerId, "@")) {
 		source.sendMessage("#ff0000Player id specifiers always start with @, found \"{s}\"", .{playerId});
 		return error.InvalidArg;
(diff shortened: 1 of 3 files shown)


Wunka:
As this fixes one command and enables a lot more and with the good amount of PRs currently I have moved this to High Priority

src/server/command/_command.zig
```diff
@@ -75,3 +75,14 @@ pub fn parseCoordinates(split: *std.mem.SplitIterator(u8, .scalar), source: *Use
 		break :blk output;
 	};
 }
+
+pub fn parsePlayerId(playerId: []const u8, source: *User) !*User {
+	const id = std.fmt.parseInt(u32, playerId, 10) catch {
```
IntegratedQuantum: Should we prefix player ids with a special character to make it easier to distinguish from normal numbers?
I would use `#`, but that is already used for formatting. So maybe `@` like is used in other applications?

src/server/server.zig
```diff
@@ -793,3 +793,15 @@ pub fn sendMessage(comptime fmt: []const u8, args: anytype) void {
 	defer main.stackAllocator.free(msg);
 	sendRawMessage(msg);
 }
+
+pub fn getUserById(id: u32) !*User {
```
IntegratedQuantum: Needs to increase the refcount, otherwise the user can't be used.

IntegratedQuantum (changes requested):

src/server/command/_command.zig
Wunka: I like that. This would also allow certain additions later like other applications have (eg. `@p` or `@e`). Is implemented!

src/server/server.zig
Wunka: which one? I am already increasing from the userlist and when I find a user also from them

IntegratedQuantum: Ah I didn't see that, please make this clear in the function name to signal to the caller that they need to decreaseRefCount once they are done.

Wunka: I added `AndIncreaseRefCount` to both helper functions name

src/server/command/permission/perm.zig
```diff
@@ -21,35 +25,54 @@ pub fn execute(args: []const u8, source: *User) void {
 	if (split.next()) |arg| {
 		if (std.ascii.eqlIgnoreCase(arg, "remove")) {
 			const helper = Helper.parseHelper(source, &split) catch return;
-			if (!source.permissions.removePermission(helper.listType, helper.permissionPath)) {
+			defer if (helper.isReference) helper.user.decreaseRefCount();
```
IntegratedQuantum: I'd suggest to add a deinit function to the helper for it.
Also to make it more clear that a deinit is needed, the creation function should be called `init`.

src/server/command/permission/perm.zig
```diff
@@ -62,13 +85,25 @@ const Helper = struct {
 			return error.InvalidArgs;
 		}
 
-		const permissionPath = split.next() orelse {
-			source.sendMessage("#ff0000Too few arguments for command /perm.", .{});
+		arg = split.next() orelse {
```
IntegratedQuantum: I prefer the old naming. Just calling it `arg` kind of makes it harder to understand the meaning in my opinion.

src/server/command/permission/perm.zig
```diff
@@ -62,13 +85,25 @@ const Helper = struct {
...
+			source.sendMessage("#ff0000Too few arguments for command /perm", .{});
 			return error.InvalidArgs;
 		};
 
-		if (permissionPath[0] != '/') {
-			source.sendMessage("#ff0000Permission paths always begin with a \"/\", got: {s}", .{permissionPath});
+		var isReference = false;
+		const user: *User = blk: {
+			if (split.next()) |next| {
```
IntegratedQuantum: I think this should use the `@` symbol to distinguish between players and the permission path, instead of checking if there is something afterwards (this is a fragile solution that will break if you add more arguments).

src/server/command/permission/perm.zig
```diff
@@ -21,35 +25,54 @@ pub fn execute(args: []const u8, source: *User) void {
...
-			source.sendMessage("#ff0000Expected either add, remove or a valid permission path, found \"{s}\"", .{arg});
 		}
 	}
 }
 
 const Helper = struct {
 	listType: ListType,
 	permissionPath: []const u8,
+	user: *User,
+	isReference: bool,
```
IntegratedQuantum: I think this is a weird name. Both are references. The real question is if you need to decrement the refcount

src/server/server.zig
```diff
@@ -793,3 +793,15 @@ pub fn sendMessage(comptime fmt: []const u8, args: anytype) void {
...
+pub fn getUserByIdAndIncreaseRefCount(id: u32) !*User {
+	const userList = getUserListAndIncreaseRefCount(main.stackAllocator);
+	defer freeUserListAndDecreaseRefCount(main.stackAllocator, userList);
+	for (userList) |user| {
+		if (user.id == id) {
+			user.increaseRefCount();
+			return user;
+		}
+	}
+	return error.UserNotFound;
```
IntegratedQuantum: So far we've used optional instead of errors when something cannot be found (and it is expected behavior that this happens).

IntegratedQuantum (changes requested):

src/server/command/permission/perm.zig
Wunka: all done. I saw that I copied a lot over for `/perm [blank]`, thats why I now made persing playerids even better with the Target struct. else every command would just copy over the `increasedRefCoun` bool. with all its logic

src/server/command/permission/perm.zig
```diff
@@ -20,34 +24,43 @@ pub fn execute(args: []const u8, source: *User) void {
...
+				return;
+			};
+			if (permissionPath[0] != '/') {
+				source.sendMessage("#ff0000Permission paths always begin with a \"/\", got: {s}", .{permissionPath});
 				return;
 			}
-			if (source.hasPermission(arg)) {
-				source.sendMessage("#00ff00User has permission for path: {s}", .{arg});
+			if (target.user.hasPermission(permissionPath)) {
+				source.sendMessage("#00ff00Player {s}#00ff00 has permission for path: {s}", .{target.user.name, permissionPath});
```
IntegratedQuantum: Also reset the other formatting options
```suggestion
				source.sendMessage("#00ff00Player {s}§#00ff00 has permission for path: {s}", .{target.user.name, permissionPath});
```

src/server/command/permission/perm.zig
```diff
@@ -20,34 +24,43 @@ pub fn execute(args: []const u8, source: *User) void {
...
 			} else {
-				source.sendMessage("#ff0000User has no permission for path: {s}", .{arg});
+				source.sendMessage("#ff0000Player {s}#ff0000 has no permission for path: {s}", .{target.user.name, permissionPath});
```
```suggestion
				source.sendMessage("#ff0000Player {s}§#ff0000 has no permission for path: {s}", .{target.user.name, permissionPath});
```

IntegratedQuantum (changes requested):

Wunka:
accepted both

IntegratedQuantum (approved):

PR #2664, opened 2026-03-04
ItemDrops be more visible

tillpp:
Items are hard to spot, because they tend to lay on the ground or be half sunken in the ground.
for example:
image
 **before**
image
**after**
image

### DIFF
--- a/src/itemdrop.zig
+++ b/src/itemdrop.zig
@@ -43,7 +43,7 @@ pub const ItemDropNetworkData = struct {
 
 pub const ItemDropManager = struct { // MARK: ItemDropManager
 	/// Half the side length of all item entities hitboxes as a cube.
-	pub const radius: f64 = 0.1;
+	pub const radius: f64 = 0.4;
 	/// Side length of all item entities hitboxes as a cube.
 	pub const diameter: f64 = 2*radius;
 


IntegratedQuantum (changes requested):
If I remember correctly the actual problem was that items rotate around one of the corners instead of the center.

I don't like that with the increased radius they appear to be floating.

tillpp:
> I don't like that with the increased radius they appear to be floating.

the floating was intentional, i think they make them stand out more and signify that they are something to pick up

tillpp:
@ikabod-kee @careeoki @Argmaster 
I would like to know what you opinion on this is?

ikabod-kee:
>  I would like to know what you opinion on this is?

This is a good change in my opinion! But also item drops should not have random rotations in general, and maybe face the camera. (that would be the best visibility)

Argmaster:
Personally I don't mind low visibility, I do mind clipping tho.

Making them have physics with more precise hitbox, so they can lay flat and lean against the wall, would be super cool, but also very expensive and potentially hard to implement.
(I would love to see that tho)

I'm not a big fan of floating, not my preferred art style.

> and maybe face the camera.

Making them face the camera may cause them to still clip through walls, no? Also not my preferred art style.

tillpp:
> Making them face the camera may cause them to still clip through walls, no?

fyi: Rotation have no effect on Hitboxes

IntegratedQuantum:
I actually like that it clips into the block a little. Maybe it could be a bit less extreme, but overall I would like to keep this.
It makes it look like it actually fell down and then got stuck there in the ground.

The current hitbox size is a bit conservative, we could make a hitbox that depends on the actual size (item vs block) and rotation of the item.

> and maybe face the camera. (that would be the best visibility)

Is this about visibility or style? If it's purely about visibility then I'd prefer an accessibility option for it, and keep the "items are stuck in blocks" look we currently have.

ikabod-kee:
> Is this about visibility or style? If it's purely about visibility then I'd prefer an accessibility option for it, and keep the "items are stuck in blocks" look we currently have.

Sort of both. I think having the items be poking out of the ground makes more sense if the items didn't slide like they were covered in vaseline and slipping on ice.

tillpp:
Not sure what the issue with this PR was,
but also turned into draft, low priority, Let's focus quantum on ECS parts

PR #2667, opened 2026-03-05
Metal synergistic effects

### LINKED ISSUE #2637: Metal synergistic effects similar to encasing a gem
I think it'd be cool to see more metal synergistic effects.

As an example:

Copper next to silver could activate an effect in silver that makes it more durable.

In real life, sterling silver (Mostly silver with some copper) is more durable than just silver alone. 
Silver is soft, copper is added to make it less soft and so it's used in lots of jewelery.

Rather than create the alloy directly, we just apply an effect to the tool properties if a metal is next to another metal (like copper adjacent to silver) so players are further encouraged to encase a gem in silver but have copper on the outside for example.

I don't think this applies to every case where you use multiple metals in a tool, but, we can create specific synergies like silver and copper that would complement #2636

H41ogen:
Closes #2637

Adds a modifier of 10% fragile to silver, and 20% fragile to gold when not encased in at least 1 copper. Gold also has slightly higher base durability.

I originally tried a durable modifier, but since most tools already have at least 2 items giving them the same modifier, it barely had an effect.

### DIFF
--- a/assets/cubyz/items/gold_ingot.zig.zon
+++ b/assets/cubyz/items/gold_ingot.zig.zon
@@ -13,7 +13,7 @@
 		.modifiers = .{
 			.{
 				.id = "fragile",
-				.strength = 0.2,
+				.strength = 0.15,
 				.restriction = .{
 					.id = .not,
 					.child = .{


TyrrelMeter:
20 percent might be a bit to much of an increase, maybe reduce it a bit.

H41ogen:
It doesn't have that much of an effect in practice, due to the stats you lose by using copper. Also, durability goes down.

H41ogen:
never mind lol

H41ogen:
Ok so I think this is a bug with the `.light` modifier. 2 5% bonuses should not combine to make 33%.

BoySanic:
This isn't entirely what I had in mind as you'd have to enclose the silver or gold with copper. I more imagined like...

image

Something to this effect. Maybe you improve durability if a silver is adjacent to a copper or something. I think it'd need a new modifier trigger other than encased.

BoySanic:
image

alternate idea: Adjacent to copper on two sides = "fused"?

H41ogen:
> This isn't entirely what I had in mind as you'd have to enclose the silver or gold with copper. I more imagined like...
> image
> 
> Something to this effect. Maybe you improve durability if a silver is adjacent to a copper or something. I think it'd need a new modifier trigger other than encased.

Is this not already how encasing works? I'm confused.

Wunka:
Just to add. @BoySanic did you envision one of the modifier types from these PRs? #2646 #2650

ikabod-kee:
I think the idea of metals being synergistic with each other is a genius idea. It's like alloys without alloys!

BoySanic:
> I think the idea of metals being synergistic with each other is a genius idea. It's like alloys without alloys!

Precisely what I thought :)

More intuitive than having the alloys outright while still allowing some metal mixing fun

BoySanic:
> > This isn't entirely what I had in mind as you'd have to enclose the silver or gold with copper. I more imagined like...
> > image
> > Something to this effect. Maybe you improve durability if a silver is adjacent to a copper or something. I think it'd need a new modifier trigger other than encased.
> 
> Is this not already how encasing works? I'm confused.

You might be right, I missed the encase only had a requirement of 1

BoySanic:
> Just to add. @BoySanic did you envision one of the modifier types from these PRs? #2646 #2650

Hmm, I don't think so.

assets/cubyz/items/gold_ingot.zig.zon
```diff
@@ -10,5 +10,16 @@
 		.colors = .{
 			0xffbc7132, 0xffda972b, 0xfff2cd53, 0xfff2cd53, 0xfffae39e,
 		},
+		.modifiers = .{
+			.{
+				.id = "powerful",
```
BoySanic: What if we had a `durability` modifier for encasing with 2 coppers? Same for silver. 

Though I'm not sure how that'll mesh with the powerful modifier. Like, will it invoke both? Seems kinda cool to me if it does but we'll want to look at balancing a little bit.

assets/cubyz/items/silver_ingot.zig.zon
```diff
@@ -10,5 +10,16 @@
 		.colors = .{
 			0xff67727f, 0xff89919e, 0xffbdc2c6, 0xffe4e6e6, 0xfffcfcfc,
 		},
+		.modifiers = .{
+			.{
+				.id = "light",
```
BoySanic: Same comment as for gold. I think a durability mod would be cool for 2x encased (sandwiching a gold or silver "meat" in between copper "bread")

assets/cubyz/items/gold_ingot.zig.zon
H41ogen: It's difficult to balance anything with a durable modifier since on a tool with 2+ wood it'll do basically nothing, but on a tool with no wood it's crazy powerful.

H41ogen:
I'm waiting for #2668 before continuing with working on silver.

IntegratedQuantum:
How does this fit into the world/lore? These effects on their own seem rather random (what does copper do to enhance the other metals? What's the mechanism here?). Same could be said about precious-encasing, but I feel like there it at least fits aesthetically: We have a lot of jewelry in the real world that's encased gems.

Where I could see multiple metals working together is for building some kind of battery out of 2 different metals and some plant stuff, maybe mushrooms, or some kind of thermoelectric effect or something like that.

BoySanic:
> what does copper do to enhance the other metals? 

In the issue I wrote (#2637) I mentioned sterling silver being an alloy of silver and copper, where the durability/strength properties of copper are combined with silver for use in jewelry as silver is a quite soft metal on its own. That said, this PR wasn't what I had in mind for the modifiers. 

The metal-metal interaction concept sounded fun to me. Alloy-like effects without having actual alloying.

IntegratedQuantum:
Oh yeah, a durability boost actually makes more sense.
On the other end of the spectrum we have also the opposite behavior, like gallium+aluminium which is super brittle, I think it would be find if you had to actively avoid certain metal combinations.

H41ogen:
> On the other end of the spectrum we have also the opposite behavior, like gallium+aluminium which is super brittle, I think it would be find if you had to actively avoid certain metal combinations.

Iron and copper don't mix well, but would that work in the game?

BoySanic:
> > On the other end of the spectrum we have also the opposite behavior, like gallium+aluminium which is super brittle, I think it would be find if you had to actively avoid certain metal combinations.
> 
> Iron and copper don't mix well, but would that work in the game?

As it currently exists, that'd be fine. That said I do see a potential tool with this synergy idea where you'd use iron for its hardness on some parts of the tool, but use silver or gold inlaid in copper to encase a gem so that silver/gold won't make the tool too soft.
image

I could see the argument that since the iron is only the "tips" of the pickaxe, it's just added on, not mixed as in metallurgy. (So, encase iron w/ 2 coppers maybe could be a negative, or encase iron with 2 silvers, or 2 golds, etc).

H41ogen:
Since #2668 was merged, I think I'll look into balancing this again when I get back home in a few days. I think the way to go might just be to give tools a mild `weak` modifier if gold/silver isn't encased in 1 copper.

H41ogen:
I think what I have now is pretty good. I gave gold a 20% fragile modifier if not encased in 1 copper, as well as slightly higher base durability, and gave silver a 10% fragile modifier if not encased in 1 copper. Let me know what you guys think.

IntegratedQuantum:
I'm really not sure of this, it makes all the tools more complex due to basically requiring at least two copper.
I think this complexity could be nice, but I think we should come back to this idea for future materials instead of applying it for silver and gold.

H41ogen:
What if we reserve the requiring copper for just gold, and add a bonus for combining gold and silver, keeping the complex stuff to after the player finds gold (z=-2k i think)?

IntegratedQuantum:
That might be worth a try. I would still introduce it slowly then (with one interaction with gold, not two, so decide if you wwant it to synergize with silver or copper)

BoySanic:
Synergize with copper, imo, since the reason I thought of this to begin with was to give copper more uses in tools.

H41ogen:
makes sense to me

IntegratedQuantum:
Could someone check how this feels in-game?

Argmaster:
1. I see this mostly as a quirk instead of actual useful feature and I don't like it this way.
2. Effect can be ignored, because it overlaps with damage and swing speed tradeoff; consider following trivial designs first one just `gold+iron+diamon`, second one `+copper`
image
image

3. This is just a plain stat change. Plain stat changes already happen when we swap copper to iron / gold to silver. Plain stat changes are not interesting, you just have to deal with them.
4. I would insist on preferring buffs over debuffs. Result is the same, but the reception is completely different. When you have  a debuff you feel bad with it, feel pressured to remove it. When you can have a buff, you feel motivated to get it, but its optional. In this case lower durability + durable would be better way to express this behavior IMO

I still think we don't need this, not in this form. We need mid-end game uses for copper / silver, but that doesn't equal making mid-game materials worse, just to squeeze the early materials in there. The whole point of moving to mid-game is to progress from early game materials.

---

Something that I think would be cool way to make material return:

Add 3rd precious material (platinum/electrum/palladium idk make it blue) and require using gold+silver+palladium in single tool to remove palladium debuff modifiers. Damage of tool like this should overshoot health and resistance of most existing materials by a lot to make it reliably one-shot most blocks. 
- Silver would be the early game one.
- Gold would be the simple/default choice most of the time.
- Palladium would be the weird one, the outlier, the exclusive, rare one. 

That doesn't automatically make silver super useful, but that's not the point, the point is to make it return after it was not useful for some time. In the end you won't be able to make all materials meta.

PR #2668, opened 2026-03-05
Fix inverse modifiers combining incorrectly

H41ogen:
Uses this:
image
instead of this:
image
- - -
This mainly fixes an issue where combining 2 small modifiers would always result in a modifier of at least around 29%, no matter how low the composite modifiers were. For example, two 5% modifiers would combine to 33%.

### DIFF
--- a/src/tool/modifiers/bad_at.zig
+++ b/src/tool/modifiers/bad_at.zig
@@ -13,7 +13,7 @@ pub fn loadData(zon: main.ZonElement) Data {
 
 pub fn combineModifiers(data1: Data, data2: Data) ?Data {
 	if (data1.tag != data2.tag) return null;
-	return .{.strength = 1.0 - 1.0/(1 + std.math.hypot(1.0/(1.0 - data1.strength) - 1, 1.0/(1.0 - data2.strength) - 1)), .tag = data1.tag};
+	return .{.strength = 1.0 - 1.0/(1.0 + std.math.hypot(1.0/(1.0 - data1.strength) - 1.0, 1.0/(1.0 - data2.strength) - 1.0)), .tag = data1.tag};
 }
 
 pub fn changeToolParameters(_: *Tool, _: Data) void {}


H41ogen:
I think the best solution would be for "good" modifiers to be positive, and "bad" modifiers to be negative. It's easy for the player to accept that "bad" modifiers might combine "differently" than "good" ones.

H41ogen:
Additionally, to keep the player from immediately assuming that amber must be overpowered (omg!!!! +100% swing speed!!!!!!), maybe amber's modifier could also be moved down to something like +75% swing speed.

IntegratedQuantum:
it should be set to 50%, the reason it even to got to 100% was a mistake on my part when I converted the internal tool property from swing speed to swing time.

H41ogen:
What was the original reason for amber affecting swing time instead of swing speed?

IntegratedQuantum:
because the underlying tool parameter was renamed to swing time

H41ogen:
This can now be done and make sense

src/tool/modifiers/bad_at.zig
```diff
@@ -13,7 +13,7 @@ pub fn loadData(zon: main.ZonElement) Data {
 
 pub fn combineModifiers(data1: Data, data2: Data) ?Data {
 	if (data1.tag != data2.tag) return null;
-	return .{.strength = 1.0 - 1.0/std.math.hypot(1.0/(1.0 - data1.strength), 1.0/(1.0 - data2.strength)), .tag = data1.tag};
+	return .{.strength = 1.0 - 1.0/(1 + std.math.hypot(1.0/(1.0 - data1.strength) - 1, 1.0/(1.0 - data2.strength) - 1)), .tag = data1.tag};
```
IntegratedQuantum: Please consistently use .0 everywhere

```suggestion
	return .{.strength = 1.0 - 1.0/(1.0 + std.math.hypot(1.0/(1.0 - data1.strength) - 1.0, 1.0/(1.0 - data2.strength) - 1.0)), .tag = data1.tag};
```

IntegratedQuantum (approved):

PR #2669, opened 2026-03-05
Send fewer item position informations  

tillpp:
Less bandwidth to be sacrificed for item position

### DIFF
--- a/src/itemdrop.zig
+++ b/src/itemdrop.zig
@@ -169,7 +169,7 @@ pub const ItemDropManager = struct { // MARK: ItemDropManager
 			j += 1;
 		}
 		const result = allocator.alloc(ItemDropNetworkData, j);
-		@memcpy( result, buffer[0..j]);
+		@memcpy(result, buffer[0..j]);
 		return result;
 	}
 
--- a/src/itemdrop.zig
+++ b/src/itemdrop.zig
@@ -154,19 +154,24 @@ pub const ItemDropManager = struct { // MARK: ItemDropManager
 	}
 
 	pub fn getPositionAndVelocityData(self: *ItemDropManager, allocator: NeverFailingAllocator) []ItemDropNetworkData {
-		const result = allocator.alloc(ItemDropNetworkData, self.size);
-		var j:usize = 0;
-		for (self.indices[0..self.size]) |i, | {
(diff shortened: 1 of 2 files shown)


src/itemdrop.zig
```diff
@@ -153,14 +154,22 @@ pub const ItemDropManager = struct { // MARK: ItemDropManager
...
 
 	pub fn getPositionAndVelocityData(self: *ItemDropManager, allocator: NeverFailingAllocator) []ItemDropNetworkData {
-		const result = allocator.alloc(ItemDropNetworkData, self.size);
-		for (self.indices[0..self.size], result) |i, *res| {
-			res.* = .{
+		const buffer = main.stackAllocator.alloc(ItemDropNetworkData, self.size);
+		defer main.stackAllocator.free(buffer);
+		var j: usize = 0;
+		for (self.indices[0..self.size]) |i| {
+			if (vec.lengthSquare(self.list.items(.pos)[i] - self.list.items(.oldPos)[i]) < 1e-4)
```
IntegratedQuantum, line 161: We need to at least send one position of the block entity being at rest. Otherwise the client interpolation may keep using a high velocity sample from the last packet which may clip the item into the block it just landed on.

Furthermore I'd suggest to explicitly set the velocity to zero if it's small enough, and then only send the items whose velocities are non-zero. This prevent a small drift (1e-4 can sum up to measurable values) from being unnoticed by the client.

src/itemdrop.zig
```diff
@@ -495,6 +506,10 @@ pub const ClientItemDropManager = struct { // MARK: ClientItemDropManager
 		self.timeDifference.addDataPoint(time);
 		var pos: [ItemDropManager.maxCapacity]Vec3d = undefined;
 		var vel: [ItemDropManager.maxCapacity]Vec3d = undefined;
+
+		@memcpy(@as(*[maxf64Capacity]f64, @ptrCast(&pos)), &self.interpolation.lastPos[self.interpolation.frontIndex]);
```
IntegratedQuantum, line 510: this is incorrect if there is a lag spike between this and the previous frame, then the item will be stuck at an old interpolated position, which may put it mid-air.

IntegratedQuantum (changes requested):

tillpp:
turned to draft, low priority right now, let focus on more Important stuff, like ECS8

PR #2672, opened 2026-03-06
ECS Part 1 | Assets.

tillpp:
First step to splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652

credit for the models @careeoki 

Assets for cubert, snail, snale (moved), mofallo and missing (in case a model was unable to be loaded)


assets/cubyz/entities/models/snale.obj
IntegratedQuantum, line 1: Moving these files has an unintended side effect: They are now addressable from blocks, which is undesirable since blocks treat UVs differently.
If you do this, please also move all the block models into a separate folder.
Alternatively we could also consider using the same structure as for textures: `cubyz/blocks/textures` `cubyz/blocks/models` `cubyz/entities/models`

tillpp: sounds good. changed

(if you merge this, dont forget to credit carrie in the merge commit)

IntegratedQuantum: I think automatically credits everyone involved in the PR

PR #2673, opened 2026-03-06
ECS Part 2 | Templates

tillpp:
(to splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652 )

added Templates for entity component and entity systems

also renamed entity.zig -> clientEntity.zig to make it easier to distingues it from the already existing Entity.zig in server/


src/entityComponent/_list.zig
```diff
@@ -0,0 +1 @@
+pub const template = @import("template.zig");
```
IntegratedQuantum: I'd suggest to call it _template to symbolize that it's a special thing and sort it at the start in the file explorer.

src/entityComponent/template.zig
```diff
@@ -0,0 +1,70 @@
...
+const blocks = main.blocks;
+const chunk_zig = main.chunk;
+const ServerChunk = chunk_zig.ServerChunk;
+const World = game.World;
+const ServerWorld = main.server.ServerWorld;
+const items = main.items;
+const ItemStack = items.ItemStack;
+const random = main.random;
+
+pub const ENTITY_COMPONENT_VERSION = 0;
```
IntegratedQuantum: WE DON'T SCREAM IN THE CUBYZ CODEBASE

src/entityComponent/template.zig
```diff
@@ -0,0 +1,70 @@
...
+
+// ############################# Client only stuff ################################
+pub const Client = struct {
+	pub fn register(id: u32, reader: *utils.BinaryReader, version: u32) void {
```
IntegratedQuantum: Why did you call this register/unregister and not load/unload?

src/entitySystem/_template.zig
```diff
@@ -0,0 +1,55 @@
...
+	pub fn init() void {}
+	pub fn deinit() void {}
+	pub fn clear() void {}
+
+	pub fn render(projMatrix: Mat4f, ambientLight: Vec3f, playerPos: Vec3d) void {
+		_ = projMatrix;
+		_ = ambientLight;
+		_ = playerPos;
+	}
+	pub fn renderInfo(projMatrix: Mat4f, ambientLight: Vec3f, playerPos: Vec3d) void {
```
IntegratedQuantum: What's the difference between render and renderInfo?

src/entitySystem/template.zig
```diff
@@ -0,0 +1,55 @@
...
+
+const entityComponent = main.entityComponent;
+
+// ############################# Client only stuff ################################
+pub const Client = struct {
+	pub fn init() void {}
+	pub fn deinit() void {}
+	pub fn clear() void {}
+
+	pub fn render(projMatrix: Mat4f, ambientLight: Vec3f, playerPos: Vec3d) void {
```
IntegratedQuantum: Should also have a deltaTime parameter to allow animating stuff

src/clientEntity.zig
IntegratedQuantum, line 1: I think it might be reasonable to create a client folder, and call it client.entity.

src/entitySystem/_template.zig
```diff
@@ -0,0 +1,55 @@
...
+		_ = projMatrix;
+		_ = ambientLight;
+		_ = playerPos;
+	}
+};
+// ############################# Server only stuff ################################
+pub const Server = struct {
+	pub fn init() void {}
+	pub fn deinit() void {}
```
IntegratedQuantum, line 55: Why does this one have no functionality at all?

src/main.zig
```diff
@@ -582,8 +584,26 @@ pub fn main() void { // MARK: main()
 	network.init() catch @panic("Failed to initialize network");
 	defer network.deinit();
 
-	if (!headless) entity.ClientEntityManager.init();
-	defer if (!headless) entity.ClientEntityManager.deinit();
+	// Entity Component System
+	if (!headless) {
+		inline for (@typeInfo(entityComponent).@"struct".decls) |decl| {
+			@field(entityComponent, decl.name).Client.init();
+		}
```
IntegratedQuantum: Please don't put these implementation details into the global main. This belongs into a client.entity.init/deinit function.

src/main.zig
```diff
@@ -11,7 +11,9 @@ pub const blocks = @import("blocks.zig");
 pub const blueprint = @import("blueprint.zig");
 pub const callbacks = @import("callbacks/callbacks.zig");
 pub const chunk = @import("chunk.zig");
-pub const entity = @import("entity.zig");
+pub const clientEntity = @import("clientEntity.zig");
+pub const entityComponent = @import("entityComponent/_list.zig");
+pub const entitySystem = @import("entitySystem/_list.zig");
```
IntegratedQuantum: These don't belong in the root namespace.
Also it's a list, so it should be plural

IntegratedQuantum: Should we maybe have a global entity.zig that handles these and their initialization?

src/game.zig
```diff
@@ -660,8 +660,14 @@ pub const World = struct { // MARK: World
 		main.items.clearRecipeCachedInventories();
 		main.sync.ClientSide.reset();
 
+		inline for (@typeInfo(main.entitySystem).@"struct".decls) |decl| {
+			@field(main.entitySystem, decl.name).Client.clear();
+		}
+		inline for (@typeInfo(main.entityComponent).@"struct".decls) |decl| {
+			@field(main.entityComponent, decl.name).Client.clear();
+		}
```
IntegratedQuantum: same here, this doesn't belong here.

src/server/server.zig
```diff
@@ -559,6 +565,12 @@ fn deinit() void {
 
 	main.sync.ServerSide.deinit();
 	main.items.Inventory.ServerSide.deinit();
+	inline for (@typeInfo(main.entityComponent).@"struct".decls) |decl| {
+		@field(main.entityComponent, decl.name).Server.deinit();
+	}
+	inline for (@typeInfo(main.entitySystem).@"struct".decls) |decl| {
+		@field(main.entitySystem, decl.name).Server.deinit();
+	}
```
IntegratedQuantum: same here

src/server/server.zig
```diff
@@ -515,6 +515,12 @@ fn init(name: []const u8, singlePlayerPort: ?u16) void { // MARK: init()
 		@panic("Could not open Server.");
 	}; // TODO Configure the second argument in the server settings.
 
+	inline for (@typeInfo(main.entityComponent).@"struct".decls) |decl| {
+		@field(main.entityComponent, decl.name).Server.init();
+	}
+	inline for (@typeInfo(main.entitySystem).@"struct".decls) |decl| {
+		@field(main.entitySystem, decl.name).Server.init();
+	}
```
IntegratedQuantum: and here

src/entitySystem/_template.zig
tillpp: RenderInfo does not run when pressing F1 (hiding UI)
RenderInfo displays the playername for example.

src/entityComponent/template.zig
tillpp: xD
i wasnt sure how to name this without shadowing the "version" parameter in function below.
should i call it _version?

src/entitySystem/_template.zig
tillpp: You mean, why there is no update function for example?
Currently there is no need right now,
So I thought we will add them once we need them.

(But you really want, I can also just add them right now)

src/entityComponent/template.zig
tillpp: changed it to entityComponentVersion

src/entityComponent/template.zig
tillpp: renamed

src/entitySystem/template.zig
tillpp: added.

tillpp:
Changed the folder structure
Is this what had in mind?

tillpp:
unfortunatly git is not smart enough to realize that i just moved code around, making this PR look bigger now then it should

summary:
(previously entity.zig) clientEntity.zig => client.entity.zig and client.EntityManager.zig
template.zig => _template.zig
ECS inits/deinit and the lists of components/system => entity.zig

src/entitySystem/_template.zig
IntegratedQuantum: Then I'd call it renderHud

src/entitySystem/_template.zig
IntegratedQuantum: yes, please. I think it would be good to have that part also there already, and it shouldn't be too difficult.

src/entity.zig
```diff
@@ -1,22 +1,12 @@
-const std = @import("std");
```
IntegratedQuantum, line 1: std should always take the first spot, even if you don't need it right now

src/entity.zig
```diff
@@ -1,22 +1,12 @@
...
+pub const components = @import("entityComponent/_list.zig");
+pub const systems = @import("entitySystem/_list.zig");
```
IntegratedQuantum: public imports come always after the private imports so you can easily skip over the private import section of the file.

src/entity.zig
```diff
@@ -25,257 +15,47 @@ pub const EntityNetworkData = struct {
...
-		ambientLight: c_int,
-	} = undefined;
-	var modelBuffer: main.graphics.SSBO = undefined;
-	var modelSize: c_int = 0;
-	var modelTexture: main.graphics.Texture = undefined;
-	var pipeline: graphics.Pipeline = undefined; // Entities are sometimes small and sometimes big. Therefor it would mean a lot of work to still use smooth lighting. Therefor the non-smooth shader is used for those.
-	pub var entities: main.utils.VirtualList(ClientEntity, 1 << 20) = undefined;
-	pub var mutex: std.Thread.Mutex = .{};
-
+pub const Client = struct {
```
IntegratedQuantum: namespaces are lowercase
```suggestion
pub const client = struct {
```

src/entity.zig
```diff
@@ -25,257 +15,47 @@ pub const EntityNetworkData = struct {
...
-			const alpha: u32 = @intFromFloat(std.math.clamp(0xff - transparency, 0, 0xff));
-			graphics.draw.setColor(alpha << 24);
-
-			var buf = graphics.TextBuffer.init(main.stackAllocator, ent.name, .{.color = 0xffffff}, false, .center);
-			defer buf.deinit();
-			const fontSize = std.mem.max(f32, &.{fontMinScreenSize, fontScreenSize/projectedPos[3]});
-			const size = buf.calculateLineBreaks(fontSize, @floatFromInt(main.Window.width*8));
-			buf.render(xCenter - size[0]/2, yCenter - size[1], fontSize);
+};
+pub const Server = struct {
```
```suggestion
pub const server = struct {
```

src/game.zig
```diff
@@ -660,8 +660,10 @@ pub const World = struct { // MARK: World
 		main.items.clearRecipeCachedInventories();
 		main.sync.ClientSide.reset();
 
+		main.entity.Client.clear();
+
 		main.threadPool.clear();
-		main.entity.ClientEntityManager.clear();
+		main.client.EntityManager.clear();
```
IntegratedQuantum: should this also be moved to entity.client.clear()?

src/game.zig
```diff
@@ -660,8 +660,10 @@ pub const World = struct { // MARK: World
 		main.items.clearRecipeCachedInventories();
 		main.sync.ClientSide.reset();
 
+		main.entity.Client.clear();
```
IntegratedQuantum: should be done after the threadpool was cleared, so there is no chance that any remaining tasks still reference entities

src/main.zig
```diff
@@ -582,8 +583,12 @@ pub fn main() void { // MARK: main()
...
 	defer network.deinit();
 
-	if (!headless) entity.ClientEntityManager.init();
-	defer if (!headless) entity.ClientEntityManager.deinit();
+	// Entity Component System
+	if (!headless) entity.Client.init();
+	defer if (!headless) entity.Client.deinit();
+
+	if (!headless) client.EntityManager.init();
+	defer if (!headless) client.EntityManager.deinit();
```
IntegratedQuantum: Should maybe also be moved to entity.client.init()?

src/main.zig
```diff
@@ -582,8 +583,12 @@ pub fn main() void { // MARK: main()
 	network.init() catch @panic("Failed to initialize network");
 	defer network.deinit();
 
-	if (!headless) entity.ClientEntityManager.init();
-	defer if (!headless) entity.ClientEntityManager.deinit();
+	// Entity Component System
```
IntegratedQuantum: not needed, the namespace already implies this, and the implementation detail (whether it's ECS or not) is irrelevant here.

src/client/entity_manager.zig
IntegratedQuantum, line 1: this is a namespace and should be called `entity_manager`

src/entityComponent/_list.zig
```diff
@@ -0,0 +1 @@
+pub const _ = @import("_template.zig");
```
IntegratedQuantum: Every entry in lists should have the same name, also `_` as a name is a bit of a weird edge case in the language.
```suggestion
pub const _template = @import("_template.zig");
```

src/entitySystem/_list.zig
```diff
@@ -0,0 +1 @@
+pub const _ = @import("_template.zig");
```
IntegratedQuantum: same here

src/entityComponent/_list.zig
tillpp: i thought maybe this would compile it, without having it as an option as a component.

src/client/entity_manager.zig
tillpp: Hey you are the one who named it this way not me! xD
but i can change it

tillpp: and done

IntegratedQuantum: that was java me. A totally different person.

src/entityComponent/_list.zig
IntegratedQuantum: In Zig things only compile if they are used. An import alone would not be enough.

src/entitySystem/_template.zig
tillpp: forgot that this was open

src/entityComponent/_template.zig
```diff
@@ -0,0 +1,70 @@
...
+const World = game.World;
+const ServerWorld = main.server.ServerWorld;
+const items = main.items;
+const ItemStack = items.ItemStack;
+const random = main.random;
+
+pub const entityComponentVersion = 0;
+
+// ############################# Client only stuff ################################
+pub const Client = struct {
```
IntegratedQuantum: Are these expected to hold any fields? Otherwise I'd say they should also be lowercase.

src/entityComponent/_template.zig
```diff
@@ -0,0 +1,70 @@
...
+	pub fn unload(id: u32) void {
+		_ = id;
+	}
+	pub fn init() void {}
+	pub fn deinit() void {}
+	pub fn clear() void {}
+};
+// ############################# Server only stuff ################################
+pub const Server = struct {
+	pub const ItemComponent = struct {
```
IntegratedQuantum:
```suggestion
	pub const ExampleComponent = struct {
```

src/entitySystem/_template.zig
```diff
@@ -0,0 +1,58 @@
...
+const World = game.World;
+const ServerWorld = main.server.ServerWorld;
+const items = main.items;
+const ItemStack = items.ItemStack;
+const random = main.random;
+
+const entityComponent = main.entityComponent;
+
+// ############################# Client only stuff ################################
+pub const Client = struct {
```
IntegratedQuantum: should be lowercase since it has no fields.

src/server/server.zig
```diff
@@ -597,6 +599,7 @@ fn getInitialEntityList(allocator: main.heap.NeverFailingAllocator) []const u8 {
 
 fn update() void { // MARK: update()
 	world.?.update();
+	main.entity.server.deinit();
```
IntegratedQuantum: that looks wrong

src/entityComponent/_template.zig
tillpp: they hold stuff like:

pub var _____Components: _____ = _____;

so i guess it should be lowercase?

IntegratedQuantum: these are declarations. If it doesn't have any fields then it should be formatted as a namespace.

src/server/server.zig
tillpp: good catch, meant .update()

src/entitySystem/_template.zig
tillpp: done

Issue #2670, opened 2026-03-05
Only send limit information about enitites


tillpp:
### Description

To limit the scope of https://github.com/PixelGuys/Cubyz/pull/2652  in the hope of getting entities done before 0.2.0 release
I put some of the less important todo points here: (to be done probably after 0.2.0 release)

- [ ] Only give Information about Entities (and their components) in a range around that player:
 Enity's and component's save() should have parameters, that specify who is requesting the information, 

a nearby player?
 a far away player? 
or is the save() called for storing it in the filesystem?

Depending on who asks for the information, the information gets limited, or not send at all.

- [ ] Entities that are to faw away should get removed on the client side. 
And come into existence if they close again. (Maybe add a Timer until they get removed client side, to stop too frequent adding and removing) 

- [ ] Only send Entitites that moved.
to reduce the burden on the network only Entites( and items ) that moved should be stored

PR #2674, opened 2026-03-06
ECS Part 3 | Entity loads/unloads components

tillpp:
Third step of splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652
(depends on Part 2)

- Entity loads/unloads components
- Base64 Class and functionality that is ~~going to be used in the next parts~~ is being used in this part

- created the beginning of an audience system,
depending on who you are, you get different information send.
.Disk = all information
.playerHimself = everything the player should now about themself (like energy and health)
.playerNearby =  information for players in proximity (  like position )
.playerFaraway = less information: (no information about position, but name and playerIndex)
Note: those examples are not implemented in this PR, they only serve as a demostration what could be doing in the future. https://github.com/PixelGuys/Cubyz/issues/2670

- Each Component has an ID, just like the protocolIDs, but unlike protocolIDs ComponentIDs get stored, so to ensure compability with mods that add their own components or removing or adding components, the entityComponentPalette is introduced.

### DIFF
--- a/src/server/server.zig
+++ b/src/server/server.zig
@@ -183,7 +183,7 @@ pub const User = struct { // MARK: User
 
 		self.worldEditData.deinit();
 
-		self.player.deinit(.server);
+		self.player().deinit(.server);
 
 		self.unloadOldChunk(.{0, 0, 0}, 0);
 		self.conn.deinit();
@@ -755,7 +755,7 @@ pub fn connectInternal(user: *User) void {
 		const zonArray = main.ZonElement.initArray(main.stackAllocator);
 		defer zonArray.deinit(main.stackAllocator);
 
-		const entityZon = user.player.save(main.stackAllocator, .playerNearby);
+		const entityZon = user.player().save(main.stackAllocator, .playerNearby);
 		entityZon.put("name", user.name);
 		entityZon.put("playerIndex", user.playerIndex);
 		zonArray.array.append(entityZon);
@@ -769,7 +769,7 @@ pub fn connectInternal(user: *User) void {
 		const zonArray = main.ZonElement.initArray(main.stackAllocator);
 		defer zonArray.deinit(main.stackAllocator);
 		for (userList) |other| {
-			const entityZon = other.player.save(main.stackAllocator, .playerNearby);
+			const entityZon = other.player().save(main.stackAllocator, .playerNearby);
 			entityZon.put("name", other.name);
 			entityZon.put("playerIndex", other.playerIndex);
 			zonArray.array.append(entityZon);


tillpp:
Rebasing is actually doable (no raytracing)

src/server/Entity.zig
```diff
@@ -15,22 +16,90 @@ health: f32 = 8,
...
 	self.pos = zon.get(Vec3d, "position", .{0, 0, 0});
 	self.vel = zon.get(Vec3d, "velocity", .{0, 0, 0});
 	self.rot = zon.get(Vec3f, "rotation", .{0, 0, 0});
 	self.health = zon.get(f32, "health", self.maxHealth);
 	self.energy = zon.get(f32, "energy", self.maxEnergy);
+	if (zon.getChildOrNull("components")) |components| {
+		const list = main.entity.components;
+		inline for (@typeInfo(list).@"struct".decls) |decl| {
+			if (components.get(?[]const u8, decl.name, null)) |base64| {
+				if (main.utils.fromBase64(main.stackAllocator, base64) catch null) |data| {
```
tillpp: It is better to move this into _template.server.loadFromBase64() function

src/server/Entity.zig
```diff
@@ -15,22 +16,90 @@ health: f32 = 8,
...
 	zon.put("velocity", self.vel);
 	zon.put("rotation", self.rot);
 	zon.put("health", self.health);
 	zon.put("energy", self.energy);
+	const components = ZonElement.initObject(allocator);
+	{
+		const list = main.entity.components;
+		inline for (@typeInfo(list).@"struct".decls) |decl| {
+			if (@field(list, decl.name).server.get(self.id)) |component| {
+				var writer = main.utils.BinaryWriter.init(allocator);
```
tillpp: And this into a _template.server.writeFromBase64() function

src/server/Entity.zig
```diff
@@ -15,22 +16,90 @@ health: f32 = 8,
...
 	return zon;
 }
+pub fn deinit(self: *@This(), comptime side: Side) void {
+	if (self.name) |name| {
+		main.globalAllocator.free(name);
+		self.name = null;
+	}
+	if (side == .ServerSide) {
+		const list = main.entity.components;
+		inline for (@typeInfo(list).@"struct".decls) |decl| {
```
tillpp: have a entity.server.unloadAll(id) function for this

src/server/Entity.zig
tillpp: no actually this would lead to worse/more complex code.
i revert my review.

src/server/Entity.zig
IntegratedQuantum: As discussed, please do binary storage for the entire component list, not just for each individual component.

tillpp: We actually only discussed the componentContent itself.
but sure.

tillpp:
Take your time reviewing it,
I will only return after mid April to Cubyz again.

src/entityComponent/_list.zig
```diff
@@ -1 +1 @@
-pub const _template = @import("_template.zig");
+pub const @"cubyz:_template" = @import("_template.zig");
```
tillpp, line 1: Mods might add their own entityComponents here

src/server/Entity.zig
```diff
@@ -15,22 +15,57 @@ health: f32 = 8,
 maxHealth: f32 = 8,
 energy: f32 = 8,
 maxEnergy: f32 = 8,
-// TODO: Name
+name: ?[]const u8 = null,
```
IntegratedQuantum, line 18: Doesn't need to happen now, but in the future this should probably be a component.

src/server/world.zig
```diff
@@ -945,11 +951,12 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		}
 		const player = &user.player;
 		if (playerData == .null) {
+			player.loadFrom(user.id, playerData, .ServerSide);
 			player.pos = @floatFromInt(self.spawn);
 
 			main.sync.setGamemode(user, self.settings.defaultGamemode);
 		} else {
-			player.loadFrom(playerData.getChild("entity"));
+			player.loadFrom(user.id, playerData.getChild("entity"), .ServerSide);
```
IntegratedQuantum: Should be `serverSide` to follow the naming conventions

src/entity.zig
```diff
@@ -15,6 +15,32 @@ pub const EntityNetworkData = struct {
...
 	rot: Vec3f,
 };
+pub const Side = enum { ClientSide, ServerSide };
+
+// Analogous to Protocols.
+const EntityComponentVTable = struct {
+	server: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) anyerror!void,
+	client: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) anyerror!void,
+};
+var receiveList: [256]?EntityComponentVTable = @splat(null);
```
IntegratedQuantum: We probably shouldn't limit this to 256 elements

src/entity.zig
```diff
@@ -15,6 +15,32 @@ pub const EntityNetworkData = struct {
 	vel: Vec3d,
 	rot: Vec3f,
 };
+pub const Side = enum { ClientSide, ServerSide };
+
+// Analogous to Protocols.
+const EntityComponentVTable = struct {
+	server: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) anyerror!void,
```
IntegratedQuantum: Please use explicit error sets instead of anyerror

src/entity.zig
```diff
@@ -67,4 +93,66 @@ pub const server = struct {
 			@field(systems, decl.name).server.update();
 		}
 	}
+	pub fn componentsToBase64(allocator: main.heap.NeverFailingAllocator, entityID: u32, audience: main.entity.AudienceInfo) main.utils.Base64 {
+		var writer = main.utils.BinaryWriter.init(allocator);
+		defer writer.deinit();
+
+		inline for (@typeInfo(main.entity.components).@"struct".decls) |decl| {
```
IntegratedQuantum, line 111: I'm not sure if this is a good idea, iterating through all components might get rather expensive.
Not something you need to change here, but maybe consider to use a VTable and e.g. a mask for which components are set.

src/entity.zig
```diff
@@ -67,4 +93,66 @@ pub const server = struct {
...
+			if (@field(main.entity.components, decl.name).server.get(entityID)) |component| {
+				var writerComponent = main.utils.BinaryWriter.init(allocator);
```
IntegratedQuantum: local allocations should use the stackAllocator

src/entity.zig
```diff
@@ -67,4 +93,66 @@ pub const server = struct {
 			@field(systems, decl.name).server.update();
 		}
 	}
+	pub fn componentsToBase64(allocator: main.heap.NeverFailingAllocator, entityID: u32, audience: main.entity.AudienceInfo) main.utils.Base64 {
+		var writer = main.utils.BinaryWriter.init(allocator);
```
IntegratedQuantum: stackAllocator

src/utils.zig
```diff
@@ -1769,8 +1769,10 @@ pub const BinaryReader = struct {
 		defer self.remaining = self.remaining[length..];
 		return self.remaining[0..length];
 	}
+	pub fn readSliceWithSize(self: *BinaryReader) ![]const u8 {
+		return try self.readSlice(try self.readVarInt(usize));
+	}
 };
-
```
IntegratedQuantum, line 1773: unintended change

src/entity.zig
```diff
@@ -67,4 +93,66 @@ pub const server = struct {
...
+
+				if (component.save(&writerComponent, audience) == .save) {
+					writer.writeVarInt(u32, @field(main.entity.components, decl.name).entityComponentID);
+					writer.writeVarInt(u32, @field(main.entity.components, decl.name).entityComponentVersion);
+					writer.writeSliceWithSize(writerComponent.data.items);
+				}
+			}
+		}
+		const base64 = main.utils.Base64.toBase64(allocator, writer.data.items);
+		return base64;
```
```suggestion
		return main.utils.Base64.toBase64(allocator, writer.data.items);
```

src/entity.zig
```diff
@@ -67,4 +93,66 @@ pub const server = struct {
...
+	}
+
+	pub fn componentsRemoveAll(id: u32) void {
```
IntegratedQuantum: this grammar is weird, maybe it should be `removeAllComponents`

src/entity.zig
```diff
@@ -67,4 +93,66 @@ pub const server = struct {
...
+		const list = main.entity.components;
+		inline for (@typeInfo(list).@"struct".decls) |decl| {
+			@field(list, decl.name).server.unload(id);
+		}
+	}
+};
+
+pub fn loadComponentsFromBase64(base64Data: []const u8, id: u32, comptime side: Side) void {
+	if (main.utils.fromBase64(main.stackAllocator, base64Data) catch null) |data| {
```
```suggestion
	const data = main.utils.fromBase64(main.stackAllocator, base64Data) catch return;
```

src/entity.zig
```diff
@@ -67,4 +93,66 @@ pub const server = struct {
...
+		defer main.stackAllocator.free(data);
+
+		var reader = main.utils.BinaryReader.init(data);
+		while (reader.remaining.len != 0) {
+			const componentID: u32 = reader.readVarInt(u32) catch @panic("Couldn't read componentID correctly.");
```
IntegratedQuantum: Since this can be called from the network, it should fail gracefully instead of crashing.
Instead maybe log the error, and then return an which can be used to e.g. disconnect the player.

Same for the other errors

src/server/Entity.zig
tillpp: yes i plan that too,
also with energy, maxEnergy, and all the other stuff, https://github.com/PixelGuys/Cubyz/issues/2814

tillpp: having created the Issue, i consider this comment resolved?

src/entity.zig
tillpp: > Not something you need to change here,

ok Issue created:
https://github.com/PixelGuys/Cubyz/issues/2815

src/entity.zig
```diff
@@ -23,14 +23,18 @@ const EntityComponentVTable = struct {
 	server: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) EntityComponentLoadError!void,
 	client: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) EntityComponentLoadError!void,
 };
-var receiveList: [256]?EntityComponentVTable = @splat(null);
+var receiveList: main.ListUnmanaged(?EntityComponentVTable) = .{};
```
IntegratedQuantum: We already know the total number of components at this point, so maybe you can just use a normal slice?

src/server/server.zig
```diff
@@ -245,7 +245,9 @@ pub const User = struct { // MARK: User
 		self.id = freeId;
 		freeId += 1;
 
-		world.?.loadPlayer(self);
+		world.?.loadPlayer(self) catch {
+			std.log.err("Error while loading player data {s} ", .{self.name});
```
IntegratedQuantum: Maybe mention that this will discard the old data.

src/entity.zig
```diff
@@ -125,23 +129,23 @@ pub const server = struct {
 	}
 };
 
-pub fn loadComponentsFromBase64(base64Data: []const u8, id: u32, comptime side: Side) void {
+pub fn loadComponentsFromBase64(base64Data: []const u8, id: u32, comptime side: Side) EntityComponentLoadError!void {
 	const data = main.utils.fromBase64(main.stackAllocator, base64Data) catch return;
```
IntegratedQuantum: Should this also try?

src/entity.zig
```diff
@@ -15,6 +15,41 @@ pub const EntityNetworkData = struct {
 	vel: Vec3d,
 	rot: Vec3f,
 };
+pub const Side = enum { clientSide, serverSide };
```
IntegratedQuantum: We actually already have a Side in sync.zig

src/entity.zig
```diff
@@ -27,28 +27,30 @@ const EntityComponentVTable = struct {
 	server: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) EntityComponentLoadError!void,
 	client: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) EntityComponentLoadError!void,
 };
-var receiveList: main.ListUnmanaged(?EntityComponentVTable) = .{};
+var receiveList: []?EntityComponentVTable = &.{};
```
IntegratedQuantum: Also I'd suggest setting this to undefined to catch incorrect usage.

src/entity.zig
```diff
@@ -27,28 +27,30 @@ const EntityComponentVTable = struct {
...
 			};
 		} else {
 			std.log.err("entity components: Duplicate list id {}.", .{id});
 		}
 	}
+	receiveList = tmpReceiveList.items;
 }
 pub fn deinitComponent() void {
-	receiveList.deinit(main.globalAllocator);
+	receiveList = &.{};
```
IntegratedQuantum: undefined

PR #2675, opened 2026-03-06
ECS Part 4 | Network Protocol for EntityComponent updates

tillpp:
fourth step of splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652
(depends on Part 3)

- adds Protocol 15 to set or reset a component


src/network/protocols.zig
```diff
@@ -960,3 +960,60 @@ pub const blockEntityUpdate = struct { // MARK: blockEntityUpdate
...
 	}
 };
+
+// format: entityID set   componentID binary
+// format: entityID reset componentID
+pub const EntityComponentUpdate = struct { // MARK: EntityComponentUpdate
+	pub const id: u8 = 15;
+	pub const asynchronous = false;
+
+	const ActionType = enum(u8) { set, reset };
```
IntegratedQuantum: For protocols please explicitly assign values to enums, so it's harder to break the order accidentally

src/network/protocols.zig
```diff
@@ -960,3 +960,60 @@ pub const blockEntityUpdate = struct { // MARK: blockEntityUpdate
...
+
+	fn clientReceive(_: *Connection, reader: *utils.BinaryReader) !void {
+		const entityID = try reader.readInt(u32);
+		const actionType: ActionType = try reader.readEnum(ActionType);
+		const componentID = try reader.readSliceWithSize();
```
IntegratedQuantum: Please send the palette index instead of strings. That should also reduce the iteration through all components below.

src/network/protocols.zig
```diff
@@ -960,3 +960,60 @@ pub const blockEntityUpdate = struct { // MARK: blockEntityUpdate
...
+			const list = main.entity.components;
+			inline for (@typeInfo(list).@"struct".decls) |decl| {
+				if (std.mem.eql(u8, decl.name, componentID)) {
+					@field(list, decl.name).client.unload(entityID);
+					break;
+				}
+			}
+		}
+	}
+	pub fn reset(conn: *Connection, entityID: u32, componentID: []const u8) void {
```
IntegratedQuantum: I would suggest to use the same names as the functions that are being called (load unload) for consistency.

src/network/protocols.zig
```diff
@@ -960,3 +960,60 @@ pub const blockEntityUpdate = struct { // MARK: blockEntityUpdate
...
+
+	const ActionType = enum(u8) { set, reset };
+
+	fn clientReceive(_: *Connection, reader: *utils.BinaryReader) !void {
+		const entityID = try reader.readInt(u32);
+		const actionType: ActionType = try reader.readEnum(ActionType);
+		const componentID = try reader.readSliceWithSize();
+
+		if (actionType == .set) {
+			const version = reader.readVarInt(u32) catch std.math.maxInt(u32);
```
IntegratedQuantum: Invalid data should result in a disconnect, since it's likely a corrupted stream or version mismatch, both of which are unrecoverable.

So please just `try` here.

src/network/protocols.zig
```diff
@@ -960,3 +960,60 @@ pub const blockEntityUpdate = struct { // MARK: blockEntityUpdate
...
+			const list = main.entity.components;
+			inline for (@typeInfo(list).@"struct".decls) |decl| {
+				if (std.mem.eql(u8, decl.name, componentID)) {
+					@field(list, decl.name).client.load(entityID, reader, version) catch unreachable;
```
IntegratedQuantum: Woah, hold up, how can you assert that data coming from the internet is valid?
If my computer blows up due to undefined behavior, I'll hold you accountable for it.

IntegratedQuantum (changes requested):

tillpp:
Lot's of changed requested, give me, some easier, some needing more redesign decisions.
You can expect this to take a longer.

src/network/protocols.zig
tillpp: ah forgot to change this after the recent change

tillpp: done

tillpp:
i hate how github makes moving code always seem like having done a lot.
This PR exploded from +40 lines to +100 again 

There is a main.entity.load/unload function that removes duplicate code between the protocol and loadComponentsFromBase64

src/entity.zig
```diff
@@ -21,36 +21,76 @@ pub const EntityComponentLoadError = error{
 	UnreadableID,
 	UnreadableVersion,
 	UnreadableComponentData,
+	unknownComponentID,
```
IntegratedQuantum: In case you didn't notice: All errors start with a capital letter.

src/network/protocols.zig
```diff
@@ -960,3 +960,51 @@ pub const blockEntityUpdate = struct { // MARK: blockEntityUpdate
 		sendServerDataUpdateToClientsInternal(pos, &ch.super, block, blockEntity);
 	}
 };
+
+// format: entityID load   componentID binary
+// format: entityID unload componentID
```
IntegratedQuantum: And that's why I generally don't like comments like this: They do not match what you've written in the code (entityId, componentId, type).

src/network/protocols.zig
```diff
@@ -960,3 +960,51 @@ pub const blockEntityUpdate = struct { // MARK: blockEntityUpdate
...
+	pub const id: u8 = 15;
+	pub const asynchronous = false;
+
+	const ActionType = enum(u8) {
+		unload = 0,
+		load = 1,
+	};
+
+	fn clientReceive(_: *Connection, reader: *utils.BinaryReader) !void {
+		const entityID = try reader.readInt(u32);
```
IntegratedQuantum: Name rule for abbreviations
```suggestion
		const entityId = try reader.readInt(u32);
```

src/network/protocols.zig
```diff
@@ -960,3 +960,51 @@ pub const blockEntityUpdate = struct { // MARK: blockEntityUpdate
...
+		const componentID = try reader.readInt(u32);
```
IntegratedQuantum: Since most of time we won't have that many ids, I think it should be a varint.

src/entity.zig
```diff
@@ -21,36 +21,76 @@ pub const EntityComponentLoadError = error{
...
 };
 // Analogous to Protocols.
 const EntityComponentVTable = struct {
-	server: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) EntityComponentLoadError!void,
-	client: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) EntityComponentLoadError!void,
+	serverLoad: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) EntityComponentLoadError!void,
+	clientLoad: *const fn (id: u32, reader: *main.utils.BinaryReader, version: u32) EntityComponentLoadError!void,
+	serverUnload: *const fn (id: u32) void,
+	clientUnload: *const fn (id: u32) void,
```
IntegratedQuantum: Something I just noticed: Please introduce a type for the entity ID soon (not in this PR though).

Also most ECS say that the id is the entity, I think we should also follow this pattern, I've already done so for block entities in #2835.

IntegratedQuantum: Also a struct for the component id would be good as well. I'd really like some type safety here.

src/entity.zig
```diff
@@ -21,36 +21,76 @@ pub const EntityComponentLoadError = error{
...
 			std.log.err("entity components: Duplicate list id {}.", .{id});
 		}
 	}
-	receiveList = tmpReceiveList.items;
+	componentList = tmpComponentList.items;
+}
+pub fn deinitComponents() void {
+	componentList = undefined;
+}
+pub fn load(comptime side: main.sync.Side, componentID: u32, entityID: u32, componentData: []const u8, componentVersion: u32) EntityComponentLoadError!void {
```
IntegratedQuantum: Within the namespace context `entity.load()` I think this name is confusing
```suggestion
pub fn loadComponent(comptime side: main.sync.Side, componentID: u32, entityID: u32, componentData: []const u8, componentVersion: u32) EntityComponentLoadError!void {
```

src/entity.zig
```diff
@@ -21,36 +21,76 @@ pub const EntityComponentLoadError = error{
...
+			},
+		}
+	} else {
+		std.log.err("unknown Component ID {} ", .{componentID});
+		return error.unknownComponentID;
+	}
 }
-pub fn deinitComponent() void {
-	receiveList = undefined;
+pub fn unload(comptime side: main.sync.Side, componentID: u32, entityID: u32) EntityComponentLoadError!void {
```
```suggestion
pub fn unloadComponent(comptime side: main.sync.Side, componentID: u32, entityID: u32) EntityComponentLoadError!void {
```

IntegratedQuantum (changes requested):

src/network/protocols.zig
tillpp: ah yes changed the order, felt more natural (and inline with other code)

src/entity.zig
tillpp: Done

IntegratedQuantum (approved):

PR #2676, opened 2026-03-06
ECS Part 5 | Load EntityModels from Assets

tillpp:
Fifth step of splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652
(depends on Part 2)

loads EntityModels from their Assets folder ( and their models, but not with lossy compressed as it would have been when using meshes.model )

### DIFF
--- a/src/assets.zig
+++ b/src/assets.zig
@@ -571,10 +571,10 @@ pub fn loadWorldAssets(assetFolder: []const u8, blockPalette: *Palette, itemPale
 	// models (Entities):
 	{
 		var modelIterator = worldAssets.entityModels.iterator();
-		rawModelData = .init(main.worldArena.allocator);
+		rawEntityModelData = .init(main.worldArena.allocator);
 		while (modelIterator.next()) |entry| {
 			std.log.debug("Registering entity model {s}", .{entry.key_ptr.*});
-			registerModelRaw(entry.key_ptr.*, entry.value_ptr.*);
+			registerEntityModelRaw(entry.key_ptr.*, entry.value_ptr.*);
 		}
 	}
 	if (!main.settings.launchConfig.headlessServer) blocks_zig.meshes.registerBlockBreakingAnimation(assetFolder);
@@ -745,7 +745,7 @@ pub fn loadWorldAssets(assetFolder: []const u8, blockPalette: *Palette, itemPale
 	worldAssets.log(.world);
 }
 pub fn registerEntityModelRaw(id: []const u8, data: []const u8) void {
-	rawModelData.put(id, main.models.Model.loadRawModelDataFromObj(main.worldArena, data)) catch unreachable;
+	rawEntityModelData.put(id, main.models.Model.loadRawModelDataFromObj(main.worldArena, data)) catch unreachable;
 }
 pub fn unloadAssets() void { // MARK: unloadAssets()
 	if (!loadedAssets) return;


src/assets.zig
```diff
@@ -106,6 +110,7 @@ pub const Assets = struct {
 			addon.readAllModels(allocator, &self.models);
 			addon.readAllZon(allocator, "particles", true, &self.particles, null);
 			addon.readAllZon(allocator, "world_presets", true, &self.worldPresets, null);
+			addon.readAllZon(allocator, "entity", true, &self.entityModel, null);
```
IntegratedQuantum: Maybe we should also name the folder entityModels? We somehow need to distinguish between entities and their models.
Or just put the zon next to the obj files inside `entity/models`

src/assets.zig
```diff
@@ -36,6 +36,7 @@ pub const Assets = struct {
 	blueprints: BytesHashMap,
 	particles: ZonHashMap,
 	worldPresets: ZonHashMap,
+	entityModel: ZonHashMap,
```
```suggestion
	entityModels: ZonHashMap,
```

tillpp: Done

src/assets.zig
```diff
@@ -523,6 +532,7 @@ pub const Palette = struct { // MARK: Palette
 };
 
 var loadedAssets: bool = false;
+pub var rawModelData: std.StringHashMap([]main.models.QuadInfo) = undefined;
```
IntegratedQuantum: Should be const
```suggestion
pub var rawModelData: std.StringHashMap([]const main.models.QuadInfo) = undefined;
```

src/assets.zig
```diff
@@ -550,12 +560,23 @@ pub fn loadWorldAssets(assetFolder: []const u8, blockPalette: *Palette, itemPale
...
+		var modelIterator = worldAssets.blockModels.iterator();
+		while (modelIterator.next()) |entry| {
+			_ = main.models.registerModel(entry.key_ptr.*, entry.value_ptr.*);
+		}
 	}
 
+	// models (Entities):
+	{
+		var modelIterator = worldAssets.entityModels.iterator();
+		rawModelData = std.StringHashMap([]main.models.QuadInfo).init(main.worldArena.allocator);
```
```suggestion
		rawModelData = .init(main.worldArena.allocator);
```

src/assets.zig
```diff
@@ -550,12 +560,23 @@ pub fn loadWorldAssets(assetFolder: []const u8, blockPalette: *Palette, itemPale
...
+		while (modelIterator.next()) |entry| {
+			std.debug.print("{s}\n", .{entry.key_ptr.*});
```
IntegratedQuantum: Maybe also say what this is. Also use the logger please
```suggestion
			std.log.debug("Registering entity model {s}", .{entry.key_ptr.*});
```

tillpp: oh lol, that was just for debugging, i forgot to remove that,
but we can also keep it to say what we have registered i guess

src/assets.zig
```diff
@@ -523,6 +532,7 @@ pub const Palette = struct { // MARK: Palette
 };
 
 var loadedAssets: bool = false;
+pub var rawModelData: std.StringHashMap([]const main.models.QuadInfo) = undefined;
```
```suggestion
pub var rawEntityModelData: std.StringHashMap([]const main.models.QuadInfo) = undefined;
```

src/assets.zig
```diff
@@ -723,7 +744,9 @@ pub fn loadWorldAssets(assetFolder: []const u8, blockPalette: *Palette, itemPale
 
 	worldAssets.log(.world);
 }
-
+pub fn registerModelRaw(id: []const u8, data: []const u8) void {
```
```suggestion
pub fn registerEntityModelRaw(id: []const u8, data: []const u8) void {
```

IntegratedQuantum (changes requested):

src/assets.zig
tillpp: makes sense

PR #2677, opened 2026-03-06
ECS Part 6 | user.player is now a function

tillpp:
Sixth step of splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652
(no dependency)

instead of user.player we use user.player() now,
for now it still uses an Entity that is attached to the user class ( inner_player ) 
but in future parts, this is going to access the EntitySystem with EntitySystem.get(user.id);

### DIFF
--- a/src/server/server.zig
+++ b/src/server/server.zig
@@ -98,7 +98,7 @@ pub const User = struct { // MARK: User
 	const simulationSize = 2*maxSimulationDistance;
 	const simulationMask = simulationSize - 1;
 	conn: *Connection = undefined,
-	inner_player: Entity = .{},
+	innerPlayer: Entity = .{},
 	timeDifference: utils.TimeDifference = .{},
 	interpolation: utils.GenericInterpolation(3) = undefined,
 	lastTime: i16 = undefined,
@@ -144,7 +144,7 @@ pub const User = struct { // MARK: User
 	permissions: permission.Permissions = undefined,
 
 	pub fn player(self: *User) *Entity {
-		return &self.inner_player;
+		return &self.innerPlayer;
 	}
 
 	pub fn initAndIncreaseRefCount(manager: *ConnectionManager, ipPort: []const u8) !*User {


tillpp:
OK i see the problem:

grafik

tillpp:
when you say "ECS component" i think you mean Entities right?
because there are no components given out in this PR.

regarding Entities, they use VirtualList, which means they will update their address sometimes (swapRemove), 
So the solution in my opinion would be a kind of updateEntity function that updates the address and everything dependent on it (like interpolation) if the swapRemove was used.

Maybe need to rethink this

IntegratedQuantum:
> when you say "ECS component" i think you mean Entities right?

No I mean ECS components, all the data in the current `Entity` struct should move to separate components (one for health, one for hunger, one for physics, one for animation data and rotation, one for interpolation data) in the future.

And more broadly the same applies to all components.

If you get the pointer inside the sparse set, meanwhile another thread maybe removes and entity, or unloads it, or loads a new one, and then change the value, the pointer may be invalid.

mainly I'm asking because I want to know your planned approach to multithreading.

tillpp:
> > when you say "ECS component" i think you mean Entities right?
> 
> No I mean ECS components, all the data in the current `Entity` struct should move to separate components (one for health, one for hunger, one for physics, one for animation data and rotation, one for interpolation data) in the future.
> 
> And more broadly the same applies to all components.
> 
> If you get the pointer inside the sparse set, meanwhile another thread maybe removes and entity, or unloads it, or loads a new one, and then change the value, the pointer may be invalid.
> 
> mainly I'm asking because I want to know your planned approach to multithreading.

Let's discuss this in an Issue:
https://github.com/PixelGuys/Cubyz/issues/2800

tillpp:
Argmaster and you do have some more complex plans it seems like,
I propose, that for the current ECS PRs, im not going to touch multithreading stuff, (since everything currently is done on one Thread anyway)

src/server/server.zig
```diff
@@ -98,7 +98,7 @@ pub const User = struct { // MARK: User
 	const simulationSize = 2*maxSimulationDistance;
 	const simulationMask = simulationSize - 1;
 	conn: *Connection = undefined,
-	player: Entity = .{},
+	inner_player: Entity = .{},
```
```suggestion
	innerPlayer: Entity = .{},
```

tillpp:
> In the future this will also change to a more component oriented approach, right?

you mean https://github.com/PixelGuys/Cubyz/issues/2814 ?

PR #2680, opened 2026-03-06
ECS Part 7 | entity Model

tillpp:
Sixth step of splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652
(requires Part 1 to 6, except 4)

Adds EntityModels


tillpp:
Have to combine it with @RanPix 's https://github.com/PixelGuys/Cubyz/pull/2733 version of entityModel.zig
Maybe i wait until you have yours merged? (Hopefully soon)
Or I merge my first

tillpp:
My and RanPix code is so interconnected, that we need some time to sort some things out,
**So this PR is on Pause - no review pls**

tillpp:
Unless @RanPix has something against this , this is now ready for merging

src/assets.zig
RanPix, line 134: is it intended to not log the newly added asset loaders? im asking because block models are not listed there as well

tillpp: We print the new EntityModels in the assets function,
Blocks also get printed too, which is probably why it isn't mentioned here.

I guess you have no other concern here, regarding the EntityModels, so this is ready for review now.

tillpp:
This is ready for review now.

src/assets.zig
```diff
@@ -119,11 +115,10 @@ pub const Assets = struct {
 			addon.readAllZon(allocator, "recipes", false, &self.recipes, null);
 			addon.readAllZon(allocator, "sbb", true, &self.structureBuildingBlocks, null);
 			addon.readAllBlueprints(allocator, "sbb", &self.blueprints);
-			addon.readAllModels(allocator, "models", &self.blockModels);
-			addon.readAllModels(allocator, "entityModels/models", &self.entityModels);
+			addon.readAllModels(allocator, "models", ".obj", &self.blockModels);
 			addon.readAllZon(allocator, "particles", true, &self.particles, null);
 			addon.readAllZon(allocator, "world_presets", true, &self.worldPresets, null);
-			addon.readAllZon(allocator, "entityModels", true, &self.entityModelDescriptions, null);
+			addon.readAllZon(allocator, "entityModels/descriptions", true, &self.entityModelDescriptions, null);
```
IntegratedQuantum: What do you think about putting the description right next to the model files, so same folder?

src/assets.zig
```diff
@@ -785,6 +779,30 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
 	}
 }
 
+pub fn readAsset(allocator: NeverFailingAllocator, assetFolder: []const u8, subPath: []const u8, id: []const u8, fileEnding: []const u8) ?[]const u8 {
+	var split = std.mem.splitScalar(u8, id, ':');
+	const mod = split.first();
+	const name = split.next() orelse return null;
```
IntegratedQuantum: I'm pretty sure we can assert that this doesn't happen → `orelse unreachable`.

src/assets.zig
```diff
@@ -785,6 +779,30 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
...
+
+	var path = std.fmt.allocPrint(main.stackAllocator.allocator, "{s}/{s}/{s}/{s}{s}", .{assetFolder, mod, subPath, name, fileEnding}) catch &.{};
```
IntegratedQuantum: Allocation with these allocators can't fail
```suggestion
	var path = std.fmt.allocPrint(main.stackAllocator.allocator, "{s}/{s}/{s}/{s}{s}", .{assetFolder, mod, subPath, name, fileEnding}) catch unreachable;
```

src/assets.zig
```diff
@@ -785,6 +779,30 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
...
+	defer main.stackAllocator.free(path);
+	std.fs.cwd().access(path, .{}) catch {
+		main.stackAllocator.free(path);
+		path = std.fmt.allocPrint(main.stackAllocator.allocator, "assets/{s}/{s}/{s}{s}", .{mod, subPath, name, fileEnding}) catch &.{};
```
```suggestion
		path = std.fmt.allocPrint(main.stackAllocator.allocator, "assets/{s}/{s}/{s}{s}", .{mod, subPath, name, fileEnding}) catch unreachable;
```

src/assets.zig
```diff
@@ -785,6 +779,30 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
...
+	};
+
+	const string = std.fs.cwd().readFileAlloc(path, allocator.allocator, .unlimited) catch |err| {
+		std.log.err("Could not open {s}/{s}{s}: {s}", .{subPath, name, fileEnding, @errorName(err)});
+		return null;
```
IntegratedQuantum: I'd prefer to return the error, instead of optional, since this is an exceptional state.

src/assets.zig
```diff
@@ -785,6 +779,30 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
...
+	const name = split.next() orelse return null;
+
+	var path = std.fmt.allocPrint(main.stackAllocator.allocator, "{s}/{s}/{s}/{s}{s}", .{assetFolder, mod, subPath, name, fileEnding}) catch &.{};
+	defer main.stackAllocator.free(path);
+	std.fs.cwd().access(path, .{}) catch {
+		main.stackAllocator.free(path);
+		path = std.fmt.allocPrint(main.stackAllocator.allocator, "assets/{s}/{s}/{s}{s}", .{mod, subPath, name, fileEnding}) catch &.{};
+	};
+
+	const string = std.fs.cwd().readFileAlloc(path, allocator.allocator, .unlimited) catch |err| {
```
IntegratedQuantum: I thought you wanted to read binary models as well?
```suggestion
	const data = std.fs.cwd().readFileAlloc(path, allocator.allocator, .unlimited) catch |err| {
```

src/assets.zig
```diff
@@ -785,6 +779,30 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
...
+}
+
 pub fn worldPresets() *const Assets.ZonHashMap {
 	return &common.worldPresets;
 }
+
+// TODO: Tempoary, will be removed in future ECS parts.
+pub fn entityModelDescriptions() *const Assets.ZonHashMap {
+	return &common.entityModelDescriptions;
+}
```
IntegratedQuantum: unused function, no point in adding it. It's wrong/insufficient anyways.

src/entityModel.zig
```diff
@@ -1,15 +1,34 @@
...
+const Vec3d = vec.Vec3d;
+const Vec3f = vec.Vec3f;
+const Vec4f = vec.Vec4f;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+const BinaryReader = main.utils.BinaryReader;
 
 pub const EntityModel = struct {
-	vao: graphics.VertexArray = undefined,
+	height: f32,
```
IntegratedQuantum, line 22: In what context do we need the height, but none of the other dimensions?

src/entityModel.zig
```diff
@@ -1,15 +1,34 @@
...
+	texturePath: []const u8,
+	id: []const u8,
+
+	vao: ?graphics.VertexArray = null,
 	indexCount: c_int,
-	texture: main.graphics.Texture,
+	defaultTexture: ?main.graphics.Texture,
```
IntegratedQuantum, line 28: Is this default to account for skins?

src/entityModel.zig
```diff
@@ -1,15 +1,34 @@
...
+	modelFile: []const u8,
```
IntegratedQuantum: The file is big, store the path instead and read it on load to temporary memory.

Actually you already have the id, so you just need to make the assets store the per-world asset folder somewhere.

src/entityModel.zig
```diff
@@ -33,22 +52,46 @@ pub const EntityModel = struct {
...
+				self.texturePath = std.fmt.allocPrint(main.globalAllocator.allocator, "{s}/{s}/entityModels/textures/{s}", .{assetFolder, mod, texture}) catch &.{};
+				std.fs.cwd().access(self.texturePath, .{}) catch {
+					main.globalAllocator.free(self.texturePath);
+					self.texturePath = std.fmt.allocPrint(main.globalAllocator.allocator, "assets/{s}/entityModels/textures/{s}", .{mod, texture}) catch &.{};
+				};
+			}
+		}
+		return self;
+	}
+	fn generateGraphics(self: *EntityModel) void {
```
IntegratedQuantum: Already prepared for procedural *generation*? :​P

```suggestion
	fn loadModelAndTexture(self: *EntityModel) void {
```

src/entityModel.zig
```diff
@@ -33,22 +52,46 @@ pub const EntityModel = struct {
...
+		self.defaultTexture = main.graphics.Texture.initFromFile(self.texturePath);
+
+		defer {
+			// We don't need the modelFile in memory anymore, so why waste memory?
+			main.globalAllocator.free(self.modelFile);
+			self.modelFile = &.{};
+		}
```
IntegratedQuantum, line 90: Not needed, see above.

src/entityModel.zig
```diff
@@ -63,20 +106,75 @@ pub const EntityModel = struct {
...
-			.vao = .init(EntityVertex, vertices, indices),
-			.texture = texture,
-			.indexCount = @intCast(indices.len),
-		};
+		self.vao = .init(Vertex, vertices, indices);
+		self.indexCount = @intCast(indices.len);
+	}
+	pub fn bind(self: *EntityModel) void {
+		if (self.vao == null) {
+			self.generateGraphics();
```
IntegratedQuantum: No lazy loading please.
It makes it hard to argue about the loading process and can cause needless lag spikes.
(This is especially true once we use Vulkan, where this would need to be synced up with the rendering and allocating video memory)

src/entityModel.zig
```diff
@@ -33,22 +52,46 @@ pub const EntityModel = struct {
...
-		const modelFile = main.files.cwd().read(main.stackAllocator, modelPath) catch |err| blk: {
-			std.log.err("Error while reading player model from path {s}: {s}", .{modelPath, @errorName(err)});
-			break :blk &.{};
-		};
-		defer main.stackAllocator.free(modelFile);
-		const quadInfos = main.models.Model.loadRawModelDataFromObj(main.stackAllocator, modelFile);
-		defer main.stackAllocator.free(quadInfos);
+	pub fn init(assetFolder: []const u8, id: []const u8, zon: ZonElement) EntityModel {
+		var self: EntityModel = undefined;
+		self.id = main.globalAllocator.dupe(u8, id);
```
IntegratedQuantum: Use worldArena for things that stay in memory for the lifetime of the world.

src/entityModel.zig
```diff
@@ -63,20 +106,75 @@ pub const EntityModel = struct {
...
 
-	pub fn deinit(self: EntityModel) void {
-		self.vao.deinit();
-		self.texture.deinit();
+pub const EntityModelIndex = struct {
+	index: u32,
+	pub fn get(self: EntityModelIndex) *EntityModel {
+		if (entityModels.items.len > self.index)
+			return &entityModels.items[self.index];
+		// should always exist because of firstEntry in entityModelPalette
```
IntegratedQuantum: There is no variable named `firstEntry`

src/entityModel.zig
```diff
@@ -63,20 +106,75 @@ pub const EntityModel = struct {
...
 	}
 };
+
+pub var reverseIndices: std.StringHashMapUnmanaged(EntityModelIndex) = .{};
+pub var entityModels: main.ListUnmanaged(EntityModel) = .{};
+
+pub fn register(assetFolder: []const u8, id: []const u8, zon: ZonElement) usize {
+	const index = entityModels.items.len;
+	const entityModel = entityModels.addOne(main.worldArena);
+	entityModel.* = EntityModel.init(assetFolder, id, zon);
```
IntegratedQuantum: Did you know that there is an append function?

src/entityModel.zig
```diff
@@ -63,20 +106,75 @@ pub const EntityModel = struct {
...
+pub fn hasRegistered(id: []const u8) bool {
+	return reverseIndices.contains(id);
+}
+
+pub fn getTypeById(id: []const u8) EntityModelIndex {
+	if (reverseIndices.get(id)) |result| {
+		return result;
+	} else {
+		std.log.err("Couldn't find entityModel {s}. Replacing it with cubyz:missing ...", .{id});
+		return EntityModelIndex{.index = 0};
```
IntegratedQuantum: This was a poor design choice in the place you copied it from, please use optional return type instead.

src/entityModel.zig
```diff
@@ -63,20 +106,75 @@ pub const EntityModel = struct {
...
+}
+pub fn reset() void {
+	for (entityModels.items) |*model| {
+		model.deinit();
+	}
+	entityModels = .{};
+	reverseIndices = .{};
+}
+
+pub fn hasRegistered(id: []const u8) bool {
```
IntegratedQuantum: Note needed, can use `getById() != null`

src/entityModel.zig
```diff
@@ -63,20 +106,75 @@ pub const EntityModel = struct {
...
+pub fn getTypeById(id: []const u8) EntityModelIndex {
+	if (reverseIndices.get(id)) |result| {
+		return result;
+	} else {
+		std.log.err("Couldn't find entityModel {s}. Replacing it with cubyz:missing ...", .{id});
+		return EntityModelIndex{.index = 0};
+	}
+}
+
+pub fn getTypeByIdOrNull(id: []const u8) ?EntityModelIndex {
```
IntegratedQuantum: duplicate

src/entityModel.zig
```diff
@@ -63,20 +106,75 @@ pub const EntityModel = struct {
...
+	}
+	entityModels = .{};
+	reverseIndices = .{};
+}
+
+pub fn hasRegistered(id: []const u8) bool {
+	return reverseIndices.contains(id);
+}
+
+pub fn getTypeById(id: []const u8) EntityModelIndex {
```
IntegratedQuantum: Isn't this a model?
```suggestion
pub fn getModelById(id: []const u8) EntityModelIndex {
```

IntegratedQuantum (changes requested):

src/entityModel.zig
tillpp: height determines Text position.
Other dimensions aren't required currently, can be added in the future if desired.

src/entityModel.zig
tillpp: Yes different skins for mobs and players.

src/entityModel.zig
tillpp: > No lazy loading please.

This is inspired by the way you handle items (being lazily loaded)
you lazy load items because the also get to be loaded on the server, where there isnt any graphics.

src/entityModel.zig
tillpp:
```zig
self.entityModelPalette = try loadPalette(arena, path, "entity_model_palette", "cubyz:missing");
errdefer self.entityModelPalette.deinit();
```

will be added in the future PR,
Maybe i should already do it here.
The 0 entityModel being the cubyz:missing

src/entityModel.zig
tillpp: grafik


omni culpa in quantum

src/entityModel.zig
tillpp: I know that you had ideas for customization with procedural models like in the crafting table.
But It will take a **long time** until then.

Giving player the ability to modify their textures now would make them happy. (right, @ikabod-kee and everyone else ? )

And in 420 years, when we finally to add procedural modeling, we can exchange this with the procedural modeling thingy.

src/assets.zig
tillpp: > unused function, no point in adding it.

grafik

entity_manager.zig line 47

tillpp: grafik

why is the top function ok, but the lower one wrong/insufficient?

src/entityModel.zig
IntegratedQuantum: item textures cannot be loaded upfront because we cannot enumerate all the tools. (but base items could yes, and they probably should)

src/assets.zig
IntegratedQuantum: because world presets cannot be part of per-world assets, as you cannot create a world after you are already in a world.

src/entityModel.zig
IntegratedQuantum: Yes, please add it here then.

src/assets.zig
tillpp: I can do this, and it is probably the better solution, since this way it is easier to remember to add a description file to every model.
But on the other hand i kind of feel like it is uglier:
grafik
(beautiful:)
grafik
If you insist i can change it, otherwise i would keep it like this.

src/entityModel.zig
tillpp: In future PRs this will stay in lifetimes of the world.
but currently exist outside the game aswell because of entity_manager.zig
(snale is being initialized at the beginning and end of the game)

I dont want to make the PR too big,
so **ModelComponent** (ECS8 #2681  **123 lines changes (only 1 new file)**) 
and **Modelrenderer** (ECS9 #2682 **151 lines changes (only 1 new file)**) 
are in separate PRs.
But if you think they are small enough, I can also already implement them here, and avoid those weird transition fixes.

src/entityModel.zig
tillpp: not duplicate but ok. (notice how one always returns a fallback just like cubyz:air in blocks)
but i can remove this anyway, can do the fallback somewhere else

src/entityModel.zig
tillpp: forgot to rename this, if you remember EntityModel used to be called EntityType a long time ago ( ECS1 ). renamed.

src/entityModel.zig
tillpp: Done.

src/entityModel.zig
tillpp: Nevermind got it without ECS8 and ECS9

src/assets.zig
IntegratedQuantum: @careeoki @ikabod-kee  you are the artists, what would you prefer?

src/assets.zig
```diff
@@ -537,12 +536,14 @@ pub const Palette = struct { // MARK: Palette
 };
 
 var loadedAssets: bool = false;
-pub var rawEntityModelData: std.StringHashMap([]const main.models.QuadInfo) = undefined;
+pub var folder: []const u8 = undefined;
```
IntegratedQuantum: What folder is this? Is it the per world asset folder? Is it the data of the new 'folder' entity that attacks the player by attemptng to fold them?
Please use a more descriptive name.

src/assets.zig
```diff
@@ -794,6 +809,25 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
 	}
 }
 
+pub fn readAsset(allocator: NeverFailingAllocator, assetFolder: []const u8, subPath: []const u8, id: []const u8, fileEnding: []const u8) ![]const u8 {
+	var split = std.mem.splitScalar(u8, id, ':');
+	const mod = split.first();
+	const name = split.next() orelse unreachable;
+
+	var path = try std.fmt.allocPrint(main.stackAllocator.allocator, "{s}/{s}/{s}/{s}{s}", .{assetFolder, mod, subPath, name, fileEnding});
```
IntegratedQuantum: allocations can't fail
```suggestion
	var path = std.fmt.allocPrint(main.stackAllocator.allocator, "{s}/{s}/{s}/{s}{s}", .{assetFolder, mod, subPath, name, fileEnding}) catch unreachable;
```

src/assets.zig
```diff
@@ -794,6 +809,25 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
...
+	defer main.stackAllocator.free(path);
+	if (!main.files.cwd().hasFile(path)) {
+		main.stackAllocator.free(path);
+		path = try std.fmt.allocPrint(main.stackAllocator.allocator, "assets/{s}/{s}/{s}{s}", .{mod, subPath, name, fileEnding});
```
```suggestion
		path = std.fmt.allocPrint(main.stackAllocator.allocator, "assets/{s}/{s}/{s}{s}", .{mod, subPath, name, fileEnding}) catch unreachable;
```

src/assets.zig
Argmaster: I assume by description you mean metadata; I insist on keeping them next to each other so they are directly next to each other with O(1) human search time in UI instead of O(N) when they are in different directories.

Argmaster: Things should be convenient first, pretty only if possible.

tillpp: > I assume by description you mean metadata; I insist on keeping them next to each other so they are directly next to each other with O(1) human search time in UI instead of O(N) when they are in different directories.
> Things should be convenient first, pretty only if possible.

As i said, i have no deep attachment to any of the two solutions, (i dont care)
If it help merge this PR faster, i can change this

src/entityModel.zig
```diff
@@ -33,22 +52,47 @@ pub const EntityModel = struct {
...
+			}
+		}
+		return self;
+	}
+
+	fn loadModelAndTexture(self: *EntityModel) !void {
+		self.deinitModelAndTexture();
+
+		const fileEnding = ".obj";
+		const file = main.assets.readAsset(main.globalAllocator, main.assets.folder, "entityModels/models", self.id, fileEnding) catch main.assets.readAsset(main.stackAllocator, main.assets.folder, "entityModels/models", "cubyz:missing", fileEnding) catch unreachable;
```
IntegratedQuantum: You are using mismatched allocators in both paths.
```suggestion
		const file = main.assets.readAsset(main.stackAllocator, main.assets.folder, "entityModels/models", self.id, fileEnding) catch main.assets.readAsset(main.stackAllocator, main.assets.folder, "entityModels/models", "cubyz:missing", fileEnding) catch unreachable;
```

src/entityModel.zig
```diff
@@ -33,22 +52,47 @@ pub const EntityModel = struct {
...
+		defer main.globalAllocator.free(file);
```
```suggestion
		defer main.stackAllocator.free(file);
```

src/entityModel.zig
```diff
@@ -33,22 +52,47 @@ pub const EntityModel = struct {
...
+					main.worldArena.free(self.texturePath);
+					self.texturePath = std.fmt.allocPrint(main.worldArena.allocator, "assets/{s}/entityModels/textures/{s}", .{mod, texture}) catch &.{};
+				};
+			}
+		}
+		return self;
+	}
+
+	fn loadModelAndTexture(self: *EntityModel) !void {
+		self.deinitModelAndTexture();
```
IntegratedQuantum: When would this happen that we load the model twice?

src/assets.zig
Argmaster: Okay, I would suggest a third one that makes it more pretty, keeps it convenient to access and makes the file names describe what is inside 

```

entityModels/
    cubert/
        model.glb
        texture.png
        metadata.zig.zon

```

src/entityModel.zig
```diff
@@ -33,22 +52,47 @@ pub const EntityModel = struct {
...
+
+		const fileEnding = ".obj";
+		const file = main.assets.readAsset(main.globalAllocator, main.assets.folder, "entityModels/models", self.id, fileEnding) catch main.assets.readAsset(main.stackAllocator, main.assets.folder, "entityModels/models", "cubyz:missing", fileEnding) catch unreachable;
+		defer main.globalAllocator.free(file);
+
+		self.defaultTexture = main.graphics.Texture.initFromFile(self.texturePath);
```
IntegratedQuantum, line 105: Also maybe not put the texture code in between loading the model file and reading the model file.

IntegratedQuantum: Maybe it should just be at the end where you set the other attributes?

src/entityModel.zig
```diff
@@ -63,20 +107,69 @@ pub const EntityModel = struct {
...
+		if (self.vao) |vao| {
+			vao.deinit();
+		}
+	}
+	pub fn bind(self: *EntityModel) void {
+		self.vao.?.bind();
+		self.defaultTexture.?.bindTo(0);
 	}
+	pub fn deinit(self: *EntityModel) void {
+		self.deinitModelAndTexture();
```
IntegratedQuantum: Why do you have two publicly accessible functions that do the same thing?

IntegratedQuantum: also deinit should be after init (you should be used to that from C++)

src/entityModel.zig
```diff
@@ -63,20 +107,69 @@ pub const EntityModel = struct {
...
+	index: u32,
+	pub fn get(self: EntityModelIndex) *EntityModel {
+		if (entityModels.items.len > self.index) {
+			const rv = &entityModels.items[self.index];
+			if (rv.isLoaded)
+				return rv;
+		}
+		// should always exist because of firstEntry in entityModelPalette
+		std.debug.assert(entityModels.items.len > 0);
+		return &entityModels.items[0];
```
IntegratedQuantum: When would we actually reach this code? I think this should be asserted here and the incorrect data (index < len) should be handled immediately where it's created.

src/entityModel.zig
```diff
@@ -63,20 +107,69 @@ pub const EntityModel = struct {
...
+	reverseIndices = .{};
+}
+
+pub fn getById(id: []const u8) ?EntityModelIndex {
+	if (reverseIndices.get(id)) |result| {
+		return result;
+	}
+	return null;
+}
+pub fn loadModelAndTexture() void {
```
```suggestion
pub fn loadModelsAndTextures() void {
```

src/assets.zig
tillpp: wouldn't this make _default.zig not work? or do _default.zig also apply to all sub folders?

IntegratedQuantum (changes requested):

src/assets.zig
tillpp: you have to read it with the filename (the way you access it)
assets.folder

tillpp: Additionaly, the name "assetsFolder" was being shadowed as a parameter

IntegratedQuantum: still ambiguous, since there are multiple asset folders

src/assets.zig
```diff
@@ -794,6 +809,25 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
 	}
 }
 
+pub fn readAsset(allocator: NeverFailingAllocator, assetFolder: []const u8, subPath: []const u8, id: []const u8, fileEnding: []const u8) ![]const u8 {
```
IntegratedQuantum: We know the per world asset folder locally, you can also make it private.
```suggestion
pub fn readAsset(allocator: NeverFailingAllocator, subPath: []const u8, id: []const u8, fileEnding: []const u8) ![]const u8 {
```

src/entityModel.zig
tillpp: good catch

src/assets.zig
Argmaster: Uh depends how you alter the implementation of _default, you would have to treat leaf folders as if they were not there, as a single file. Recursively applying defaults would be cool to tho (no idea why would that be useful, but sure, multi level inheritance)

IntegratedQuantum: Will we even have enough attributes for defaults to make sense here?

src/entityModel.zig
tillpp: legacy, an solution me MNHS and me came up with was to load the missing model instead in case the assets were faulty.
We scraped that idea, now the entire entityModel counts as not Loaded if it failed.
TL;DR i   remove this

src/entityModel.zig
tillpp: i put it in front, more fitting, since the setting of the attributes at the end are still part of the model loading

src/entityModel.zig
tillpp: > also deinit should be after init (you should be used to that from C++)

deinit in case of old model already exisiting.
This is removed anyway. i.e. doesnt matter anymore.

src/assets.zig
Argmaster: Depends. How do we express texture variants of a model? Surely we would want some variety across moffalos?

Edit:

Probably irrelevant, we won't really duplicate the metadata files. I don't think there will be any important use case, maybe beyond grouping models with different coordinate system orientation, if we support those.

src/entityModel.zig
tillpp: grafik
grafik

src/assets.zig
Argmaster: Eh maybe we should do the simple thing of just doing it the old way (ie. Nested metadata and textures folders like we do for blocks and items) to merge it fast, rethink and refactor later, preferably applying changes to all relevant assets? We will create obvious inconsistency if we change this now.

src/entityModel.zig
IntegratedQuantum: Different functions, different names

tillpp: it's the same function

tillpp: oh i see what you mean

src/assets.zig
tillpp: > Eh maybe we should do the simple thing of just doing it the old way (ie. Nested metadata and textures folders like we do for blocks and items) to merge it fast, rethink and refactor later, preferably applying changes to all relevant assets? We will create obvious inconsistency if we change this now.

didnt fully understand you, did you say, we let it be like this and refactor a better solution in the future?

tillpp: grafik
So this?

tillpp: ok descriptions folder movement removed, 
This PR now doesnt touch the location of the files in the assets.
Thus this can be resolved.

if someone is unhappy if they it is right now, you can create a separate PR or Issue for that

Argmaster: Yeah

tillpp: > Though I think there is possibly a use-case for reusing the same model file for different model descriptions.

done

assets/cubyz/entityModels/cubert.zig.zon
```diff
@@ -1,5 +1,5 @@
 .{
 	.model = "cubyz:cubert",
-	.texture = "cubert.png",
```
IntegratedQuantum, line 3: Don't need to do this now, but these should probably also use ids instead of file paths, as is done for block textures

src/assets.zig
```diff
@@ -573,15 +577,27 @@ pub fn loadWorldAssets(assetFolder: []const u8, blockPalette: *Palette, itemPale
 		}
 	}
 
-	// models (Entities):
+	// EntityModels:
 	{
-		var modelIterator = worldAssets.entityModels.iterator();
-		rawEntityModelData = .init(main.worldArena.allocator);
-		while (modelIterator.next()) |entry| {
+		// First blocks from the palette to enforce ID values.
```
```suggestion
		// First models from the palette to enforce ID values.
```

src/assets.zig
```diff
@@ -573,15 +577,27 @@ pub fn loadWorldAssets(assetFolder: []const u8, blockPalette: *Palette, itemPale
...
+		for (entityModelPalette.palette.items) |entityModelId| {
+			std.log.debug("Registering entity model {s}", .{entityModelId});
+			_ = main.entityModel.register(assetFolder, entityModelId, worldAssets.entityModelDescriptions.get(entityModelId) orelse .null);
+		}
+		// Then all the blocks that were missing in palette but are present in the game.
```
```suggestion
		// Then all the models that were missing in palette but are present in the game.
```

src/entityModel.zig
```diff
@@ -33,22 +52,51 @@ pub const EntityModel = struct {
...
+		self.indexCount = 0;
 
-		const vertices = main.stackAllocator.alloc(EntityVertex, quadInfos.len*4);
+		// get TexturePath
+		{
+			self.texturePath = &.{};
+			var split = std.mem.splitScalar(u8, entityModelId, ':');
+			const mod = split.first();
+			if (zon.get(?[]const u8, "defaultTexture", null)) |texture| {
+				self.texturePath = std.fmt.allocPrint(main.worldArena.allocator, "{s}/{s}/entityModels/textures/{s}", .{assetFolder, mod, texture}) catch &.{};
```
```suggestion
				self.texturePath = std.fmt.allocPrint(main.worldArena.allocator, "{s}/{s}/entityModels/textures/{s}", .{assetFolder, mod, texture}) catch unreachable;
```

src/entityModel.zig
```diff
@@ -33,22 +52,51 @@ pub const EntityModel = struct {
...
+				main.files.cubyzDir().dir.access(self.texturePath, .{}) catch {
+					main.worldArena.free(self.texturePath);
+					self.texturePath = std.fmt.allocPrint(main.worldArena.allocator, "assets/{s}/entityModels/textures/{s}", .{mod, texture}) catch &.{};
```
```suggestion
					self.texturePath = std.fmt.allocPrint(main.worldArena.allocator, "assets/{s}/entityModels/textures/{s}", .{mod, texture}) catch unreachable;
```

src/entityModel.zig
```diff
@@ -63,20 +111,65 @@ pub const EntityModel = struct {
...
 	}
-
-	pub fn bind(self: EntityModel) void {
-		self.vao.bind();
-		self.texture.bindTo(0);
+	pub fn bind(self: *EntityModel) void {
+		self.vao.?.bind();
+		self.defaultTexture.?.bindTo(0);
 	}
+	pub fn deinit(self: *EntityModel) void {
```
IntegratedQuantum: As said before, deinit should be after init
https://github.com/PixelGuys/Cubyz/pull/2680#discussion_r3067918482

src/entityModel.zig
```diff
@@ -63,20 +111,65 @@ pub const EntityModel = struct {
...
-	pub fn deinit(self: EntityModel) void {
-		self.vao.deinit();
-		self.texture.deinit();
+pub const EntityModelIndex = struct {
+	index: u32,
+	pub fn get(self: EntityModelIndex) *EntityModel {
+		std.debug.assert(entityModels.items.len > self.index);
+		const rv = &entityModels.items[self.index];
+		if (rv.isLoaded)
+			return rv;
```
IntegratedQuantum: I still think this should not be allowed. These index types should not be doing more than directly accessing the thing.

If something couldn't be loaded, then it should be given the fallback texture+model during load time, and not every frame during rendering.

IntegratedQuantum (changes requested):

src/entityModel.zig
IntegratedQuantum: Since this PR is otherwise mostly done, I could also be convinced to turn this into an issue

src/assets.zig
tillpp: now you know where i copied that code from

assets/cubyz/entityModels/cubert.zig.zon
tillpp: you mean like cubyz:cubert.png ?

src/entityModel.zig
tillpp: OH you meant **DIRECTLY** after

IntegratedQuantum: yes

src/entityModel.zig
tillpp: done

tillpp: (no issue needed)

assets/cubyz/entityModels/cubert.zig.zon
IntegratedQuantum: Just cubyz:cubert

tillpp: the blocks also use .png because you cant know what file format the use

tillpp: oh wait, was that changed?

IntegratedQuantum: yes it was changed

IntegratedQuantum: I got annoyed by the funky x.png_emission.png names.

src/entityModel.zig
```diff
@@ -169,6 +169,9 @@ pub fn getById(id: []const u8) ?EntityModelIndex {
 	if (reverseIndices.get(id)) |result| {
 		return result;
 	}
+	if (reverseIndices.get("cubyz:missing")) |result| {
+		return result;
+	}
```
IntegratedQuantum, line 178: Defaulting to missing should be handled at the callsite. (that's the point for returning optional)

IntegratedQuantum (changes requested):

src/entityModel.zig
tillpp: will lead to a lot of duplicate code

IntegratedQuantum: Then add a helper function for it `orelse {std.log.err("bla", .{}); entityModel.default()}`

So it's only one more line among error handling that you are already supposed to be doing at the callsite.

tillpp: done

IntegratedQuantum (approved):

PR #2681, opened 2026-03-06
ECS Part 8 | ModelComponent

tillpp:
Eights step of splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652
(requires Part 7)

Adds the Entity components for models

Assigns you a random entityModel from {cubert,snale}, but you can't actually see it until ECS9
default player models can be edited in the world settings

Beginning of using main.entity.Entity instead of u32
https://github.com/PixelGuys/Cubyz/issues/2878


tillpp:
I am so confused, why does it compile on my pc locally but not on github

src/entityComponent/_template.zig
```diff
@@ -34,13 +34,13 @@ pub const entityComponentVersion = 0;
...
-	pub fn load(entityId: u32, reader: *utils.BinaryReader, version: u32) main.entity.EntityComponentLoadError!void {
-		_ = entityId;
+	pub fn load(entity: main.entity.Entity, reader: *utils.BinaryReader, version: u32) main.entity.EntityComponentLoadError!void {
+		_ = entity;
 		_ = reader;
 		_ = version;
 	}
-	pub fn unload(entityId: u32) void {
-		_ = entityId;
+	pub fn unload(entity: main.entity.Entity) void {
```
IntegratedQuantum: Could you do this in a separate PR please?

IntegratedQuantum (changes requested):

src/entityComponent/_template.zig
tillpp: If i didnt had done this, your review would be "can you enforce the main.entity.Entity" everywhere else too :joy:

IntegratedQuantum: yes, the right course of action would have been to make the PR to change main.entity.Entity first :P

tillpp: Dont be ridiculous! I can't do an ECS7.5!

src/entity.zig
```diff
@@ -23,8 +23,12 @@ pub const EntityComponentLoadError = error{
 	UnreadableComponentData,
 	UnknownComponentId,
 };
-// Analogous to Protocols.
+pub const Entity = enum(u32) {
+	noValue = std.math.maxInt(u32),
+	_,
+};
 pub const EntityComponentId = u32;
+// Analogous to Protocols.
```
IntegratedQuantum: Does this really make sense? I don't think this comment has any value, since it is likely that the implementations will diverge over time anyways (in fact they already have diverged a bit).

src/entityComponent/_list.zig
```diff
@@ -1 +1,2 @@
+pub const @"cubyz:model" = @import("model.zig");
 pub const @"cubyz:_template" = @import("_template.zig");
```
IntegratedQuantum, line 1: I would put the template first, with an empty line after, to emphasize that this is special and to make it easier to find.

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+const vec = main.vec;
+const Mat4f = vec.Mat4f;
+const Vec3d = vec.Vec3d;
+const Vec3f = vec.Vec3f;
+const Vec4f = vec.Vec4f;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+
+const BinaryReader = main.utils.BinaryReader;
+
+pub var entityComponentID: u32 = undefined;
```
IntegratedQuantum:
```suggestion
pub var entityComponentID: main.entity.EntityComponentId = undefined;
```

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+pub const entityComponentVersion = 0;
+
+// ############################# Client only stuff ################################
+pub const client = struct {
+	const RenderComponent = struct {
```
IntegratedQuantum: But it's a model component?

Also should we even name these? The name is already implicit from the filename.

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+		entity: u32, // entity
```
IntegratedQuantum: Not necessary, the SparseSet already has this data.

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+
+	pub fn init() void {
+		renderComponents = .{};
+	}
+	pub fn deinit() void {
+		renderComponents.deinit(main.globalAllocator);
+	}
+	pub fn clear() void {
+		renderComponents.deinit(main.globalAllocator);
+		renderComponents = .{};
```
IntegratedQuantum: Why not use clear and retain the capacity?
```suggestion
		renderComponents.clear();
```

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+// ############################# Client only stuff ################################
+pub const client = struct {
+	const RenderComponent = struct {
+		entity: u32, // entity
+		entityModel: main.entityModel.EntityModelIndex, // model
+	};
+	pub var renderComponents: main.utils.SparseSet(RenderComponent, main.entity.Entity) = undefined;
+
+	pub fn init() void {
+		renderComponents = .{};
```
IntegratedQuantum: Could initialize it statically.
```suggestion
	pub var renderComponents: main.utils.SparseSet(RenderComponent, main.entity.Entity) = .{};

	pub fn init() void {
```

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+	}
+	pub fn deinit() void {
+		renderComponents.deinit(main.globalAllocator);
+	}
+	pub fn clear() void {
+		renderComponents.deinit(main.globalAllocator);
+		renderComponents = .{};
+	}
+	pub fn load(entity: u32, reader: *utils.BinaryReader, version: u32) main.entity.EntityComponentLoadError!void {
+		_ = version;
```
IntegratedQuantum: Always check the version, to be able to reject corrupted data. (maybe this could be added to the template as well, so we don't forget this in the future)

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+		const entityModel = reader.readInt(u32) catch return;
```
IntegratedQuantum: Could be a varint.

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+			std.log.err("entity {} couldn't be unloaded", .{entity});
+		};
+	}
+};
+
+// ############################# Server only stuff ################################
+
+pub const server = struct {
+	pub const RenderComponent = struct {
+		entity: u32, // entity
```
IntegratedQuantum: not needed, same as above

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+		entityModel: main.entityModel.EntityModelIndex, // model
+		pub fn save(self: RenderComponent, writer: *utils.BinaryWriter, audience: main.entity.AudienceInfo) main.entity.ComponentSaveBehaviour {
+			_ = audience;
+			writer.writeInt(u32, self.entityModel.index);
```
IntegratedQuantum: varint

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+			return .save;
+		}
+	};
+	var renderComponents: main.utils.SparseSet(RenderComponent, main.entity.Entity) = undefined;
+	pub fn init() void {
+		renderComponents = .{};
+	}
+	pub fn deinit() void {
```
IntegratedQuantum, line 72: Just noticed: Why does the client have a clear method, but the server doesn't?

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+		renderComponents.deinit(main.globalAllocator);
+	}
+	pub fn loadFromData(entity: u32, reader: *utils.BinaryReader, version: u32) main.entity.EntityComponentLoadError!void {
+		_ = version;
```
IntegratedQuantum: check the version

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+		const entityModel = reader.readInt(u32) catch return;
```
IntegratedQuantum: varint

src/entityComponent/model.zig
```diff
@@ -0,0 +1,106 @@
...
+
+		try loadByIndex(entity, main.entityModel.EntityModelIndex{.index = entityModel});
```
IntegratedQuantum:
```suggestion
		try loadByIndex(entity, .{.index = entityModel});
```

src/server/server.zig
```diff
@@ -249,6 +249,10 @@ pub const User = struct { // MARK: User
 		self.id = freeId;
 		freeId += 1;
 
+		if (world.?.playerEntityModels.items.len != 0) {
+			const defaultModel = world.?.playerEntityModels.items[main.random.nextInt(u32, &main.seed)%world.?.playerEntityModels.items.len];
```
IntegratedQuantum: nextIntBounded

src/server/world.zig
```diff
@@ -632,6 +646,26 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+					const entityModel = value.as(?[]const u8, null) orelse {
+						std.log.err("playerEntityModels in world.zig.zon must be an Array of strings.", .{});
+						continue;
+					};
+					self.playerEntityModels.append(main.globalAllocator, main.globalAllocator.dupe(u8, entityModel));
+				}
+			}
+		}
+		if (self.playerEntityModels.items.len == 0) {
+			for ([_][]const u8{"cubyz:snale", "cubyz:cubert"}) |entityModel| {
```
IntegratedQuantum: Instead of hardcoding this, how about adding a tag to the model?

src/server/world.zig
```diff
@@ -454,6 +454,8 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 
 	biomeChecksum: i64 = 0,
 
+	playerEntityModels: main.ListUnmanaged([]const u8),
```
IntegratedQuantum: I think this (together with the loading code) belongs inside the entityModel file where the rest of the model loading happens, furthermore this should contain the EntityModelIndex directly instead of going through the id.

IntegratedQuantum (changes requested):

src/entity.zig
tillpp: comment was just there to convince you about this implementation, not needed anymore

src/entityComponent/model.zig
tillpp: > maybe this could be added to the template as well, so we don't forget this in the future

yeah maybe, you should make an issue for that.

src/entityComponent/model.zig
tillpp: client.init and deinit get called at the beginning/end of the entire Game. So they need a .clear when the world is switched.
server.init and deinit is called when you join a world or leave a world, so it never needs a clear()

This design wasnt by me, it was inherited by entity_manager which was replaced by my entity.client. I just stuck with the design that was already there:
grafik


To make both consistent we could either:
A) remove client.clear(); and use .deinit() instead
B) make server.init()/deinit() get called at the beginning of the game too, and use server.clear() when deiniting the serverWorld

src/entityComponent/model.zig
tillpp: do we want to retain capacity when switching to another world?
I think it might be best to remove the client.clear function entirely

src/game.zig
```diff
@@ -704,9 +705,13 @@ pub const World = struct { // MARK: World
 		try assets.loadWorldAssets(path, self.blockPalette, self.itemPalette, self.proceduralItemPalette, self.biomePalette, self.entityModelPalette, self.entityComponentPalette);
 		Player.id = zon.get(u32, "player_id", std.math.maxInt(u32));
 		Player.inventory = ClientInventory.init(main.globalAllocator, Player.inventorySize, .serverShared, .{.playerInventory = Player.id}, .{});
-		try Player.loadFrom(zon.getChild("player"));
 		self.playerBiome = .init(main.server.terrain.biomes.getPlaceholderBiome());
 		main.audio.setMusic(self.playerBiome.raw.preferredMusic);
+
+		main.Window.c.glfwMakeContextCurrent(main.Window.window);
+		main.entityModel.loadModelsAndTexture();
```
tillpp: @RanPix needs the entityModel fully loaded before they get used in EntityComponents for his Joint PR in the future
For this purpose, just for the handshake, the network thread on the client shortly get access to opengl to finish loading the entities, before the player.

tillpp: for more consistency, we could take all of the asset initialization from world.init  into the finishHandshake. (the selected area in the screenshot)

image

tillpp: another alternative would be lazy loading the models, but i don't think you like that.

IntegratedQuantum: I will not accept such a hack in an unrelated PR. If this is really needed (which I strongly doubt, but I haven't seen the code yet), then it should be introduced in the PR that actually adds the code that needs it.

tillpp: Yeah kind of expected that answer. Sorry @RanPix

tillpp: reverted

assets/cubyz/entityModels/cubert.zig.zon
```diff
@@ -2,5 +2,6 @@
 	.model = "cubyz:cubert",
 	.defaultTexture = "cubyz:cubert",
 	.height = 2,
+	.isPlayerModel = true,
```
Argmaster: We could and probably should add and use tag field for that?

IntegratedQuantum: Yeah, that is a good idea.

src/entityComponent/model.zig
```diff
@@ -0,0 +1,101 @@
...
+
+const BinaryReader = main.utils.BinaryReader;
+
+pub var entityComponentID: main.entity.EntityComponentId = undefined;
+pub const entityComponentVersion = 0;
+
+// ############################# Client only stuff ################################
+pub const client = struct {
+	const Component = struct {
+		entityModel: main.entityModel.EntityModelIndex, // model
```
IntegratedQuantum:
```suggestion
		entityModel: main.entityModel.EntityModelIndex,
```

src/entityComponent/model.zig
```diff
@@ -0,0 +1,101 @@
...
+	pub fn get(entity: u32) ?*Component {
+		return components.get(@enumFromInt(entity));
+	}
+};
+
+// ############################# Server only stuff ################################
+
+pub const server = struct {
+	pub const Component = struct {
+		entityModel: main.entityModel.EntityModelIndex, // model
```
IntegratedQuantum:
```suggestion
		entityModel: main.entityModel.EntityModelIndex,
```

src/entityComponent/model.zig
```diff
@@ -0,0 +1,101 @@
...
+	pub fn init() void {}
+	pub fn deinit() void {
+		components.deinit(main.globalAllocator);
+	}
+	pub fn clear() void {
+		components.clear();
+	}
+	pub fn load(entity: u32, reader: *utils.BinaryReader, version: u32) main.entity.EntityComponentLoadError!void {
+		if (version != 0)
+			return main.entity.EntityComponentLoadError.InvalidComponentVersion;
```
IntegratedQuantum: No need to specify the error set, to Zig all errors with the same name are identical.
```suggestion
			return error.InvalidComponentVersion;
```

src/entityComponent/model.zig
```diff
@@ -0,0 +1,101 @@
...
+
+		const entityModel = reader.readVarInt(u32) catch return main.entity.EntityComponentLoadError.UnreadableComponentData;
```
IntegratedQuantum:
```suggestion
		const entityModel = reader.readVarInt(u32) catch return error.UnreadableComponentData;
```

src/entityComponent/model.zig
```diff
@@ -0,0 +1,101 @@
...
+	pub fn init() void {
+		components = .{};
+	}
+	pub fn deinit() void {
+		components.deinit(main.globalAllocator);
+	}
+	pub fn loadFromData(entity: u32, reader: *utils.BinaryReader, version: u32) main.entity.EntityComponentLoadError!void {
+		if (version != 0)
+			return error.InvalidComponentVersion;
+		const entityModel = reader.readVarInt(u32) catch return main.entity.EntityComponentLoadError.UnreadableComponentData;
```
IntegratedQuantum:
```suggestion
		const entityModel = reader.readVarInt(u32) catch return error.UnreadableComponentData;
```

src/server/server.zig
```diff
@@ -249,6 +249,10 @@ pub const User = struct { // MARK: User
 		self.id = freeId;
 		freeId += 1;
 
+		if (main.entityModel.playerEntityModels.items.len != 0) {
+			const defaultModel = main.entityModel.playerEntityModels.items[main.random.nextIntBounded(u32, &main.seed, @truncate(main.entityModel.playerEntityModels.items.len))];
```
IntegratedQuantum: Why not use the correct type to avoid casting?
```suggestion
			const defaultModel = main.entityModel.playerEntityModels.items[main.random.nextIntBounded(usize, &main.seed, main.entityModel.playerEntityModels.items.len)];
```

IntegratedQuantum (changes requested):

IntegratedQuantum:
Also please rebase to the new Zig version, so we don't get any accidental errors.

src/server/server.zig
tillpp: grafik
Didn't work.

src/entityModel.zig
```diff
@@ -76,6 +76,12 @@ pub const EntityModel = struct {
 		self.indexCount = 0;
 		self.coordinateSystem = zon.get(CoordinateSystem, "coordinateSystem", .right_handed_z_up);
 
+		if (zon.getChildOrNull("isPlayerModel")) |isPlayerModel| {
+			if (isPlayerModel.as(bool, false)) {
```
```suggestion
		if (zon.get(bool, "isPlayerModel", false)) {
```

src/entityModel.zig
```diff
@@ -277,13 +283,15 @@ pub const EntityModelIndex = struct {
...
+
 pub var reverseIndices: std.StringHashMapUnmanaged(EntityModelIndex) = .{};
 pub var entityModels: main.ListUnmanaged(EntityModel) = .{};
 
-pub fn register(assetFolder: []const u8, entityModelId: []const u8, zon: ZonElement) usize {
-	const index = entityModels.items.len;
-	entityModels.append(main.worldArena, EntityModel.init(assetFolder, zon));
-	reverseIndices.put(main.worldArena.allocator, entityModelId, EntityModelIndex{.index = @truncate(index)}) catch unreachable;
+pub fn register(assetFolder: []const u8, entityModelId: []const u8, zon: ZonElement) EntityModelIndex {
+	const index = EntityModelIndex{.index = @truncate(entityModels.items.len)};
```
IntegratedQuantum: Use intCast so we have a chance at catching mistakes in safe modes.
truncate should only be used if you really want to throw away the remaining bits.
```suggestion
	const index = EntityModelIndex{.index = @intCast(entityModels.items.len)};
```

src/server/server.zig
```suggestion
			const defaultModel = main.entityModel.playerEntityModels.items[main.random.nextIntBounded(u32, &main.seed, @intCast(main.entityModel.playerEntityModels.items.len))];
```

IntegratedQuantum: that's annoying. It seems like I only ~~stole that~~ wrote that to support 48 bit random numbers.

IntegratedQuantum (changes requested):
some small things, then we are good to go here.

src/entityModel.zig
```diff
@@ -76,6 +76,19 @@ pub const EntityModel = struct {
 		self.indexCount = 0;
 		self.coordinateSystem = zon.get(CoordinateSystem, "coordinateSystem", .right_handed_z_up);
 
+		var isPlayerModel = false;
+		if (zon.getChildOrNull("tags")) |tags| {
+			for (tags.toSlice()) |tagZon| {
```
IntegratedQuantum: Please use `main.Tag.loadTagsFromZon` so we only have one consistent way of loading tags for the entire codebase.

src/entityModel.zig
```diff
@@ -76,6 +76,19 @@ pub const EntityModel = struct {
...
+				const tag = tagZon.as([]const u8, "invalid");
+				if (std.mem.eql(u8, tag, "playerModel")) {
```
IntegratedQuantum: with the above you can define a `playerModel` field in the enum.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #2682, opened 2026-03-07
ECS Part 9 | ModelRenderer

tillpp:
Ninth step of splittig up the 3000 lines big PR: https://github.com/PixelGuys/Cubyz/pull/2652
(requires Part 8)

**QUANTUM DON'T BE SCARED OF THE BIG SCARY LOC NUMBER!**
**THIS PR FOR THE MOST PART ONLY MOVES CODE!**

``entity_manager.zig`` => ``modelRenderer.zig``


src/entitySystem/modelRenderer.zig
```diff
@@ -0,0 +1,160 @@
...
+const ServerWorld = main.server.ServerWorld;
+const items = main.items;
+const ItemStack = items.ItemStack;
+const random = main.random;
+
+const entityComponent = main.entity.components;
+
+// ############################# Client only stuff ################################
+pub const client = struct {
+	var pipeline: graphics.Pipeline = undefined; // Entities are sometimes small and sometimes big. Therefor it would mean a lot of work to still use smooth lighting. Therefor the non-smooth shader is used for those.
```
IntegratedQuantum: I don't think the comment is really relevant anymore

src/entitySystem/modelRenderer.zig
```diff
@@ -0,0 +1,160 @@
...
+
+		c.glUniformMatrix4fv(uniforms.projectionMatrix, 1, c.GL_TRUE, @ptrCast(&projMatrix));
+		c.glUniform3fv(uniforms.ambientLight, 1, @ptrCast(&ambientLight));
+		c.glUniform1f(uniforms.contrast, 0.12);
+
+		for (main.client.entity_manager.entities.items()) |*ent| {
+			if (ent.id == game.Player.id) // don't render local player
+				continue;
+
+			const component = entityComponent.@"cubyz:model".client.components.get(@enumFromInt(ent.id)) orelse continue;
```
IntegratedQuantum: This system should only iterate through the entities that have a model (using the SparseSet), instead of iterating through all entities, and checking if each of them has the component.

src/entitySystem/modelRenderer.zig
```diff
@@ -0,0 +1,160 @@
...
+		const screenUnits = @as(f32, @floatFromInt(main.Window.height))/1024;
+		const fontBaseSize = 128.0;
+		const fontMinScreenSize = 16.0;
+		const fontScreenSize = fontBaseSize*screenUnits;
+
+		for (main.client.entity_manager.entities.items()) |*ent| {
+			if (ent.id == game.Player.id) // don't render local player
+				continue;
+
+			const component = entityComponent.@"cubyz:model".client.components.get(@enumFromInt(ent.id)) orelse continue;
```
IntegratedQuantum: What about entities that have a name but not a model? You only need this to adjust the position based on the model.

src/entitySystem/modelRenderer.zig
```diff
@@ -0,0 +1,160 @@
...
+			const entModel = component.entityModel.get();
+
+			const pos3d = ent.getRenderPosition() - playerPos;
+			const pos4f = Vec4f{
+				@floatCast(pos3d[0]),
+				@floatCast(pos3d[1]),
+				@floatCast(pos3d[2] + entModel.height - 0.9),
```
IntegratedQuantum: Note that the entity position is in the center, so this should probably use half the height.

IntegratedQuantum (changes requested):

src/entitySystem/modelRenderer.zig
```diff
@@ -0,0 +1,165 @@
...
+		main.client.entity_manager.mutex.lock();
+		defer main.client.entity_manager.mutex.unlock();
+		pipeline.bind(null);
+
+		c.glUniformMatrix4fv(uniforms.projectionMatrix, 1, c.GL_TRUE, @ptrCast(&projMatrix));
+		c.glUniform3fv(uniforms.ambientLight, 1, @ptrCast(&ambientLight));
+		c.glUniform1f(uniforms.contrast, 0.12);
+
+		for (entityComponent.@"cubyz:model".client.components.dense.items, 0..) |component, index| {
+			const id = entityComponent.@"cubyz:model".client.components.denseToSparseIndex.items[index];
```
IntegratedQuantum: By the way there is a little trick you can do here. Though it would probably make sense to create and alias for the common path to keep the line shorter.
```suggestion
		for (entityComponent.@"cubyz:model".client.components.dense.items, entityComponent.@"cubyz:model".client.components.denseToSparseIndex.items[index]) |component, id| {
```

src/entitySystem/modelRenderer.zig
tillpp: had to make the code a bit more complicated (used SparseSet for clientSide entities instead of virtualList).

src/client/entity_manager.zig
```diff
@@ -28,158 +28,61 @@ var uniforms: struct {
 	ambientLight: c_int,
 } = undefined;
 
-var pipeline: graphics.Pipeline = undefined; // Entities are sometimes small and sometimes big. Therefor it would mean a lot of work to still use smooth lighting. Therefor the non-smooth shader is used for those.
-pub var entities: main.utils.VirtualList(main.client.Entity, 1 << 20) = undefined;
+pub var entities: main.utils.SparseSet(main.client.Entity, main.entity.Entity) = undefined;
```
IntegratedQuantum: I'm sure if we should change it to a SparseSet, since all entities have this "component". I'd prefer if we keep it as is for now, and then decide on a benchmark later (once we have enough entities to make a representative benchmark).

Also you cannot even use a SparseSet, since by design it reorders the values in memory, however (due to poor design decisions in the interpolation system) entities require stable pointers for now, so only a VirtualList, which guarantees stable pointers can be used.

IntegratedQuantum (changes requested):

src/client/entity_manager.zig
tillpp: > I'm sure if we should change it to a SparseSet, since all entities have this "component". I'd prefer if we keep it as is for now, and then decide on a benchmark later (once we have enough entities to make a representative benchmark).

with the previous virtual list you get O(n) for get(id), thats why i used sparset here

IntegratedQuantum: That could also be achieved with a map (array) from id to index

tillpp: seems like a batch.
I think it is best to pause this PR, until "the due to poor design decisions in the interpolation system" are fixed.
Btw why did you do this that way?

IntegratedQuantum: I did it this way so that the interpolation thing can store the result by pointer.

Unless you want to fix that, you'd wait for a while.

tillpp: Done

tillpp: _Cubert time_

tillpp:
this is ready for merge!
have you forgotten about thir PR?

src/client/entity_manager.zig
```diff
@@ -28,43 +28,36 @@ var uniforms: struct {
 	ambientLight: c_int,
 } = undefined;
 
-var pipeline: graphics.Pipeline = undefined; // Entities are sometimes small and sometimes big. Therefor it would mean a lot of work to still use smooth lighting. Therefor the non-smooth shader is used for those.
 pub var entities: main.utils.VirtualList(main.client.Entity, 1 << 20) = undefined;
+pub var idMapping: std.AutoHashMap(u32, *main.client.Entity) = undefined;
```
IntegratedQuantum: ids are generally small numbers, I think a hashmap for the mapping makes little sense.
```suggestion
pub var idMapping: main.List(?*main.client.Entity) = undefined;
```

src/client/entity_manager.zig
```diff
@@ -73,108 +66,29 @@ fn update() void {
...
 }
-
+pub fn getEntity(id: u32) ?*main.client.Entity {
+	return idMapping.get(id);
+}
 pub fn removeEntity(id: u32) void {
 	mutex.lock();
 	defer mutex.unlock();
+
 	for (entities.items(), 0..) |*ent, i| {
```
IntegratedQuantum: Please also use the mapping to replace this loop

src/client/entity_manager.zig
```diff
@@ -28,43 +28,36 @@ var uniforms: struct {
 	ambientLight: c_int,
 } = undefined;
```
IntegratedQuantum: You also moved this, so I think it can be removed.

src/entitySystem/modelRenderer.zig
```diff
@@ -0,0 +1,163 @@
...
+			const modelMatrix = (Mat4f.identity()
+				.mul(Mat4f.translation(Vec3f{
+					@floatCast(pos[0]),
+					@floatCast(pos[1]),
+					@floatCast(pos[2] - entModel.height/2),
+				}))
+				.mul(Mat4f.rotationZ(-ent.rot[2])));
+			const modelViewMatrix = game.camera.viewMatrix.mul(modelMatrix);
+			c.glUniformMatrix4fv(uniforms.viewMatrix, 1, c.GL_TRUE, @ptrCast(&modelViewMatrix));
+			c.glDrawElements(c.GL_TRIANGLES, 6*entModel.indexCount, c.GL_UNSIGNED_INT, null);
```
IntegratedQuantum: The 6* seems wrong. if the entModel.indexCount, is the indexCount, then it should be already premultiplied.

IntegratedQuantum (changes requested):

src/entitySystem/modelRenderer.zig
tillpp: i copied the code in march 7th from that spot to the other. appearently in the meantime the original code changed

Tried to fix all changed i could find

src/client/entity_manager.zig
tillpp: not possible, since removeEntity requires the index.
So i used ``main.List(?u32)`` instead

and done. _Cubert Time_

src/client/entity_manager.zig
```diff
@@ -73,112 +59,60 @@ fn update() void {
...
+	if (idMapping.items.len <= id)
+		return;
+	const index: u32 = idMapping.items[id] orelse return;
+	const ent = entities.items()[index];
+
+	// remove id
+	{
+		idMapping.items[id] = null;
+
+		// resize array
```
IntegratedQuantum: Why resize it, the memory is never reclaimed anyways?

IntegratedQuantum (changes requested):

src/client/entity_manager.zig
tillpp: oh, reverted. _Cubert time?_

IntegratedQuantum (approved):

PR #2691, opened 2026-03-07
Adjust some things in the Root

ikabod-kee:
- Made Phantasmal Void biomes have the same fog as regular Void biomes.
- Adjusted the size of boulders.


IntegratedQuantum:
I think the problem isn't the rare biome chance, but rather that we do not have enough variation in common biomes.
I'm sure this will change over time, so I'd prefer if rare biomes are consistently kept at 0.01 everywhere.

ikabod-kee:
> I think the problem isn't the rare biome chance, but rather that we do not have enough variation in common biomes. I'm sure this will change over time, so I'd prefer if rare biomes are consistently kept at 0.01 everywhere.

Very well! I'll change it then

PR #2692, opened 2026-03-07
Updating Encased [Modifier 1 of 4]

Crepestrom:
From #2650 
-Encased Rework
Encased-visual
you can now define a range for the encased modifier
default is 1

### DIFF
--- a/src/tool/modifiers/restrictions/encased.zig
+++ b/src/tool/modifiers/restrictions/encased.zig
@@ -15,7 +15,7 @@ const Encased = struct {
 pub fn satisfied(self: *const Encased, tool: *const Tool, x: i32, y: i32) bool {
 	var count: usize = 0;
 	const lowBound = 0;
-	const highBound = self.range * 2 + 1;
+	const highBound = self.range*2 + 1;
 	for (lowBound..highBound) |dx| {
 		for (lowBound..highBound) |dy| {
 			const checkedX = x + @as(i32, @intCast(dx - self.range));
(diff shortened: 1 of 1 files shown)


Argmaster:
I think it might be confusing when set to anything other than 1. Consider that tools are not using full 5x5 grid, which means encase with higher values will rarely be possible.

Crepestrom:
when set to 5 it effectivly asks how many of a tag is in a tool
default is 1 so it wont change any behaviors unless wanted
i just wanted some fancy modifiers for my addon

src/tool/modifiers/restrictions/encased.zig
```diff
@@ -26,10 +31,16 @@ pub fn loadFromZon(allocator: NeverFailingAllocator, zon: ZonElement) *const Enc
...
 	};
 	return result;
 }
 
 pub fn printTooltip(self: *const Encased, outString: *main.List(u8)) void {
-	outString.print("encased in {} .{s}", .{self.amount, self.tag.getName()});
+	if (self.range == 1) {
+		outString.print("encased in {} .{s}", .{self.amount, self.tag.getName()});
+	}
+	if (!(self.range == 1)) {
```
IntegratedQuantum:
```suggestion
	} else {
```

IntegratedQuantum (changes requested):
I would like to put this in hold until #2818 is addressed. I think even the current encased modifier is already quite difficult to understand, the range is even less clear (is it a circle or a square or a diamond?)

Wbjitscool:
Apparently there's some conflicts occuring atm

PR #2693, opened 2026-03-07
Added OnDiagonal Modifier [Modifier 2 of 4]

Crepestrom:
From #2650 
-onDiagonal
diagonal-visual
Default range is 0 (this covers the entire available crafting space on diagonals)

Testing Addon
[diagonal.zip](https://github.com/user-attachments/files/26884990/diagonal.zip)


src/tool/modifiers/restrictions/onDiagonal.zig
```diff
@@ -0,0 +1,50 @@
...
+const main = @import("main");
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const ModifierRestriction = main.items.ModifierRestriction;
+const Tool = main.items.Tool;
+const ZonElement = main.ZonElement;
+
+const Encased = struct {
+	tag: main.Tag,
+	amount: usize,
+	range: usize,
```
IntegratedQuantum: Please use an optional and if not present check the entire size of the tool. Tools may not be limited to 5×5 in the future.

src/tool/modifiers/restrictions/onDiagonal.zig
```diff
@@ -0,0 +1,50 @@
...
+	const lowBound = 0;
+	const highBound = self.range*2 + 1;
+	for (lowBound..highBound) |dx| {
+		const checkedX = x + @as(i32, @intCast(dx - self.range));
+		const checkedY = y + @as(i32, @intCast(dx - self.range));
+		if ((tool.getItemAt(checkedX, checkedY) orelse continue).hasTag(self.tag)) count += 1;
+	}
+	for (lowBound..highBound) |dx| {
+		const checkedX = x + @as(i32, @intCast(dx - self.range));
+		const checkedY = y - @as(i32, (@intCast(dx - self.range)));
```
IntegratedQuantum:
```suggestion
		const checkedY = y - @as(i32, @intCast(dx - self.range));
```

src/tool/modifiers/restrictions/onDiagonal.zig
```diff
@@ -0,0 +1,50 @@
...
+		if (!(dx == 0)) {
```
IntegratedQuantum:
```suggestion
		if (dx != 0) {
```

src/tool/modifiers/restrictions/_list.zig
```diff
@@ -3,3 +3,4 @@ pub const @"and" = @import("and.zig");
 pub const encased = @import("encased.zig");
 pub const not = @import("not.zig");
 pub const @"or" = @import("or.zig");
+pub const OnDiagonal = @import("onDiagonal.zig");
```
IntegratedQuantum:
```suggestion
pub const onDiagonal = @import("onDiagonal.zig");
```

IntegratedQuantum (changes requested):

IntegratedQuantum:
please fix the formatting

Crepestrom:
Requested Changes have been made
and also copied to OnOrthogonal

src/tool/modifiers/restrictions/onDiagonal.zig
```diff
@@ -0,0 +1,61 @@
...
+};
+
+pub fn satisfied(self: *const OnDiagonal, tool: *const Tool, x: i32, y: i32) bool {
+	var count: usize = 0;
+	const gridSize: usize = tool.craftingGrid.len;
+	var rangeChecked: usize = 0;
+	if (self.range > gridSize) {
+		rangeChecked = gridSize;
+	} else {
+		if (self.range == 0) {
```
IntegratedQuantum: Please use an optional instead of magic value 0.
```zig
const rangeChecked = @min(self.range orelse gridSize, gridSize);
```

src/tool/modifiers/restrictions/onDiagonal.zig
```diff
@@ -0,0 +1,61 @@
...
+			if ((tool.getItemAt(checkedX, checkedY) orelse continue).hasTag(self.tag)) count += 1;
+		}
+	}
+	return count >= self.amount;
+}
+
+pub fn loadFromZon(allocator: NeverFailingAllocator, zon: ZonElement) *const OnDiagonal {
+	const result = allocator.create(OnDiagonal);
+	result.* = .{
+		.tag = main.Tag.find(zon.get([]const u8, "tag", "not specified")),
```
IntegratedQuantum: Please print an error in the missing case.
```suggestion
		.tag = main.Tag.find(zon.get(?[]const u8, "tag", null) orelse {...std.log.err(...)...}),
```

IntegratedQuantum (changes requested):

src/tool/modifiers/restrictions/onDiagonal.zig
Crepestrom: this just added more errors

Crepestrom: i dont know what that std log err is or why its covered in ...

IntegratedQuantum: You are supposed to replace ... with the actual code. This was meant to be pseudo code.

H41ogen:
rip formatting

src/tool/modifiers/restrictions/onDiagonal.zig
Crepestrom: Ive finally gotten around to it

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
```diff
@@ -0,0 +1,52 @@
...
+pub fn satisfied(self: *const OnDiagonal, proceduralItem: *const ProceduralItem, x: i32, y: i32) bool {
+	var count: usize = 0;
+	const gridSize: usize = proceduralItem.craftingGrid.len;
+	const rangeChecked = @min(self.range orelse gridSize, gridSize);
+	const lowBound = 0;
+	const highBound = rangeChecked*2 + 1;
+	for (lowBound..highBound) |dx| {
+		const checkedX = x + @as(i32, @intCast(dx - rangeChecked));
+		const checkedY = y + @as(i32, @intCast(dx - rangeChecked));
+		if ((proceduralItem.getItemAt(checkedX, checkedY) orelse continue).hasTag(self.tag)) count += 1;
```
IntegratedQuantum, line 24: Missing the `dx != 0` check from below

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
```diff
@@ -0,0 +1,52 @@
...
+
+pub fn satisfied(self: *const OnDiagonal, proceduralItem: *const ProceduralItem, x: i32, y: i32) bool {
+	var count: usize = 0;
+	const gridSize: usize = proceduralItem.craftingGrid.len;
+	const rangeChecked = @min(self.range orelse gridSize, gridSize);
+	const lowBound = 0;
+	const highBound = rangeChecked*2 + 1;
+	for (lowBound..highBound) |dx| {
+		const checkedX = x + @as(i32, @intCast(dx - rangeChecked));
+		const checkedY = y + @as(i32, @intCast(dx - rangeChecked));
```
IntegratedQuantum: This looks wrong, using dx for y. Also `dx` convention is for delta values, but only `dx - rangedCheck` is the actual delta.
I'd suggest to rename it.

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
```diff
@@ -0,0 +1,52 @@
...
+};
+
+pub fn satisfied(self: *const OnDiagonal, proceduralItem: *const ProceduralItem, x: i32, y: i32) bool {
+	var count: usize = 0;
+	const gridSize: usize = proceduralItem.craftingGrid.len;
+	const rangeChecked = @min(self.range orelse gridSize, gridSize);
+	const lowBound = 0;
+	const highBound = rangeChecked*2 + 1;
+	for (lowBound..highBound) |dx| {
+		const checkedX = x + @as(i32, @intCast(dx - rangeChecked));
```
IntegratedQuantum: This looks like a crash
```suggestion
		const checkedX = x + @as(i32, @intCast(dx)) - rangeChecked;
```

IntegratedQuantum (changes requested):
Please also test this on your own by adding it to a test item and checking if it works. Just by looking at it, I already see a case where it will definitely crash.

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
Crepestrom: no
thats so it can count itself just like the encased modifier
the second one has a check so it doesnt double count itself

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
Crepestrom: changed it

Crepestrom:
yipee it works

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
```diff
@@ -0,0 +1,53 @@
+const std = @import("std");
+
+const main = @import("main");
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const ModifierRestriction = main.items.ModifierRestriction;
+const ProceduralItem = main.items.ProceduralItem;
+const ZonElement = main.ZonElement;
+
+const On_diagonal = struct {
```
IntegratedQuantum, line 9: This is not a name convention I have seen before.
```suggestion
const OnDiagonal = struct {
```

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
```diff
@@ -0,0 +1,53 @@
...
+	tag: main.Tag,
+	amount: usize,
+	range: ?usize,
```
IntegratedQuantum, line 12: Honestly I'm not too happy with this range parameter, which is also one of the reasons why I have neglected your PR series.

The range is really clunky in the UI and in my opinion there is little value to it.
We still should allow it of course, but in my opinion it would be better solved by allowing arbitrary custom patterns instead of giving each restriction an optional range parameter.

IntegratedQuantum (changes requested):

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
Crepestrom: renamed

src/proceduralItem/modifiers/restrictions/on_diagonal.zig
Crepestrom: sure i can remove it

Crepestrom: i will look to make a custom pattern modifier

Crepestrom: the range paramet is gone now if you wanna check it

PR #2695, opened 2026-03-07
Adding atRange Modifier [Modifier 4 of 4]

Crepestrom:
From #2650
-At range
at range visualizer large
default range is 2
Like encased but only at that specific range


IntegratedQuantum:
Should be allowed through custom restrictions (see #3092)

PR #2699, opened 2026-03-08
Cheese

TyrrelMeter:
Cheese


careeoki:
I'm currently working on a unique block palette for sky islands, I would prefer if we wait until after we have those before adding more biomes.

I also think we should avoid reusing the same tree SBBs for sky islands (instead create new, more whimsical trees)
The player has already seen these trees on the surface, they should be rewarded with new stuff for getting up here.

TyrrelMeter:
got it. I guess I'll close the PR for now then.

PR #2705, opened 2026-03-10
New save creation menu v2

H41ogen:
Rework of #2145 

I decided to take another crack at one of my first Cubyz projects: improving the save creation menu. For now, I've left out the addons page to make this as small of a change as possible. I think this new menu might be nice for when permissions are added, for setting default permissions and stuff.

Screenshot_20260309_210927
Screenshot_20260309_210953

This is the lowest of low priority to look at tbh.


Crepestrom:
let us make folders for our worlds
and color code them / use symbols for each world

H41ogen:
That's out of the scope of this pr tbh. could be nice to have later, and this could definitely help

Wunka:
I like this 👍 
The only thing I don't like is the implementation. I think there can be improvements. I think when I am back at home I will try to do another implementation and when it works will do a PR on your branch.

H41ogen:
Sounds good, I'm open to ideas

Wunka:
> Sounds good, I'm open to ideas

I created a PR to your branch. Its just an abstraction. When I first thought about it. It seemed way more complicated in my head.

IntegratedQuantum:
Have you considered using collapsible sections instead of tabs? I think that could get you more quickly where you want without having to cycle through ten tabs. Furthermore it would allow recursive sections, which would make it easier to extend the feature set in one category.

H41ogen:
I'll work on adding some kind of option to VerticalList to make it collapsible when i have the drive to, I like that idea.

PR #2706, opened 2026-03-10
Velocity-based chunk generation priority

### LINKED ISSUE #2600: Momentum-based chunk priority
### Description

At the moment, falling down for long periods of time, specially in multiplayer, is not a fun activity for lower-end computers.

Image

prior to the above image, i have accumulated 800+ seconds of falling in a multiplayer session, but for most of the time, my momentum was frozen for as long as chunks took time to load, which wouldn't be a problem of its own, but the problem is that it would attempt to load everything around that is never going to be visible due to the sheer depth of the fall (over 16 thousand voxels deep).

that said, i believe that, during terminal velocity, or somewhere close, both the client and the server should start prioritizing the incoming path of the fall in pace with the velocity direction, and maybe skip rendering chunks altogether if the fall is happening under the surface, and instead focus on a single line unless an opening is immediately available in the hole (i'm guessing the future implementation of SDFs might make it easier for finetuning an implementation like this?)

thank you for reading <3

H41ogen:
In the future it might be nice to also take into account the direction of the player, but just adding the velocity makes flying at decently high speeds feel a lot nicer.

Closes #2600

### DIFF
--- a/src/renderer/mesh_storage.zig
+++ b/src/renderer/mesh_storage.zig
@@ -773,7 +773,7 @@ pub fn updateMeshes(targetTime: std.Io.Timestamp) void { // MARK: updateMeshes()
 					defer mutex.lock();
 					continue;
 				}
-				const priority = pos.getPriority(playerPos + playerVel);
+				const priority = pos.getPriority(playerPos, playerVel);
 				if (priority > closestPriority) {
 					closestPriority = priority;
 					closestIndex = i;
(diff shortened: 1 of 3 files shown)


src/renderer/mesh_storage.zig
```diff
@@ -772,7 +773,7 @@ pub fn updateMeshes(targetTime: std.Io.Timestamp) void { // MARK: updateMeshes()
 					defer mutex.lock();
 					continue;
 				}
-				const priority = pos.getPriority(playerPos);
+				const priority = pos.getPriority(playerPos + playerVel);
```
IntegratedQuantum: Please pass the velocity as a second parameter, to allow changing the time it looks ahead in one central place.

H41ogen: To `updateMeshes`?

H41ogen: oh never mind.

H41ogen:
well that's far out of the scope of what I know how to do, I was thinking it'd be a lot simpler

PR #2710, opened 2026-03-10
Show distance below player names

H41ogen:
It's a bit hard to tell the distance to another player; I think this would be nice, at least as an optional toggle.

The code isn't amazingly written since I'm not sure if this is wanted.


Wunka:
this was hit by the great entity merge conflicts.

IntegratedQuantum:
I'd rather not have this, players shouldn't have access to such precise numbers by default in my opinion.
We can rediscuss this once accessories are on the table.

PR #2712, opened 2026-03-10
tablist gui Component

Wunka:
This adds add Tablist gui Component
After seeing #2705 I thought that the tabs should have an abstraction but to not make #2705 too big @H41ogen said I should seperate this into this PR.

Tested can this be with my tablist branch https://github.com/Wunka/Cubyz/tree/tablist with which I first wanted to make a PR onto #2705

This component only handels the rendering, buttonPressing etc of the tabs themself. A header or other things to control which tab is open is for now in the hands of the component user

### DIFF
--- a/src/gui/components/TabList.zig
+++ b/src/gui/components/TabList.zig
@@ -1,0 +1,110 @@
+const std = @import("std");
+
+const main = @import("main");
+const graphics = main.graphics;
+const draw = graphics.draw;
+const TextBuffer = graphics.TextBuffer;
+const vec = main.vec;
+const Vec2f = vec.Vec2f;
+
+const gui = @import("../gui.zig");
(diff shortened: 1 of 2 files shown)


H41ogen:
You should add an image showing it off

BoySanic:
I put it in high priority because it's blocking #2705

Wunka:
Just a heads up. When I am back at my pc I will remove `setValues()` as its logic is replaced by `finish()` I just forgot to delete it. This does not change anything else so this PR can still be reviewed.

IntegratedQuantum (changes requested):
As noted in #2705, I would prefer a collapsible section. Please close, or provide other reasons for why we should have it.

Wunka:
I currently cannot think about another place where this could be useful. so closed

Issue #2505, opened 2026-01-19
Replacing branches with decaying branches in bulk


careeoki:
### Description

There's no nice way to do this right now:
https://discord.com/channels/443805812390100992/702167480654757991/1462876418592604274

PR #2717, opened 2026-03-12
Sky islands block palette

careeoki:
Adds the following blocks for sky islands:
- Sky grass & vegetation
- Aerosoil
- Nimbusite
- Nimbusite bricks, tile, and wall

image
image

Also small biome changes:
- Less trees (since they look much more out of place)
- Tweaked fog colour (i will probably continue to revisit this)

Sky grass vegetation is currently just a recolor. I will change this in the future.


careeoki:
> awesome, now we just need some fitting trees.

which i will do as soon as #2505 is solved :-)

IntegratedQuantum:
oh, right

PR #2721, opened 2026-03-15
Cactus Planks

masterminer176:
Adds Cactus Plank Blocks as a material made from Cactus Logs/Branches. As well as Fences, Signs, and Chests.
Primarily because I found it odd that cactus had Most of the DNA of the wood blocks, and yet no planks. Also makes for a good excuse to have a Green Woodtype.
image
Balancing might be a little wacky so feel free to critique/change that if needed.

### DIFF
--- a/assets/cubyz/blocks/planks/cactus.zig.zon
+++ b/assets/cubyz/blocks/planks/cactus.zig.zon
@@ -1,5 +1,5 @@
 .{
-	.blockhealth = 3.75
+	.blockhealth = 3.75,
 	.item = .{
 		.material = .{
 			.durability = 28,


BoySanic:
@careeoki @ikabod-kee Artsy :)

ikabod-kee:
Cactus wood is a cool idea, but how many wood types is too much? I'm not sure. I wouldn't be mad if this was added, though.

As for the textures, a touch-up from Carrie would probably be needed.

ikabod-kee:
Let me go in the game to check the balancing.

H41ogen:
i feel like it'd be fun for every "log" to have a whole wood type

masterminer176:
> Cactus wood is a cool idea, but how many wood types is too much? I'm not sure. I wouldn't be mad if this was added, though.

Honestly, in my opinion, if we tried to add any more "naturally" coloured woods then we'd have too many probably. Hence why I tried to go for something different.
In this case I don't think it's overcrowding to have green wood where there otherwise wouldn't be.
> As for the textures, a touch-up from Carrie would probably be needed.

By all means go ahead. I have my own sensibilities as a sprite artist and it would definitely help for cohesion's sake if Carrie did her Final Pass on it :+1:

careeoki:
Cactus doesn't have planks because it's not a wood type. I think it's more interesting if cacti and mushrooms are their own little category of block sets.
I'm not against more cactus blocks, but it should be something more unique than just more planks. Cacti are not trees, so we should represent them accordingly.

It also poses some design questions- if you can get wood from deserts now, they become more hospitable which doesn't feel right for deserts. Obviously we aren't a survival game but I think its more interesting if biomes have varying amounts/kinds of resources.

ikabod-kee:
Carrie's true. I didn't want to say anything because I am a famous party pooper but I think it'd be best to have wood be wood and cactus be cactus. Also, come to think of it, having cactus wood would just let players craft that instead of dealing with the (future) modifiers that come with cactus, that being thorns.

ikabod-kee:
I feel bad denying this because it's your first PR, but after giving it a second thought I don't think this should be merged.

H41ogen:
i think there should be `cubyz:cactus_planks` but they shouldn't look like or be called cactus planks

masterminer176:
I mean you did provide pretty good reasons for why not, so I understand where you're both coming from.
Plus, worst case scenario, I did make an Addon for it already, so the effort's not in vain.
Either way you did make a good point of a new type of blocks for these, so I'll respect it :+1:

That said, should I close the request?

ikabod-kee:
Yes, feel free to close it. :) Thanks!

masterminer176:
Alright fair enough :saluting_face:

PR #2727, opened 2026-03-16
Fail compile if tagged version does not match build.zig version

### LINKED ISSUE #2508: The release CI script should check if the version of the compiled game matches the tag name.
This mistake happened both for version 0.0.1 and version 0.1.1 which incorrectly print 0.0.0 and 0.1.0 in the console.
This is a human error and should be checked in the CI.

I guess another option would be to automatically generate the version from the tag, if we are in a release build.

Note that this should ignore everything after a `-` in the tag name, that part is useful for pre-releases.

BoySanic:
Fixes #2508 

Not entirely sure the best way to test this, or how we want to handle the default case where someone doesn't provide a version number

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,6 @@ jobs:
       - name: Create release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          tag: ${{ github.ref_name }}
         run: |
           gh release create "$tag" \
               --repo="$GITHUB_REPOSITORY" \


BoySanic:
-Dversion provided:
image
image
No -Dversion provided:
image
image

BoySanic:
> While this does produce the correct version number for the pre-built binaries, it will produce the wrong version you download the sources and self-compile.
> 
> Instead of using the version number, please compare it with the one from the argument (mind that it can contain extra stuff at the end, like -pre1).

Hmm, not sure I get it.
Fail the compile if the provided version (in arguments) doesn't match the version in build.zig?
I'm not sure what the compare you're suggesting should do otherwise

IntegratedQuantum:
yes, fail the compilation if the tag version doesn't match the version in the build.zig

BoySanic:
I'll try to simplify this a bit, but this at least functions

BoySanic:
With what I wrote currently, if I provide the build script with `-Dversion 0.2.0-pre1` it'll still build with `0.2.0-dev` (debug_linux.sh) but it won't fail when the tag specifies a different suffix.

Right now I treat it as though the version we care about is whatever comes before the 1st `-` in the version number. It ignores everything after the first `-` assuming it's irrelevant to the version number

BoySanic:
Okay this is a lot cleaner than my initial attempts. Should be good to go

build.zig
```diff
@@ -180,6 +180,27 @@ pub fn build(b: *std.Build) !void {
 	const options = b.addOptions();
 	const isRelease = b.option(bool, "release", "Removes the -dev flag from the version") orelse false;
 	const version = b.fmt("0.2.0{s}", .{if (isRelease) "" else "-dev"});
+	if (b.option([]const u8, "version", "tagged version for CI")) |tagVersion| {
+		const tagVersionUpperbound: usize = if (std.mem.count(u8, tagVersion, "-") > 0) std.mem.indexOfScalar(u8, tagVersion, '-').? - 1 else tagVersion.len - 1;
+		const versionUpperbound: usize = if (std.mem.count(u8, version, "-") > 0) std.mem.indexOfScalar(u8, version, '-').? - 1 else version.len - 1;
+		var tagSplit = std.mem.splitSequence(u8, tagVersion[0..tagVersionUpperbound], ".");
+		var versionSplit = std.mem.splitSequence(u8, version[0..versionUpperbound], ".");
```
IntegratedQuantum: Why not use std.SemanticVersion.order() to compare them?

BoySanic: I wasn't aware of it :p I'll give it a shot

BoySanic: This mostly works, but I'll still need my `tagVersionUpperbound` and `versionUpperbound` to remove anything prepended with `-` at the end because:

image

IntegratedQuantum: In that case you can also manually compare the fields of SemanticVersion instead of using their compare function.

BoySanic: Certainly cleaner. All done

BoySanic: > In that case you can also manually compare the fields of SemanticVersion instead of using their compare function.

I can try that. It may not be any cleaner than my solution though

BoySanic: I would actually argue against manually comparing the fields since we'd be special casing things that come after the `-` in each field rather than just ignoring them outright.
As-implemented it doesn't even parse the version beyond the `-`, so we can use the builtin order function

build.zig
```diff
@@ -180,6 +180,16 @@ pub fn build(b: *std.Build) !void {
 	const options = b.addOptions();
 	const isRelease = b.option(bool, "release", "Removes the -dev flag from the version") orelse false;
 	const version = b.fmt("0.2.0{s}", .{if (isRelease) "" else "-dev"});
+	if (b.option([]const u8, "version", "tagged version for CI")) |tagVersion| {
```
```suggestion
	if (b.option([]const u8, "version", "used by the CI to check if the git tag and game version match")) |tagVersion| {
```

build.zig
```diff
@@ -180,6 +180,16 @@ pub fn build(b: *std.Build) !void {
...
+		const tagVersionUpperbound: usize = if (std.mem.count(u8, tagVersion, "-") > 0) std.mem.indexOfScalar(u8, tagVersion, '-').? else tagVersion.len;
```
```suggestion
		const tagVersionUpperbound: usize = std.mem.indexOfScalar(u8, tagVersion, '-') orelse tagVersion.len;
```

build.zig
```diff
@@ -180,6 +180,16 @@ pub fn build(b: *std.Build) !void {
...
+		const versionUpperbound: usize = if (std.mem.count(u8, version, "-") > 0) std.mem.indexOfScalar(u8, version, '-').? else version.len;
```
```suggestion
		const versionUpperbound: usize = std.mem.indexOfScalar(u8, version, '-') orelse version.len;
```

build.zig
```diff
@@ -180,6 +180,16 @@ pub fn build(b: *std.Build) !void {
...
+		const tagParsed = try std.SemanticVersion.parse(tagVersion[0..tagVersionUpperbound]);
+		const versionParsed = try std.SemanticVersion.parse(version[0..versionUpperbound]);
+		if (std.SemanticVersion.order(tagParsed, versionParsed) != .eq) {
+			std.log.err("Tagged version {s} does not match version in build.zig: {s}", .{tagVersion, version});
```
```suggestion
			std.log.err("Provided version {s} does not match version in build.zig: {s}", .{tagVersion, version});
```

BoySanic:
Done

.github/workflows/release.yml
IntegratedQuantum: Please also add this to the temporarily disabled line

BoySanic: Done

IntegratedQuantum: I don't think you can use `$tag` above, since it's only created down here.

IntegratedQuantum: it would be good if you could test the CI script by running it on your master branch in your fork.

BoySanic: You were indeed correct. That should be fixed now, but feel free to wait on the "one that should work" to finish:

https://github.com/BoySanic/Cubyz/releases/tag/0.2.1
https://github.com/BoySanic/Cubyz/releases/tag/0.2.0

BoySanic: https://github.com/BoySanic/Cubyz/actions/runs/23160995319 - 0.2.1
https://github.com/BoySanic/Cubyz/actions/runs/23160985254 - 0.2.0

BoySanic: Image

Well, I didn't create the release properly, I guess. I'm not sure how to do that. I just created the release and tag together and the CI ran.

Anyway I think maybe I can remove the 2nd reference to `tag` in the 2nd env block. Not sure it's even separate other than limiting the scope of what commands have that environment variable

PR #2728, opened 2026-03-16
Branch rebalance

### LINKED ISSUE #2573: Branches are kind of bad for what they cost, this is also potentially confusing to newcomers

BoySanic:
Fixes #2573

I buffed glimmergill's `durable` modifier here too because I felt if I give branches a durable modifier that's slightly higher than planks, glimmergill should stand out.

I can also see an argument for toadstool to be buffed but I didn't touch that for the moment.

### DIFF
--- a/assets/cubyz/blocks/branch/bolete.zig.zon
+++ b/assets/cubyz/blocks/branch/bolete.zig.zon
@@ -20,7 +20,7 @@
 			.modifiers = .{
 				.{
 					.id = "durable",
-					.strength = 0.4,
+					.strength = 0.45,
 				},
 			},
 		},
--- a/assets/cubyz/blocks/branch/toadstool.zig.zon
+++ b/assets/cubyz/blocks/branch/toadstool.zig.zon
@@ -20,7 +20,7 @@
 			.modifiers = .{
 				.{
 					.id = "durable",
-					.strength = 0.4,
+					.strength = 0.45,
 				},
 			},
 		},


BoySanic:
Birch plank:
image
Birch branch (old):
image

Birch branch (new):
image

assets/cubyz/blocks/branch/baobab.zig.zon
```diff
@@ -12,18 +12,18 @@
 	.texture5 = "cubyz:branch/baobab/cross",
 	.item = .{
 		.material = .{
-			.durability = 20,
-			.massDamage = 0.4,
+			.durability = 25,
+			.massDamage = 1.2,
 			.hardnessDamage = 0.4,
```
IntegratedQuantum: should be 0.8

assets/cubyz/blocks/branch/baobab.zig.zon
```diff
@@ -12,18 +12,18 @@
...
-			.swingSpeed = 6.3,
+			.swingSpeed = 4.0,
```
IntegratedQuantum: Please add some variation in these numbers that matches the planks, e.g. birch planks are faster than baobab planks, but your branches are not.

IntegratedQuantum (changes requested):
Please also do the comparison of a full wood pickaxe

BoySanic:
Just to try it, birch branch only:
Old:
image

New:
image

Birch branches for the handle (3), planks for the head:
Old:
image

New:
image

BoySanic:
I'll try some of the branches with more mass damage to see the extreme end of the scale, but birch branch pickaxe can break copper ore with this iteration of changes

BoySanic:
image
slate head, birch branch handle. Same damage as plank handle with slate head. similar (but still higher) durability to full-branch pick. But you get more damage with slate to get better ores.

IntegratedQuantum:
> but birch branch pickaxe can break copper ore with this iteration of changes

Maybe to counteract this it would be good to increase hardness damage, that should at least debuff the pure branch pickaxe.

And if you engineer your way to a copper breaking pickaxe made from wood, then that wouldn't be terrible either I guess.

BoySanic:
I can increase hardness damage sure. 

I was also going to suggest that pure-branch picks being able to break copper isn't completely game breaking, as they have worse durability (slightly) than slate and far worse mass damage.

It's possible someone might skip from wood to copper this way, but I think this would be a one-time event for someone seeing as slate is so much better for far less resources when you're already caving and don't have access to more branches to replace your branch pickaxe

IntegratedQuantum:
Oh sorry, I meant decrease

IntegratedQuantum:
I'm still waiting for your to decrease the hardness of the branches, while not strictly required for balancing, I think this makes intuitive sense, as the bark is less hard.

Issue #2729, opened 2026-03-16
Non 0 refcount on user while testing with two local players in one world


Wunka:
### Steps to Reproduce and Observed Behavior

I can not pinpoint where this is coming from, and before I go into madness here is an issue:

Steps to Reproduce:
1. With the first client create a world (testing or not doens't influence the outcome)
2. Join the world with the first client
3. join `localhost` with the second client.
4. leave the world with the second client / kick the second client with the first client... (both options cause this. I was testing my #2723 implementation when I noticed this)
5. Leave the world with the first client

Observed Behavior:
```shell
thread 5552 panic: reached unreachable code
/home/Wunka/Documents/Cubyz/master/compiler/zig/lib/std/debug.zig:417:14: 0x1470add in assert (Cubyz)
    if (!ok) unreachable; // assertion failure
             ^
/home/Wunka/Documents/Cubyz/master/src/server/server.zig:161:19: 0x1654959 in deinit (Cubyz)
  std.debug.assert(self.refCount.load(.monotonic) == 0);
                  ^
/home/Wunka/Documents/Cubyz/master/src/server/server.zig:546:14: 0x158c1a0 in deinit (Cubyz)
  user.deinit();
             ^
/home/Wunka/Documents/Cubyz/master/src/server/server.zig:663:14: 0x154976b in startFromExistingThread (Cubyz)
 defer deinit();
             ^
/home/Wunka/Documents/Cubyz/master/src/server/server.zig:657:25: 0x1906abf in startFromNewThread (Cubyz)
 startFromExistingThread(name, port);
                        ^
/home/Wunka/Documents/Cubyz/master/compiler/zig/lib/std/Thread.zig:558:13: 0x1821a9e in callFn__anon_128433 (Cubyz)
            @call(.auto, f, args);
            ^
/home/Wunka/Documents/Cubyz/master/compiler/zig/lib/std/Thread.zig:830:30: 0x17294cf in entryFn (Cubyz)
                return callFn(f, args_ptr.*);
                             ^
???:?:?: 0x7f1a1e6dd979 in ??? (/usr/lib/libc.so.6)
???:?:?: 0x7f1a1e7612bb in ??? (/usr/lib/libc.so.6)
./debug_linux.sh: line 31:  5451 Aborted                    (core dumped) ./zig-out/bin/Cubyz
```

Expected Behavior:
No crash...

IntegratedQuantum:
[...]

Issue #2594, opened 2026-02-24
Enable manage player window for headless servers


Wunka:
### Description

With #2530 "moderators" could now be given the right permissions to kick people, but for that there first needs a way to do that without the manage player window. 
So we either enable the window for people with certain permissions or add commands that do the same.

IntegratedQuantum:
> So we either enable the window for people with certain permissions or add commands that do the same.

I think we should have both

H41ogen:
first we'd need to be able to read stdin, which i dont think we have yet lol

Wunka:
I think a good way to achive both is to: 
1. create the /kick command. 
2. rewrite the manage player window to use the command in the background. 
3. after #2588 make the window only appear for clients who have the permission (before for everyone but the kick button would not work because of permissions).

If no one has anything against this idea. I would create sub-issues for each step

IntegratedQuantum:
Yes, I like this idea. Though we should try to reduce some of the feddback in that case, so it only prints the errors into the terminal.

Wunka:
> Yes, I like this idea. Though we should try to reduce some of the feddback in that case, so it only prints the errors into the terminal.

I don't completely agree on that point. I do agree that we should limit messages like "User ran ... command" but in my opinion things like "User has no permission to use /kick" error messages should still appear in the chat. Also things like "User ... has been kicked". 
Why should a player who just wants to kick someone and doesn't know why the Button doesn't work look into the Terminal / logs?

IntegratedQuantum:
That's what I said:

> so it only prints the errors

Wunka:
> so it only prints the errors into the terminal.

Ups. Sorry. I misunderstood the "into the terminal", in that sense nothing would printed into the chat

Argmaster:
I don't know the current status quo but after some Rust (game) experience I would say that information about someone being kicked or banned should be global, to work as a wall of shame. There is nothing better than seeing a cheater being banned after they just killed you. 

In headless I would want to see a full global chat feed for practical reasons (chat moderation w/o being on the server), but in long term It would be important to be able to separate global chat messages from command I/O (through sockets) for web UIs which will eventually appear.

Also I think in long term it might be important to preserve a full activity log of players for at least few minutes into the past to allow retrospective review of suspicious activity. Out of scope, but might affect the choices here.

I think Linux logs somewhere attempts to use commands that require root access somewhere, or at least PCs at my university always threatened to do so 😅

PR #2731, opened 2026-03-16
add /kick command

### LINKED ISSUE #2723: /kick command
proposed syntax: `/kick @<playerId>`

This command should do the same as the manage player window.
(as its also supposed to be later used for the window)

Wunka:
I have sadly problems with testing because of #2729, but from what I can see the command is working. 
Closes: #2723
Here I send to the general chat that a player has been kicked. But it can of course be discussed if it should happen.  (Taken form https://github.com/PixelGuys/Cubyz/issues/2594#issuecomment-4066723964)


IntegratedQuantum (approved):
Seems to work fine for self-kicking at least.

Issue #2419, opened 2025-12-20
joints and links for entity models


Wunka:
To apply inverse kinematic or other types of animation #87 we need some kind of skeleton for our models.

This issue is first for collecting ideas. 

in my (little) knowledge we could either go with a different format then .obj for the models
These formats could then be formats that include the skeleton.

Or we could do a separate definition of the skeleton with all joints and links (.zon?). This would mean we don't have to abandon the .obj models

Bloxxel64:
i'd say whatever we do we need to have blockbench support, perhaps we should support .json models?

Bloxxel64:
i could look into how the formatting works

Wunka:
What do you guys think of something like this where we define the joints and bones through a zon file and then assign the faces of the models to bones to then later transform them. Here I used the `controls` for this. The groups come then from the obj files and .`all` means all faces from the group:
```
.{
	.joints = {
		.{
			.name = "root",
			.position = .{0.0, 1.0, 0.0},
		},
		.{
			.name = "head",
			.position = .{0.0, 2.0, 0.0},
		},
	},
	.bones = {
		.{
			.from = "root",
			.to = "head",
			.controls = .{
				.eyestalks = .all,
				.head = .all,
			},
		},
	},
}
```

We would need a little more dafa for each joint and bone when we want to apply inverse kinematics like how much the joints can move and which are fixed certain ways but that can come later.

RanPix:
keep in mind that we are probably going to use gltf format for entity models in future which allows us to store them in binary representation with them also having animations included https://github.com/PixelGuys/Cubyz-libs/pull/6 https://github.com/PixelGuys/Cubyz/pull/2106

Wunka:
> keep in mind that we are probably going to use gltf format for entity models in future which allows us to store them in binary representation with them also having animations included [PixelGuys/Cubyz-libs#6](https://github.com/PixelGuys/Cubyz-libs/pull/6) [#2106](https://github.com/PixelGuys/Cubyz/pull/2106)

That is of course much better. I thought we wanted to stay with .obj for the models. Then this issue will probably be resolved when the model of the snals is switched

PR #2733, opened 2026-03-17
Switch entity ssbo quad models to triangle meshes and draw them with normal draw calls

RanPix:
needed as we are switching from obj models to gltf/glb meshes which would allow us to easily load meshes from this new format as it is only triangle format, and have node hierarchy and animations

progress towards #2419 and #87

### DIFF
--- a/src/entityModel.zig
+++ b/src/entityModel.zig
@@ -35,7 +35,7 @@ pub const EntityModel = struct {
 
 	pub fn initFromObj(modelPath: []const u8, texturePath: []const u8) EntityModel {
 		const modelFile = main.files.cwd().read(main.stackAllocator, modelPath) catch |err| blk: {
-			std.log.err("Error while reading player model: {s}", .{@errorName(err)});
+			std.log.err("Error while reading player model from path {s}: {s}", .{modelPath, @errorName(err)});
 			break :blk &.{};
 		};
 		defer main.stackAllocator.free(modelFile);


src/models.zig
```diff
@@ -899,3 +900,85 @@ pub fn uploadModels() void {
...
+
+	const EntityVertex = struct {
+		pos: [3]f32,
+		normal: [3]f32,
+		uv: [2]f32,
+		textureSlot: u32,
+		opaqueInLod: u32 = 0,
+	};
+
+	pub fn initFromQuads(quadInfos: []main.models.QuadInfo) EntityModel {
```
IntegratedQuantum: I don't think the callsite should know this (current) implementation detail. In the future this probably get the triangles directly through gltf anyways, right?

Please give it the path or file contents instead.

src/models.zig
```diff
@@ -899,3 +900,85 @@ pub fn uploadModels() void {
 	quadSSBO = graphics.SSBO.initStatic(QuadInfo, quads.items);
 	quadSSBO.bind(4);
 }
+
+pub const EntityModel = struct {
+	vao: c_uint,
+	vbo: c_uint,
+	ebo: c_uint,
+	size: u32,
```
IntegratedQuantum: Would it make sense to also store the texture here?

src/models.zig
```diff
@@ -899,3 +900,85 @@ pub fn uploadModels() void {
...
+		normal: [3]f32,
+		uv: [2]f32,
+		textureSlot: u32,
+		opaqueInLod: u32 = 0,
+	};
+
+	pub fn initFromQuads(quadInfos: []main.models.QuadInfo) EntityModel {
+		var vertices = main.stackAllocator.alloc(EntityVertex, quadInfos.len*4);
+		defer main.stackAllocator.free(vertices);
+		var indices: []u32 = main.stackAllocator.alloc(u32, quadInfos.len*6);
```
IntegratedQuantum:
```suggestion
		const indices: []u32 = main.stackAllocator.alloc(u32, quadInfos.len*6);
```

src/models.zig
```diff
@@ -899,3 +900,85 @@ pub fn uploadModels() void {
...
+	const EntityVertex = struct {
+		pos: [3]f32,
+		normal: [3]f32,
+		uv: [2]f32,
+		textureSlot: u32,
+		opaqueInLod: u32 = 0,
+	};
+
+	pub fn initFromQuads(quadInfos: []main.models.QuadInfo) EntityModel {
+		var vertices = main.stackAllocator.alloc(EntityVertex, quadInfos.len*4);
```
IntegratedQuantum:
```suggestion
		const vertices = main.stackAllocator.alloc(EntityVertex, quadInfos.len*4);
```

src/models.zig
```diff
@@ -899,3 +900,85 @@ pub fn uploadModels() void {
...
+		for (quadInfos) |quad| {
+			inline for (0..4) |i| {
+				const v = cur + @as(u32, @intCast(i));
+				vertices[v].normal = quad.normal;
+				vertices[v].textureSlot = quad.textureSlot;
+				vertices[v].opaqueInLod = quad.opaqueInLod;
+				vertices[v].pos = quad.corners[i];
+				vertices[v].uv = quad.cornerUV[i];
+			}
+			cur += 4;
```
IntegratedQuantum: Why not do `cur += 1` in the inner loop?

src/models.zig
```diff
@@ -899,3 +900,85 @@ pub fn uploadModels() void {
...
+		c.glGenBuffers(1, &ebo);
+
+		const vertSize = @sizeOf(EntityVertex);
+
+		c.glBindBuffer(c.GL_ARRAY_BUFFER, vbo);
+		c.glBufferData(c.GL_ARRAY_BUFFER, @intCast(vertices.len*vertSize), @ptrCast(vertices), c.GL_STATIC_DRAW);
+		c.glBindBuffer(c.GL_ELEMENT_ARRAY_BUFFER, ebo);
+		c.glBufferData(c.GL_ELEMENT_ARRAY_BUFFER, @intCast(indices.len*@sizeOf(u32)), @ptrCast(indices), c.GL_STATIC_DRAW);
+
+		c.glVertexAttribPointer(0, 3, c.GL_FLOAT, c.GL_FALSE, vertSize, @ptrFromInt(0));
```
IntegratedQuantum: I would suggest to use `@fieldParentPtr` to extract this information from the struct, instead of hardcoding these offsets. This makes it easier to catch mistakes.

src/models.zig
```diff
@@ -899,3 +900,85 @@ pub fn uploadModels() void {
...
+	vbo: c_uint,
+	ebo: c_uint,
+	size: u32,
+
+	const EntityVertex = struct {
+		pos: [3]f32,
+		normal: [3]f32,
+		uv: [2]f32,
+		textureSlot: u32,
+		opaqueInLod: u32 = 0,
```
IntegratedQuantum: This is irrelevant, it is an optimization for blocks, entities won't need this.

src/models.zig
```diff
@@ -899,3 +900,85 @@ pub fn uploadModels() void {
...
+	vao: c_uint,
+	vbo: c_uint,
+	ebo: c_uint,
+	size: u32,
+
+	const EntityVertex = struct {
+		pos: [3]f32,
+		normal: [3]f32,
+		uv: [2]f32,
+		textureSlot: u32,
```
IntegratedQuantum: this one also isn't used in the shader

IntegratedQuantum (changes requested):

src/models.zig
RanPix: its used inside the inner loop and messes up the indexing

src/models.zig
RanPix: i think that should be in a different pr, where we add multiple entity models #2641

src/models.zig
RanPix: `@intFromPtr(&v.normal) - @intFromPtr(@as(*EntityVertex, @fieldParentPtr("normal", &v.normal)))`

like that?

src/models.zig
IntegratedQuantum: From what I see it would not make any functional difference

src/models.zig
IntegratedQuantum: Yeah, or maybe we can somehow get Zig to accept a null pointer as the input.

RanPix: i remember it said that if the pointer is invalid it will be undefined behaviour as it tries to get the offset from the pointer or whatever

src/models.zig
RanPix: well i tried that and it crashed instantly

src/models.zig
RanPix: looks good now?

src/models.zig
IntegratedQuantum: Actually I think the other way might be better, this should work:
`&@as(*allowzero EntityVertex, @ptrFromInt(0)).normal`

src/models.zig
IntegratedQuantum: How?

src/models.zig
RanPix: nvm i think that might be a good idea

src/models.zig
RanPix:
```zig
inline for (0..4) |i| {
  const v = cur + @as(u32, @intCast(i));
  vertices[v].normal = quad.normal;
  vertices[v].pos = quad.corners[i];
  vertices[v].uv = quad.cornerUV[i];
  cur += 1;
}
```

```
thread 11316 panic: index out of bounds: index 264, len 264
D:\Files\Dev\Zig\Cubyz\src\models.zig:934:13: 0x7ff6838c6eb9 in initFromObj (Cubyz_zcu.obj)
    vertices[v].normal = quad.normal;
            ^
```
does it really matter where its placed? also if you place it in the inside loop then its going to mess up the indices

src/models.zig
```diff
@@ -899,3 +900,94 @@ pub fn uploadModels() void {
...
+			.size = @intCast(indices.len),
+			.texture = main.client.entity_manager.missingModelTexture,
+		};
+	}
+
+	pub fn deinit(self: EntityModel) void {
+		c.glDeleteVertexArrays(1, &self.vao);
+		c.glDeleteBuffers(1, &self.vbo);
+		c.glDeleteBuffers(1, &self.ebo);
+		self.texture.deinit();
```
IntegratedQuantum: This is a double free is self.texture is missingModelTexture.

src/models.zig
```diff
@@ -899,3 +900,94 @@ pub fn uploadModels() void {
...
+		const lut = [_]u32{0, 2, 1, 1, 2, 3};
+		for (0..indices.len) |i| {
+			indices[i] = @as(u32, @intCast(i))/6*4 + lut[i%6];
+		}
+
+		return uploadMeshAndGetModel(vertices, indices);
+	}
+
+	pub fn loadTextureFromFile(self: *EntityModel, path: []const u8) void {
+		self.texture = main.graphics.Texture.initFromFile(path);
```
IntegratedQuantum: This should be done in the init function as well to avoid having it set to undefined in-between.

src/models.zig
IntegratedQuantum: To avoid this Id suggest to just reload the missing texture for every model when the texture was not found.

src/client/entity_manager.zig
```diff
@@ -46,17 +45,10 @@ pub fn init() void {
...
-		std.log.err("Error while reading player model: {s}", .{@errorName(err)});
-		break :blk &.{};
-	};
-	defer main.stackAllocator.free(modelFile);
-	const quadInfos = main.models.Model.loadRawModelDataFromObj(main.stackAllocator, modelFile);
-	defer main.stackAllocator.free(quadInfos);
-	modelBuffer = .initStatic(main.models.QuadInfo, quadInfos);
-	modelBuffer.bind(11);
-	modelSize = @intCast(quadInfos.len);
+	missingModelTexture = main.graphics.Texture.initFromFile("assets/cubyz/entities/textures/missing.png");
```
IntegratedQuantum: this is leaking

src/models.zig
```diff
@@ -899,3 +900,94 @@ pub fn uploadModels() void {
...
+	texture: main.graphics.Texture = undefined,
+
+	const EntityVertex = struct {
+		pos: [3]f32,
+		normal: [3]f32,
+		uv: [2]f32,
+	};
+
+	pub fn initFromObj(allocator: main.heap.NeverFailingAllocator, path: []const u8) EntityModel {
+		const modelFile = main.files.cwd().read(allocator, path) catch |err| blk: {
```
IntegratedQuantum: Local allocations should always use the stackAllocator.
An allocator should only be passed into the function if the result needs to be allocated.

src/client/entity_manager.zig
```diff
@@ -157,7 +151,7 @@ pub fn render(projMatrix: Mat4f, ambientLight: Vec3f, playerPos: Vec3d) void {
 			.mul(Mat4f.rotationZ(-ent.rot[2])));
 		const modelViewMatrix = game.camera.viewMatrix.mul(modelMatrix);
 		c.glUniformMatrix4fv(uniforms.viewMatrix, 1, c.GL_TRUE, @ptrCast(&modelViewMatrix));
-		c.glDrawElements(c.GL_TRIANGLES, 6*modelSize, c.GL_UNSIGNED_INT, null);
+		c.glDrawElements(c.GL_TRIANGLES, @intCast(model.size), c.GL_UNSIGNED_INT, null);
```
IntegratedQuantum: I'd suggest to store it with the right type to avoid the cast here.

Also instead of size I'd suggest to call it indexCount, since size can mean several different things for models.

src/models.zig
```diff
@@ -899,3 +900,94 @@ pub fn uploadModels() void {
...
 }
+
+pub const EntityModel = struct {
+	vao: c_uint,
+	vbo: c_uint,
+	ebo: c_uint,
+	size: u32,
+	texture: main.graphics.Texture = undefined,
+
+	const EntityVertex = struct {
```
IntegratedQuantum: Should be extern since you want to share it with OpenGL.

src/models.zig
```diff
@@ -899,3 +900,94 @@ pub fn uploadModels() void {
...
+		const quadInfos = main.models.Model.loadRawModelDataFromObj(allocator, modelFile);
+		defer allocator.free(quadInfos);
+
+		const vertices = allocator.alloc(EntityVertex, quadInfos.len*4);
+		defer allocator.free(vertices);
+		const indices: []u32 = allocator.alloc(u32, quadInfos.len*6);
+		defer allocator.free(indices);
+
+		for (quadInfos, 0..quadInfos.len) |quad, i| {
+			inline for (0..4) |j| {
```
IntegratedQuantum: Doesn't need to be inline

src/models.zig
```diff
@@ -899,3 +900,94 @@ pub fn uploadModels() void {
...
+				const v = (i*4) + @as(u32, @intCast(j));
```
IntegratedQuantum: unnecessary cast, i*4 is already usize, so is j
```suggestion
				const v = i*4 + j;
```

src/models.zig
```diff
@@ -899,3 +900,94 @@ pub fn uploadModels() void {
...
+
+		return uploadMeshAndGetModel(vertices, indices);
+	}
+
+	pub fn loadTextureFromFile(self: *EntityModel, path: []const u8) void {
+		self.texture = main.graphics.Texture.initFromFile(path);
+	}
+
+	fn uploadMeshAndGetModel(vertices: []EntityVertex, indices: []u32) EntityModel {
+		var vao: c_uint = 0;
```
IntegratedQuantum: Please use the new `graphics.VertexArray` struct that I just introduced in #2811 to make this code a little more vulkan-friendly.

src/models.zig
RanPix: the function that loads the texture from file returns a default missing texture from image, is that ok?

src/client/entity_manager.zig
```diff
@@ -131,15 +120,17 @@ pub fn render(projMatrix: Mat4f, ambientLight: Vec3f, playerPos: Vec3d) void {
...
+	model.vao.bind();
 	c.glUniformMatrix4fv(uniforms.projectionMatrix, 1, c.GL_TRUE, @ptrCast(&projMatrix));
-	modelTexture.bindTo(0);
 	c.glUniform3fv(uniforms.ambientLight, 1, @ptrCast(&ambientLight));
 	c.glUniform1f(uniforms.contrast, 0.12);
 
 	for (entities.items()) |ent| {
 		if (ent.id == game.Player.id) continue; // don't render local player
 
+		model.texture.bindTo(0);
```
IntegratedQuantum: I'd suggest to have a single model.bind()

RanPix:
done

src/entityModel.zig
```diff
@@ -0,0 +1,82 @@
+const std = @import("std");
+
+const main = @import("main");
+const graphics = main.graphics;
+const c = graphics.c;
+
+pub const EntityModel = struct {
```
IntegratedQuantum, line 7: Do you intend to have any other structs in this file? If not I'd suggest to turn this into a file as a struct.

src/entityModel.zig
```diff
@@ -0,0 +1,82 @@
...
+				.location = 2,
+				.format = c.VK_FORMAT_R32G32_SFLOAT,
+				.offset = @offsetOf(@This(), "uv"),
+			},
+		};
+	};
+
+	pub fn initFromObj(modelPath: []const u8, texturePath: []const u8) EntityModel {
+		const modelFile = main.files.cwd().read(main.stackAllocator, modelPath) catch |err| blk: {
+			std.log.err("Error while reading player model: {s}", .{@errorName(err)});
```
IntegratedQuantum: Always mention the path to make it easier for addon creators to find the issue.
```suggestion
			std.log.err("Error while reading player model from path {s}: {s}", .{modelPath, @errorName(err)});
```

src/entityModel.zig
tillpp: He might not, but I do.

tillpp: https://github.com/tillpp/Cubzio/blob/17fce9b81ed3e7ce57c32ba12da2180a614b36df/src/entityModel.zig

here is a peek, of what i am planning for this file at the end of my ECS series

RanPix:
done

IntegratedQuantum (approved):

PR #2739, opened 2026-03-18
Fix particle and item drop collision

HTRMC:
## Fix particle/item collision behaving weird on surfaces

Particles and dropped items were resolving collisions in the wrong axis order,
which caused them to snap to block edges instead of landing properly.

Now we resolve Z gravity the gravity axis first, so they fall and settle
correctly. Also zero out velocity on collision so they don't keep pushing into blocks.

Added some friction when they hit the ground so they don't slide forever.

Tested by breaking blocks and dropping items they now land and stay put.

### DIFF
--- a/src/itemdrop.zig
+++ b/src/itemdrop.zig
@@ -350,7 +350,7 @@ pub const ItemDropManager = struct { // MARK: ItemDropManager
 
 	fn updateEnt(self: *ItemDropManager, chunk: *ServerChunk, pos: *Vec3d, vel: *Vec3d, deltaTime: f64) void {
 		const collision = main.game.collision;
-		const hitBox = collision.Box{ .min = @splat(-radius), .max = @splat(radius) };
+		const hitBox = collision.Box{.min = @splat(-radius), .max = @splat(radius)};
 		if (collision.collides(.server, .x, 0, pos.*, hitBox) != null) {
 			self.fixStuckInBlock(chunk, pos, vel, deltaTime);
 			return;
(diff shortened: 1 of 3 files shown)


tillpp:
> Particles and dropped items were resolving collisions in the wrong axis order,
> which caused them to snap to block edges instead of landing properly.
> 
> Now we resolve Z gravity the gravity axis first, so they fall and settle
> correctly. 
> 



not sure what issue you are refering to, can you send a video recording that shows the problem or link the issue?

HTRMC:
> > Particles and dropped items were resolving collisions in the wrong axis order,
> > which caused them to snap to block edges instead of landing properly.
> > Now we resolve Z gravity the gravity axis first, so they fall and settle
> > correctly.
> 
> not sure what issue you are refering to, can you send a video recording that shows the problem or link the issue?

If you pull the block breaking particle branch u can see that collisions for particles are completly bugging out and they snap to the edges of blocks. ive had a similar issue in my own engine which i resolved by doing the gravity axis first which is Z for cubyz

src/itemdrop.zig
```diff
@@ -358,15 +358,19 @@ pub const ItemDropManager = struct { // MARK: ItemDropManager
...
+		inline for ([_]comptime_int{2, 0, 1}) |i| {
+			const move = vel.*[i]*deltaTime;
 			if (main.game.collision.collides(.server, @enumFromInt(i), move, pos.*, hitBox)) |box| {
 				if (move < 0) {
 					pos.*[i] = box.max[i] + radius;
 				} else {
 					pos.*[i] = box.min[i] - radius;
 				}
 				vel.*[i] = 0;
+				if (i == 2 and move < 0) {
```
HTRMC: Its for the ground friction i was thinking it would look weird if a particle had wall friction

HTRMC:
> Before making a PR to fix something please research for issues first: #1663 This has exactly the kind of hacky short-term workarounds I wanted to avoid with my comment there: [#1663 (comment)](https://github.com/PixelGuys/Cubyz/issues/1663#issuecomment-3118763744)

My apologies for this, I wasn't aware of this comment. I'm willing to look into it would this pr be the right place for that?

HTRMC:
> this doesn't go far enough, the collision code is still duplicated, and the implementation is incomplete, surface friction is not considered.
> 
> In my opinion there should be a single function, that takes position and velocity and some config (to disable eyepos calculations) and updates the values accordingly, otherwise these implementations will again diverge once we add anything new (→#1728)
> 
> Also this PR is already getting big. In order to go there you will likely need to make multiple refactoring rounds first, I'd suggest to also check out codemobs PRs on the matter to avoid making the same mistakes.

Started the refactor in #2759. extracting PhysicsState and InputState structs so the physics functions don't depend on Player globals. Follow up PR will migrate item drops and particles to use the shared code path.

PR #2744, opened 2026-03-19
Floating Island Stalagtites

H41ogen:
Right now, the floating islands are very round and flat, so I increased the cave noise strength there and added a bunch of stalactites. Since the stalactites cover up a decent amount of the bottom of the islands, I also made jade 50% more common.
- - -
image
Screenshot_20260319_124501
Screenshot_20260319_124521

### DIFF
--- a/assets/cubyz/biomes/cave/sky_islands/_defaults.zig.zon
+++ b/assets/cubyz/biomes/cave/sky_islands/_defaults.zig.zon
@@ -90,5 +90,30 @@
 			.height = 1,
 			.height_variation = 0,
 		},
+		.{
+			.id = "cubyz:stalagmite",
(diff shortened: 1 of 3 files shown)


careeoki (changes requested):
Personally I like the more round look the islands have right now, I think something like 30% as many stalagmites as you have here would look good. Ideally, you should add these without needing to change the rarity of jade.

ZigguratV:
Could stalagmite islands be something like a biome? Also is there a technical limitation preventing them from containing jade?

Argmaster:
> Could stalagmite islands be something like a biome?

+1

careeoki:
Stalactites were added to the islands in #2912 , so either close this PR or turn this into a separate biome which is extra spiky.

PR #2757, opened 2026-03-20
Better notification when password is incorrect

### LINKED ISSUE #2750: Better notification when password is incorrect
"the game encountered errors", pointing at "AuthenticationFailed" in the logs is not particularly helpful for the average user.

H41ogen:
Fixes #2750

image

### DIFF
--- a/src/gui/windows/authentication/unlock.zig
+++ b/src/gui/windows/authentication/unlock.zig
@@ -33,7 +33,9 @@ fn apply() void {
 		if (err == error.AuthenticationFailed) {
 			incorrectPasswordLabel.updateText("#ff0000Incorrect password.");
 		} else {
-			incorrectPasswordLabel.updateText("#ff0000Encrypted authentication data is corrupted: {s}. Please login with your Account Code again.", .{@errorName(err)});
+			const formattedError = std.fmt.allocPrint(main.stackAllocator.allocator, "#ff0000Authentication data is corrupted: {s}", .{@errorName(err)}) catch unreachable;
+			defer main.stackAllocator.free(formattedError);
+			incorrectPasswordLabel.updateText(formattedError);
 		}
 		return;
 	};


src/gui/windows/authentication/unlock.zig
```diff
@@ -24,11 +24,20 @@ const padding: f32 = 8;
...
+var incorrectPassword = false;
+
 fn apply() void {
 	var failureText: main.List(u8) = .init(main.stackAllocator);
 	defer failureText.deinit();
 	const accountCode = main.settings.storedAccount.decryptFromPassword(textComponent.currentString.items, &failureText) catch |err| {
-		std.log.err("Encountered error while decrypting password: {s}", .{@errorName(err)});
+		if (err == error.AuthenticationFailed) {
+			incorrectPassword = true;
+			onClose();
```
IntegratedQuantum: This is a data corruption, onClose will clean up the components, which are currently being iterated, as soon as the function returns, you are accessing invalid memory from e.g. the Button that called this function.

Instead store the label in a global and change only the label text.

IntegratedQuantum:
also please link the issue this fixes.

src/gui/windows/authentication/unlock.zig
H41ogen: fixed

src/gui/windows/authentication/unlock.zig
```diff
@@ -24,11 +24,17 @@ const padding: f32 = 8;
...
+
 fn apply() void {
 	var failureText: main.List(u8) = .init(main.stackAllocator);
 	defer failureText.deinit();
 	const accountCode = main.settings.storedAccount.decryptFromPassword(textComponent.currentString.items, &failureText) catch |err| {
-		std.log.err("Encountered error while decrypting password: {s}", .{@errorName(err)});
+		if (err == error.AuthenticationFailed) {
+			incorrectPasswordLabel.updateText("#ff0000Incorrect password.");
+		} else {
+			std.log.err("Encountered error while decrypting password: {s}", .{@errorName(err)});
```
IntegratedQuantum: This means that the stored data got corrupted. Since we already have a label for this, I think we should print it onto it, something like this:
`"#ff0000Encrypted authentication data is corrupted: {s} Please login with your Account Code again.", .{@errorName(err)}`

IntegratedQuantum:
please fix CI failure, you need to allocPrint the text before passing it to the label.

src/gui/windows/authentication/unlock.zig
H41ogen: This has the problem of now stretching across multiple lines

IntegratedQuantum: Well, you could also add a helper function to the label that formats the input to a temporary.

IntegratedQuantum (approved):
I fixed the remaining issues due to the closeness of the release

H41ogen:
That's fair, I have a lot on my plate right now so I wouldn't have gotten to this

PR #2758, opened 2026-03-20
Change player and chest inventories to 24 slots to match the hotbar

H41ogen:
image

- - -
This could also allow in the future for cool things like scrolling rows of your inventory through your hotbar.

Related to #1282


H41ogen:
I am not saying the current inventory is too small - just that its width should match the hotbar.

IntegratedQuantum:
Please note https://github.com/PixelGuys/Cubyz/issues/1282#issuecomment-2779012916

It's easy to make the inventory bigger, but it's very difficult to make it smaller without dropping all your stuff, that's why this balancing should wait.

H41ogen:
It's not balancing. It's literally just making the inventory size make sense.

IntegratedQuantum:
It is something that affects balancing. Yes I agree the current inventory size makes no sense visually, but there are also several other slots that we need to put somewhere, and the layout hasn't been decided yet.

PR #2762, opened 2026-03-20
Load entity meshes from gltf/glb files

### LINKED ISSUE #2830: Convert .obj to .glb
Since we are going to use .glb soon, could someone convert the .obj in assets/entityModels/models to .glb?
But please keep the .obj in there too for now, for legacy.

RanPix:
Requires #2733 and #2680 

Progress towards #2419 and #87

progress towards #2777
fixes #2830

So far only works in blockbench coordinate system (right handed, y-up, z-forward) converting it to our system, so for now you have to be sure you export your mesh in the right system if its from another program. Doesn't load textures so they are still stored in game assets. This is a really simple start that i will build up on in future PRs.

A few questions still remain:
- ~~Do we want to keep the legacy .obj loading?~~ (no, because in a later pr when i add joints i wont be able to add them and it will brake stuff, and thats generally useless)
- ~~Do we want to load the file data ourselves and then feed it into the cgltf or just let cgltf handle that~~ (turns out there is no problem with that so its best to do that so that we can have local allocations)
- Do we want to override cgltf allocation functions?
- ~~Do we want to try handle other possible edge cases that might happen when exporting from other applications? (like different coordinate systems and stuff)~~ (yes #2777)


RanPix:
ready for review

src/models.zig
```diff
@@ -899,3 +904,191 @@ pub fn uploadModels() void {
...
+			},
+		};
+	};
+
+	pub fn initFromGltf(modelPath: []const u8, texturePath: []const u8) !EntityModel {
+		// TODO: consider overriding cgltf_memory_options functions
+		var options: gltf.cgltf_options = .{};
+		var data: *gltf.cgltf_data = undefined;
+
+		// const file = main.files.cwd().read(main.stackAllocator, path) catch |err| blk: {
```
tillpp: commented code
I think Quantum will flag this

RanPix: its because im still not sure if its better to load the data ourselves and pass it to cgltf or let cgltf do the job

RanPix:
ready for review

src/entityModel.zig
```diff
@@ -33,40 +40,138 @@ pub const EntityModel = struct {
...
-			.indexCount = @intCast(indices.len),
+			.indexCount = 0,
+		};
+	}
+
+	fn getHierarchyMatrix(node: gltf.cgltf_node) Mat4f {
+		var currentMat = Mat4f.translation(Vec3f{
+			node.translation[0],
+			node.translation[2],
+			node.translation[1],
```
IntegratedQuantum: Please fix #2777 first, so we don't end up having two model formats that load in different orientations.

IntegratedQuantum (changes requested):

tillpp:
Can we merge the ECS 7 8 and 9 first before merging this one?
I think that would be the easier way of merging stuff

src/entityModel.zig
RanPix: then that requires #2680 to be merged

RanPix:
ready for review

assets/cubyz/entityModels/missing.zig.zon
```diff
@@ -2,4 +2,6 @@
 	.model = "cubyz:missing",
 	.defaultTexture = "cubyz:missing",
 	.height = 1,
+	.coordinateSystem = .right_handed_y_up,
+	.swapTriangleWinding = true,
```
IntegratedQuantum: If you have to do this, it probably means that you got the coordinate system wrong. I'm pretty sure all model editors use the same winding order.

src/entityModel.zig
```diff
@@ -106,33 +128,161 @@ pub const EntityModel = struct {
...
-		const indices: []u32 = main.stackAllocator.alloc(u32, quadInfos.len*6);
-		defer main.stackAllocator.free(indices);
-
-		for (quadInfos, 0..quadInfos.len) |quad, i| {
-			for (0..4) |j| {
-				const v = i*4 + j;
-				vertices[v].normal = quad.normal;
-				vertices[v].pos = quad.corners[j];
-				vertices[v].uv = quad.cornerUV[j];
+		const file = try main.assets.readAsset(main.globalAllocator, "entityModels/models", self.modelId.?, ".glb");
```
IntegratedQuantum: For local allocations, always use the stackAllocator
```suggestion
		const file = try main.assets.readAsset(main.stackAllocator, "entityModels/models", self.modelId.?, ".glb");
```

src/entityModel.zig
```diff
@@ -106,33 +128,161 @@ pub const EntityModel = struct {
...
+		defer main.globalAllocator.free(file);
+
+		var options: gltf.cgltf_options = .{};
+		var data: *gltf.cgltf_data = undefined;
+
+		var result = gltf.cgltf_parse(&options, @ptrCast(file.ptr), @intCast(file.len), @ptrCast(&data));
+		if (result == gltf.cgltf_result_file_not_found or result == gltf.cgltf_result_io_error) {
```
IntegratedQuantum: Please handle all errors with `result != cgltf_result_success`

src/entityModel.zig
```diff
@@ -106,33 +128,161 @@ pub const EntityModel = struct {
...
+		var indices = main.List(u32).init(main.stackAllocator);
+		defer indices.deinit();
+		var baseVertex: u32 = 0;
+
+		for (data.nodes, 0..data.nodes_count) |node, _| {
+			if (node.mesh != null) {
+				const finalMat = getHierarchyMatrix(node, self.coordinateSystem);
+
+				const primitives = node.mesh.*.primitives;
+				for (primitives, 0..node.mesh.*.primitives_count) |primitive, _| {
```
```suggestion
				for (primitives) |primitive| {
```

src/entityModel.zig
```diff
@@ -106,33 +128,161 @@ pub const EntityModel = struct {
...
+							gltf.cgltf_attribute_type_normal => normalAttr = attribAccessor,
+							gltf.cgltf_attribute_type_texcoord => uvAttr = attribAccessor,
+							else => continue,
+						}
+					}
+
+					for (0..positionAttr.count) |v| {
+						var p: [3]f32 = undefined;
+						_ = positionAttr.float(v, @ptrCast(&p), 3);
+						const p2 = convertCoordinateSystemVec(.{p[0], p[1], p[2]}, self.coordinateSystem);
```
IntegratedQuantum: Does this not work?
```suggestion
						const p2 = convertCoordinateSystemVec(p, self.coordinateSystem);
```

src/entityModel.zig
```diff
@@ -106,33 +128,161 @@ pub const EntityModel = struct {
...
+						const pos: vec.Vec4f = finalMat.mulVec(.{p2[0], p2[1], p2[2], 1});
+						vertSlice[v].pos = .{pos[0], pos[1], pos[2]};
```
```suggestion
						vertSlice[v].pos = vec.xyz(pos);
```

src/entityModel.zig
```diff
@@ -106,33 +128,161 @@ pub const EntityModel = struct {
...
+
+						var normal: [3]f32 = undefined;
+						_ = normalAttr.float(v, @ptrCast(&normal), 3);
+						vertSlice[v].normal = convertCoordinateSystemVec(.{normal[0], normal[1], normal[2]}, self.coordinateSystem);
```
```suggestion
						vertSlice[v].normal = convertCoordinateSystemVec(normal, self.coordinateSystem);
```

src/vec.zig
```diff
@@ -175,6 +195,56 @@ pub const Mat4f = struct { // MARK: Mat4f
 		};
 	}
 
+	pub fn rotationQuat(quat: Vec4f) Mat4f {
```
IntegratedQuantum, line 201: If you copied this then please reference the source.

src/vec.zig
```diff
@@ -10,6 +10,26 @@ pub const Vec4i = @Vector(4, i32);
...
+	v: Vec4f,
+	comptime x: Vec4fComponent,
+	comptime y: Vec4fComponent,
+	comptime z: Vec4fComponent,
+	comptime w: Vec4fComponent,
+) Vec4f {
+	return @shuffle(f32, v, undefined, [4]i32{@intFromEnum(x), @intFromEnum(y), @intFromEnum(z), @intFromEnum(w)});
+}
+
+pub inline fn andInt(v0: anytype, v1: anytype) @TypeOf(v0, v1) {
```
IntegratedQuantum: Do not make poorly named copied helper public

IntegratedQuantum: Also this should be next to the function that's using it.

IntegratedQuantum (changes requested):

src/entityModel.zig
RanPix: this check is not needed anyways

RanPix: nvm it is

src/entityModel.zig
IntegratedQuantum: If primitives is a multi-item pointer, you can use `primitives[0..count]`

src/entityModel.zig
```diff
@@ -63,6 +75,12 @@ pub const EntityModel = struct {
 		self.vao = null;
 		self.indexCount = 0;
 
+		const coordSystemName = zon.get([]const u8, "coordinateSystem", @tagName(CoordinateSystem.right_handed_z_up));
```
IntegratedQuantum: You can use `zon.get(CoordinateSystem, ...`

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #2764, opened 2026-03-21
Improve fly, ghost, and hyperSpeed toggle logic

H41ogen:
If there was a reason for the logic to be like that, let me know.


IntegratedQuantum:
The reason is that I find early returns more readable especially in more complex functions.
While it doesn't make a lot of difference here, since the scope is small, I also see no reason to change this.

Argmaster:
[Avoid nesting in your code whenever possible](https://youtu.be/CFRhGnuXG-4?si=wZQ-CldDAvhdS2dF)

PR #2769, opened 2026-03-22
Ashen biomes

H41ogen:
I don't know if people want this so I'm not putting much effort into it yet. Right now there's the `ashen/base` biome, which is now where the volcano spawns. The biome rarities are all off for now since this is just the initial idea.

In the future, I'd like to make the following sub-biomes:

- [ ] a biome similar to the magma cave
- [ ] ashen forest, the ashen logs can be crafted into coal. maybe the ashen logs are just coal blocks?
- [ ] lava lake
- [ ] an area that's starting to regrow, the ground is still mostly ash but there's some patches of grass blocks and grass vegetation, as well as lots of flowers


careeoki:
an ash block exists in #2464 by the way

H41ogen:
oh cool, i'll use that then

Issue #2529, opened 2026-01-28
Change how the workbench is synchronized and remove the shared `Inventory.Type`


IntegratedQuantum:
After #2506 there are only two shared types: `.workbench` and `.normal`
The workbench should be reworked to be more client-centric:
The server-shared inventory should only have 25 slots, the output slot should be client-only
To synchronize crafting a new inventory command should be added, as was done in #2506 for inventory crafting.
This will reduce the number of edge cases to check for in the networking code.

Wunka:
After #2530 is done, I think I will work on this next. 
The sync Command will in my first iteration (maybe we find something better) look something like this:
```
const CraftTool = struct {
	destinations: Inventories,
	craftingGrid: Inventory,
	tool: Tool,
};
```
The ClientType union gets then also an addition: `workbench: InventoryId // link to the crafting grid inventory`

IntegratedQuantum:
> tool: Tool,

I don't think the tool is needed, as it is derived from the crafting grid.

Wunka:
I did notice something now... You want to remove the share Inventory Type, but for what I see it will still be needed (for `canPutIntoWorkbench` and `shouldDepositToUserOnClose`). The crafting grid is still special and needs also on the server side special handling. So I think we will not be able to remove it completely. 

Except if we generalize all of that... Like making `shouldDepositToUserOnClose` a bool and make `canPutIntoWorkbench` into a `canPutInto` Callback? For now the craftingGrid would be the only one having the callback but I could help in the future for other things.

IntegratedQuantum:
Yes, I think generalizing it is the way to go. In the future e.g. different kinds of tools may even have different kinds of material requirements. And yeah, shouldDepositToUserOnClose also makes sense as an independent property.
Or maybe just use the already existing `onLastCloseCallback` for that.

Wunka:
I just want to mention this, as this is in the 0.2.0 milestone. I am currently not working on this. generalizing the functions. (which some need to be synced to the server from what I understand) brought me a giant headache and I cannot currently get my head around how I could best implement that. Maybe I will come back and at least write down my specific problems I had (I can't currently recall which the specific problems) but if someone else wants to tackle this they are open to.

Wunka:
[...]

PR #2770, opened 2026-03-22
move shouldDepositOnClose to callbacks

Wunka:
Step toward #2529 

I don't know where I should put the callback function and how its name should be, so input there would be nice.
Also I introduced the user parameter to the callback. I think this fits better then using the id to find the user and then get the position of the player.

### DIFF
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -248,12 +248,20 @@ pub const ServerSide = struct { // MARK: ServerSide
 				}
 				return error.Invalid;
 			},
-			.workbench => |id| {
-				if (id != user.id) {
-					std.log.err("Player {s} tried to access the inventory of another player.", .{user.name});
-					return error.Invalid;
-				}
-				callbacks.onLastCloseCallback = &workbenchCloseCallback;
+			.workbench => {
+				const workbenchCloseCallback = struct {
+					fn callback(callbackSource: Source, callbackUser: *main.server.User) void {
+						switch (callbackSource) {
+							.workbench => {
+								const workbenchInventory = getInventoryFromSource(callbackSource) orelse @panic("Could not find workbench Inventory");
+								const playerInventory = ServerSide.getInventory(callbackUser, callbackUser.inventory.?) orelse @panic("Could not find player Inventory");
+								sync.ServerSide.executeCommand(.{.depositOrDrop = .initWithInventories(&.{playerInventory}, workbenchInventory, callbackUser.player.pos)}, null);
+							},
+							else => unreachable,
+						}
+					}
+				};
+				callbacks.onLastCloseCallback = &workbenchCloseCallback.callback;
 			},
 			.other => {},
 			.alreadyFreed => unreachable,
@@ -346,17 +354,6 @@ pub const ServerSide = struct { // MARK: ServerSide
 		}
 		if (itemStack.amount == 0) itemStack.item = .null;
 	}
-
-	fn workbenchCloseCallback(source: Source, user: *main.server.User) void {
-		switch (source) {
-			.workbench => |id| {
-				const workbenchInventory = getInventoryFromSource(source) orelse @panic("Could not find workbench Inventory");
-				const playerInventory = getInventoryFromSource(.{.playerInventory = id}) orelse @panic("Could not find player inventory");
-				sync.ServerSide.executeCommand(.{.depositOrDrop = .initWithInventories(&.{playerInventory}, workbenchInventory, user.player.pos)}, null);
-			},
-			else => unreachable,
-		}
-	}
 };
 
 pub fn getInventory(id: InventoryId, side: sync.Side, user: ?*main.server.User) ?Inventory {
@@ -386,7 +383,7 @@ pub const Source = union(SourceType) {
 	playerInventory: u32,
 	hand: u32,
 	blockInventory: Vec3i,
-	workbench: u32,
+	workbench: void,
 	other: void,
 };
 
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -251,14 +251,10 @@ pub const ServerSide = struct { // MARK: ServerSide
 			.workbench => {
 				const workbenchCloseCallback = struct {
 					fn callback(callbackSource: Source, callbackUser: *main.server.User) void {
-						switch (callbackSource) {
-							.workbench => {
-								const workbenchInventory = getInventoryFromSource(callbackSource) orelse @panic("Could not find workbench Inventory");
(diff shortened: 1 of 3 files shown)


src/Inventory.zig
```diff
@@ -341,6 +346,17 @@ pub const ServerSide = struct { // MARK: ServerSide
 		}
 		if (itemStack.amount == 0) itemStack.item = .null;
 	}
+
+	fn workbenchCloseCallback(source: Source, user: *main.server.User) void {
+		switch (source) {
+			.workbench => |id| {
+				const workbenchInventory = getInventoryFromSource(source) orelse @panic("Could not find workbench Inventory");
+				const playerInventory = getInventoryFromSource(.{.playerInventory = id}) orelse @panic("Could not find player inventory");
```
IntegratedQuantum: Wait, but on the client it sends the player id, and not the inventory id.

src/Inventory.zig
```diff
@@ -341,6 +346,17 @@ pub const ServerSide = struct { // MARK: ServerSide
 		}
 		if (itemStack.amount == 0) itemStack.item = .null;
 	}
+
+	fn workbenchCloseCallback(source: Source, user: *main.server.User) void {
```
IntegratedQuantum: Can we not use the source id, to retrieve the user?

src/Inventory.zig
```diff
@@ -354,21 +370,23 @@ pub fn getInventory(id: InventoryId, side: sync.Side, user: ?*main.server.User)
...
 	blockInventory = 5,
+	workbench = 6,
 	other = 0xff, // TODO: List every type separately here.
 };
 pub const Source = union(SourceType) {
 	alreadyFreed: void,
 	playerInventory: u32,
 	hand: u32,
 	blockInventory: Vec3i,
+	workbench: u32,
```
IntegratedQuantum: I'd suggest to use a struct, we may want to store other stuff in there in the future (→tool type) and the u32 really is unreadable.
```suggestion
	workbench: struct { playerId: u32 },
```

IntegratedQuantum (changes requested):
> I don't know where I should put the callback function and how its name should be

Alright, so in the far future, the flow of opening chests and workbenches will look something like this:
User right clicks → sends onclick sync command to server → executes onclick callback which opens the inventory and sends it back to the client
So ideally the inventory callbacks would then be declared in the same file as the onclick callback.

But we don't have any of that yet, so for now I'd suggest to put it inside the createInventory function (as a local `struct {fn workbenchCloseCallback(...)...}workbenchCloseCallback`) to make it sure that it is moved along when that refactor happens.

As for the name, I think it's fine.

src/Inventory.zig
Wunka: yes, and that is what I also wanted to happen. `getInventoryFromSource(.{.playerInventroy` uses the playerId

src/Inventory.zig
Wunka: we could, but in my opinion this callback is called when the last user closes the inventory, so why not give the callback the user that did that? 
I know its kind of weird because we now have two ways the user is send.

Another idea you could do with that is renaming the source to something like `temporary: void` and then whoever closes the inventory last would get the deposit. 
This could then be used for any temporary inventories even for those who can be opened by multiple people at the same time

IntegratedQuantum: For now workbench inventories are player specific, so I'd say we postpone this discussion until we have a use-case for it.

src/Inventory.zig
IntegratedQuantum: Oh right, I'd still prefer if you'd just use `source.inventory` directly.

Wunka: so `source.inventory` = playerInventory and `getInventoryFromSource(source)` = workbench? I think I am confused.

IntegratedQuantum: Sorry, I meant user.inventory

Wunka: done I think, maybe I still misunderstood something. But now I was able to make the source `void`.

src/Inventory.zig
```diff
@@ -251,12 +248,27 @@ pub const ServerSide = struct { // MARK: ServerSide
 				}
 				return error.Invalid;
 			},
+			.workbench => {
+				const workbenchCloseCallback = struct {
+					fn callback(callbackSource: Source, callbackUser: *main.server.User) void {
+						switch (callbackSource) {
+							.workbench => {
```
IntegratedQuantum: I'd suggest to just `std.debug.assert(callbackSource == .workbench);` instead of this giant switch.

src/Inventory.zig
```diff
@@ -251,12 +248,27 @@ pub const ServerSide = struct { // MARK: ServerSide
...
+								const workbenchInventory = getInventoryFromSource(callbackSource) orelse @panic("Could not find workbench Inventory");
```
IntegratedQuantum: This will give you the first inventory with the workbench source type, since it doesn't contain enough data to distinguish between ones from different players. That's why you need to keep the player id in the source.

IntegratedQuantum (changes requested):

src/Inventory.zig
Wunka: that makes sense! I used the struct then as you before suggested

src/Inventory.zig
```diff
@@ -251,12 +248,23 @@ pub const ServerSide = struct { // MARK: ServerSide
 				}
 				return error.Invalid;
 			},
+			.workbench => {
+				const workbenchCloseCallback = struct {
```
IntegratedQuantum: namespace calling convention
```suggestion
				const workbench_close_callback = struct {
```

src/Inventory.zig
```diff
@@ -354,21 +363,23 @@ pub fn getInventory(id: InventoryId, side: sync.Side, user: ?*main.server.User)
 pub const Callbacks = struct {
 	onUpdateCallback: ?*const fn (Source) void = null,
 	onFirstOpenCallback: ?*const fn (Source) void = null,
-	onLastCloseCallback: ?*const fn (Source) void = null,
+	onLastCloseCallback: ?*const fn (Source, *main.server.User) void = null,
```
IntegratedQuantum: I think the reason for not passing e.g. the user, was because these callbacks may also be shared with the client-side at some point.

As said we don't need it here since there is only a single user per workbench.

IntegratedQuantum (changes requested):

src/Inventory.zig
Wunka: Removd and instead got the user over the userlist

IntegratedQuantum (approved):
To avoid duplicate work: Are you going to keep working on #2529 after this?

Wunka:
I will look into it today how feasible it seems now for me.

PR #2771, opened 2026-03-22
Update glass_forest.zig.zon

Crepestrom:
makes it look better
image

mixes some cobblestone with the slate on the ground


H41ogen:
upload a picture here

careeoki (approved):
Yeah, I don't really have any problems with this.

PR #2773, opened 2026-03-22
Store the block type + data as a string in blueprint palettes

### LINKED ISSUE #2477: Migrations don't work on ore data

Ore rotation stores host block information in its data value. That information is not affected by migrations so blueprints which used such ore-block combination get broken.

Image

Related-to: #1930

IntegratedQuantum:
Instead of only storing the block type, the blueprint palette now also stores the data, e.g.:
```
Blueprint block palette:
palette[0]: cubyz:moss:cubyz:aerosoil
palette[1]: cubyz:aerosoil
palette[2]: cubyz:moss:cubyz:permafrost
Saved clipboard to blueprint file: test.blp
```

fixes #2477

@Argmaster I would like your review here, since you wrote the blueprint code, and since I think it would be rather annoying if anything breaks with this.

I tested this by loading and pasting some mossy aerosoil into an old world (which due to its age has aerosoil mapped to a different index).

### DIFF
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -258,7 +258,7 @@ pub fn finishBlocks(zonElements: Assets.ZonHashMap) void {
 		registerOpaqueVariant(i, zonElements.get(_id[i]) orelse continue);
 		registerCallbacks(i, zonElements.get(_id[i]) orelse continue);
 	}
-	blueprint.registerVoidBlock(parseBlock("cubyz:void", .{}));
+	blueprint.registerVoidBlock(parseBlock("cubyz:void"));
 	meshes.finishTextureLoading();
 }
 
@@ -284,7 +284,7 @@ const ParseBlockConfig = struct {
 
 fn parseBlockData(fullBlockId: []const u8, data: []const u8, comptime config: ParseBlockConfig) ?u16 {
 	if (std.mem.containsAtLeastScalar(u8, data, 1, ':')) {
-		const oreChild = parseBlock(data, config);
+		const oreChild = parseBlockWithOptions(data, config);
 		if (oreChild.data != 0) {
 			std.log.warn("Error while parsing ore block data of '{s}': Parent block data must be 0.", .{fullBlockId});
 		}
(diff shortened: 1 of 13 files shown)


src/blocks.zig
```diff
@@ -292,12 +296,15 @@ fn parseBlockData(fullBlockId: []const u8, data: []const u8) ?u16 {
 	};
 }
 
-pub fn parseBlock(data: []const u8) Block {
+pub fn parseBlock(data: []const u8, comptime config: ParseBlockConfig) Block {
```
Argmaster: Why do we need that extra argument? Now you had to edit every call to this function, what's the benefit? U use it once.
I think there is a naming pattern in zig standard library for having function with extra arguments and one without those arguments?

IntegratedQuantum: I found `...WithOptions` in the allocator interface.

mods/cubyz/rotation/ore.zig
```diff
@@ -72,3 +72,7 @@ pub fn onBlockBreaking(_: main.items.Item, _: Vec3f, _: Vec3f, currentData: *Blo
 	currentData.typ = currentData.data;
 	currentData.data = 0;
 }
+
+pub fn formatBlockData(block: Block, _list: *main.List(u8)) void {
```
Argmaster, line 76: I think this should be symmetrical in the end and it should follow `save`/`load` naming pattern (or is it `save` or `store`, we have some consistency issues here no?). For now just complying with the naming would be enough since we aren't loading anything really, `load` would be basically #2793

Argmaster: Hm, or is it more of a ` format`/`parse`?
I guess the name might be fine 🤷

IntegratedQuantum: Yeah, I intended this to be format/parse pair.
Though it is saved in the blueprint, it is (at least on the parsing addons side) also meant for human-readable stuff.

PR #2774, opened 2026-03-22
Introduce new Sync command CraftTool and remove Inventory.Type

### LINKED ISSUE #2529: Change how the workbench is synchronized and remove the shared `Inventory.Type`

Wunka:
Closes #2529

- [x] write first iteration of the Sync command
- [x] split workbench inventories
- [x] point workbench crafting to the new sync command
- [x] remove old workbench sync code
- [x] remove `inventory.type`

### DIFF
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -396,7 +396,7 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 		serverShared: void,
 		creative: void,
 		crafting: *const main.items.Recipe,
-		workbenchResult: *ClientInventory,
+		workbenchResult: ClientInventory,
 	};
 	super: Inventory,
 	type: ClientType,
@@ -504,7 +504,7 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 		std.debug.assert(source.type == .workbenchResult);
 		for (destinations) |inv| std.debug.assert(inv.type == .serverShared);
 
-		main.sync.ClientSide.executeCommand(.{.craftTool = .init(destinations, source.type.workbenchResult.*)});
+		main.sync.ClientSide.executeCommand(.{.craftTool = .init(destinations, source.type.workbenchResult)});
 	}
 
 	pub fn placeBlock(self: ClientInventory, slot: u32) void {
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -605,9 +605,11 @@ pub const CanHoldReturn = union(enum) {
 
 pub fn canHold(self: Inventory, sourceStack: ItemStack) CanHoldReturn {
 	if (sourceStack.amount == 0) return .yes;
+	if (self.source == .workbench and !sync.Command.canPutIntoWorkbench(sourceStack.item)) return .{.remainingAmount = sourceStack.amount};
 
 	var remainingAmount = sourceStack.amount;
-	for (self._items) |*destStack| {
+	for (self._items, 0..) |*destStack, destSlot| {
+		if (self.source == .workbench and self.source.workbench.toolIndex.slotInfos()[destSlot].disabled) continue;
 		if (std.meta.eql(destStack.item, sourceStack.item) or destStack.item == .null) {
 			const amount = @min(sourceStack.item.stackSize() - destStack.amount, remainingAmount);
 			remainingAmount -= amount;
@@ -761,9 +763,11 @@ pub const Inventories = struct { // MARK: Inventories
 		var selectedEmptyInv: ?Inventory = null;
 
 		outer: for (self.inventories) |dest| {
+			if (dest.source == .workbench and !sync.Command.canPutIntoWorkbench(item)) continue;
 			var emptySlot: ?u32 = null;
 			var hasItem = false;
 			for (dest._items, 0..) |*destStack, destSlot| {
+				if (dest.source == .workbench and dest.source.workbench.toolIndex.slotInfos()[destSlot].disabled) continue;
 				if (destStack.item == .null and emptySlot == null) {
 					emptySlot = @intCast(destSlot);
 					if (selectedEmptySlot == null) {
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -68,6 +68,15 @@ pub const ClientSide = struct {
 		main.utils.assertLocked(&main.sync.ClientSide.mutex);
 		return serverToClientMap.get(serverId);
 	}
+
+	fn getInventoryByClientId(clientId: InventoryId) ?Inventory {
+		main.utils.assertLocked(&main.sync.ClientSide.mutex);
+		var it = serverToClientMap.valueIterator();
+		while (it.next()) |inv| {
+			if (inv.id == clientId) return inv.*;
+		}
+		return null;
+	}
 };
 
 pub const ServerSide = struct { // MARK: ServerSide
@@ -503,13 +512,13 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 	pub fn craftTool(source: ClientInventory, destinations: []const ClientInventory) void {
 		std.debug.assert(source.type == .workbenchResult);
 		for (destinations) |inv| std.debug.assert(inv.type == .serverShared);
-		const workbenchResultInv = blk: {
+		const workbenchInv = blk: {
 			main.sync.ClientSide.mutex.lock();
 			defer main.sync.ClientSide.mutex.unlock();
-			break :blk getInventory(source.type.workbenchResult, .client, null);
+			break :blk ClientSide.getInventoryByClientId(source.type.workbenchResult);
 		} orelse return;
 
-		main.sync.ClientSide.executeCommand(.{.craftTool = .init(destinations, workbenchResultInv)});
+		main.sync.ClientSide.executeCommand(.{.craftTool = .init(destinations, workbenchInv)});
 	}
 
 	pub fn placeBlock(self: ClientInventory, slot: u32) void {
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -427,6 +427,13 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 			carried.fillFromCreative(0, dest.getItem(destSlot));
 			return;
 		}
+		if (dest.type == .workbenchResult) {
+			dest.craftTool(&.{carried});
+			return;
+		}
+		if (carried.type == .workbenchResult) {
+			carried.craftTool(&.{dest});
+		}
 		std.debug.assert(dest.type == .serverShared);
 		main.sync.ClientSide.executeCommand(.{.depositOrSwap = .{.dest = .{.inv = dest.super, .slot = destSlot}, .source = .{.inv = carried.super, .slot = 0}}});
 	}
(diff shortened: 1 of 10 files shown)


Wunka:
I didn't think this would be this straight forward but here we are. 
One thing again most of the `updateResult` code exists at two places. Here again the question how could best a helper function for this look and where do I put it?. 

Also I have a not nice solution of getting the `craftingGridInv` from the window instead over for example the `ClientInventoryType`.  Here I am also open for suggestions.

As always when something like this is done with the inventories. If someone wants to try to break this please do! I don't want to introduce new bugs through this (can probably be not prevented but either way)

src/sync.zig
```diff
@@ -977,19 +955,8 @@ pub const Command = struct { // MARK: Command
...
-					ctx.execute(.{.move = .{
-						.dest = self.dest,
-						.source = self.source,
-						.amount = 1,
-					}});
-					ctx.cmd.removeToolCraftingIngredients(ctx.allocator, self.source.inv, ctx.side);
-				}
-				return;
-			}
+			if (self.source.inv.source == .workbench and self.source.inv.source.workbench.toolIndex.slotInfos()[self.source.slot].disabled) return;
```
IntegratedQuantum: Now that the assertion is gone, we have to assume that `dest` can also theoretically be a workbench, so it needs the checks there as well.

Also this particular check (taking half from a disabled slot) is pointless, since you won't have items in these slots in the first place.

src/sync.zig
```diff
@@ -905,7 +884,7 @@ pub const Command = struct { // MARK: Command
 					return;
 				}
 			}
-			if (self.source.inv.type == .workbench and !canPutIntoWorkbench(self.dest)) return;
+			if (self.source.inv.source == .workbench and !canPutIntoWorkbench(self.dest)) return;
```
IntegratedQuantum: Also misses the check for being able to put stuff into the slot.

src/sync.zig
```diff
@@ -1032,10 +999,8 @@ pub const Command = struct { // MARK: Command
 
 		fn run(self: Drop, ctx: Context) error{serverFailure}!void {
 			if (self.source.ref().item == .null) return;
-			if (self.source.inv.type == .workbench and self.source.slot != 25 and self.source.inv.type.workbench.slotInfos()[self.source.slot].disabled) return;
-			if (self.source.inv.type == .workbench and self.source.slot == 25) {
-				ctx.cmd.removeToolCraftingIngredients(ctx.allocator, self.source.inv, ctx.side);
-			}
+			if (self.source.inv.source == .workbench and self.source.inv.source.workbench.toolIndex.slotInfos()[self.source.slot].disabled) return;
```
IntegratedQuantum: pointless

src/sync.zig
```diff
@@ -1068,7 +1033,7 @@ pub const Command = struct { // MARK: Command
 		amount: u16 = 0,
 
 		fn run(self: FillFromCreative, ctx: Context) error{serverFailure}!void {
-			if (self.dest.inv.type == .workbench and (self.dest.slot == 25 or self.dest.inv.type.workbench.slotInfos()[self.dest.slot].disabled)) return;
+			if (self.dest.inv.source == .workbench and self.dest.inv.source.workbench.toolIndex.slotInfos()[self.dest.slot].disabled) return;
```
IntegratedQuantum: Also needs a check for `canPutIntoWorkbench`

src/sync.zig
```diff
@@ -1197,12 +1160,7 @@ pub const Command = struct { // MARK: Command
...
 		pub fn run(self: DepositOrDrop, ctx: Context) error{serverFailure}!void {
-			for (self.destinations.inventories) |dest| {
-				std.debug.assert(dest.type == .normal);
-			}
-			var sourceItems = self.source._items;
-			if (self.source.type == .workbench) sourceItems = self.source._items[0..25];
-			for (sourceItems, 0..) |*sourceStack, sourceSlot| {
+			for (self.source._items, 0..) |*sourceStack, sourceSlot| {
 				if (sourceStack.item == .null) continue;
 				const remainingAmount = self.destinations.putItemsInto(ctx, sourceStack.amount, .{.move = .{.inv = self.source, .slot = @intCast(sourceSlot)}});
```
IntegratedQuantum, line 1165: Also needs to check if it can put stuff into workbench now that you removed the assertions.

IntegratedQuantum: (or actually do them in `putItemsInto`)

src/sync.zig
```diff
@@ -1298,9 +1253,6 @@ pub const Command = struct { // MARK: Command
 		}
 
 		fn run(self: CraftFrom, ctx: Context) error{serverFailure}!void {
-			for (self.destinations.inventories) |dest| if (dest.type != .normal) return;
-			for (self.sources.inventories) |source| if (source.type != .normal) return;
-
 			if (self.destinations.canHold(.{.item = .{.baseItem = self.recipe.resultItem}, .amount = self.recipe.resultAmount}) != .yes) return;
```
IntegratedQuantum, line 1256: same for canHold

src/sync.zig
```diff
@@ -1350,13 +1302,73 @@ pub const Command = struct { // MARK: Command
...
+			}
+			var hash = std.hash.Crc32.init();
+			for (availableItems) |item| {
+				if (item != null) {
+					hash.update(item.?.id());
+				} else {
+					hash.update("none");
+				}
+			}
+			const tool = Item{.tool = main.items.Tool.initFromCraftingGrid(availableItems, hash.final(), self.craftingGrid.source.workbench.toolIndex)};
```
IntegratedQuantum: I think we should put this logic into a `Tool.initFromInventory` function.

src/sync.zig
```diff
@@ -1350,13 +1302,73 @@ pub const Command = struct { // MARK: Command
...
+
+			if (self.destinations.canHold(.{.item = tool, .amount = 1}) != .yes) return;
```
IntegratedQuantum: Needs to deinit the tool before returning to avoid leaks.

IntegratedQuantum (changes requested):

Wunka:
all done

src/sync.zig
```diff
@@ -977,19 +956,8 @@ pub const Command = struct { // MARK: Command
...
-					ctx.execute(.{.move = .{
-						.dest = self.dest,
-						.source = self.source,
-						.amount = 1,
-					}});
-					ctx.cmd.removeToolCraftingIngredients(ctx.allocator, self.source.inv, ctx.side);
-				}
-				return;
-			}
+			if (self.dest.inv.source == .workbench and self.dest.inv.source.workbench.toolIndex.slotInfos()[self.dest.slot].disabled) return;
```
IntegratedQuantum: also needs the check for canPutIntoWorkbench

src/Inventory.zig
```diff
@@ -426,6 +427,14 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 			carried.fillFromCreative(0, dest.getItem(destSlot));
 			return;
 		}
+		if (dest.type == .workbenchResult) {
+			dest.craftTool(&.{carried});
+			return;
+		}
+		if (carried.type == .workbenchResult) {
```
IntegratedQuantum: Ideally this shouldn't happen, since this is the client side, please use an assert(carried.type == .serverShared) here.

IntegratedQuantum (changes requested):

src/sync.zig
Wunka: both done

IntegratedQuantum (changes requested):
While testing I noticed that you can't shift-click from the result slot. Please implement this as well.

Wunka:
> While testing I noticed that you can't shift-click from the result slot. Please implement this as well.

I had to move the `CraftTool` logic to the same logic where also normal crafting is to avoid crafting as much as possible (like some other game is doing) but its done

IntegratedQuantum (changes requested):
Shift-click crafting works, but everything else fails. Please test more thoroughly next time.

Wunka:
> Shift-click crafting works, but everything else fails. Please test more thoroughly next time.

I am very sorry about that, normally I do test and I don't know why I didn't, it works now.

One thing I don't know to handle is right clicking. before it was allowed and would get the item, now it doesn't as crafting normally with right clicking also doesn't work. 
Its not hard to bring back, I just want an answer, do we want it or should this maybe later do something else? I think for crafting its not done because normally right clicking is for taking half which is of course not allowed when crafting

IntegratedQuantum:
I think we should have consistent behavior for all take-only crafting slots, and at the end of the day not being able to right click probably won't matter anyways.

src/Inventory.zig
```diff
@@ -498,6 +500,13 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 		main.sync.ClientSide.executeCommand(.{.craftFrom = .init(destinations, &.{source}, craftingInv.type.crafting)});
 	}
 
+	pub fn craftTool(source: ClientInventory, destinations: []const ClientInventory) void {
+		std.debug.assert(source.type == .workbenchResult);
+		for (destinations) |inv| std.debug.assert(inv.type == .serverShared);
+
+		main.sync.ClientSide.executeCommand(.{.craftTool = .init(destinations, main.gui.windowlist.workbench.craftingGridInv)});
```
IntegratedQuantum: This dependence on a specific window implementation is a problem. What if a mod wants to make their own crafting station with a different interface?

Instead I'd suggest to reference the corresponding crafting grid in the inventory type, as was done for the recipe.

IntegratedQuantum (changes requested):

src/Inventory.zig
Wunka: done

src/Inventory.zig
```diff
@@ -396,13 +396,14 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 		serverShared: void,
 		creative: void,
 		crafting: *const main.items.Recipe,
+		workbenchResult: *ClientInventory,
```
IntegratedQuantum: should be stored by value to avoid possibly referencing this after the window was closed but before the last pending operation was confirmed.

IntegratedQuantum (changes requested):

src/Inventory.zig
Wunka: I knew it... I was not sure but you confirmed it. Changed!

IntegratedQuantum: but wait, we can't do that because it's a recursion, so I guess maybe just store the client inventory id then.

Wunka: saved per inventory id, but I got a mutex not locked assertion. So I needed to lock it before getting it. Maybe I did something wrong there. Have not yet worked much with the mutexes

IntegratedQuantum (changes requested):
Something weird is going on here. If I open a workbench, shift-craft something, and then close and reopen it, I cannot craft anything.

Wunka:
> Something weird is going on here. If I open a workbench, shift-craft something, and then close and reopen it, I cannot craft anything.

Its because the `getInventory` function uses the `ServerInventoryId` on the client side... Can I even get the `ServerInventoryId` on client side? 
I think I will just introduce a function that gets the client Inventory with the `clientInventoryId`

Wunka:
thanks for catching that! 
I have introduced the function and it seems to work now!

IntegratedQuantum (approved):

PR #2775, opened 2026-03-22
Highlands

careeoki:
A temperate equivalent to the snowcapped hill, a flatish biome raised up to 100 blocks.
Its terrain is broken up by large spiky rocks jutting out from the surface.
image

image
viewed from a grassland
image


Crepestrom:
couuld you mix in some dirt patches?

IntegratedQuantum (changes requested):
I like the jagged (with stalagmites) hills and overall it's a nice scenery, but I feel like something is missing here, maybe another subbiome to add some contrast, or maybe a few structures (a stone circle would come to mind)

careeoki:
Added a new thicket subbiome, which has some trees. You can also see trees in the base biome occasionally.

image

I also made rock circles:

image

And if you dig at the center of them, you can find some goodies.
image

IntegratedQuantum (approved):
I think the random pine trees really make a difference.
Also I love the idea of putting little ore deposits in the center of the stone circles.

Issue #2720, opened 2026-03-14
Empty lanterns/lamps


careeoki:
### Description

(as discussed on the playsession)

Instead of having separate blocks for sulfur and coal lanterns, have 1 "lantern frame/empty lantern" block that can have different fuels inserted into it to light them different colors.
Then we can have copper, gold, silver, uranium lanterns without needing separate blocks for all the possible combinations.

and this can be applied to lamps as well

ikabod-kee:
Frame blocks. Sounds. familiar. Good idea. Sounds fmailair. god mida. smells fmailr. #frame

https://github.com/PixelGuys/Cubyz/issues/1206

IntegratedQuantum:
the important difference to frame block is that lamps are not transparent

IntegratedQuantum:
sadly this requires #1821

OneAvargeCoder193:
this would require that each of the metals that can be used as fuel to somehow have another property that is defined by the lamp somehow, because there needs to be a check to see if the fuel is valid, so you dont put pebbles as fuel into the lamp.

IntegratedQuantum:
We just a way of secretly transforming the fuel into another block type on placement that matches the model and has a handmade texture for the lantern/lamps.

careeoki:
I'm thinking it would make more sense to use a torch to light the lanterns rather than the fuel. That way it's a bit cheaper and more clear what can be used to light lanterns.

PR #2780, opened 2026-03-25
Lumiflora & Lumi torches

careeoki:
Lumiflora is a vibrant flower exclusive to sky islands. It emits a dim pink glow.
image
image
With 1 lumiflora and 1 plank, you can craft 8 lumi torches. Now you can light up sky islands and their hypothetical caves without coal (which doesn't appear on sky islands)
It might sound easy picking flowers to craft torches, but remember that once you run out of them on one island, you'll have to get yourself to another island, which is not so easy.
image

There's no lumi lanterns or lamps because I would like to wait until #2720 
Lumiflora replaces castelleja on sky islands. Eventually I would like to replace daffodils with a unique flower too.


ikabod-kee:
Having it craft into yet another torch seems kind of boring, but it works for now. I think it'd be cooler it could just _be_ the torch, but I understand its scarcity.

Later on, we could have the Lumiflora bloom only in dark places and night!

careeoki:
> Having it craft into yet another torch seems kind of boring, but it works for now. I think it'd be cooler it could just be the torch, but I understand its scarcity.

If it was as bright as a torch, it would light up the sky islands more than I would like. And you would need a ton of flowers to light up a cave, way more than should reasonably be on the islands.
I don't think it's a bad thing to have more torches. It means we could have more variety with lanterns and lamps in the future.
Torches are a lot more versatile in builds, and with a separate flower and torch, we get more options in terms of brightness of lights.

ikabod-kee:
> If it was as bright as a torch, it would light up the sky islands more than I would like.

Yeah, obviously we wouldn't want that.

TyrrelMeter:
> If it was as bright as a torch, it would light up the sky islands more than I would like.

Maybe make it so that you could combine a flower with say a glow Crystal to make a brighter version.

du82:
I really like both the torch and flower varieties! More variety in blocks and textures is good! It works with so many different types of bulids!

TyrrelMeter:
> I really like both the torch and flower varieties! More variety in blocks and textures is good! It works with so many different types of bulids!

Honestly there are only two torch varieties so it's fine to add more. It isn't like we have an abundance of them.

Wbjitscool:
yeah

IntegratedQuantum (approved):

PR #2785, opened 2026-03-27
Add mushroom poles

IntegratedQuantum:
These odd wooden poles started appearing in the world. Not sure what they are, they are made out of wood and have 3 mushrooms at the top, so I decided to call them mushroom poles.

In the world they always appear in groups of at least 3 of them in a row, and most of the time they have boletes at the top.
Though rarely different mushrooms were sighted.
Screenshot at 2026-03-27 20-34-51
Screenshot at 2026-03-27 20-10-45

alongside I also fixed a bug with inherit rotation.


TyrrelMeter:
What is the lore explanation for this?

TyrrelMeter:
And will there be varieties for all woods?

IntegratedQuantum:
> What is the lore explanation for this?

Same reasons as the triclops statue.

For those few who have access to it: https://github.com/IntegratedQuantum/Cubyz-Lore/issues/4

For everyone else: That's for you to decide.

> And will there be varieties for all woods?

No, I decided on mahagony because I think it fits best due to its color being a bit less ordinary.

TyrrelMeter:
I agree, Mahogany does look good for it.

Wunka:
Is that the Quantum statue? the top kinda looks like the the greek letter on quantums profile picture

And maybe the statue couldn't agree on its position, which is why it is at three positions at once

ikabod-kee:
> power lines... traffic lights... hmmmm...

You mean mushroom poles and triclops? :P

PR #2792, opened 2026-03-30
Runtime formatting and logging

IntegratedQuantum:
This is something I've been wanting to try for a long time, and with some trickery the results are surprisingly good: No major changes in the interface, decent compile time performance gains.

The biggest surprise to me was that I can mostly reusing the formatting infrastructure from the standard library with only the formatting specifier being needed at comptime.

This should be merged after the release in case there are some edge cases where I broke something. Note that with runtime formatting, we'll only get errors at runtime about incorrect formatting specifiers. But I think this is worth the improvement to compile times:

. | Before | After | Binary size
--- | --- | --- | ---
Debug + llvm |16.0 s | 13.8 s | - 0.3 MB
Debug + x86_64 | 6.9 s | 4.9 s | + 3.9 MB (the whole binary is over 600 MB, probably for incremental)
ReleaseSafe | 86.6 s | 82.6 s | - 0.9 MB
ReleaseFast | 94.4 s | 90.0 s | - 1.1 MB

Over 2 seconds saved in debug mode. And this is just the beginning: We can reuse this for bufPrint and allocPrint and other print throughout the code. Overall I'd expect we can probably save another few seconds from the llvm build times and maybe 300 ms from the x86_64 build time.

Remaining work:
- [x] Wait until 0.2.0
- [x] support positional arguments

### DIFF
--- a/src/fmt.zig
+++ b/src/fmt.zig
@@ -181,7 +181,7 @@ pub noinline fn format(writer: *std.Io.Writer, formatString: []const u8, args: [
 				.width = placeholder.width,
 			};
 			const arg = argState.nextArg(placeholder.argPos) orelse {
-				std.log.err("Not enough arguments for format string: '{s}', attempted to use argument {}, only has {} arguments", .{formatString, placeholder.argPos, args.len});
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
Time to get those compile-times down

PR #2795, opened 2026-03-31
Add vine decay (take 2)

Argmaster:
https://github.com/user-attachments/assets/9270c47b-60eb-4b05-9590-98d18eb967af

Related to: #1930 
Related to: #2486
Resolves: https://github.com/PixelGuys/Cubyz/pull/1930#pullrequestreview-3601984128


Wbjitscool:
looks good

TyrrelMeter:
Do we get to collect vine blocks broken by vine decay?

Argmaster:
Current implementation doesn't drop any items to avoid creating huge mess while decaying trees. Intended way to collect vines would be to use sickle and break the vines manually.

PR #2796, opened 2026-03-31
Stone pillars

careeoki:
Pillar blocks for all 10 stone types.
image

They look like... pillars. They have unique top textures that can provide some finer details in your stone walls.
Also existing limestone building blocks have been tweaked

todo:
- [x] final pass
- [x] limestone wall


H41ogen:
this is fire

H41ogen:
it'd be nice if they could be like logs, but i understand if that's too much work

TyrrelMeter:
Will the tops be forced to be there, or can I manipulate it to just have the pillar texture?

careeoki:
> it'd be nice if they could be like logs, but i understand if that's too much work

Log rotation doesn't really work with geometric patterns. But yeah I aint doing allat

careeoki:
> Will the tops be forced to be there, or can I manipulate it to just have the pillar texture?

They work like this:
image

TyrrelMeter:
I see now, I just assumed it switched textures not that those were just rotated.

H41ogen:
> Log rotation doesn't really work with geometric patterns.

what do you mean?

careeoki:
> what do you mean?

Instead of having separate textures for every possible state, log rotation rotates textures (which is good). 
It's perfectly fine for organic textures but doesn't work for textures with distinct patterns.
image

Crepestrom:
eh kinda
if you make your patterns with specific rules it works
you just have to make sure your pattern can connect with a mirrored version of itself

Crepestrom:
image
you can see in this screenshot that some patterns do work with logs

du82:
I really like these textures!

H41ogen:
I just don't want to add these with this rotation now and regret it like with logs

Argmaster:
Idk, when I think about pillars, those from ancient Greek come to my mind and they were not going on L or T shapes

rgbAvnix:
I'm going to give an opinion that may be controversial. It's more expansive than this feature in particular.

So far, this game has taken the Minecraft ethos of "different base block = different pattern on decorative blocks." I think this should not be the case, or at least much less than it is now.

This game has an ethos of not arbitrarily restricting the player. We already have the chiseling system, which goes against Minecraft's ethos of restricting partial blocks to arbitrary shapes, and the procedural crafting system, which lets you make a wide variety of tool designs. I think we should continue this ethos here, and (at least mostly) avoid restricting patterns to specific arbitrary blocks.

It's not a super big deal with these pillars, but it's a much bigger deal with the Tile blocks. In Minecraft, there is a block type called "polished", but in Cubyz this pattern has been implemented as a special case of Tile, which means it does not exist for all stone types.

Again, not specific to this PR, but I see a pattern here and my opinion is that going forward it should stop.

If we want there to be patterns specific to a particular stone type, I think it should be restricted to a single type of decorative block (called "Ornate" or something) rather than affecting all the types of craftable stone blocks in the game.

ikabod-kee:
> If we want there to be patterns specific to a particular stone type, I think it should be restricted to a single type of decorative block (called "Ornate" or something) rather than affecting all the types of craftable stone blocks in the game.

Maybe paint would fix this problem.

TyrrelMeter:
Is paint a planned feature?

ikabod-kee:
I don't think we ever even discussed the concept of paint at any point in Cubyz..

Argmaster:
I wonder if we could do something similar to what ores+tools do and provide gray blocks with patterns which have data that points to other base block. No idea how to reliably sample colors tho and making them all block entities is terrible idea.
Maybe we could provide dedicated palette field to block definition like we do materials and only those blocks which have that palette field could be used as a base block to create an ornate block.
Then you can have a workbench that converts given base block into ornate with base block set to given base block.

Argmaster:
What I described above would mean that we have to finally figure out how to make rotation mode involved in block hp

ikabod-kee:
I would absolutely love to have block HP (finally)

I do think that maybe we should focus on the PRs we currently have though for 0.3.0

careeoki:
> It's not a super big deal with these pillars, but it's a much bigger deal with the Tile blocks. In Minecraft, there is a block type called "polished", but in Cubyz this pattern has been implemented as a special case of Tile, which means it does not exist for all stone types.

Tiles could reasonably be split into a polished full block and a block of smaller tiles like limestone or terracotta.
But there has to be some limit to how many different stone building block variants we have.

> If we want there to be patterns specific to a particular stone type, I think it should be restricted to a single type of decorative block (called "Ornate" or something) rather than affecting all the types of craftable stone blocks in the game.

What particular existing blocks would you consider "ornate?"

I think it's important for each stone to have its own unique characteristics. Vintage Story for example, has many stone types, but they are all basically recolours, making all the same shapes of bricks and etc. To me, that's much less interesting/inspiring.

careeoki:
> Maybe paint would fix this problem.

I have very strong opinions against paint. I think it makes building a lot less approachable, since there's just so many different possible combinations (decision paralysis)
Also, all the existing blocks are not made with paint in mind. We would have to remove all the chalk and cloth variants for example.
I would have designed cubyz' blocks completely different if paint was planned.

rgbAvnix:
> Tiles could reasonably be split into a polished full block and a block of smaller tiles like limestone or terracotta.
> But there has to be some limit to how many different stone building block variants we have.

Below is a picture of all the current stone block sets.

The patterns I would recommend all stone types to have:
"Polished" (as in Basalt, Glacite, Slate, and Sandstone Tiles)
"Large Bricks" (as in Marble, Slate, and Limestone Bricks)
"Small Bricks" (as in Limestone and Terracotta Bricks)
"Square Bricks" (as in Glacite Bricks)
"Tiles" (as in Limestone, Nimbusite, and Terracotta Tiles)
"Pillars" (as in the Marble, Slate, Basalt and Limestone Pillars here)

> What particular existing blocks would you consider "ornate?"

When I mentioned "ornate" blocks, I was thinking of something like Minecraft's "chiselled" blocks or Hytale's "ornate" and "decorative" blocks. A special pattern/accent block that has a different design for each base material. I don't think any of the existing block designs really fall in that category. The Nimbusite and Sandstone pillars you designed are the closest to what I imagine ornate blocks looking like (but without being pillars).

For current block designs that don't fall into one of the patterns, I guess for some of them I would recommend just changing the design slightly to make them fit: Rotate basalt bricks to make a "large bricks", add more divisions to Nimbusite bricks to make a "small bricks", and standardize the sizes of the pieces of Ferrock bricks and tiles (and possibly tone down the color contrast a bit) to make them into "small bricks" and "tiles", respectively. I wouldn't change the Voidstone blocks because they seem like they are weird on purpose (but I would add a "polished" variant.)

Screenshot 2026-04-03 160012

careeoki:
> The patterns I would recommend all stone types to have:

I'd agree with most of this, but I would consider glacite and nimbusite bricks to just be small bricks.

> For current block designs that don't fall into one of the patterns, I guess for some of them I would recommend just changing the design slightly to make them fit: Rotate basalt bricks to make a "large bricks", add more divisions to Nimbusite bricks to make a "small bricks", and standardize the sizes of the pieces of Ferrock bricks and tiles 

I strongly believe it is important for different stone sets to have unique shapes and designs. It's way less interesting to have the exact designs for all the stone sets. It's also better for accessibility if you're able to distinguish blocks by their shape rather than just their color.

> and standardize the sizes of the pieces of Ferrock bricks and tiles 

The whole point of the ferrock set is to be messy and uneven.

> When I mentioned "ornate" blocks, I was thinking of something like Minecraft's "chiselled" blocks or Hytale's "ornate" and "decorative" blocks. 

Yeah, I would like something like that at some point.

rgbAvnix:
> I'd agree with most of this, but I would consider glacite and nimbusite bricks to just be small bricks.

I really like the "square bricks" pattern on Glacite bricks (and minecraft's mud bricks) and am disappointed by its lack of prevalence in both games, so I advocate for having it be one of the base patterns in Cubyz.

> I strongly believe it is important for different stone sets to have unique shapes and designs. It's way less interesting to have the exact designs for all the stone sets.

This is where we disagree, I guess. But the designs wouldn't have to be *exactly* the same anyway (like sandstone and marble bricks do not have the exact same design, but they do have the same pattern). There could also be multiple "ornate" variants for some stone types (Hytale has two for each stone type, but the number doesn't have to be standardized)

H41ogen:
> Idk, when I think about pillars, those from ancient Greek come to my mind and they were not going on L or T shapes

i have the feeling pillars will be used for lots of purposes other than as pillars. having non-straight shapes would be nice for that

TyrrelMeter:
Honestly I like the differences in the different blocks.

H41ogen:
me too. having all the stone sets be the same is boring, and plus each one being different gives them personality

careeoki:
> you can see in this screenshot that some patterns do work with logs 

Yes, some patterns work. But then they don't if you want to add shading:
In general, I don't want to limit what pillar designs I can do just to make it work as a log.

image

rgbAvnix:
From the previous discussion here, I also want to note that it's important to have "2x2 tile" blocks with aligned 8x8 pixel tiles because they are the only block pattern that lines up with the edge of chiseled block shapes.

IntegratedQuantum (changes requested):
I think it would make sense to mirror the bottom texture so it lines up
Screenshot at 2026-04-11 19-53-11

careeoki:
All the asymmetrical designs have bottom textures now

image

Wbjitscool:
nice

IntegratedQuantum (approved):

PR #2798, opened 2026-04-01
use kick inside manage player window

### LINKED ISSUE #2724: use /kick command inside the manage player window
After: #2723, the manage player window should use the command instead of doing the kicking itself.
Also every player should be able to see the window. So that every player who has permission to kick is able to per the window. Restricting access to the window will then be done by #2725 
This enables #2594.

Wunka:
Closes #2724 

Changes: 
- Rename manage players window to just players window
- On client only instances the players window uses the kick command
- Put the players window directly into the pause window instead of having it inside the invite window
- Show the manage player window to client only instances

One difference between the windows is still that the player opening the window is not shown for the client-side only version, while for the client-server version the player is in the list.

I would be ok with both versions that's why I take the question to the reviewers.


src/gui/windows/pause.zig
```diff
@@ -21,9 +21,7 @@ fn reorderHudCallbackFunction() void {
 }
 pub fn onOpen() void {
 	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
-	if (main.server.world != null) {
-		list.add(Button.initText(.{0, 0}, 128, "Invite Player", gui.openWindowCallback("invite")));
-	}
+	list.add(Button.initText(.{0, 0}, 128, "Manage Players", gui.openWindowCallback("manage_players")));
```
IntegratedQuantum: Maybe we should call it just players, since not everyone will be allowed to manage them.

As for the invite screen: I think it should stay accessible from this inventory directly, since it is not so obvious that you would get there from manage players.

src/gui/windows/manage_players.zig
```diff
@@ -21,14 +21,40 @@ pub var window = GuiWindow{
...
+		if (entityCount == 0) {
+			list.add(Label.init(.{0, 0}, 200, "No players to manage", .left));
+			break :blk;
+		}
+
+		const old = main.settings.showIdWithName;
+		main.settings.showIdWithName = true;
+		defer main.settings.showIdWithName = old;
+		for (main.client.entity_manager.entities.items()) |ent| {
+			if (ent.name.len == 0) continue;
```
IntegratedQuantum: Ideally you'd check for the presence of a player index.

src/gui/windows/players.zig
```diff
@@ -21,14 +21,40 @@ pub var window = GuiWindow{
...
+	if (main.server.world == null) blk: {
+		entityCount = main.client.entity_manager.entities.len;
+		if (entityCount == 0) {
+			list.add(Label.init(.{0, 0}, 200, "No players to manage", .left));
+			break :blk;
+		}
+
+		const old = main.settings.showIdWithName;
+		main.settings.showIdWithName = true;
+		defer main.settings.showIdWithName = old;
```
IntegratedQuantum: This is a hack, I don't like it. Instead you could just create a separate formatting function and call it with an Io.Writer.

IntegratedQuantum (changes requested):

src/gui/windows/manage_players.zig
Wunka: Do you mean with `@hasDecl`? I know that the playerIndex will be its own component. but currently its still just a variable inside the `entity`

IntegratedQuantum: No, by making it optional. We are close to entity support, and you shouldn't be allowed to kick named entities, which would instead kick the player with id 0.

Wunka: Oh that makes sense. Yeah sorry for that misunderstanding

src/gui/windows/players.zig
Wunka: I have not found the right writer I should use... but I have found this alt method from the WriterGate PR.

src/gui/windows/pause.zig
Wunka: Done.

src/gui/windows/players.zig
```diff
@@ -0,0 +1,72 @@
...
+	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
+	blk: {
+		entityCount = main.client.entity_manager.entities.len;
+		if (entityCount == 0) {
+			list.add(Label.init(.{0, 0}, 200, "No players", .left));
+			break :blk;
+		}
+
+		for (main.client.entity_manager.entities.items()) |ent| {
+			if (ent.name.len == 0 or ent.playerIndex == null) continue;
```
IntegratedQuantum: We also want to be able to kick players with the empty name.

src/gui/windows/players.zig
```diff
@@ -0,0 +1,72 @@
...
+	const command = std.fmt.allocPrint(main.globalAllocator.allocator, "kick @{d}", .{playerIndex}) catch unreachable;
+	main.sync.ClientSide.executeCommand(.{.chatCommand = .{.message = command}});
+}
+
+pub fn onOpen() void {
+	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
+	blk: {
+		entityCount = main.client.entity_manager.entities.len;
+		if (entityCount == 0) {
+			list.add(Label.init(.{0, 0}, 200, "No players", .left));
```
IntegratedQuantum: There is always at least one player who observes the menu
```suggestion
			list.add(Label.init(.{0, 0}, 200, "No other players", .left));
```

src/gui/windows/manage_players.zig
```diff
@@ -1,77 +0,0 @@
...
-		userList = main.globalAllocator.alloc(*main.server.User, main.server.connectionManager.connections.items.len);
-		for (main.server.connectionManager.connections.items, 0..) |connection, i| {
-			userList[i] = connection.user.?;
-			userList[i].increaseRefCount();
-			const row = HorizontalList.init();
-			if (connection.user.?.name.len != 0) {
-				row.add(Label.init(.{0, 0}, 200, connection.user.?.name, .left));
-				row.add(Button.initText(.{0, 0}, 100, "Kick", .initWithPtr(kick, connection)));
-			} else {
-				const ip = std.fmt.allocPrint(main.stackAllocator.allocator, "{f}", .{connection.remoteAddress}) catch unreachable;
```
IntegratedQuantum, line 44: Incomplete connections should also be observable and cancelable by the server owner.

IntegratedQuantum (changes requested):

src/gui/windows/manage_players.zig
Wunka: I brought it back. This had the same problem with the zero length name being skipped. but I now skip the the player himself. So it also now has the "no other player" info when needed.

Wunka: I hope I understood the `isConnected` check here is the right one. I don't know what I could else check (I am not that knowledgeable on the network code

src/gui/windows/players.zig
```diff
@@ -0,0 +1,115 @@
...
+pub var window = GuiWindow{
+	.contentSize = Vec2f{128, 256},
+	.closeIfMouseIsGrabbed = true,
+};
+
+const padding: f32 = 8;
+var userList: []*main.server.User = &.{};
+var entityCount: u32 = 0;
+
+fn kickPerConn(conn: *main.network.Connection) void {
```
IntegratedQuantum:
```suggestion
fn kickByConnection(conn: *main.network.Connection) void {
```

src/gui/windows/players.zig
```diff
@@ -0,0 +1,115 @@
...
+	conn.disconnect();
+}
+
+fn kickPerIndex(playerIndex: usize) void {
```
IntegratedQuantum:
```suggestion
fn kickByPlayerIndex(playerIndex: usize) void {
```

src/gui/windows/players.zig
```diff
@@ -0,0 +1,115 @@
...
+		}
+	} else {
+		main.server.connectionManager.mutex.lock();
+		defer main.server.connectionManager.mutex.unlock();
+		std.debug.assert(userList.len == 0);
+		userList = main.globalAllocator.alloc(*main.server.User, main.server.connectionManager.connections.items.len);
+		for (main.server.connectionManager.connections.items, 0..) |connection, i| {
+			userList[i] = connection.user.?;
+			userList[i].increaseRefCount();
+			if (userList[i].id == main.game.Player.id and connection.isConnected()) continue;
```
IntegratedQuantum, line 62: irrelevant check
```suggestion
			if (userList[i].id == main.game.Player.id) continue;
```

src/gui/windows/players.zig
```diff
@@ -0,0 +1,115 @@
...
+			const row = HorizontalList.init();
+			if (connection.isConnected()) {
```
IntegratedQuantum: This is a lower level state, the connection is connected when it has done the initial connection handshake and is ready to transmit game data (like the player name).
You can use the `handShakeState` to check the progress of the game handshake where it exchanges the name.

IntegratedQuantum (changes requested):

src/gui/windows/players.zig
Wunka: not irrelevant from my testing. If I invite an player (without them joining to make the "Cancel" appear) the player has the same id. so to prevent them from not showing I check the connection

IntegratedQuantum: Oh I see.

src/gui/windows/players.zig
Wunka: done. now it only shows the ip if the handshake is below the `userData` state. Also applied your suggestions with the function names

src/gui/windows/players.zig
```diff
@@ -61,16 +61,16 @@ pub fn onOpen() void {
 			userList[i].increaseRefCount();
 			if (userList[i].id == main.game.Player.id and connection.isConnected()) continue;
 			const row = HorizontalList.init();
-			if (connection.isConnected()) {
+			if (@intFromEnum(connection.handShakeState.load(.monotonic)) >= @intFromEnum(main.network.Connection.HandShakeState.userData)) {
```
IntegratedQuantum: Should be `>`, since the protocol sets the value before actually processing the data.
Furthermore since it's unclear how this will change with the ECS (the entity might get created in a later step of the handshake), and generally avoid problems on future changes, I'd suggest to just check if the handshake is complete.

IntegratedQuantum (changes requested):

src/gui/windows/players.zig
Wunka: done

IntegratedQuantum (approved):
Yay, I can finally kick the host

PR #2801, opened 2026-04-02
Whitelisting1

Mabeeck:
Pull request for whitelisting functionality on the server side.  
No commands will be included in this PR.  
Does some progress with #2566.

I might still need some help to handle rejecting the connection better.

- [ ] Fix: `[error]: Encountered error while opening world: DisconnectedByServer` on client, when connection is rejected
- [ ] Prevent player file creation for rejected connection requests

### DIFF
--- a/src/network.zig
+++ b/src/network.zig
@@ -1710,11 +1710,13 @@ pub const Connection = struct { // MARK: Connection
 
 	pub fn receive(self: *Connection, data: []const u8) void {
 		self.tryReceive(data) catch |err| {
-			std.log.err("Got error while processing received network data: {s}", .{@errorName(err)});
-			if (@errorReturnTrace()) |trace| {
-				std.log.info("{f}", .{main.fmt.FormatErrorTrace{.stackTrace = trace.*}});
+			if (err != error.Denied) {
+				std.log.err("Got error while processing received network data: {s}", .{@errorName(err)});
+				if (@errorReturnTrace()) |trace| {
+					std.log.info("{f}", .{main.fmt.FormatErrorTrace{.stackTrace = trace.*}});
+				}
+				std.log.debug("Packet data: {any}", .{data});
 			}
-			std.log.debug("Packet data: {any}", .{data});
 			self.disconnect();
 		};
 	}
--- a/src/network/protocols.zig
+++ b/src/network/protocols.zig
@@ -54,7 +54,9 @@ pub fn onReceive(conn: *Connection, protocolIndex: u8, data: []const u8) !void {
 
 	var reader = utils.BinaryReader.init(data);
 	protocolReceive(conn, &reader) catch |err| {
-		std.log.debug("Got error while executing protocol {} with data {any}", .{protocolIndex, data});
+		if (err != error.Denied) {
+			std.log.debug("Got error while executing protocol {} with data {any}", .{protocolIndex, data});
+		}
 		return err;
 	};
 
@@ -170,6 +172,8 @@ pub const handShake = struct { // MARK: handShake
 					if (main.server.world.?.mode != .singleplayer) {
 						const keys = zon.getChild("keys");
 						try conn.user.?.identifyFromKeysAndName(name, keys);
+						conn.user.?.initPlayer();
+						if (!main.server.world.?.playerMayJoin(conn.user.?)) return error.Denied;
 
 						var writer: utils.BinaryWriter = .init(main.stackAllocator);
 						defer writer.deinit();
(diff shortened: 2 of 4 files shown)


src/network/protocols.zig
```diff
@@ -58,7 +58,9 @@ pub fn onReceive(conn: *Connection, protocolIndex: u8, data: []const u8) !void {
 	} else {
 		var reader = utils.BinaryReader.init(data);
 		protocolReceive(conn, &reader) catch |err| {
-			std.log.debug("Got error while executing protocol {} with data {any}", .{protocolIndex, data});
+			if (!std.mem.eql(u8, @errorName(err), "Denied")) {
```
Wunka:
```suggestion
			if (err != error.Denied) {
```

src/server/whitelist.zig
```diff
@@ -0,0 +1,37 @@
...
+		if (user.isLocal) return true;
+
+		const mayJoin = user.mayJoin;
+		return switch (mayJoin) {
+			.default => !self.*.whitelisted,
+			.whitelisted => true,
+			.blacklisted => false,
+		};
+	}
+};
```
Wunka:
```suggestion
pub const JoinFilter = struct {
	whitelisted: bool = false,

	pub const mayJoinState = enum { default, whitelisted, blacklisted };

	pub fn load(worldData: ZonElement) JoinFilter {
		self.*.whitelisted = worldData.get(bool, "whitelisted", false);
	}

	pub fn playerMayJoin(self: JoinFilter, user: *main.server.User) bool {
		if (user.isLocal) return true;
		
		return switch (user.mayJoin) {
			.default => !self.whitelisted,
			.whitelisted => true,
			.blacklisted => false,
		};
	}
};
```
If you don't want to add anything more complex the `whitelist` can just be a `bool` inside `ServerWorld` and then with the method `playerMayJoin` also moved there this entire struct is not needed

src/server/world.zig
```diff
@@ -449,6 +450,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 
 	playerDatabase: std.StringHashMapUnmanaged(usize) = .{},
 	nextPlayerIndex: std.atomic.Value(usize) = .init(0),
+	joinFilter: *JoinFilter = undefined,
```
Wunka:
```suggestion
	joinFilter: JoinFilter = undefined,
```
with the changes to JoinFilter

src/server/world.zig
```diff
@@ -542,6 +545,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		self.itemPalette.deinit();
 		self.toolPalette.deinit();
 		self.biomePalette.deinit();
+		self.joinFilter.deinit();
```
Wunka:
```suggestion
```

src/server/world.zig
```diff
@@ -497,6 +499,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		errdefer main.assets.unloadAssets();
 
 		const worldData = try dir.readToZon(arena, "world.zig.zon");
+		self.joinFilter = JoinFilter.init(main.worldArena, worldData);
```
Wunka:
```suggestion
		self.joinFilter = .load(worldData);
```

src/network.zig
```diff
@@ -1600,11 +1599,13 @@ pub const Connection = struct { // MARK: Connection
 
 	pub fn receive(self: *Connection, data: []const u8) void {
 		self.tryReceive(data) catch |err| {
-			std.log.err("Got error while processing received network data: {s}", .{@errorName(err)});
-			if (@errorReturnTrace()) |trace| {
-				std.log.info("{f}", .{std.debug.FormatStackTrace{.stack_trace = trace.*, .tty_config = .no_color}});
+			if (!std.mem.eql(u8, @errorName(err), "Denied")) {
```
Wunka:
```suggestion
			if (err != error.Denied) {
```
the same as the other `if`

Wunka (changes requested):
The struct is probably from the times when it had a few more fields and this is still a draft so don't take my requested changes to hard. 

I have not yet tested your code (it still a draft I don't require it to already work perfectly)

src/server/whitelist.zig
Mabeeck: > The struct is probably from the times when it had a few more fields

You got that right. I'll remove it, it is no longer needed.

Wunka:
>  Get rid of `[info]: Disconnected` (x2) in server logs when rejecting a connection

I think this happens not only in your PR. So if you can get rid of it :tada: but don't focus on it

yel0h:
Hey @Mabeeck, since this PR has been inactive for a few months now, I'm picking up the issue over at #3365.

PR #2803, opened 2026-04-02
We never unmaped Inventories on Client Side

Wunka:


IntegratedQuantum (approved):
I would like to have less looping through hashmaps, but this is kind of game-breaking, so I made an issue instead #2804

PR #2805, opened 2026-04-02
Triclops Statue redesign

rgbAvnix:
I redesigned the Triclops Statue to look better and more realistic. All variants work the same as before.
Screenshot 2026-04-02 185747


tillpp:
here is the real one for comparison:
grafik

careeoki (changes requested):
I do think this looks better. Though, while you're changing this, I think it would benefit from some height variation. You could do this by separating the `base` into a `pole `and `head`, and giving the `pole` having some variants of different heights.

rgbAvnix:
Hopefully I updated the PR correctly. I added some height variation:

Screenshot 2026-04-04 131158

Incidentally, this means the head can now rotate on the pole, but it makes no difference because the base is symmetrical.

ikabod-kee:
image

If you could, can you make the roots not replace solid blocks? Everything else is perfect, just this one change is needed.

rgbAvnix:
> If you could, can you make the roots not replace solid blocks?

They replaced solid blocks before this redesign as well. I don't think it's possible to set a single part of a structure to degradable while having the rest replace all blocks.

ikabod-kee:
> They replaced solid blocks before this redesign as well. 

Whoops, I guess I made a mistake there then XD
Does anybody know if parts of structures can be marked as degradable?

rgbAvnix:
I'm pretty sure they can't because .degradable is not specified in an SBB file, it's specified in the biome that generates the SBB. So the same structure can be degradable or not depending on what the biome specifies, but that carries over to all the child SBBs. So for individual child structures to have a different .placeMode they would have to override the .placeMode specified in the biome file.

ikabod-kee (approved):
Hmmm, that sucks. I guess we can either wait until that feature's added or we could just merge it as is and fix it later. I vote on the latter.

careeoki (approved):

IntegratedQuantum (approved):

IntegratedQuantum:
If you are looking for ideas what to work on next, a multi-headed hydraclops version in the same design might be fitting.

PR #2806, opened 2026-04-03
show Index even if the the name has a length of zero and rename the showIdWithName to showPlayerIndexWithName

Wunka:


src/client/entity_manager.zig
```diff
@@ -95,7 +95,7 @@ pub fn renderNames(projMatrix: Mat4f, playerPos: Vec3d) void {
 	const fontScreenSize = fontBaseSize*screenUnits;
 
 	for (entities.items()) |ent| {
-		if (ent.id == game.Player.id or ent.name.len == 0) continue; // don't render local player
+		if (ent.id == game.Player.id or (ent.name.len == 0 and !settings.showIdWithName)) continue; // don't render local player
```
IntegratedQuantum: I'd suggest to split this into two lines, the second part really has nothing to do with the comment.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #2807, opened 2026-04-03
More runtime formatting

IntegratedQuantum:
bufPrint and allocPrint at runtime

Should reduce compile time and binary size, and reduces `catch unreachable`

follow-up on #2792

Perfomance:
. | Before | After | Binary size
--- | --- | --- | ---
Debug + llvm | 13.75 s | 13.3 s | - 0.3 MB
Debug + x86_64 | 4.85 s | 4.75 s | + 12.4 MB (the whole binary is over 600 MB, probably for incremental)
ReleaseSafe | 82.8 s | 81.2 s | - 0.4 MB
ReleaseFast | 90.0 s | 85.5 s | - 0.8 MB

Remaining work:
- [x] Make performance comparisons
- [ ] Check if it catches mistakes at compile-time when running tests
- [ ] There is a problem with vectors and precision

### DIFF
--- a/src/assets.zig
+++ b/src/assets.zig
@@ -378,8 +378,8 @@ fn registerItem(assetFolder: []const u8, id: []const u8, zon: ZonElement) !void 
 	var replacementTexturePath: []const u8 = &.{};
 	defer main.stackAllocator.free(replacementTexturePath);
 	if (zon.get(?[]const u8, "texture", null)) |texture| {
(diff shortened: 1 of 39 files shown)


IntegratedQuantum:
I have decided to discard this for now, since the x86 backend is now usable, it doesn't provide a meaningful compile-time performance improvement, and there are many edge-cases that aren't handled correctly (e.g. printing characters)

PR #2808, opened 2026-04-03
add playerIndex option to spawn command

Wunka:


src/server/command/spawn.zig
```diff
@@ -5,22 +5,26 @@ const User = main.server.User;
 
 const command = @import("_command.zig");
 
-pub const description = "Get or set the player spawn point";
+pub const description = "Get or set a players spawn point";
```
IntegratedQuantum:
```suggestion
pub const description = "Get or set a player's spawn point";
```

IntegratedQuantum (changes requested):
Please fix conflicts with the PR that added the world spawn setting.

Wunka:
fixed conflicts

IntegratedQuantum (changes requested):
It crashes when I type `/spawn`

Wbjitscool:
That's not good

Wunka:
> It crashes when I type `/spawn`

mhhh before `userIndex[0] == '@'` worked even on userIndex wich had length 0... this probably needs to be investigated if such an if exists somewhere else and could crash the game

Wunka:
Nope. I was wrong. I didn't work the entire time... wtf. but its now fixed.

IntegratedQuantum (changes requested):
`/spawn @0` crashes now

Wunka:
I think I understand now completly where this came from. 
When the iterator has no results instead of giving `null` it gives a string with `len` 0 as the first result. This brought both bugs, as both Target and the general spawn command didn't check this behavior

IntegratedQuantum (approved):

PR #2812, opened 2026-04-04
Add default implementation to some of the modifiers virtual functions

Argmaster:
I am looking to add a ~~callback~~ hook (possibly even few in long term) to the modifier vtable and it gets impractical to define all of them every time, so in this pull request I am proposing to add default implementations for some of the callbacks so we need less boilerplate:
- `changeToolParameters`
- `changeBlockDamage`


src/items.zig
```diff
@@ -162,6 +162,24 @@ const Modifier = struct {
...
+			return .{
+				.changeToolParameters = @ptrCast(if (@hasDecl(ModifierStruct, "changeToolParameters")) &ModifierStruct.changeToolParameters else &VTable.Defaults.changeToolParameters),
+				.changeBlockDamage = @ptrCast(if (@hasDecl(ModifierStruct, "changeBlockDamage")) &ModifierStruct.changeBlockDamage else &VTable.Defaults.changeBlockDamage),
+				.combineModifiers = @ptrCast(&ModifierStruct.combineModifiers),
+				.printTooltip = @ptrCast(&ModifierStruct.printTooltip),
+				.loadData = @ptrCast(&ModifierStruct.loadData),
+				.priority = ModifierStruct.priority,
+			};
+		}
 	};
```
Argmaster, line 183: Can someone smart tell me why am I getting a segfault here :C

Argmaster: I was able to eliminate the segfault by making `initFromModifierStruct` `inline` and  `comptime`

IntegratedQuantum:
Please fix merge conflicts

IntegratedQuantum (approved):

PR #2817, opened 2026-04-04
Ferrock Hills

TyrrelMeter:
This pull request adds a new biome that has a flat clay expanse with large ferrock subiomes that tower above the ground.
Cubyz 0 3 0-dev 4_4_2026 12_13_35 PM
Cubyz 0 3 0-dev 4_4_2026 12_12_45 PM
Cubyz 0 3 0-dev 4_4_2026 12_01_48 PM
Cubyz 0 3 0-dev 4_4_2026 12_01_40 PM


H41ogen:
clay typically isn't just on the ground--it dries quickly into brittle rock

also this def needs more decorations, it's very bland rn

i like the idea a lot tho

TyrrelMeter:
For point A, I don't really care if the block is typically "on the ground" I think it looks cool and the alternatives would look worse, and for the second point, it does need more decorations. I guess I could up the cactus and dried grass chance.

Argmaster:
Please make dry trees (just branches and logs w/o leaves)

careeoki (changes requested):
Looks alright so far, but I would like to see this concept fleshed out further.

TyrrelMeter:
No longer interested in the PR.

PR #2819, opened 2026-04-04
Ok, so there is a bug in vine decay

Argmaster:
and I have no idea where.
But surely somewhere in the booleans which are negated over and over.
So lets not do that and make the logic simpler.
Here.


Argmaster:
Yeah and it works now btw

Issue #2734, opened 2026-03-17
Third person entity held item display


RanPix:
### Description

its the same as you can see your own held item but for other entities that aren't you

requires #2419

PR #2824, opened 2026-04-05
Entity model node system

### LINKED ISSUE #2419: joints and links for entity models

RanPix:
requires #2762, #3262 and #2682
fixes #2419 
progress towards #87, #2734 and #726

https://github.com/user-attachments/assets/1b1b399b-07b0-496a-b24b-2e48bb305abb


tillpp:
I think its better to have the current animation state be stored in a component, instead of client/Entity.

so i would say, this requires: https://github.com/PixelGuys/Cubyz/pull/2681 ECS8

RanPix:
by animation state you mean the node list?

src/entityModel.zig
```diff
@@ -29,7 +29,7 @@ pub const EntityModel = struct {
 
 		// TODO: add a matrix and a dirty flag
 		parent: ?u16 = null,
-	}; // hi
+	};
```
tillpp: hi

RanPix: hi
silly gatogif

Wunka: hi

RanPix:
ready for review

assets/cubyz/entityModels/cubert.zig.zon
```diff
@@ -5,5 +5,6 @@
 	.model = "cubyz:cubert",
 	.defaultTexture = "cubyz:cubert",
 	.height = 2,
-	.coordinateSystem = .right_handed_z_up,
+	.isPlayerModel = true,
```
IntegratedQuantum: unintended change

assets/cubyz/entityModels/snale.zig.zon
```diff
@@ -5,5 +5,6 @@
 	.model = "cubyz:snale",
 	.defaultTexture = "cubyz:snale",
 	.height = 2,
+	.isPlayerModel = true,
```
IntegratedQuantum: unintended change

assets/cubyz/entityModels/textures/snale.png
IntegratedQuantum, line 1: Could you explain what changed here? I see no difference

assets/cubyz/shaders/entity_vertex.vert
```diff
@@ -35,7 +37,7 @@ vec3 calcLight(uint fullLight) {
 void main() {
 	normal = inNormal;
 
-	vec4 mvPos = viewMatrix*vec4(inPos, 1);
+	vec4 mvPos = viewMatrix * nodeMatrices[inNodeID] * vec4(inPos, 1);
```
IntegratedQuantum: Please follow the formatting style
```suggestion
	vec4 mvPos = viewMatrix*nodeMatrices[inNodeID]*vec4(inPos, 1);
```

src/client/Entity.zig
```diff
@@ -77,6 +79,25 @@ pub fn update(self: *@This(), time: i16, lastTime: i16) void {
...
 	self.rot[1] = @floatCast(self.interpolatedValues.outPos[4]);
 	self.rot[2] = @floatCast(self.interpolatedValues.outPos[5]);
+
+	if (main.entity.components.@"cubyz:model".client.get(self.id)) |modelComp| {
+		const model = modelComp.entityModel.get();
+
+		const head = model.nodeReverse.get("Head");
+		if (model.nodeReverse.get("Eyestalks")) |eyestalksId| {
+			const stalkRot = self.rot[0]*0.25;
+			const headRot = self.rot[0]*0.75;
```
IntegratedQuantum: Instead of hardcoding it here, could this information be stored in the model somehow? Does it allow to add named variables to the nodes, so you could add to each node how much head rotation it gets?

assets/cubyz/shaders/entity_vertex.vert
```diff
@@ -13,6 +14,7 @@ layout(location = 0) uniform mat4 projectionMatrix;
 layout(location = 1) uniform mat4 viewMatrix;
 layout(location = 2) uniform vec3 ambientLight;
 layout(location = 3) uniform uint light;
+layout(location = 6) uniform mat4 nodeMatrices[20];
```
IntegratedQuantum: Instead of limiting this to 20 nodes, please use an SSBO. (also if we wanted to have so many uniforms in Vulkan, we'd need to use a UBO, which I'm sure you don't want to touch)

You can use the LargeBuffer to allocator memory for each entity, then you can use a uniform to specify the offset into the buffer.

src/client/Entity.zig
```diff
@@ -77,6 +79,25 @@ pub fn update(self: *@This(), time: i16, lastTime: i16) void {
 	self.rot[0] = @floatCast(self.interpolatedValues.outPos[3]);
 	self.rot[1] = @floatCast(self.interpolatedValues.outPos[4]);
 	self.rot[2] = @floatCast(self.interpolatedValues.outPos[5]);
+
+	if (main.entity.components.@"cubyz:model".client.get(self.id)) |modelComp| {
+		const model = modelComp.entityModel.get();
+
+		const head = model.nodeReverse.get("Head");
```
IntegratedQuantum: Why is it called reverse? reverse of what?

src/entityComponent/model.zig
```diff
@@ -25,6 +25,9 @@ pub const entityComponentVersion = 0;
 pub const client = struct {
 	const Component = struct {
 		entityModel: main.entityModel.EntityModelIndex,
+		hasLoaded: bool = false,
```
IntegratedQuantum: This is not used

src/entityComponent/model.zig
```diff
@@ -25,6 +25,9 @@ pub const entityComponentVersion = 0;
...
+		nodes: [20]main.entityModel.EntityModel.Node = undefined,
```
IntegratedQuantum: Please don't add an arbitrary limit, you are also wasting memory if you do it like that.

src/entityComponent/model.zig
```diff
@@ -25,6 +25,9 @@ pub const entityComponentVersion = 0;
...
+		matrices: [20]Mat4f = undefined,
```
IntegratedQuantum: I'm pretty sure you need to recalculate them every frame anyways (otherwise child nodes of e.g. the head won't be updated properly) , so why even store them?

src/entityModel.zig
```diff
@@ -27,11 +27,41 @@ pub const EntityModel = struct {
 	texturePath: []const u8,
 	modelId: ?[]const u8,
 
+	nodeReverse: std.StringHashMap(u16) = undefined,
+	nodes: NodeList = undefined,
+	nodeCount: u8,
```
IntegratedQuantum: Does gltf really only allow for 255 nodes?

src/entityModel.zig
```diff
@@ -27,11 +27,41 @@ pub const EntityModel = struct {
...
+
 	vao: ?graphics.VertexArray = null,
 	indexCount: c_int,
 	defaultTexture: ?main.graphics.Texture,
 	coordinateSystem: CoordinateSystem,
 
+	pub const maxNodesCount = 20;
+	pub const NodeList = [maxNodesCount]Node;
+	pub const MatrixList = [maxNodesCount]Mat4f;
```
IntegratedQuantum: unused

src/entityModel.zig
```diff
@@ -27,11 +27,41 @@ pub const EntityModel = struct {
...
+		pub fn getHierarchyMatrix(self: Node, nodes: NodeList) Mat4f {
+			var mat = self.originMat.mul(Mat4f.translation(self.pos));
+			mat = mat.mul(Mat4f.rotationQuat(self.rot));
+			mat = mat.mul(Mat4f.scale(self.scale));
+
+			if (self.parent == null) {
+				return mat;
+			}
+
+			return nodes[self.parent.?].getHierarchyMatrix(nodes).mul(mat);
```
IntegratedQuantum: This is a very inefficient way of computing all nodes. I'd suggest to use memoization to avoid recalculating parent matrices multiple times each frame, or better use a top-down approach to calulate them, starting at the root, and calculate them using child indices.

src/entityModel.zig
```diff
@@ -27,11 +27,41 @@ pub const EntityModel = struct {
...
+	pub const MatrixList = [maxNodesCount]Mat4f;
+
+	pub const Node = struct {
+		pos: Vec3f = @splat(0),
+		rot: Vec4f = Vec4f{0, 0, 0, 1},
+		scale: Vec3f = @splat(1),
+
+		originMat: Mat4f,
+
+		parent: ?u16 = null,
```
IntegratedQuantum: Why are you using different types for the node index?

src/entityModel.zig
```diff
@@ -43,6 +73,7 @@ pub const EntityModel = struct {
 		pos: [3]f32,
 		normal: [3]f32,
 		uv: [2]f32,
+		nodeID: c_uint,
```
```suggestion
		nodeId: c_uint,
```

src/entityModel.zig
```diff
@@ -60,6 +91,11 @@ pub const EntityModel = struct {
 				.format = c.VK_FORMAT_R32G32_SFLOAT,
 				.offset = @offsetOf(@This(), "uv"),
 			},
+			.{
+				.location = 3,
+				.format = c.VK_FORMAT_R32_UINT,
+				.offset = @offsetOf(@This(), "nodeID"),
```
```suggestion
				.offset = @offsetOf(@This(), "nodeId"),
```

src/vec.zig
```diff
@@ -114,6 +114,48 @@ pub fn rotate2d(self: anytype, angle: @typeInfo(@TypeOf(self)).vector.child, cen
 	} + center;
 }
 
+// copied from zmath library (MIT Liscence) : https://github.com/zig-gamedev/zmath/blob/9f7beb0753bd5cf885285dda8b00361c87c5b6b3/src/root.zig#L3570
+fn sincos32(v: f32) [2]f32 {
```
IntegratedQuantum: Please just use `@sin` and `@cos` or if you really think we need it, you can use sincos from math.h

I'd rather not have too much foreign code in here.

src/vec.zig
```diff
@@ -114,6 +114,48 @@ pub fn rotate2d(self: anytype, angle: @typeInfo(@TypeOf(self)).vector.child, cen
...
+	cosv = @mulAdd(f32, cosv, y2, -0.0013888378);
+	cosv = @mulAdd(f32, cosv, y2, 0.041666638);
+	cosv = @mulAdd(f32, cosv, y2, -0.5);
+	cosv = sign*@mulAdd(f32, cosv, y2, 1.0);
+
+	return .{sinv, cosv};
+}
+
+// copied from zmath library (MIT Liscence) :  https://github.com/zig-gamedev/zmath/blob/9f7beb0753bd5cf885285dda8b00361c87c5b6b3/src/root.zig#L2985
+pub fn quatFromAxisAngle(axis: Vec3f, angle: f32) Vec4f {
```
IntegratedQuantum: Please add a `Quat` type that contains a vector, so we have at least some type safety.
also add a `MARK` for it as well.

src/entityModel.zig
```diff
@@ -164,63 +213,96 @@ pub const EntityModel = struct {
...
+					const idx = indicesAccessor.read_index(i);
+					indicesSlice[i] = @as(u32, @intCast(idx)) + baseVertex;
+				}
 
-						var uv: [2]f32 = undefined;
-						_ = uvAttr.read_float(v, @ptrCast(&uv), 2);
-						vertSlice[v].uv = .{uv[0], 1 - uv[1]};
+				var positionAttr: gltf.cgltf_accessor = undefined;
+				var normalAttr: gltf.cgltf_accessor = undefined;
+				var uvAttr: gltf.cgltf_accessor = undefined;
```
IntegratedQuantum: Could you please make a seperate PR for just the indentation change you did here?
This is just awful to review in github.

IntegratedQuantum (changes requested):

src/client/Entity.zig
```diff
@@ -77,6 +79,25 @@ pub fn update(self: *@This(), time: i16, lastTime: i16) void {
...
+			const headRot = self.rot[0]*0.75;
+			modelComp.nodes[eyestalksId].rot = vec.quatFromAxisAngle(Vec3f{1, 0, 0}, stalkRot);
+			modelComp.matrices[eyestalksId] = modelComp.nodes[eyestalksId].getHierarchyMatrix(modelComp.nodes);
+
+			const headId = head.?;
+			modelComp.nodes[headId].rot = vec.quatFromAxisAngle(Vec3f{1, 0, 0}, headRot);
+			modelComp.matrices[headId] = modelComp.nodes[headId].getHierarchyMatrix(modelComp.nodes);
+		} else if (head) |headId| {
+			modelComp.nodes[headId].rot = vec.quatFromAxisAngle(Vec3f{1, 0, 0}, self.rot[0]);
+			modelComp.matrices[headId] = modelComp.nodes[headId].getHierarchyMatrix(modelComp.nodes);
```
tillpp: Is it possible to move this code somewhere else?
I would prefer the Entity.zig be component agnostic

tillpp (changes requested):
I appreciate the effort, but request a change:

assets/cubyz/entityModels/textures/snale.png
RanPix: nothing, just an artifact of debugging, reverted

assets/cubyz/entityModels/cubert.zig.zon
RanPix: it is intended because i updated the model to be correct

src/vec.zig
RanPix: do you want a vector alias or a struct with a vec4?

src/client/Entity.zig
RanPix: maybe we could add custom data type definitions to the model type zon file? so that we could create custom variables like "rotationLimitMultiplier"
```
.nodeVariables = .{
    .Head = .{
          .rotationLimitMultiplier = .{ .f32 = 0.75}
     }
}
```

i would also like to solve the problem of using strings to get the nodes and maybe later also for getting those variables if we add them because this is error prone and slow, but im not sure how yet

src/client/Entity.zig
RanPix: can i create some kind of a a model processor system and move it there currently?

src/vec.zig
IntegratedQuantum: Multiplication and stuff works differently on it right? Then I'd like to have a struct.

src/client/Entity.zig
IntegratedQuantum: Does gltf have some sort of custom data fields? How do other people solve this, surely we are not the only people to have this problem.

RanPix: gltf has extras data section, but im not sure how different programs fill this out, havent seen this being filled with anything but i will look again if its possible to do at laest in blockbench and blender

RanPix: also even if it is possible to get the extras data from a mesh i think this is still not good as we cannot the the named variable data inside the mesh file in case we need it, only if we import it back in some mesh editor but that doesnt sound good tbh

RanPix: so i checked, blender has that kind of a thing and it stores the data in extras, but the thing is gltf spec says that extras are stored as json, so we would need to bring back json parsing lol, and also blockbench doesnt have anything like that as far as i can see

IntegratedQuantum: Alright, then we keep it as is I guess.

src/entityComponent/model.zig
RanPix: i need to store those matrices separately because i need to upload the raw transform matrix data to gpu, if i just upload the nodes then we would be wasting memory on the gpu

RanPix: but i still can create a large buffer for both nodes and matrices to not waste memory with 20 nodes for each model i could just not bind the node data to the ssbo

src/entityComponent/model.zig
RanPix: so you want to allocate a node list for each component on ram?

IntegratedQuantum: I don't see any other choice.

src/entitySystem/_list.zig
```diff
@@ -1,2 +1,3 @@
 pub const _template = @import("_template.zig");
+pub const nodeProcessor = @import("nodeProcessor.zig");
```
IntegratedQuantum: These lists should be sorted alphabetically to prepare for auto-generating.
Since you do have a sensitive order of operations here, you should either put the dependent code paths into one system, or add a way to define the order of how systems are executed (which we will probably need in the future anyways).

src/entitySystem/nodeProcessor.zig
```diff
@@ -0,0 +1,98 @@
...
+				component.nodes[eyestalksId].setRot(vec.quatFromAxisAngle(Vec3f{1, 0, 0}, stalkRot));
+
+				const headId = head.?;
+				component.nodes[headId].setRot(vec.quatFromAxisAngle(Vec3f{1, 0, 0}, headRot));
+			} else if (head) |headId| {
+				component.nodes[headId].setRot(vec.quatFromAxisAngle(Vec3f{1, 0, 0}, ent.rot[0]));
+			}
+
+			for (component.nodes, 0..) |*node, i| {
+				if ((node.parent != null and component.nodes[node.parent.?].version > node.parentVersion)) {
```
IntegratedQuantum: Instead of version management and dirty flags I'd suggest to just regenerate everything every frame. In the end we will probably have an idle animation that has every limb move slightly all the time, which means there is little benefit from caching these values. And if this ever becomes a bottleneck then the way to go would be putting it on the GPU anyways, where branches aren't particularly efficient.

src/entitySystem/nodeProcessor.zig
```diff
@@ -0,0 +1,98 @@
+const std = @import("std");
```
IntegratedQuantum: In my opinion the matrix calculation code belongs inside the modelRenderer, all other rendering operations calculate their matrices by themselves right before uploading them to the GPU.

The only thing that makes sense outside of it, would be the head rotation code, but until we have a way to define ordering of systems, I'd suggest to just put it inside modelRenderer as well.

IntegratedQuantum (changes requested):
I'd suggest to split this into 3 parts:
1. Introduce a `Quat` type as mentioned in https://github.com/PixelGuys/Cubyz/pull/2824/changes#r3184026802
2. Dynamically calculate the matrices instead of baking them in the model, but don't add any rotation/translation parameters yet, only use the nodePivots.
3. Add the remaining parameters and do the head/eye rotation

RanPix:
> Dynamically calculate the matrices instead of baking them in the model, but don't add any rotation/translation parameters yet, only use the nodePivots.

do you mean to ditch matrix baking entirely or only for the nodes themselves?

IntegratedQuantum:
> do you mean to ditch matrix baking entirely or only for the nodes themselves?

What do you mean by that? Is there any other baking going on?

RanPix:
>    What do you mean by that? Is there any other baking going on?

it bakes any mesh/primitive that is a child of a node, to save on amount of nodes processed and stored but the nodes themselves are dynamic

IntegratedQuantum:
Then we should keep baking it

RanPix:
> Add the remaining parameters and do the head/eye rotation

also what do you mean by that? are you talking about that suggestion i made to add node parameters to the model .zon file?

src/vec.zig
RanPix: #3139

IntegratedQuantum:
> also what do you mean by that? are you talking about that suggestion i made to add node parameters to the model .zon file?

I mean all the parameters in the Node struct except from the pivot matrix, the names of the nodes. Step 2 should only unbake the matrices, without any other new parameters.

RanPix:
> I mean all the parameters in the Node struct except from the pivot matrix, the names of the nodes. Step 2 should only unbake the matrices, without any other new parameters.

so then i will delete all of the transform parameters `pos, rot, scale`, parent is needed to calculate children properly, but do i move the hierarchy update logic to component load function so that it doesnt run all the time for now? or its fine to keep it updating in model renderer system? there isnt much to split really

IntegratedQuantum:
> or its fine to keep it updating in model renderer system?

Yes, that's the idea to put it already in the place where it needs to be, so that step 3 doesn't move any major pieces of code.

RanPix:
ready for review

src/game.zig
```diff
@@ -298,13 +298,12 @@ pub const World = struct { // MARK: World
...
 
-		main.Window.setMouseGrabbed(true);
-
-		main.blocks.meshes.generateTextureArray();
-		main.particles.ParticleManager.generateTextureArray();
-		main.models.uploadModels();
-		main.entityModel.loadModelsAndTexture();
+		try self.finishHandshake(zon);
+		main.network.protocols.handShake.assetsLoadedCondition.signal();
+		main.network.protocols.handShake.hasFinishedLoadingAssets = true;
```
IntegratedQuantum: Didn't you say you would make a separate PR for this?

RanPix: https://github.com/PixelGuys/Cubyz/pull/3262

RanPix:
fixed merge conflicts

RanPix:
requires #3262 before merge

src/client/Entity.zig
```diff
@@ -14,6 +14,7 @@ const Mat4f = vec.Mat4f;
 const Vec3d = vec.Vec3d;
 const Vec3f = vec.Vec3f;
 const Vec4f = vec.Vec4f;
+const EntityModel = main.entityModel.EntityModel;
```
IntegratedQuantum: unintended change

src/entitySystem/modelRenderer.zig
```diff
@@ -110,12 +115,30 @@ pub const client = struct {
 		_ = deltaTime;
 		main.client.entity_manager.mutex.lock();
 		defer main.client.entity_manager.mutex.unlock();
+
+		// TODO: move to a separate system after we get custom system ordering
```
IntegratedQuantum: TODO: Make or mention your TODOs in an issue

src/entityComponent/model.zig
```diff
@@ -27,6 +27,17 @@ pub const entityComponentVersion = 0;
 pub const client = struct {
 	const Component = struct {
 		entityModel: main.entityModel.EntityModelIndex,
+
+		bufferAllocation: graphics.SubAllocation = .{.len = 0, .start = 0},
+		nodes: []EntityModel.Node = undefined,
```
IntegratedQuantum: Now Node only stores its parent, and that data shouldn't change, so I think it would be better just just access it through the EntitiyModelIndex instead of storing a pointless copy here.

src/entityModel.zig
```diff
@@ -109,6 +137,10 @@ pub const EntityModel = struct {
 	}
 
 	fn cloneMetaData(self: *EntityModel) EntityModel {
+		const newNodes = main.worldArena.alloc(Node, self.nodes.len);
+		@memcpy(newNodes, self.nodes);
+		const newNodePivots = main.worldArena.alloc(Mat4f, self.nodePivots.len);
+		@memcpy(newNodePivots, self.nodePivots);
```
IntegratedQuantum: There is a function called allocator.dupe for just this

src/entityModel.zig
```diff
@@ -69,6 +86,17 @@ pub const EntityModel = struct {
 		self.indexCount = 0;
 		self.coordinateSystem = zon.get(CoordinateSystem, "coordinateSystem") orelse .right_handed_z_up;
 
+		self.nodeIndexMap = .init(main.worldArena.allocator);
+		self.nodes = main.worldArena.alloc(Node, 0);
+		self.nodePivots = main.worldArena.alloc(Mat4f, 0);
```
```suggestion
		self.nodePivots = &.{};
```

src/entityModel.zig
```diff
@@ -69,6 +86,17 @@ pub const EntityModel = struct {
 		self.indexCount = 0;
 		self.coordinateSystem = zon.get(CoordinateSystem, "coordinateSystem") orelse .right_handed_z_up;
 
+		self.nodeIndexMap = .init(main.worldArena.allocator);
+		self.nodes = main.worldArena.alloc(Node, 0);
```
```suggestion
		self.nodes = &.{};
```

src/entityModel.zig
```diff
@@ -52,9 +62,16 @@ pub const EntityModel = struct {
...
 			},
+			.{
+				.location = 3,
+				.format = c.VK_FORMAT_R32_UINT,
+				.offset = @offsetOf(@This(), "nodeId"),
+			},
 		};
 	};
 
+	const NodeRemap = struct { depth: u16, gltfNodeIdx: u32 };
```
IntegratedQuantum: No unnecessary abbreviations
```suggestion
	const NodeRemap = struct { depth: u16, gltfNodeIndex: u32 };
```

src/entityModel.zig
```diff
@@ -69,6 +86,17 @@ pub const EntityModel = struct {
 		self.indexCount = 0;
 		self.coordinateSystem = zon.get(CoordinateSystem, "coordinateSystem") orelse .right_handed_z_up;
 
+		self.nodeIndexMap = .init(main.worldArena.allocator);
+		self.nodes = main.worldArena.alloc(Node, 0);
+		self.nodePivots = main.worldArena.alloc(Mat4f, 0);
+		self.nodeCount = 0;
+
+		if (zon.getChildOrNull("isPlayerModel")) |isPlayerModel| {
+			if (isPlayerModel.as(bool) orelse false) {
```
```suggestion
		if (zon.get(bool, "isPlayerModel") orelse false) {
```

src/entityModel.zig
```diff
@@ -157,9 +193,56 @@ pub const EntityModel = struct {
 		defer indices.deinit(main.stackAllocator);
 		var baseVertex: u32 = 0;
 
+		var nodeDepthRemap: main.List(NodeRemap) = .empty;
```
IntegratedQuantum, line 197: The functionality for NodeRemap is scattered through half the file, yet only used inside this function. Please declare it together with its compare function locally inside this function.

src/entityModel.zig
```diff
@@ -157,9 +193,56 @@ pub const EntityModel = struct {
...
+
+		self.nodes = main.worldArena.alloc(Node, nodeCount);
+		self.nodePivots = main.worldArena.alloc(Mat4f, nodeCount);
+
+		for (nodeDepthRemap.items, 0..) |nodeRemap, i| {
+			const node = data.nodes[nodeRemap.gltfNodeIdx];
+
+			const nameC = std.mem.span(node.name);
+			const name = main.globalArena.alloc(u8, nameC.len);
+			@memcpy(name, nameC);
```
IntegratedQuantum: dupe(u8, std.mem.span(...));

src/entityModel.zig
```diff
@@ -157,9 +193,56 @@ pub const EntityModel = struct {
...
+			self.nodeIndexMap.put(name, @intCast(i)) catch unreachable;
```
IntegratedQuantum: Does the specification say that the name has to be unique? What if it isn't?

src/entityModel.zig
```diff
@@ -157,9 +193,56 @@ pub const EntityModel = struct {
...
+		}
+
 		for (data.nodes[0..data.nodes_count]) |node| {
 			if (node.mesh != null) {
-				const finalMat = getHierarchyMatrix(node, self.coordinateSystem);
+				var finalMat = Mat4f.translation(self.coordinateSystem.convertVec(node.translation, @splat(0)));
+				finalMat = finalMat.mul(Mat4f.rotationQuat(self.coordinateSystem.convertQuat(node.rotation)));
+				finalMat = finalMat.mul(Mat4f.scale(self.coordinateSystem.convertScale(node.scale)));
+
+				const parentNodeID = if (node.parent) |p| self.nodeIndexMap.get(std.mem.span(p.*.name)).? else 0;
```
IntegratedQuantum: Can this happen in well-specified models? If not I'd suggest to print an error.

IntegratedQuantum (changes requested):

src/entityModel.zig
RanPix: i was looking for "clone", forgot about "dupe" xd

src/entitySystem/modelRenderer.zig
RanPix: now good? #3342

src/entityModel.zig
RanPix: it has to be unique yes, but in the 3d editors you cannot name something with the same name

src/entityModel.zig
RanPix: in theory all primitives should be contained inside a parent node so yeah having an error would be better

src/entityComponent/model.zig
RanPix: can i  leave current node struct as empty so that i dont have to readd it again later?

IntegratedQuantum: Do you plan to add more fields? Then it's fine

RanPix: yes, this is going to have positions

IntegratedQuantum: I would still suggest to separate the variable from the static functionality, I doubt you'd ever change the parent index, right?

RanPix: yeah im going to separate the parent index from this struct and store it as a separate array in the model struct, but because of this the struct is going to be empty for some time untill i add node movement

RanPix:
done

assets/cubyz/entity_models/models/cubert.glb
IntegratedQuantum, line 1: Why are these here now?

src/entityComponent/model.zig
```diff
@@ -27,6 +27,15 @@ pub const entityComponentVersion = 0;
 pub const client = struct {
 	const Component = struct {
 		entityModel: main.entityModel.EntityModelIndex,
+
+		bufferAllocation: graphics.SubAllocation = .{.len = 0, .start = 0},
+		matrices: []Mat4f = undefined,
```
IntegratedQuantum, line 32: Do we even need these here? It seems that you overwrite them each frame and then upload them, and they are not used for anything else.

IntegratedQuantum (changes requested):

assets/cubyz/entity_models/models/cubert.glb
RanPix: git automerged it for me after https://github.com/PixelGuys/Cubyz/pull/3338

src/entityComponent/model.zig
RanPix: yes because moving matrices to the entity model would be annoying and i would need to then re add all of the things back again to the component and make it work in future, i cant really do it any other way nicely except for only uploading the node matrix data once when the component is created

assets/cubyz/entity_models/models/cubert.glb
IntegratedQuantum: Can you make it go away? I don't want it to accidentally create new files on merge.

src/entityComponent/model.zig
IntegratedQuantum: I don't really understand this. Why do you need the matrices after uploading them to the GPU?

assets/cubyz/entity_models/models/cubert.glb
RanPix: what do you mean go away? its fine there is no any new files or folders, it just moved everything to that folder properly

Image

Image

IntegratedQuantum: I'm just confused because it shows these as additions in the ui.

src/entityComponent/model.zig
RanPix: the matrices will be needed for things like holding items in the hand, maybe even accessories, and i guess i could just create a matrix array  and calculate all matrices every time for each entity but that would require a reallocation for each entity which i imagine would be annoying to work with

assets/cubyz/entity_models/models/cubert.glb
RanPix: i think its just a modification icon, how it was before because i merged with master it just moved everything there and all files i touched there were changed, and it doesnt show that any textures were added so it shows that it is correct

src/entityModel.zig
```diff
@@ -69,6 +84,16 @@ pub const EntityModel = struct {
...
 		self.coordinateSystem = zon.get(CoordinateSystem, "coordinateSystem") orelse .right_handed_z_up;
 
+		self.nodeIndexMap = .init(main.worldArena.allocator);
+		self.nodes = &.{};
+		self.nodeParents = &.{};
+		self.nodePivots = &.{};
+		self.nodeCount = 0;
+
+		if (zon.get(bool, "isPlayerModel") orelse false) {
+			playerEntityModels.append(main.worldArena, index);
```
IntegratedQuantum: Why is this done twice?

src/entityModel.zig
```diff
@@ -157,9 +190,64 @@ pub const EntityModel = struct {
 		defer indices.deinit(main.stackAllocator);
 		var baseVertex: u32 = 0;
 
+		const NodeRemap = struct {
+			depth: u16,
+			gltfNodeIndex: u32,
```
IntegratedQuantum: Everywhere else you use a u16 for the node index, I'd suggest to use one here to to avoid needless casting

IntegratedQuantum (changes requested):

src/entityModel.zig
RanPix: no idea

RanPix:
done

IntegratedQuantum:
nevermind

IntegratedQuantum (approved):

PR #2825, opened 2026-04-05
worldedit selection helper function + improvements to error feedback

### LINKED ISSUE #2816: Add a helper function to get worldedit positions
Currently this code needs to be in every command that edits the selection:
```zig
const pos1 = source.worldEditData.selectionPosition1 orelse {
	return source.sendMessage("#ff0000Position 1 is not set.", .{});
};
const pos2 = source.worldEditData.selectionPosition2 orelse {
	return source.sendMessage("#ff0000Position 2 is not set.", .{});
};

const posStart: Vec3i = @min(pos1, pos2);
const posEnd: Vec3i = @max(pos1, pos2);
```
extracted from #2797

H41ogen:
Closes #2816

### DIFF
--- a/src/server/command/_command.zig
+++ b/src/server/command/_command.zig
@@ -117,3 +117,17 @@ pub const Target = struct {
 		if (self.increasedRefCount) self.user.decreaseRefCount();
 	}
 };
+
+pub fn getSelectionBounds(source: *User) error{PositionNotSet}![2]main.vec.Vec3i {
+	const pos1 = source.worldEditData.selectionPosition1 orelse blk: {
+		source.sendMessage("#ff0000Position 1 is not set.", .{});
+		break :blk null;
+	};
+	const pos2 = source.worldEditData.selectionPosition2 orelse blk: {
+		source.sendMessage("#ff0000Position 2 is not set.", .{});
+		break :blk null;
+	};
+	if (pos1 == null or pos2 == null) return error.PositionNotSet;
+
+	return .{@min(pos1.?, pos2.?), @max(pos1.?, pos2.?)};
+}
--- a/src/server/command/worldedit/copy.zig
+++ b/src/server/command/worldedit/copy.zig
@@ -6,6 +6,8 @@ const User = main.server.User;
 const Block = main.blocks.Block;
 const Blueprint = main.blueprint.Blueprint;
 
+const command = @import("../_command.zig");
+
 pub const description = "Copy selection to clipboard.";
 pub const usage = "/copy";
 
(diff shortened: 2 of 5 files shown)


src/server/command/_command.zig
```diff
@@ -117,3 +117,17 @@ pub const Target = struct {
 		if (self.increasedRefCount) self.user.decreaseRefCount();
 	}
 };
+
+pub fn getSelectionBounds(source: *User) error{PositionNotSet}![2]main.vec.Vec3i {
```
Argmaster, line 121: I would prefer to return a struct here instead of a vector to keep things explicit:

```zig
const Selection = struct {
	minPos: Vec3i,
	maxPos: Vec3i,
}
```

Its a bit silly, but I think we should distinguish between that arbitrary pos1/pos2 and the normalized min/max positions.

Argmaster: Alternatively we could just change variable names in caller, but that struct will automatically enforce consistent names in callers.

Argmaster: Actually we don't need to min/max it - `Blueprint.capture` does that anyway no?

Argmaster: I would like this interface improved anyway tho; its a little bit out of scope, but it would be nice to change `Blueprint.capture` to accept `Selection` that has `fn init(pos1, pos2)` and that `init` min/maxes the positions.

> src/blueprint.zig:77

```zig
pub fn capture(allocator: NeverFailingAllocator, selection: Selection) CaptureResult {
```

`Selection` struct should be placed inside `blueprint.zig`

```zig
const Selection = struct {
	minPos: Vec3i,
	maxPos: Vec3i,

	fn init(pos1, pos2) Selection {
		...	
	}
}
```

Then we can remove min/max-ing from `capture`

```zig
/// Remove:
		const startX = @min(pos1[0], pos2[0]);
		const startY = @min(pos1[1], pos2[1]);
		const startZ = @min(pos1[2], pos2[2]);

		const endX = @max(pos1[0], pos2[0]);
		const endY = @max(pos1[1], pos2[1]);
		const endZ = @max(pos1[2], pos2[2]);
```

And just go

```zig
const startX, const startY, const startZ = selection.minPos;
const endX , const endY , const endZ = selection.maxPos;
```

We can also add `fn size` to `Selection` and remove it from `capture` body (`src/blueprint.zig:86`)

IntegratedQuantum (changes requested):
I agree with @Argmaster: This should be a struct with min and max for better readability. 
You don't need to do all the interface changes he suggested here, but at least the newly introduced function should return a struct.

Argmaster:
Will you work on that @H41ogen anytime soon? I might take this over since it blocks #3043

Argmaster:
Too late, already did the changes

H41ogen:
tis why i closed it

PR #2826, opened 2026-04-05
Add cave layers

IntegratedQuantum:
This moves (most) height restrictions of biomes to layers instead. Biomes are now assigned to layers using tags. This allows having multiple distinct layers of caves broken up by other biomes (→#2464 #1463).

Furthermore this will make cave layers more addon friendly. e.g. an addon could add their own layer between e.g. the caves and the root. The engine would then automatically shift down the layers below by the size of the new layer.

Note: The depth hint only decides the order of the layers and is not representative of the actual depth. The actual depth is computed using the layer heights of all layers above it (or below for sky layers).

Root and sky layers are currently set to 10000 to allow addons to place other things above/below it. However as long as no such layers exist, the root/sky will continue until the integer limit.

To @ikabod-kee and @careeoki I'll put this in requires artistic review for you to look at the new way to specify cave biome height and the new layers.

### DIFF
--- a/src/server/terrain/biomes.zig
+++ b/src/server/terrain/biomes.zig
@@ -298,6 +298,13 @@ pub const Biome = struct { // MARK: Biome
 			.maxSubBiomeCount = zon.get(f32, "maxSubBiomeCount", std.math.floatMax(f32)),
 			.tags = Tag.loadTagsFromZon(main.worldArena, zon.getChild("tags")),
 		};
+		if (self.isCave) {
+			for (self.tags) |tag| {
+				if (std.mem.endsWith(u8, tag.getName(), "_layer")) break;
+			} else {
+				std.log.err("Cave biome {s} is missing a '_layer' tag to assign it to a cave layer.", .{id});
+			}
+		}
 		if (minRadius > maxRadius) {
 			std.log.err("Biome {s} has invalid radius range ({d}, {d})", .{self.id, minRadius, maxRadius});
 		}
--- a/assets/cubyz/biomes/cave/mushroom/_defaults.zig.zon
+++ b/assets/cubyz/biomes/cave/mushroom/_defaults.zig.zon
@@ -1,5 +1,6 @@
 .{
 	.isCave = true,
+	.tags = .{.cave_layer},
 	.fogDensity = 2,
 	.music = "cubyz:heart-of-the-beast",
 	.structures = .{
--- a/assets/cubyz/biomes/cave/mushroom/big/_defaults.zig.zon
+++ b/assets/cubyz/biomes/cave/mushroom/big/_defaults.zig.zon
@@ -1,7 +1,7 @@
 .{
 	.isCave = true,
+	.tags = .{.cave_layer},
 	.maxHeight = -512,
-	.minHeight = -48250,
 	.music = "cubyz:la_kubisa_caverna",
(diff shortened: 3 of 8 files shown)


Crepestrom:
im not sure i understand could you give a bit more detail on how it works?

Crepestrom:
DepthHint may not be the best name for the variable
something like BaseDepth or similar may work better

Crepestrom:
also what happens when layers overlap?

IntegratedQuantum:
> DepthHint may not be the best name for the variable
something like BaseDepth or similar may work better

I wanted to emphasize that it's a hint and does not necessarily reflect the actual height

> also what happens when layers overlap?

That can't happen. One of them will be placed first, and the other one will be placed after the first one's layerHeight

ikabod-kee:
Ooooh hell yes this is sick. Can layers overlap, or is that not allowed?

assets/cubyz/biomes/cave/sky_islands/_defaults.zig.zon
```diff
@@ -1,6 +1,6 @@
 .{
 	.isCave = true,
-	.minHeight = 10000,
+	.tags = .{.sky_island_layer},
```
Argmaster, line 3: If I were to, by mistake, leave `.tags` empty, it would be nice to get an error.

Argmaster: Or rather, if I were to not assign a biome to any layer, that might happen even if there are tags

IntegratedQuantum: I think there are valid use-cases for having biomes that are not assigned to any layer. An addon may want to add biomes to layers introduced by other addons, while still allowing you to use it standalone.

Argmaster: Well, for that I would say to provide an extra tag `.no_cave_layer_required` or alike, so we make sure this is not done by mistake.

Btw I assume that biomes which are not part of a cave layer do not generate, right?

IntegratedQuantum: Either way good catch. I actually did miss some biomes.

I decided to require cave layers to end with `_layer`, this makes it easy to detect mistakes even when non-layer tags are present.

src/server/terrain/biomes.zig
```diff
@@ -298,6 +298,13 @@ pub const Biome = struct { // MARK: Biome
 			.maxSubBiomeCount = zon.get(f32, "maxSubBiomeCount", std.math.floatMax(f32)),
 			.tags = Tag.loadTagsFromZon(main.worldArena, zon.getChild("tags")),
 		};
+		if (self.isCave) {
+			for (self.tags) |tag| {
+				if (std.mem.endsWith(u8, tag.getName(), "_layer")) break;
```
Argmaster, line 303: Okay, but do we reserve a `no_layer` or similar tag for marking the biomes explicitly unassigned?

IntegratedQuantum: No, There is currently no reason to have an unassigned cave biome, since it wouldn't spawn under any circumstances. If you want to temporarily disable a biome you can also set the chance to 0.

PR #2831, opened 2026-04-06
Crimson wasteland revamp

Crepestrom:
image

image

image

The crimson wasteland has been made more pretty


Wunka:
This PR has formatting errors. please fix them, I will still move you to artistic review but before you get further these need to be fixed

TyrrelMeter:
I honestly don't think this makes the Crimson Wastelands better, only more dangerous. It is better off as a "wasteland" area, desolate not filled with lava.

PR #2834, opened 2026-04-06
Add `devouring` modifier that eats drops and heals player

Argmaster:
This pull request adds `devouring` modifier that, when present on a tool, heals player for a fraction of health of blocks broken.
To achieve this behavior in flexible way it was necessary to introduce new callback to modifier system called `onBlockUpdate` invoked when player causes a block update.

`devouring` modifier was unconditionally added to ruby. Currently it requires breaking around 3 stone blocks to heal for half a heart. Breaking higher quality blocks (ones with higher resistance / health) heals more. 

Some thoughts:

- it is consistent with planned combat based modifier `life-steal`.
- blocks broken with pickaxe that has this modifier do not drop items, player cannot infinitely heal themselves without exploring

Something to consider:

- further lower the healing effect, 3 stone blocks do not have much value, while half a heart heal is a huuuge thing. The issue is that effect is quite hard to notice even now, since the change has to be quite substantial to appear on health bar.
- make the healing work only on ores to require more exploration and more sacrifice to heal yourself, probably mutually exclusive with point above

Requires: #2892
Requires: #2812
Related-to: #172


RanPix:
i would suggest making it so that the more expensive items you break the more it heals you, but im not sure how would we determine the value

Argmaster:
> i would suggest making it so that the more expensive items you break the more it heals you, but im not sure how would we determine the value

This is usually positively correlated with health of the block broken. Its hard to establish the value of an item otherwise.

RanPix:
> This is usually positively correlated with health of the block broken. Its hard to establish the value of an item otherwise.

i would imagine adding a field in the item's data .zon file like `.value` or `.tier`, that might be used for other features in the future?

ZigguratV:
Thing is that block valuable-ty seems to be mostly related to defence and health. Creating `.value` or `.tier` might not be required as of now. Healing based on max hp of the enemy defeated could also make intuitive sense for weapon ruby property!
Having rubies provide strong healing in exchange for blocks sounds like a good effect. 3 blocks per half a heart is good now but in future we don't know how good wearable sources of healing will be, or how strong healing consumables will be. I would assume passive healing effects would drain energy/hunger. Balancing without knowing how PvE or PvP will work is difficult

src/proceduralItem/modifiers/devouring.zig
```diff
@@ -0,0 +1,34 @@
...
+	return .{.strength = std.math.hypot(data1.strength, data2.strength)};
+}
+
+pub fn onBlockUpdate(blockUpdate: main.sync.Command.UpdateBlock, ctx: main.sync.Command.Context, shouldDropSourceBlockOnSuccess: *bool, data: Data) void {
+	if (ctx.gamemode != .survival) return;
+	if (blockUpdate.oldBlock.typ == blockUpdate.newBlock.typ) return;
+
+	ctx.execute(.{.addHealth = .{
+		.target = ctx.user,
+		.health = data.strength*blockUpdate.oldBlock.blockHealth()*blockUpdate.oldBlock.blockResistance()*0.1,
```
IntegratedQuantum: What happens if block resistance is zero (quite a common value), shouldn't we still at least heal a little bit in that case?

src/proceduralItem/modifiers/devouring.zig
```diff
@@ -0,0 +1,34 @@
...
+	return .{.strength = @max(0, zon.get(f32, "strength", 0))};
+}
+
+pub fn combineModifiers(data1: Data, data2: Data) ?Data {
+	return .{.strength = std.math.hypot(data1.strength, data2.strength)};
+}
+
+pub fn onBlockUpdate(blockUpdate: main.sync.Command.UpdateBlock, ctx: main.sync.Command.Context, shouldDropSourceBlockOnSuccess: *bool, data: Data) void {
+	if (ctx.gamemode != .survival) return;
+	if (blockUpdate.oldBlock.typ == blockUpdate.newBlock.typ) return;
```
IntegratedQuantum, line 20: When the block was broken, but has a different required tool, then we heal, but don't consume durability.

src/game.zig
```diff
@@ -392,6 +392,7 @@ pub const DamageType = enum(u8) {
 	fall = 2,
 	heat = 3,
 	spiky = 4,
+	modifier = 5,
```
IntegratedQuantum: Maybe it should be called item damage?

src/items.zig
```diff
@@ -159,21 +160,30 @@ const Modifier = struct {
...
 		changeProceduralItemParameters: *const fn (proceduralItem: *ProceduralItem, data: Data) void,
 		changeBlockDamage: *const fn (damage: f32, block: main.blocks.Block, data: Data) f32,
+		onBlockUpdate: *const fn (blockUpdate: main.sync.Command.UpdateBlock, ctx: main.sync.Command.Context, shouldDropSourceBlockOnSuccess: *bool, data: Data) void,
 		printTooltip: *const fn (outString: *main.List(u8), data: Data) void,
 		loadData: *const fn (zon: ZonElement) Data,
 		priority: f32,
 
 		const Defaults = struct {
-			pub fn changeProceduralItemParameters(_: *ProceduralItem, _: Data) void {}
+			pub fn changeProceduralItemParameters(_: ProceduralItem, _: Data) void {}
```
IntegratedQuantum: How does that even compile?

src/items.zig
```diff
@@ -159,21 +160,30 @@ const Modifier = struct {
...
+				_ = data;
+				return;
+			}
 		};
 
 		inline fn initFromModifierStruct(comptime ModifierStruct: type) VTable {
 			return comptime .{
 				.changeProceduralItemParameters = @ptrCast(if (@hasDecl(ModifierStruct, "changeProceduralItemParameters")) &ModifierStruct.changeProceduralItemParameters else &VTable.Defaults.changeProceduralItemParameters),
 				.changeBlockDamage = @ptrCast(if (@hasDecl(ModifierStruct, "changeBlockDamage")) &ModifierStruct.changeBlockDamage else &VTable.Defaults.changeBlockDamage),
+				.onBlockUpdate = @ptrCast(if (@hasDecl(ModifierStruct, "onBlockUpdate")) &ModifierStruct.onBlockUpdate else &VTable.Defaults.onBlockUpdate),
```
IntegratedQuantum: Could you make the pointercast inside the if, so it doesn't cast the default?

src/items.zig
```diff
@@ -159,21 +160,30 @@ const Modifier = struct {
...
-			pub fn changeProceduralItemParameters(_: *ProceduralItem, _: Data) void {}
+			pub fn changeProceduralItemParameters(_: ProceduralItem, _: Data) void {}
 			pub fn changeBlockDamage(damage: f32, _: main.blocks.Block, _: Data) f32 {
 				return damage;
 			}
+			pub fn onBlockUpdate(blockUpdate: main.sync.Command.UpdateBlock, ctx: main.sync.Command.Context, shouldDropSourceBlockOnSuccess: *bool, data: Data) void {
+				_ = blockUpdate;
+				_ = ctx;
+				_ = shouldDropSourceBlockOnSuccess;
+				_ = data;
```
IntegratedQuantum: Use `_` in the parameter list, we don't need the typename here, it's already shown in the VTable.

IntegratedQuantum (changes requested):

src/game.zig
Argmaster: Renamed to `devouring` with funny message

src/items.zig
Argmaster: Pointer cast

src/items.zig
Argmaster: moved.

src/items.zig
Argmaster: ok

src/proceduralItem/modifiers/devouring.zig
Argmaster: Added `ProceduralItem` to signature, added check for isEffectiveOn

src/proceduralItem/modifiers/devouring.zig
Argmaster: I would ignore it, but since it was requested, I added `@abs(x) + 1` so its always above 0

src/items.zig
IntegratedQuantum: Please also do it in this line

Argmaster: yeah

src/game.zig
```diff
@@ -400,6 +401,7 @@ pub const DamageType = enum(u8) {
 			.fall => main.server.sendMessage("{s}§#ffffff died of fall damage", .{name}),
 			.heat => main.server.sendMessage("{s}§#ffffff burned to death", .{name}),
 			.spiky => main.server.sendMessage("{s}§#ffffff experienced death by 1000 needles", .{name}),
+			.devouring => main.server.sendMessage("{s}§#ffffff has been devoured by his own item", .{name}),
```
careeoki:
```suggestion
			.devouring => main.server.sendMessage("{s}§#ffffff has been devoured by their own item", .{name}),
```
don't gender the death messages

IntegratedQuantum (changes requested):
As for balancing: I think we should only activate the effect when encased in 6 (more than 4) precious, following the pattern of the other gems, but increasing the number since the effect is more impactful. Otherwise you can just stick a ruby on the handle and get that massive boost at the cost of just a little durability.

Also another thought I had: We have two independent actions here, consuming block drops and healing the player. I think these should be separate modifiers, so addon creators can compose them (to e.g. produce a cactus modifier that damages you, but does not eat your drops)

Argmaster:
> As for balancing: I think we should only activate the effect when encased in 6 (more than 4) precious, following the pattern of the other gems, but increasing the number since the effect is more impactful. 

I can agree with 4, but I am opposed to 6 as it occupies too many slots in crafting grid. Ruby itself is rare, has poor stats and requires traversing large distances. IMO it would be more interesting if you were forced to use this to devour valuable resources and balance it this way instead of focusing on the pickaxe itself.

 I would rather lower the value of healing and/or add fragile modifier instead of making it harder to use so you can break less blocks with it.

> Otherwise you can just stick a ruby on the handle and get that massive boost at the cost of just a little durability.

If you don't like your item drops then sure.

**Something I though about was to make the effect random and make the chance depend on encasing, eg. 10% by default, +10% for every precious.** That would be somewhat unique and would make it possible to just have it on a pickaxe instead of carrying ~~two~~ pickaxes (or even more, if you are going to main something specific with `good-at`).**

Again, I think composability is hurting us here, generic descriptions are problematic as they are soulless, context independent and often they almost fit but not really.

> Also another thought I had: We have two independent actions here, consuming block drops and healing the player. I think these should be separate modifiers, so addon creators can compose them (to e.g. produce a cactus modifier that damages you, but does not eat your drops)

Firstly, how do I name them?

- changes health when breaks a block
- doesn't drop items when breaks a block

Secondly, I think the current identity of the modifier is important part of the UX of it. 
It sounds cool as a single thing, it eats items to heal you. 
If you make it generic, the connection disappears and there is no longer an explanation why it heals you and why it doesn't drop blocks, it just does those two things. From vampire gem it becomes does-a-lot-of-things-gem.
More generic, less cool and bloats tooltip. I think a big part of the modifiers it that they present well. 

@careeoki what do you think about UX here.

Similarly I would rather have `spiky` modifier for the cactus than some generic damaging modifier.

Argmaster:
Honestly healing isn't really something you can exploit if there is proper healing in game. Players will be running around at 100% most of the time and no one will run mining blocks during a fight to heal themselves (if they do, that means rubies need nerfs asap). Healing should be cheap and abundant (by non-pickaxe means), unless you are making a realistic survival game, so that half a heart every few blocks broken shouldn't be a game changer. It should mean that you can naturally heal what you loose in caves by falling few too many blocks every now and then. 

This is exactly why I would rather have the healing come from more valuable resources, so you actually have to sacrifice something, not just hit few stone blocks. 
At the stage of the game when you get rubies you will have plenty of resources to make a pickaxe and carry it around, you will have plenty of means to heal and only thing you won't have is the patience to break blocks with low damage pickaxe. Or really break blocks in general, unless there is a good reason to do so.

careeoki:
> @carrie what do you think about UX here

Yeah, it would look weird to have two special modifiers on one gem. We can say "eats blocks to heal you" in one sentence, rather than "destroys item drops" and "heals you when breaking blocks" in two modifiers 
The first one is more intuitive, and the connection between the two actions is obvious.

I do get why we'd want it generic though, I can imagine other modifiers wanting to eat blocks or damage the player.
Maybe we could have some sort of system to compose modifiers out of seperate components in a `.zon`, like `devouring` would be:
```
.tooltip = "Consume block drops to heal player for {strength}% of block health and resistance"
.onBreak = "cubyz:destroy_item_drops", "cubyz:change_player_health"
```
Or whatever it would look like, idk, you get the idea. But then we can reuse these smaller pieces while still being able to make modifiers specialized to a certain material.

> Similarly I would rather have `spiky` modifier for the cactus than some generic damaging modifier.

Yeah, I agree.

careeoki:
Trying this out ingame, I think the effect could be weaker. I can mine 40 limestone or so to heal back all my health. On the surface, I feel bad for wrecking the natural terrain and wasting the blocks, but in a cave I wouldn't care. 
By the time you reach ruby you already have a ton of extra stone fodder blocks, so it doesn't really matter if you sacrifice some for your health.

I think it would be cooler if the modifier made you go out of your way to find higher-health blocks to heal (which often would be ores)
"Do I mine this iron, or should I sacrifice it to heal myself?"

IntegratedQuantum:
I also agree with what carrie says, just mining a bunch of stone seems kind of easy, and if you have a good_at modifier for that stone it's also going to be quite quick.
Maybe we should limit healing to blocks with a newly introduced `ore` tag? That would also give an interesting use-case for all the (at the point you have ruby) useless ores like copper and sulfur.

With that it would also make more sense to separate the modifiers `devours all block drops` and `heals you when mining ore blocks`

Argmaster:
> devours all block drops

Why all if it only activates on ore?

> heals you when mining ore blocks

That is a very specific modifier. Do I literally call the file `heals_when_mining_ore.zig`? That's a little bit ridiculous. 

Hot take: If you want modifiers that are interesting, they cannot be generic.

Argmaster:
Anyway, I think this modifier is too controversial to keep it in one PR with `onBlockUpdate` callback addition, its blocking it for no reason, I will separate them.

PR #2837, opened 2026-04-06
Fix utf8 encoding bug in the authentication

HansAschenloher:
The game crashes, when entering UTF8 chars in the Account Code field.
This is because the error text threats all chars as as one byte.
I replaced the byte iterator with an UTF8 iterator and a modified error message.

PS: This is basically my hello world in zig, so check extra carefully

### DIFF
--- a/src/network/authentication.zig
+++ b/src/network/authentication.zig
@@ -169,15 +169,14 @@ pub const AccountCode = struct {
 		defer std.crypto.secureZero(u8, result.items);
 
 		const trimmed = std.mem.trim(u8, text, &std.ascii.whitespace);
-		var unicodeIterator =
-			std.unicode.Utf8Iterator{.bytes = trimmed, .i = 0};
+		var unicodeIterator = std.unicode.Utf8View.initUnchecked(trimmed).iterator();
 
-		while (unicodeIterator.nextCodepoint()) |cp| {
-			if (cp > 0x7F) {
-				printInvalidCharError(failureText, cp);
+		while (unicodeIterator.nextCodepoint()) |codepoint| {
+			if (codepoint > 0x7F) {
+				printInvalidCharError(failureText, codepoint);
 				continue;
 			}
-			const char: u8 = @intCast(cp);
+			const char: u8 = @intCast(codepoint);
 			if (std.ascii.isAlphabetic(char)) {
 				result.appendAssumeCapacity(std.ascii.toLower(char));
 			} else if (std.ascii.isWhitespace(char)) {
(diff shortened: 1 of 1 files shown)


src/network/authentication.zig
```diff
@@ -159,22 +159,33 @@ pub const PublicKey = union(KeyTypeEnum) {
...
 		var result: main.ListUnmanaged(u8) = .initCapacity(main.stackAllocator, text.len);
 		defer result.deinit(main.stackAllocator);
 		defer std.crypto.secureZero(u8, result.items);
 
 		const trimmed = std.mem.trim(u8, text, &std.ascii.whitespace);
+		var unicodeIterator =
+			std.unicode.Utf8Iterator{.bytes = trimmed, .i = 0};
 
-		for (trimmed) |char| {
+		while (unicodeIterator.nextCodepoint()) |cp| {
```
IntegratedQuantum: There is no reason to abbreviate it here
```suggestion
		while (unicodeIterator.nextCodepoint()) |codepoint| {
```

src/network/authentication.zig
```diff
@@ -159,22 +159,33 @@ pub const PublicKey = union(KeyTypeEnum) {
...
+	}
+
 	pub fn initFromUserInput(text: []const u8, failureText: *main.List(u8)) AccountCode {
 		var result: main.ListUnmanaged(u8) = .initCapacity(main.stackAllocator, text.len);
 		defer result.deinit(main.stackAllocator);
 		defer std.crypto.secureZero(u8, result.items);
 
 		const trimmed = std.mem.trim(u8, text, &std.ascii.whitespace);
+		var unicodeIterator =
+			std.unicode.Utf8Iterator{.bytes = trimmed, .i = 0};
```
IntegratedQuantum: I don't like broken lines, especially not when the line is so short anyways.

Also I think the intended way is to use an Utf8View to init this:
```suggestion
		var unicodeIterator = std.unicode.Utf8View.initUnchecked(trimmed).iterator();
```

IntegratedQuantum (changes requested):

src/network/authentication.zig
HansAschenloher: Thank you for the quick review.
Do you happen to know where the Utf8View vs Utf8Iterator thing is documented?
Cause I just started to learn zig, and oriented me at some other code in this repo (e.g. src/graphics.zig:708).
The [zig std lib documentation](https://ziglang.org/documentation/master/std/#std.unicode) doesn't say much about it.

Wunka: The iterator iterates over a Utf8View. So doing it over the Utf8View while not really changing things is just cleaner. 
std: https://ziglang.org/documentation/master/std/#std.unicode.Utf8View.iterator

IntegratedQuantum: For me it's mainly the fact that Zig would provide a default value for the index if it's meant to be used like this. So I looked to find other ways you can initialize this iterator, and the only thing I found was `Utf8View`

IntegratedQuantum (approved):

PR #2839, opened 2026-04-06
Allow cave layers to specify the density of caves within. This allows adding caves to sky islands

IntegratedQuantum:
The density here is in number of cave spawns per chunk, default (and the previous value for the underground) is 1.0/32.0.

I'm not 100% sure about these parameters, I kind of went with what felt right.

### DIFF
--- a/assets/cubyz/biomes/cave/void/void_roots.zig.zon
+++ b/assets/cubyz/biomes/cave/void/void_roots.zig.zon
@@ -1,6 +1,6 @@
 .{
 	.isCave = true,
-	.tags = .{.root_layer},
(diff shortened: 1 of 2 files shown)


IntegratedQuantum:
I also just updated the root transition as well.

ikabod-kee:
I will check out the root

ikabod-kee:
It seems fine, besides the rare biomes such as void caverns, void crystals, and void phantasmal spawning at the very start of the Root. Sort of loses some of its mystique that way.

IntegratedQuantum:
I could make the transition layer only contain the base root biomes.

ikabod-kee:
That'd be good!

PR #2843, opened 2026-04-06
Add .glb versions of entity models

### LINKED ISSUE #2830: Convert .obj to .glb

bjeff123:
Converts the 5 .obj files in assets/entityModels/models to .glb format using obj2gltf. Original .obj files kept for legacy.

Closes #2830


bjeff123:
I noticed PR #2762 also fixes #2830 and mentions it currently requires blockbench coordinate system. I converted these using obj2gltf directly from the original .obj files — not sure if the coordinate system matches. Happy to re-export if needed or close this if it's covered by #2762.

RanPix:
in my PR i made the models be in correct coordinate system except for the snale, and i also removed the models in there and im not sure if i should keep them since in the issue till said to leave them but we are not loading them anymore in my PR, so i would just wait for a bit for till to return and then sort this out

tillpp:
> in my PR i made the models be in correct coordinate system except for the snale, and i also removed the models in there and im not sure if i should keep them since in the issue till said to leave them but we are not loading them anymore in my PR, so i would just wait for a bit for till to return and then sort this out

I assume it will take time before your PR is merged, which is why i said to keep the .obj as legacy,
but if you have the asset .glb change in the same PR as the code .glb change, it isnt necessary anymore to keep the .obj

@bjeff123 Thanks for your effort, sorry it was for nothing.

bjeff123:
no worries @tillpp - I'm in the process of learning how to contribute, so it wasn't all for nothing. Really stoked about this project (that I came across just a few days ago).

PR #2850, opened 2026-04-07
[AAA] rename Tool to ProceduralItem

Wunka:
This PR is supposed to start the AAA series (Armor and Accessories #604) 

Its goal is to make a few names more generic so things don't get confusing.


src/game.zig
```diff
@@ -688,14 +688,14 @@ pub const World = struct { // MARK: World
 		errdefer self.biomePalette.deinit();
 		self.itemPalette = try assets.Palette.init(main.globalAllocator, zon.getChild("itemPalette"), null);
 		errdefer self.itemPalette.deinit();
-		self.toolPalette = try assets.Palette.init(main.globalAllocator, zon.getChild("toolPalette"), null);
-		errdefer self.toolPalette.deinit();
+		self.proceduralItemPalette = try assets.Palette.init(main.globalAllocator, zon.getChild("proceduralItemPalette"), null);
```
Argmaster: This requires a world version bump (we have a world version field right?) and backwards support + migration for old saves.

src/server/world.zig
```diff
@@ -489,8 +489,8 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		self.itemPalette = try loadPalette(arena, path, "item_palette", null);
 		errdefer self.itemPalette.deinit();
 
-		self.toolPalette = try loadPalette(arena, path, "tool_palette", null);
-		errdefer self.toolPalette.deinit();
+		self.proceduralItemPalette = try loadPalette(arena, path, "procedural_item_palette", null);
```
Argmaster: This requires backwards compatibility.

Argmaster (changes requested):

Wunka:
As discussed on discord. This PR for now only changes in code names. No save file migrations or anything like that is done

A follow up issue with PR will then handle the file migrations like renaming the folders

src/game.zig
Wunka: This is loaded from the world zon which the server sends which I also changed. Should I revert that too?

Argmaster: Uh if it's just server-client communication then no, I might have gone overboard here

Argmaster: I may need to review this on PC because I am not sure. If this is taken from a file, even if the server is an intermediary here, it should remain an old name

Argmaster: Ok, after reading the implementation I come to a conclusion that this is taken directly from save directory and thus should remain with the old name. See `utils.Compression.pack` and implementation of `network.protocols.handShake.serverReceive`

Wunka: reverted

IntegratedQuantum (approved):

PR #2864, opened 2026-04-10
ECS Part 10 | Entity Manager (server side)

tillpp:
add the server side entitySystem

Requires: #2889


tillpp:
There is something i can not test by myself.
I need 3 players to do the following:

```
the host hosts.
player2 joins
player3 joins.

player 2 leaves.
can player3 still walk/fly/do stuff afterwards, without crashing or getting stuck. (also from the other players view)
```

Thanks for anyone willing to test this.

tillpp:
ok tested with 3 instances and test worked

src/server/EntityManager.zig
```diff
@@ -0,0 +1,90 @@
...
+const BinaryReader = main.utils.BinaryReader;
+const BinaryWriter = main.utils.BinaryWriter;
+const Blueprint = main.blueprint.Blueprint;
+const Mask = main.blueprint.Mask;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const CircularBufferQueue = main.utils.CircularBufferQueue;
+const sync = main.sync;
+const server = main.server;
+
+pub var entities: main.utils.SparseSet(server.Entity, main.entity.Entity) = undefined;
```
tillpp: notice to my future self: use virtual list, since quantums interpolation variable implicitly assumes that the memory address never changes.

tillpp (commented):
request change:

src/server/EntityManager.zig
tillpp: done

src/server/Entity.zig
```diff
@@ -18,6 +18,8 @@ maxEnergy: f32 = 8,
 name: ?[]const u8 = null,
 id: u32 = 0,
 
+interpolation: main.utils.GenericInterpolation(3) = undefined,
```
IntegratedQuantum: Only players need this, and it honestly may not even make sense anymore, given that the future trajectory is for the server to simulate the client physics as well.

I'm also honestly not sure if it even did anything other than add more latency to the system. I think it would only even be noticable for low-fps players.

tillpp: are you saying to just remove interpolation?

IntegratedQuantum: Yes, but please check you find any noticable problems, particularly when one of the clients has lower fps (let's say maybe 30)

tillpp: Done.
removing interpolation can be done in a different PR.

src/server/entity_manager.zig
IntegratedQuantum, line 1: This is a namespace, so it should be called `entity_manager`

src/server/entity_manager.zig
```diff
@@ -0,0 +1,111 @@
...
+	const ent = &entities.items()[index];
+
+	// remove id
+	idMapping.items[@intFromEnum(entityId)] = null;
+
+	// remove entity
+	{
+		std.debug.assert(ent.id == entityId);
+		ent.deinit(.server);
+		_ = entities.swapRemove(index);
```
IntegratedQuantum: Not sure if you make use of that anymore, but this breaks the stability guarantee of the VirtualList.

src/server/entity_manager.zig
```diff
@@ -0,0 +1,111 @@
...
+	if (freedList.items.len > 0) {
+		entityId = freedList.swapRemove(0);
+	} else {
+		entityId = @enumFromInt(freeId);
+		freeId += 1;
+	}
+
+	// get a new memory address
+	const index: u32 = entities.len;
+	var ent = entities.addOne();
```
IntegratedQuantum: Why not just reuse the entity id as the index?
You should never need to iterate through all entities (well until you store them in chunks you do, but until that is set up, adding tombstones would be reasonable).

src/server/entity_manager.zig
```diff
@@ -0,0 +1,111 @@
...
+		if (index != entities.len) {
+			idMapping.items[@intFromEnum(entities.items()[index].id)] = index;
+		}
+		freedList.addOne(main.globalAllocator).* = entityId;
+	}
+}
+
+pub fn getEntitiesNearbyInfo(allocator: main.heap.NeverFailingAllocator) main.ZonElement {
+	const zonArray = main.ZonElement.initArray(allocator);
+	for (entities.items()) |*entity| {
```
IntegratedQuantum, line 95: This is inefficient. Instead entities should be stored in the chunks, so when you want to get entities nearby you just iterate through the chunks in the players simulation distance. This would also make it easier to discover when new entities enter/leave the range, you'd only need to look at it entities that moved across chunks or entities from chunks that come into range.

IntegratedQuantum (changes requested):

Argmaster:
So, a rule of a thumb, a manager is fits-all name. Manager manages something and you must agree that all code manage something, so its repeating the obvious. From brief look at the code, I came to conclusion that this is just public/global API and state of ECS so it should be called this way; ecs.zig. No managers required, lets eliminate excessive management costs in our codebase :))))

tillpp:
> So, a rule of a thumb, a manager is fits-all name. Manager manages something and you must agree that all code manage something, so its repeating the obvious. From brief look at the code, I came to conclusion that this is just public/global API and state of ECS so it should be called this way; ecs.zig. No managers required, lets eliminate excessive management costs in our codebase :))))

this is the wrong place to argue about this. I am following a already existing naming convention in the codebase, if you want a different name create an Issue  :))))

(server entity manager is the server version of client entity Manager)

Argmaster:
> this is the wrong place to argue about this. 

Fine, let's go [discord](https://discord.com/channels/443805812390100992/702167480654757991/1516702126192197725).

tillpp:
> > this is the wrong place to argue about this.
> 
> Fine, let's go [discord](https://discord.com/channels/443805812390100992/702167480654757991/1516702126192197725).

try reading again:

> if you want a different name create an **Issue** :))))

src/server/entity_manager.zig
tillpp: storing entities into chunks is planned for a future PR.
This PR is just a cleanup of existing code, not trying to add new features.

IntegratedQuantum:
I did make one now: #3249

src/server/entity_manager.zig
IntegratedQuantum: No I don't. I must have been tired when I said that. I was probably arguing against adding a "renderComponent".

Since most entities won't have one, the name should be moved into its own component as mentioned in #3241.

tillpp: > No I don't. I must have been tired when I said that. I was probably arguing against adding a "renderComponent".

it was here on github as a Review.
But that doesn't matter anyway since this is the server entities and not the client entities. got them mixed up, so deleted the comment

src/server/entity_manager.zig
tillpp: (yeah i still use the stability guarantee of virtuallist.)

done.

src/server/server.zig
```diff
@@ -662,13 +664,14 @@ fn update() void { // MARK: update()
 	var entityData: main.ListManaged(main.entity.EntityNetworkData) = .init(main.stackAllocator);
 	defer entityData.deinit();
 
-	for (userList) |user| {
-		const id = user.id; // TODO
+	for (entity_manager.getAll()) |*ent| {
+		if (!ent.used) continue;
+		const id = ent.id; // TODO (why is this todo here?)
```
IntegratedQuantum: You just solved it, congratulations, I think this belonged to the TODO on freeId.
```suggestion
		const id = ent.id;
```

src/server/server.zig
```diff
@@ -103,7 +104,7 @@ pub const User = struct { // MARK: User
 	conn: *Connection = undefined,
 	innerPlayer: Entity = .{},
 	timeDifference: utils.TimeDifference = .{},
-	interpolation: utils.GenericInterpolation(3) = undefined,
+	interpolation: main.utils.GenericInterpolation(3) = undefined,
```
IntegratedQuantum: Unintended change
```suggestion
	interpolation: utils.GenericInterpolation(3) = undefined,
```

src/server/server.zig
```diff
@@ -604,6 +605,7 @@ fn deinit() void {
 	main.sync.server.deinit();
 	main.items.Inventory.server.deinit();
 	main.entity.server.deinit();
+	entity_manager.deinit();
```
IntegratedQuantum: Since #3136 the components assume that all entities are deinited before entity.server.deinit clears the components.

src/server/server.zig
```diff
@@ -103,7 +104,7 @@ pub const User = struct { // MARK: User
 	conn: *Connection = undefined,
 	innerPlayer: Entity = .{},
```
IntegratedQuantum: If I understand correctly this is no longer interacted with, since you changed the `player` method. Please remove it.

src/server/entity_manager.zig
```diff
@@ -0,0 +1,99 @@
...
+	for (entities.items()) |*ent| {
+		if (ent.used) {
+			ent.deinit(.server);
+		}
+	}
+	entities.deinit();
+	freedList.deinit(main.globalAllocator);
+}
+
+pub fn getAll() []server.Entity {
```
IntegratedQuantum: So no one is tempted to use this, could you also add a helper function for its only use-case?

src/server/entity_manager.zig
```diff
@@ -0,0 +1,99 @@
...
+	return entities.items();
+}
+
+pub fn addEntity() main.entity.Entity {
```
IntegratedQuantum, line 43: Functions like this are not thread safe.
Please add assertions that we are on the server thread to all functions that create/destroy or query entities.

IntegratedQuantum (changes requested):

src/server/server.zig
tillpp: good catch

src/server/entity_manager.zig
tillpp: you mean with "sync.threadContext.assertCorrectContext(.server);" ?
Done.

src/server/entity_manager.zig
```diff
@@ -0,0 +1,125 @@
...
+const BinaryWriter = main.utils.BinaryWriter;
+const Blueprint = main.blueprint.Blueprint;
+const Mask = main.blueprint.Mask;
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const CircularBufferQueue = main.utils.CircularBufferQueue;
+const sync = main.sync;
+const server = main.server;
+
+var entities: main.utils.VirtualList(server.Entity, 1 << 24) = undefined;
+var freedList: main.List(main.entity.Entity) = undefined;
```
IntegratedQuantum: I keep stumbling over this, it feels so much like a typo.
I think it should either be `freeList` or `freeIdList`.

src/server/server.zig
```diff
@@ -555,9 +554,10 @@ fn init(name: []const u8, singlePlayerPort: ?u16) void { // MARK: init()
 		@panic("Could not open Server.");
 	}; // TODO Configure the second argument in the server settings.
 
+	main.sync.server.init();
+	entity_manager.init();
 	main.entity.server.init();
 	main.items.Inventory.server.init();
-	main.sync.server.init();
```
IntegratedQuantum, line 560: Could you explain why you had so move the sync system init/deinit?

src/server/server.zig
```diff
@@ -555,9 +554,10 @@ fn init(name: []const u8, singlePlayerPort: ?u16) void { // MARK: init()
 		@panic("Could not open Server.");
 	}; // TODO Configure the second argument in the server settings.
 
+	main.sync.server.init();
+	entity_manager.init();
```
IntegratedQuantum: Inits should happen in the opposite order as deinits, so this hsould be after Inventory.server.init.

IntegratedQuantum (changes requested):

src/server/server.zig
tillpp: since enitity_manager needs to know what thread it is in. I think its best to have it the first thing to be inited and the last thing to be deinited from this block

src/server/entity_manager.zig
tillpp: yeah i too think it's a typo

src/server/server.zig
tillpp: oh just realized that the connectionManager thread also accesses the server entity_manager (PlayerPosition protocol)

so i think it might be better to use a mutex instead of the assert

IntegratedQuantum: That's a side-effect of not properly syncing the player position through the sync system.

IntegratedQuantum: Since you probably won't add that in this PR, I'd suggest to keep a pointer to the player entity in `User` until this is resolved properly.

Just adding a mutex is also fragile, since it doesn't protect the actual access of the Entity data.

Issue #2802, opened 2026-04-02
allowCheats option should be replaced with a default permission group


IntegratedQuantum:
This would allow elevating permissions of the server owner beyond that if needed.

Wunka:
Wow, that seems like the perfect solution. @tillpp and I talked about maybe doing two default permission Groups or staying with one and saying that no cheats meant no cheats. 

But as allowCheats is not really used for anything else. I agree that the permisison system should just replace allowCheats

PR #2865, opened 2026-04-10
ECS Part 11 | Avatar Command

tillpp:
Part of the big https://github.com/PixelGuys/Cubyz/pull/2652 ECS PR

THIS IS HOW EVERYTHING STARTED! with the /avatar command!

type : ``/avatar`` to see what you are
type : ``/avatar cubyz:modelId`` to switch modelId


tillpp:
Ready for review now

src/server/command/entity/avatar.zig
```diff
@@ -0,0 +1,43 @@
...
+		} else {
+			source.sendMessage("#ff0000EntityModelId {s} doesnt exist", .{entityModelId});
+		}
+		// transmit
+		if (model.server.get(source.id)) |rc| {
+			var binaryWriter = main.utils.BinaryWriter.init(main.stackAllocator);
+			defer binaryWriter.deinit();
+			if (rc.save(&binaryWriter, .playerNearby) == .save) {
+				for (main.server.connectionManager.connections.items) |conn| {
+					main.network.protocols.EntityComponentUpdate.load(conn, source.id, model.entityComponentID, model.entityComponentVersion, binaryWriter.data.items);
```
IntegratedQuantum: In my opinion the user of components should not care about when and with whom to sync changes, it should be handled by the component.

Do you already have a plan on how to sync ECS changes with the client? Here (as long as the avatar has no physical impact) it's fine, but as soon as it comes to e.g. effects that impact health or your inventory, then it must be synced using the sync system.

src/server/command/entity/avatar.zig
```diff
@@ -0,0 +1,43 @@
+const std = @import("std");
+
+const main = @import("main");
+const User = main.server.User;
+
+pub const description = "Change your avatar";
+pub const usage = "/avatar <entityTypeID>";
```
IntegratedQuantum: Please document both usages

IntegratedQuantum (changes requested):
Please use the new argument parser (see #3073)

src/server/command/entity/avatar.zig
Argmaster: +1
ECS should have a layer of abstraction over sync.
Ideally the flow would be
`get entity component -> write` With minimal number of arguments (2: entity id + component id; 1: component value)
From that point ECS internally figures out everything else, including how to serialize, cache, when to sync.
In general ECS should be black box beyond getting and setting components.

tillpp: yeah i tried to follow the existing code style, i.e. doing networking explicitly for everything
I thought you guys might have disliked abstraction.

but if you guys have nothing against it, i guess we can add it

tillpp: > In my opinion the user of components should not care about when and with whom to sync changes, it should be handled by the component.

There is the problem: EntityComponents are set during initPlayer(), but initPlayer() is done before sendServerPlayerData(), which means the client receives componentData before the palettes are loaded! the result: a crash.

there are multiple possible solution to this.
For example: 
1. having a "notify" parameter in the entityComponent functions that set the value, to prevent sending any component data before palettes are received.

I dislike this solution, because it adds noise for component implementers and adds a lot of if() conditions.

2. having a "transmitChange" function.
This way you explicity state that the change should be transmitted.
Downside is, that people might forget to write transmitChange() after having changed a value.

3. more complex solutions that would require to radically change how cubyz does it's handshake by splitting palettes and playerdata. And sending **only** the palettes first,  and then the playerdata.  
**Out of this scope for this small PR.** (currently palettes and player information are send mixed together) 


Edit: i have implemented 2. , if you want to know how it looks like in code ,you can checkout the current PR

src/entityComponent/model.zig
```diff
@@ -77,16 +78,9 @@ pub const server = struct {
 			return error.InvalidComponentVersion;
 		const entityModel = reader.readVarInt(u32) catch return error.UnreadableComponentData;
 
-		try loadByIndex(entity, .{.index = entityModel});
-	}
-	pub fn loadByID(entity: u32, entityModelID: []const u8) main.entity.EntityComponentLoadError!void {
```
tillpp, line 82: removed never used functions

src/server/command/entity/avatar.zig
Argmaster: p. 3. makes most sense to me when you describe it like this.
Especially since p. 2. poisons every use of component editing for sake of single edge case, one that is hardcoded into the engine.

IntegratedQuantum: > There is the problem: EntityComponents are set during initPlayer(), but initPlayer() is done before sendServerPlayerData(), which means the client receives componentData before the palettes are loaded!

But initPlayer doesn't send anything to the client, it only initializes the server side entity. How can this cause issues?

tillpp: > But initPlayer doesn't send anything to the client,

my message is a response to  `"In my opinion the user of components should not care about when and with whom to sync changes, it should be handled by the component."`. 
if we do this then initPlayer would send data to the client.

IntegratedQuantum: Ah makes sense. Though in my opinion it should only send data to the clients that are already fully connected. At this points `users.append(user);` has not happened, so it wouldn't send it to the new client.
Instead the data of the player itself would be sent in `zonObject.put("player", conn.user.?.player().save(main.stackAllocator, .playerHimself));` which gets unpacked after the palettes are loaded.

tillpp: > Ah makes sense. Though in my opinion it should only send data to the clients that are already fully connected. At this points users.append(user); has not happened, so it wouldn't send it to the new client.

Well, before i wrote my 3 points, i did already try to implement the "network is hidden behind ECS" things, and before the Palettes get received by the new clients, the component get received by the new client.

IntegratedQuantum: I don't understand how that could happen. Where does that load operation come from? If it's in initPlayer, then how would it know the `User`, when the server hasn't even added it to the list yet?

tillpp: oh i see, connectionManager.connections
grafik
are all users, even the "unconnected" once.
an implicit assumption that wasnt in my memory yet

tillpp: ok i used the original idea.
With this ECS11 is ready for rereview now.
using Argmaster new Argparser and abstracting networking into the EntityComponent.
making the transmission use the sync system can be done in another PR: https://github.com/PixelGuys/Cubyz/issues/3140

src/entityComponent/model.zig
```diff
@@ -77,25 +78,19 @@ pub const server = struct {
...
+			.entityModel = .{.index = entityModel},
+		});
 	}
 	pub fn unload(entity: u32) void {
 		components.remove(@enumFromInt(entity)) catch {};
 	}
 	pub fn put(entity: u32, renderComponent: Component) void {
 		const ptr = components.get(@enumFromInt(entity)) orelse components.add(main.globalAllocator, @enumFromInt(entity));
 		ptr.* = renderComponent;
+		main.entity.server.transmitChange(root, entity);
```
IntegratedQuantum: I would call this `registerChange`, instant transmission is not something we should keep long term.

I would also prefer to pass `@This()` directly. Aliasing it as `root` makes it unclear where it comes from/what it is in my opinion.

src/server/command/_list.zig
```diff
@@ -25,3 +25,5 @@ pub const set = @import("worldedit/set.zig");
 pub const mask = @import("worldedit/mask.zig");
 pub const replace = @import("worldedit/replace.zig");
 pub const toggledecay = @import("worldedit/toggledecay.zig");
+
+pub const avatar = @import("entity/avatar.zig");
```
IntegratedQuantum: Please keep it alphabetically sorted, entity comes before permission

src/server/server.zig
```diff
@@ -526,7 +527,7 @@ const updateTime: std.Io.Duration = .fromNanoseconds(1000000000/20);
 
 pub var world: ?*ServerWorld = null;
 var userMutex: main.utils.Mutex = .{};
-var users: main.ListManaged(*User) = undefined;
+pub var users: main.ListManaged(*User) = undefined;
```
IntegratedQuantum: Please use corresponding getter function `getUserListAndIncreaseRefCount`

src/server/server.zig
```diff
@@ -810,6 +812,7 @@ pub fn connectInternal(user: *User) void {
 	main.network.protocols.entity.send(user.conn, initialList);
 	main.stackAllocator.free(initialList);
 	sendMessage("{s}§#ffff00 joined", .{user.name});
+	user.permissions.addPermission(.white, "/command/avatar");
```
IntegratedQuantum, line 815: This doesn't work properly due to #2802, for now I'd prefer to remove it since it should be handled through permission groups instead of adding them in a random location.

IntegratedQuantum (changes requested):

src/server/server.zig
tillpp: @Wunka  told me to add this! until he is finished with his permission system stuff

src/entityComponent/model.zig
tillpp: i tried @This()  first, but didnt work since it reference to the class "server" instead of the file.

src/server/server.zig
Wunka: Its a solution until the default permission group is there. (I am actually waiting on a review for permission groups, soooo thats why this waiting. Or atleast I think I did everthing?)

src/server/command/_list.zig
tillpp: meanwhile:
grafik

Wunka: we need a linter rule for that I think. All `_list.zig` should be sorted (sorted inside groups. and groups are split through `\n\n`

src/entityComponent/model.zig
IntegratedQuantum: Then call it `Self` instead of `root`, which is a more common convention for `@This()`.

tillpp: self is a common convention for This()
but in this case it is the root of the file. 

but i can change it anyway

src/server/server.zig
tillpp: i consider this resolved then.
did make an issue to remove it when wunka is done with his permission system: https://github.com/PixelGuys/Cubyz/issues/3143

IntegratedQuantum (approved):

PR #2872, opened 2026-04-11
Round tool stats display to 1 decimal

### LINKED ISSUE #2696: Apply rounding to tool damage numbers (?)
### Description

Is there a reason why they shouldn't be integers? I mean, it'd be much friendlier to casual players if they were, and people would be less likely to min-max their tools to get "the perfect build" with what materials they have. You can still show the pure values in the interface, but I feel in the tool itself, it should be rounded.

GasSpace630:
This PR replaces #2871 which fixes #2696 
The previous PR contained unrelated changes, so it was closed.
This version only includes the formatting changes.

### DIFF
--- a/src/items.zig
+++ b/src/items.zig
@@ -844,8 +844,8 @@ pub const Tool = struct { // MARK: Tool
 		self.tooltip.clearRetainingCapacity();
 		self.tooltip.print(
 			\\{s}
-			\\{d:.2} swings/s
-			\\Damage: {d:.2}
+			\\{d:.1} swings/s
+			\\Damage: {d:.1}
 			\\Durability: {}/{}
 		, .{
 			self.type.id(),


src/items.zig
```diff
@@ -496,6 +496,9 @@ const ToolPhysics = struct { // MARK: ToolPhysics
 			tool.maxDurability = 0;
 			tool.durability = 1;
 		}
+		tool.damage = @round(tool.damage*10)/10;
+		tool.swingSpeed = @round(tool.swingSpeed*10)/10;
+		tool.maxDurability = @round(tool.maxDurability);
```
IntegratedQuantum: it is rounded 10 lines above

IntegratedQuantum (changes requested):
Also please rebase, there was a refactor that affects this.

src/items.zig
```diff
@@ -481,8 +481,8 @@ const ProceduralItemPhysics = struct { // MARK: ProceduralItemPhysics
 			sum *= property.resultScale;
 			proceduralItem.getProperty(property.destination orelse continue).* += sum;
 		}
-		if (proceduralItem.damage < 1) proceduralItem.damage = 1/(2 - proceduralItem.damage);
-		if (proceduralItem.swingSpeed < 1) proceduralItem.swingSpeed = 1/(2 - proceduralItem.swingSpeed);
```
IntegratedQuantum: Why did you remove this? This is required to prevent negative values.

IntegratedQuantum (changes requested):

IntegratedQuantum:
also CI failed, please run the formatter

IntegratedQuantum (changes requested):
Now you are only applying rounding to small and negative values, and they remain negative.

src/items.zig
IntegratedQuantum: I'm asking you again: Why did you remove the old checks?

IntegratedQuantum (changes requested):

src/items.zig
ghost: dont worry i did because it would reduce noise and maybe make the stats cleaner
and about the negative values
at line:509 it just clamps it to 0.1

i can change this if this isnt the desired behavoiur.

IntegratedQuantum: The reason for this code to exist was to allow negative values and let them still impact the stats, even for large negative values. A clamp does not allow this.

ghost: Ohh... i didnt know we wanted -ve values, my bad 😅

ghost: wait so if we have -ve value for damage does that heal the enemy?

IntegratedQuantum: No. The previous code basically mapped all negative values to positives following this function:
Screenshot at 2026-04-23 16-29-14

IntegratedQuantum (changes requested):
Please apply fixes after #2891

IntegratedQuantum:
What you've done now is remove the code that rounds the actual values, instead of fixing it up with #2891.

I am annoyed by how many times I've had to review this PR, so I decided to quickly do this myself in 8b4d843dedf1017269e0749d40ab32fc27b878c9

For the future I suggest to learn how to use `git rebase`, and try to be more careful with how you address conflicts. Furthermore please try to at least look at the changes you have made on github at least once to catch these mistakes yourself, instead of me having to point them out to you.

PR #2874, opened 2026-04-11
Use Vec3i for BlockUpdate and UpdateBlock

Argmaster:
Preparation for some work on block update messages;

Reduces side of block update signatures and calls by using Vec3i instead of 3 separate integers. No negative impact is expected and there is less code to write and read.

Apparantly I also merged part of `chunk_meshing.updateBlock` signature from separate parameters into `mesh_storage.BlockUpdate` for the same reason.

### DIFF
--- a/src/renderer/mesh_storage.zig
+++ b/src/renderer/mesh_storage.zig
@@ -45,21 +45,17 @@ var lastRD: u16 = 0;
 var mutex: std.Thread.Mutex = .{};
 
 pub const BlockUpdate = struct {
-	x: i32,
-	y: i32,
-	z: i32,
+	pos: Vec3i,
 	newBlock: blocks.Block,
 	blockEntityData: []const u8,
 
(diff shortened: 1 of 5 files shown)


src/renderer/mesh_storage.zig
```diff
@@ -45,21 +45,17 @@ var lastRD: u16 = 0;
...
 
 pub const BlockUpdate = struct {
-	x: i32,
-	y: i32,
-	z: i32,
+	pos: Vec3i,
 	newBlock: blocks.Block,
 	blockEntityData: []const u8,
 
 	pub fn init(pos: Vec3i, block: blocks.Block, blockEntityData: []const u8) BlockUpdate {
```
Argmaster, line 52: This also makes it redundant right?

Argmaster: Its used once; I will stop with changes here tho, as this is fairly complete and we can continue incrementally from here.

IntegratedQuantum (approved):

PR #2875, opened 2026-04-11
ECS Patch 1 | name freeing when Entity.clone

tillpp:
When splitting up a big PR, mini mistakes are going to happen.
Because it's hard to see what PR part contains what part of the code.
In this Patch series, i correct them.

This is PR deleted the old name of the Entity when it is set to be equal to another, and copying it's name

### DIFF
--- a/src/server/Entity.zig
+++ b/src/server/Entity.zig
@@ -38,9 +38,7 @@ pub fn loadFrom(self: *@This(), id: u32, zon: ZonElement, comptime side: main.sy
 }
 pub fn clone(self: *@This(), copy: *@This()) void {
 	const originalID = copy.id;
-	if (copy.name) |name| {
-		main.globalAllocator.free(name);
-	}
+	std.debug.assert(copy.name == null);
 	copy.* = self.*;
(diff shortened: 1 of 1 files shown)


src/server/Entity.zig
```diff
@@ -38,6 +38,9 @@ pub fn loadFrom(self: *@This(), id: u32, zon: ZonElement, comptime side: main.sy
 }
 pub fn clone(self: *@This(), copy: *@This()) void {
 	const originalID = copy.id;
+	if (copy.name) |name| {
```
IntegratedQuantum: I'm not sure how it's used, but I don't think we should allow this state in the first place.
You should only clone into an object that empty/unused.

IntegratedQuantum (changes requested):

src/server/Entity.zig
tillpp: https://github.com/tillpp/Cubzio/blob/02fbbea92e841e3684fb62a06881319a0414a4ed/src/server/command/summon.zig

tillpp: used assert instead

IntegratedQuantum (approved):

PR #2881, opened 2026-04-11
Set the value of a slider when pressing on the bar outside the button

matjojo:
This should also be implemted into the discrete slider, but before the code, the style, and the feature are approved it's probably best to keep it only in this one.

### DIFF
--- a/src/gui/components/ContinuousSlider.zig
+++ b/src/gui/components/ContinuousSlider.zig
@@ -111,12 +111,27 @@ pub fn updateHovered(self: *ContinuousSlider, mousePosition: Vec2f) main.callbac
 	return .ignored;
 }
 
+inline fn getBarPos(self: *ContinuousSlider) Vec2f {
+	return Vec2f{1.5*border + self.button.size[0]/2, self.button.pos[1] + self.button.size[1]/2 - border};
+}
+
+inline fn getBarSize(self: *ContinuousSlider) Vec2f {
+	const range: f32 = self.size[0] - 3*border - self.button.size[0];
+	return .{range, 2*border};
+}
+
 pub fn mainButtonPressed(self: *ContinuousSlider, mousePosition: Vec2f) main.callbacks.Result {
-	if (GuiComponent.contains(self.button.pos, self.button.size, mousePosition - self.pos)) {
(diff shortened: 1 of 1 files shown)


src/gui/components/ContinuousSlider.zig
```diff
@@ -111,12 +111,27 @@ pub fn updateHovered(self: *ContinuousSlider, mousePosition: Vec2f) main.callbac
 	return .ignored;
 }
 
+inline fn bar_pos(self: *ContinuousSlider) Vec2f {
```
Argmaster: Function names follow `camelCase` naming convention.
Function names should start with a verb.

matjojo:
Did some more testing after the renames, this does not appear to work on the mouse sensitivity slider in the controls menu.

src/gui/components/ContinuousSlider.zig
```diff
@@ -111,12 +111,27 @@ pub fn updateHovered(self: *ContinuousSlider, mousePosition: Vec2f) main.callbac
...
-		if (self.button.mainButtonPressed(mousePosition - self.pos) == .handled) {
+	const mousePositionRelativeToSelf = mousePosition - self.pos;
+
+	if (GuiComponent.contains(self.button.pos, self.button.size, mousePositionRelativeToSelf)) {
+		if (self.button.mainButtonPressed(mousePositionRelativeToSelf) == .handled) {
 			self.mouseAnchor = mousePosition[0] - self.button.pos[0];
 			return .handled;
 		}
+	} else if (GuiComponent.contains(self.getBarPos(), self.getBarSize(), mousePositionRelativeToSelf)) {
+		self.mouseAnchor = self.button.size[0]/2;
```
IntegratedQuantum: I think this might be the issue
```suggestion
		self.mouseAnchor = self.pos[0] + self.button.size[0]/2;
```

matjojo: Good spot. It now works on every ContinuousSlider in the settings. If you like the impl like this we can either merge now or implement it in the DiscreteSlider as well.

IntegratedQuantum (approved):
> implement it in the DiscreteSlider as well.

Please do so in a separate PR. I prefer PRs small, to avoid conflicts

PR #2883, opened 2026-04-11
make the player spawn point an optional.

### LINKED ISSUE #2873: The player spawn position should not be stored if it's the same as the world spawn
This makes it difficult for servers to change their world spawn.

Wunka:
Closes: #2873 

I am also thinking about addidtions to the spawn command like
/spawn remove
which would remove the player specific spawnpoint again.

### DIFF
--- a/src/server/server.zig
+++ b/src/server/server.zig
@@ -114,7 +114,7 @@ pub const User = struct { // MARK: User
 	lastRenderDistance: u16 = 0,
 	lastPos: Vec3i = @splat(0),
 	gamemode: std.atomic.Value(main.game.Gamemode) = .init(.creative),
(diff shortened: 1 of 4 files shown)


src/server/server.zig
```diff
@@ -494,6 +494,10 @@ pub const User = struct { // MARK: User
 		};
 	}
 
+	pub fn getSpawnPoint(user: *User) Vec3d {
```
Argmaster: Its spawn pos everywhere, make it `getSpawnPos`, hm?

Wunka: renamed

IntegratedQuantum (approved):

PR #2884, opened 2026-04-11
add renderBlock but not ancient

OneAvargeCoder193:
lets you render blocks with model matrices


OneAvargeCoder193:
i also did the thing quanturm asked which was to have it use a function instead of calculating it manually

IntegratedQuantum (changes requested):
This is too big. Could you please split it into smaller steps so it's easier to follow and has less conflict potential? Rendering stuff is something that I would like to put more work into for #102 
I'd suggest:
1 PR that just moves what's in graphics to a separate function in renderer
1 PR that introduces the getLight function
1 PR that adds the modelMatrix to the shaders and stuff.

PR #2890, opened 2026-04-12
PVP | Make killing actually doable - WIP

tillpp:
Sidequest to the ECS series.


tillpp:
I will Pause this, until Wunka's AAA series is done, so i dont get too many merge conflicts while working on this.

Argmaster:
Hey could you also move your other inactive PRs (if there are any) into draft state until we get ECS sorted?

tillpp:
They should be, saplings for example is only seem inactive because lighting updates are too slow, but appearently QD is working on that. And my Eating PR requries accesoruies that wunka is working on right now

PR #2891, opened 2026-04-12
[AAA] Part 2 properties

Wunka:
Instead of saving the properties directly in the struct. we save it inside an array of `f32` this allows easy adding of new values (for example through modding)

you can get the value by using `getProperty(ProceduralItemProperty)`
and you can modify a value by using `setProperty(ProceduralItemProperty, value)` (can only be used in and before the `evaluateProceduralItem` step

### DIFF
--- a/src/gui/windows/workbench.zig
+++ b/src/gui/windows/workbench.zig
@@ -124,9 +124,9 @@ pub fn render() void {
 	const offsetY = 4*ItemSlot.sizeWithBorder;
 	const fontSize = 16;
 
-	main.graphics.draw.print("{s}{} durability", .{if (currentResult.proceduralItem.maxDurability != 0) "#ffffff" else "#ff0000", @as(usize, @intFromFloat(currentResult.proceduralItem.maxDurability))}, offsetX, offsetY, fontSize, .left);
-	main.graphics.draw.print("#ffffff{d:.1} swings/s", .{currentResult.proceduralItem.swingSpeed}, offsetX, offsetY + fontSize, fontSize, .left);
-	main.graphics.draw.print("#ffffff{d:.1} damage", .{currentResult.proceduralItem.damage}, offsetX, offsetY + 2*fontSize, fontSize, .left);
+	main.graphics.draw.print("{s}{} durability", .{if (currentResult.proceduralItem.getProperty(.maxDurability) != 0) "#ffffff" else "#ff0000", @as(usize, @intFromFloat(currentResult.proceduralItem.getProperty(.maxDurability)))}, offsetX, offsetY, fontSize, .left);
+	main.graphics.draw.print("#ffffff{d:.1} swings/s", .{currentResult.proceduralItem.getProperty(.swingSpeed)}, offsetX, offsetY + fontSize, fontSize, .left);
+	main.graphics.draw.print("#ffffff{d:.1} damage", .{currentResult.proceduralItem.getProperty(.damage)}, offsetX, offsetY + 2*fontSize, fontSize, .left);
 }
 
 pub fn onOpen() void {
--- a/src/items.zig
+++ b/src/items.zig
@@ -629,7 +629,7 @@ pub const ProceduralItemType = struct { // MARK: ProceduralItemType
 const ProceduralItemProperty = enum {
 	damage,
 	maxDurability,
-	/// swings per second
+	/// how long it takes before the next swing happens
 	swingSpeed,
 
 	fn fromString(string: []const u8) ?ProceduralItemProperty {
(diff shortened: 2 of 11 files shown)


src/items.zig
```diff
@@ -844,10 +834,12 @@ pub const ProceduralItem = struct { // MARK: ProceduralItem
...
 
-	fn getProperty(self: *ProceduralItem, prop: ProceduralItemProperty) *f32 {
-		switch (prop) {
-			inline else => |field| return &@field(self, @tagName(field)),
-		}
+	pub fn getProperty(self: *ProceduralItem, prop: ProceduralItemProperty) f32 {
+		return self.properties[@intFromEnum(prop)];
+	}
+
+	pub fn getPropertyPtr(self: *ProceduralItem, prop: ProceduralItemProperty) *f32 {
```
Argmaster: Hm, I see.
Well, @IntegratedQuantum what do you think about changing this to `setProperty` so its impossible to create dangling pointers?

src/items.zig
```diff
@@ -629,6 +627,7 @@ pub const ProceduralItemType = struct { // MARK: ProceduralItemType
 const ProceduralItemProperty = enum {
 	damage,
 	maxDurability,
+	/// swings per second
```
Argmaster: Actually that is an inverse of swings per second.

src/items.zig
Wunka: I just used the already existing `getProperty` method and renamed it, because we don't always need the pointer. but yeah I would also be ok with `setProperty`

IntegratedQuantum: I think the bigger risk than dangling pointers, is that people may think it's safe to overwrite this after the tool initialization (which it isn't). As for whether to use getPtr or set, I think we should go with what fits the use case, if we do a lot of *= operations, then getPtr seems to produce nicer looking arithmetic.

Argmaster: Ok, different take, with setProperty you can add locking without altering every call. More future proof.
I still feel like someone will store that pointer somewhere.
I am not convinced by following syntax sugar, zig usually doesn't do that in it's design.
You do you :)))
(I can't resolve this, you have to mark it resolved)

IntegratedQuantum: > add locking

If we start to multithread tool creation, then surely we'll have taken the wrong path somewhere.

> I still feel like someone will store that pointer somewhere.

People could do the same things by storing the `Tool*` as well.

----

Either way, my main concern, attempting to modify these values after tool creation, still stands, so I won't mark it as resolved.

Argmaster: > Either way, my main concern, attempting to modify these values after tool creation, still stands, so I won't mark it as resolved.

So, what do you expect?
`ProceduralItemBuilder`?

IntegratedQuantum: A bool and an assert would be enough.

Wunka: switched to a `setProperty` method which has an assert that `finishedPropertyEvaluation` is `false` (this is set true at the end of `evaluateProceduralItem`

src/items.zig
Wunka: changed text

src/gui/windows/workbench.zig
```diff
@@ -124,9 +124,9 @@ pub fn render() void {
 	const offsetY = 4*ItemSlot.sizeWithBorder;
 	const fontSize = 16;
 
-	main.graphics.draw.print("{s}{} durability", .{if (currentResult.proceduralItem.maxDurability != 0) "#ffffff" else "#ff0000", @as(usize, @intFromFloat(currentResult.proceduralItem.maxDurability))}, offsetX, offsetY, fontSize, .left);
-	main.graphics.draw.print("#ffffff{d:.1} swings/s", .{currentResult.proceduralItem.swingSpeed}, offsetX, offsetY + fontSize, fontSize, .left);
-	main.graphics.draw.print("#ffffff{d:.1} damage", .{currentResult.proceduralItem.damage}, offsetX, offsetY + 2*fontSize, fontSize, .left);
+	main.graphics.draw.print("{s}{} durability", .{if (currentResult.proceduralItem.getProperty(.maxDurability) != 0) "#ffffff" else "#ff0000", @as(usize, @intFromFloat(currentResult.proceduralItem.getProperty(.maxDurability)))}, offsetX, offsetY, fontSize, .left);
+	main.graphics.draw.print("#ffffff{d:.1} swings/s", .{currentResult.proceduralItem.getProperty(.swingSpeed)}, offsetX, offsetY + fontSize, fontSize, .left);
+	main.graphics.draw.print("#ffffff{d:.1} damage", .{currentResult.proceduralItem.getProperty(.damage)}, offsetX, offsetY + 2*fontSize, fontSize, .left);
```
IntegratedQuantum, line 129: Note for the future: This should display all properties that are set

IntegratedQuantum (approved):

PR #2893, opened 2026-04-12
Adds the golbo model to cubyz

Crepestrom:
meant to be a alternate player model to snale
golbo
image


ikabod-kee:
I'll leave this PR to Carrie as she's the one who deals with in-game textures. Personally I find it to be charming, but the texture might need some slight adjustments.

careeoki:
Regardless of any art opinions I think new player models should wait until after ECS allows choosing your player model, and maybe after mnhs' animations PR, so we can set up the models to work properly with bones and the item nodes the first time, rather than having to update them.

I have some strong opinions about this character design and the model. I'm being pretty hard on this because the default character selection is very important, likely one of the first things players will see, and I want to make sure they are all very strong designs. 

- I think all player characters should have clothes of some sort. This makes them look more like "people" and look less like enemies. I think this is one of my main problems with this design, this looks more like a random enemy than a player. Even Cubert, while being mostly naked, has a mantle, and the metal parts of his body can resemble pants.
- This is something that I could change my mind on, but I think we should stick to primarily organic player characters, with Cubert being the odd one out (and cubert is technically a formless spirit, not just a wooden man.) I think a character that _is_ a mushroom rather than just having mushrooms on them might be interesting. Also, our players will be able to eat, so the player characters should be able to eat food somehow.
- The general proportions are similar to Cubert, mostly the torso. I want all of the player characters to have very distinct body shapes. Also, the proportions are awkward looking, the arms and legs are too short.
- The neck is problematic- generally I don't want necks at this scale, and I don't think it looks good here. I can tell the only reason you did it is so the mushroom would fit. Even still, the mushroom is crammed against the back of the head. This would make animating their head very difficult as it would constantly clip into the mushroom.
- I don't find the name Golbo very interesting. Also, when googling "golbo" the first results are for a character from a youtube series. When doing a made-up name, make sure to check if others have done it before.

- I would have to completely redo the texture, but that's not the important part here.

- The face is definitely the strongest part of this design, I do like it.

None of this is to say this design is completely bad- just that in this specific circumstances I have specific ideas of what these player characters should be. I think the concept could work, but it would need significantly more work. I could attempt a redesign of my own at some point, but not right now, I'm actually working on another player character which I should be able to show off soon.

tillpp:
If not as a playermodel, how about hostile mob?

careeoki:
I would prefer if mobs were designed with function before form. We should think of what kind of niche a monster fills in the game and design it around that, rather than working backwards from an existing design.

ZigguratV:
Mycelium is organic. Golbo might be the entire mushroom(s), mycelium and all. Something that at least *resembles* clothes could be a fun addition texture wise. If the neck will cause animation issues maybe that part of the model would need rethinking, I guess that experimentation requires those PRs to look at

PR #2896, opened 2026-04-12
Fix control binding the escape key.

### LINKED ISSUE #2257: Pressing escape during binding process closes the menu yet still binds it

I assume this comes from one of the recent updates, But when pressing escape when it has "..." (waiting for button press) does close the entire bind menu, but it does update the bind, it seems unintentional that the menu closes

HansAschenloher:
When pressing the escape key while in a control binding process, then the control should get bound to escape and the settings window should still be open.
This commit disables closing the settings window while binding a key.

fixes: #2257

### DIFF
--- a/src/graphics/Window.zig
+++ b/src/graphics/Window.zig
@@ -649,6 +649,11 @@ pub fn setNextGamepadListener(listener: ?*const fn (?GamepadAxis, c_int) void) !
 	nextGamepadListener = listener;
 }
 
+pub fn resetNextInputListenters() void {
+	nextGamepadListener = null;
+	nextKeypressListener = null;
+}
+
 fn updateCursor() void {
 	if (grabbed) {
 		c.glfwSetInputMode(window, c.GLFW_CURSOR, c.GLFW_CURSOR_DISABLED);
--- a/src/graphics/Window.zig
+++ b/src/graphics/Window.zig
@@ -80,7 +80,7 @@ pub const Gamepad = struct {
 					if ((newState.buttons[btn] == 0) and (oldState.buttons[btn] != 0)) {
 						nextGamepadListener.?(null, @intCast(btn));
 						nextGamepadListener = null;
-						break;
+						return;
 					}
 				}
 			}
@@ -91,7 +91,7 @@ pub const Gamepad = struct {
 					if (newAxis != 0 and oldAxis == 0) {
 						nextGamepadListener.?(.{.axis = @intCast(axis), .positive = newState.axes[axis] > 0}, -1);
 						nextGamepadListener = null;
-						break;
+						return;
 					}
 				}
 			}
--- a/src/graphics/Window.zig
+++ b/src/graphics/Window.zig
@@ -489,7 +489,6 @@ pub const Key = struct { // MARK: Key
 	}
 
 	fn action(self: *Key, typ: enum { press, release, repeat }, isGrabbed: bool, mods: Modifiers, textKeyPressedInTextField: bool) void {
-		if (hasNextInputListenter()) return;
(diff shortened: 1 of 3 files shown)


IntegratedQuantum (changes requested):
The same issue applies to all other keybinds that can be pressed while in the settings.
So please apply this fix in a central location where it applies to all keybinds.

src/graphics/Window.zig
```diff
@@ -489,6 +489,7 @@ pub const Key = struct { // MARK: Key
 	}
 
 	fn action(self: *Key, typ: enum { press, release, repeat }, isGrabbed: bool, mods: Modifiers, textKeyPressedInTextField: bool) void {
+		if (hasNextInputListenter()) return;
```
IntegratedQuantum: Doing this here has a side effect: If I click on a button to bind a gamepad key, without having a controller installed, I'll get stuck unable to interact with anything. I'd suggest to move the check to functions that handle the listeners.

IntegratedQuantum (changes requested):

src/graphics/Window.zig
HansAschenloher: Sorry, I totally missed this.
While fixing this I discovered another bug.
When clicking a button to bind a gamepad key, without having a controller installed, then it is not possible to bind it to a keyboard key again and you are cannot rebind the key.
I suggest resetting the binding process when toggling from keyboard to gamepad and vise versa.

I included this fix in this pr.
I have no Gamepad so please check if everything works as expected.

Wunka: You should probably also reset the binding process when closing / opening the window. Currently when set to gamepad mode and trying to set a key, but if you then close the window by for example pressing any keyboard key or clicking with the mouse outside the window. The binding process is not reset.

DellieDelta:
Tested the latest commit with a controller, and rebinding keys works. Some keys do close the settings menu though when binding them, these are X and B. And binding the thumbstick scrolls the setting menu. But the same bugs are present on the master version as well :)

IntegratedQuantum (changes requested):
Alright then please also fix the issue mentioned in https://github.com/PixelGuys/Cubyz/pull/2896#discussion_r3143220675

HansAschenloher:
> Tested the latest commit with a controller, and rebinding keys works. Some keys do close the settings menu though when binding them, these are X and B. And binding the thumbstick scrolls the setting menu. But the same bugs are present on the master version as well :)

Thanke you for testing it :)
Ok I looked over the controller part again, and it seems like the key binding is `on-release`. So the normal key actions get triggered before the binning event, that will then abort the update function.
I have no controller to test it, I suggest to create a new issue for the controller part.

If someone want to look into it, here is the [relevant code](https://github.com/PixelGuys/Cubyz/blob/master/src/graphics/Window.zig#L78-L129)

src/gui/windows/controls.zig
```diff
@@ -95,7 +102,7 @@ fn unbindKey(keyPtr: usize) void {
 	needsUpdate = true;
 }
 
-pub fn onOpen() void {
+pub fn initWindow() void {
```
IntegratedQuantum: should be private now that you renamed it

src/gui/windows/controls.zig
```diff
@@ -126,18 +137,28 @@ pub fn onOpen() void {
 	gui.updateWindowPositions();
 }
 
-pub fn onClose() void {
+pub fn deinitWindow() void {
```
IntegratedQuantum: same here

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

Issue #2863, opened 2026-04-10
Add a statistics menu including a Day Counter


GasSpace630:
### Description

It would be pretty rewarding to see how many days you played on a world.
And people could make something like _100 days Cubyz playthrough_ and stuff.

Should not be done in the F3 menu, it is intended for debugging, not for player statistics.

tillpp:
Mabye something more general: Having statistics

Wbjitscool:
yeah the java cubyz version had this in F3 mode I wouldn't mind seeing it back in again

PR #2897, opened 2026-04-12
added day-counter on the F3 debug menu

GasSpace630:
From : #2863

### DIFF
--- a/src/gui/windows/debug.zig
+++ b/src/gui/windows/debug.zig
@@ -54,6 +54,8 @@ pub fn render() void {
 			player.hyperSpeed.load(.monotonic),
 		}, 0, y, 8, .left);
 		y += 8;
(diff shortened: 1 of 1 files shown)


GasSpace630:
Note: controlling the time using time commands will reset the counter

codemob-dev:
I don't think this makes very much sense as something to show in the debug menu, it doesn't have any value for debugging. It should really be implemented as an actual user-facing statistic, perhaps as part of a more general statistics system.

GasSpace630:
i just wanted to see how many in-game days i played on my survival world.

may be a statistic board will be nice to see, like in mc
you can see how many blocks you placed and destroyd, the mobs you killed and stuff

this would begood to have untill somthing like that gets added
maybe after the UI update?

Wbjitscool:
also it should stop when you do /time stop if possible

IntegratedQuantum:
I agree with codemob here, this should not be done in the F3 menu (or any other debug windows) as it has no value for debugging

Wbjitscool:
hmm we could add it as something else then maybe on a clock if that gets added a craftable clock?

Issue #2489, opened 2026-01-13
The great world interface refactor


IntegratedQuantum:
The goal is to create a shared world interface for all kinds of actions that is independent of the side we are on. This could be used to simulate all kinds of block interactions on client and server.

Ideally this would also tie into the Sync system, since we need to be able to undo our local changes.

extracted from #2487

PR #2898, opened 2026-04-12
[AAA] Part 3 Item callbacks

Wunka:
This PR will add ItemCallbacks
For now this will only contain `onLeftClick`

Requires: #2891 

What needs to be done before this PR is ready:

- [x] add the execution of the callback at all relevant places
- [x] move breaking of blocks in a callback
- [x] move special wand mechanic into a callback
- [x] remove testing callback


Wunka:
@Argmaster I will have to look over the code again to remove some things like the temporary zon creation. But I would like your input for the general structure atleast.

src/items.zig
```diff
@@ -652,15 +662,9 @@ pub const ProceduralItem = struct { // MARK: ProceduralItem
 	seed: u32,
 	type: ProceduralItemTypeIndex,
 
-	damage: f32,
+	properties: [@typeInfo(ProceduralItemProperty).@"enum".fields.len]f32 = @splat(0),
 
 	durability: u32,
```
Argmaster, line 672: I think we should consider replacing durability field with `timesUsed` or something alike, basically reversed durability, so when a tool gets balance changes it doesn't show up funny. Would have to be clamped at zero and tool with zero durability would have to break instantly.

ofc out of scope

src/items.zig
```diff
@@ -1046,6 +1075,14 @@ pub const Item = union(ItemType) { // MARK: Item
 			inline else => |item| item.hashCode(),
 		};
 	}
+
+	pub fn onLeftClick(self: Item) ItemUsedCallback {
```
Argmaster, line 1087: I think I would prefer if we followed the pattern we used before where `Item` functions only call into identical functions of `ProceduralItem` and `BaseItem`. This guarantees that when implementations diverge, changes will be limited to those structs.
Its a bit more explicit, but I think its for a good reason.

src/callbacks/callbacks.zig
```diff
@@ -10,12 +10,24 @@ pub const ServerBlockCallback = Callback(struct { block: Block, chunk: *main.chu
 
 pub const BlockTouchCallback = Callback(struct { entity: *main.server.Entity, source: Block, blockPos: Vec3i, deltaTime: f64 }, @import("block/touch/_list.zig"));
 
+pub const ItemUsedCallback = Callback(struct {
+	item: main.items.Item,
+	target: union(enum) {
+		block: struct { block: Block, blockPos: Vec3i },
+		entity: *main.server.Entity,
```
Argmaster: I think these may not be mutually exclusive.
A scythe, or a large axe may do AOE damage, what then? (Hitting multiple blocks or entities and blocks)
Should we even do that inside the callback or outside of it?
Additionally it may be very useful to get the `source` player as a parameter and source `inventory` too.
Although that last thing has no use for now, so that's something to ignore for now.

src/renderer.zig
```diff
@@ -1039,90 +1036,21 @@ pub const MeshSelection = struct { // MARK: MeshSelection
 		}
 	}
 
-	pub fn breakBlock(inventory: main.items.Inventory.ClientInventory, slot: u32, deltaTime: f64) void {
+	pub fn breakBlock(inventory: main.items.Inventory.ClientInventory, slot: u32, _deltaTime: f64) void {
```
Argmaster: I think you should eliminate the chain of jumps before this call.
Currently you are jumping `game.breakBlock` -> `Inventory.breakBlock` -> `renderer.MeshSelection.breakBlock` -> `<onLeftClick callback>` for no reason IMO I think it should just call left click callback directly from `game.breakBlock`.

src/renderer.zig
```diff
@@ -1039,90 +1036,21 @@ pub const MeshSelection = struct { // MARK: MeshSelection
...
+		var deltaTime = _deltaTime;
 		if (selectedBlockPos) |selectedPos| {
```
Argmaster, line 1095: I think breaking animation should be handled inside the callback. You may want to extract it as helper function, but it should be called from inside the callback. There might be tools that don't remove breaking animation.

src/renderer.zig
```diff
@@ -893,15 +893,12 @@ pub const MeshSelection = struct { // MARK: MeshSelection
...
-	var currentBlockProgress: f32 = 0;
-	var currentSwingProgress: f32 = 0;
-	var currentSwingTime: f32 = 0;
+	pub var lastSelectedBlockPos: Vec3i = undefined;
 	var selectionMin: Vec3f = undefined;
 	var selectionMax: Vec3f = undefined;
 	var selectionFace: chunk.Neighbor = undefined;
 	var lastPos: Vec3d = undefined;
-	var lastDir: Vec3f = undefined;
+	pub var lastDir: Vec3f = undefined;
```
Argmaster: I don't think it should be made public; ~~this should be a parameter to callback. Preferably inside a struct alongside `lastPos` as `playerOrientation` or alike.~~ Callback should receive current player position and direction and should figure out storing that. Then the callback should figure out storing that, no?

src/renderer.zig
```diff
@@ -1039,90 +1036,21 @@ pub const MeshSelection = struct { // MARK: MeshSelection
...
-
-			if (newBlock != block) {
-				updateBlockAndSendUpdate(inventory, slot, selectedPos, block, newBlock);
-			}
+			if (inventory.getItem(slot).onLeftClick().run(.{.item = inventory.getItem(slot), .target = .{.block = .{.block = block, .blockPos = selectedPos}}, .mod = .{}, .deltaTime = deltaTime}) == .handled) return;
 		}
 	}
 
-	fn updateBlockAndSendUpdate(source: main.items.Inventory.ClientInventory, slot: u32, pos: Vec3i, oldBlock: blocks.Block, newBlock: blocks.Block) void {
+	pub fn updateBlockAndSendUpdate(source: main.items.Inventory.ClientInventory, slot: u32, pos: Vec3i, oldBlock: blocks.Block, newBlock: blocks.Block) void {
```
Argmaster: No you don't. This function makes no sense as part of public API of renderer. Renderer does rendering, it doesn't send updates. Move it or call `executeCommand`.

src/renderer.zig
```diff
@@ -893,15 +893,12 @@ pub const MeshSelection = struct { // MARK: MeshSelection
 	var posBeforeBlock: Vec3i = undefined;
 	var neighborOfSelection: chunk.Neighbor = undefined;
 	pub var selectedBlockPos: ?Vec3i = null;
-	var lastSelectedBlockPos: Vec3i = undefined;
-	var currentBlockProgress: f32 = 0;
-	var currentSwingProgress: f32 = 0;
-	var currentSwingTime: f32 = 0;
+	pub var lastSelectedBlockPos: Vec3i = undefined;
```
Argmaster: This should be managed by callback implementation IMO. `MeshSelection` may have its own use cases for it (does it?) but even then it shouldn't share it with everyone.

src/callbacks/callbacks.zig
Argmaster: Honestly I would remove the target completely and instead provide tools to query the possible target from `MeshSelection` if the callback cares to do so. This means the params need to be extended with everything required for such query.

src/renderer.zig
Wunka: that means callbacks would need a finsish method right? How would they else know that the they can remove the animation

src/items.zig
Wunka: is that not what I am already doing? The only wierd thing is the `null` but for that I have to fallback to the default behavior

src/items.zig
Wunka: Wouldn't that mean that for example if a player has an item wich has like 300 `timesUsed` and we reduce through an update the `maxUsage` to something below 300, then the item would just break instantly on joining the world.

src/items.zig
Argmaster: I would expect 

```
                        .baseItem => |item| item.onLeftClick(),

```

And call into callback inside impl of onLeftClick
Not sure tho, I had a look at other implementations after making this comment and they are not that consistent.

src/items.zig
Argmaster: We break items only when they are used.
But yes, general intention is to make the item asap.

src/items.zig
Wunka: Ah. I understand now. Yeah its not consistent, but I will change it.

src/renderer.zig
Argmaster: No? I think no.
This is clearing braking animation when you start braking different block.
Since this callback calls into updateBlock sync call,  you can do stuff before and after it too, so you can also clear it after brekitn a block.

src/items.zig
Wunka: changed

src/renderer.zig
Wunka: but the game.breakBlock has no access to `selectedPos` I think the call order is ok.

Argmaster: Eh alright, then I guess there is unfortunately a reason. I will have a look at this when I have time to maybe find a better way around this.

src/callbacks/callbacks.zig
Wunka: I removed it. callbacks now get the `selectedBlockPos` which they (if it is not null) can be used to query the block. This also already helped calls because the selection wand doesn't care about the block only its position

src/renderer.zig
Wunka: is now given over callback

src/renderer.zig
Wunka: breakBlock seems to use it without reason, in the cases it uses it, it should be the same as `selectedPos`. I atleast didn't notice any difference in breaking animation.

src/renderer.zig
Wunka: I have to look how I will exactly handle this.

src/callbacks/callbacks.zig
```diff
@@ -4,18 +4,28 @@ const main = @import("main");
...
 const Vec3i = vec.Vec3i;
+const Vec3f = vec.Vec3f;
 
 pub const ClientBlockCallback = Callback(struct { block: Block, chunk: *main.chunk.Chunk, blockPos: Vec3i }, @import("block/client/_list.zig"));
 pub const ServerBlockCallback = Callback(struct { block: Block, chunk: *main.chunk.ServerChunk, blockPos: main.chunk.BlockPos }, @import("block/server/_list.zig"));
 
 pub const BlockTouchCallback = Callback(struct { entity: *main.server.Entity, source: Block, blockPos: Vec3i, deltaTime: f64 }, @import("block/touch/_list.zig"));
 
+pub const ItemUsedCallback = Callback(struct {
+	item: main.items.Item,
```
IntegratedQuantum: Should pass the inventory slot, so you can run useDurability and similar commands on it.

IntegratedQuantum (changes requested):
I'm not sure where to start here, but I think this is a bit of a step in the wrong direction.

First up, all player actions should be synced with the server to avoid cheating, so the item callback should be executed in the context of a sync command (see also #2381).
This is also important, since otherwise actions can happen out of sync (e.g. the way it's currently implemented the selection wand action and a world edit command may happen out of order since one is happening in the render thread, while the other is deferred to the next server tick as part of the sync system)

Secondly, all the animation state should be shared through an inventory component (@RanPix @tillpp maybe you already have something planned for this?), furthermore the current swinging system is planned to work differently, instead of only progressing when you keep the button pressed, a started swing should end even if you release the button (see also #1137).
In total I'd suggest to for now keep block breaking untouched, but only run it if there is no other successful callback.

Lastly, together with the points above: The callback should not know the deltaTime.

IntegratedQuantum:
Also it might be worth considering to also pass the name of the keyboard key, this would make it easier to add other keybinds (at least right click and middle click are on the table for the base game)

Argmaster:
I am not sure if I understand.
I assumed we are aiming for similar interface as `touch` callback, where we would call into sync system as we need.
With this assumptions I cannot agree that this PR is a step in wrong direction, as only thing it does is moving existing logic into our callback framework, which I find very much aligned with **what I imagined**.
Yes, the logic of using tools requires a complete overhaul, as mentioned in our discussion https://github.com/PixelGuys/Cubyz/pull/2892#pullrequestreview-4134364939 but changing **that is waaaay beyond what one PR can do** (we already have 300 lines of changes). IMO its starting angle as good as any other.

In the long run, I imagined that those callbacks will be client side key stroke handlers and they will invoke sync system.
In a subsequent PRs we would add block damage tracking and corresponding sync commands. After that we would refactor `breakBlock.zig` so all that remain in it would be logic that prevents us from sending sync commands every time player spams the left click button. Malicious client can still DDOS  us with those, so we will have to check it properly on server side, but we don't want to DDOS ourselves by default. I assumed similar logic will be needed every now and then. 

Some more stuff that might want to be outside sync:
- UI opening
- not-implemented call to server
- different sync commands depending on modifiers

IntegratedQuantum:
touch callback also must be synced in the future together will all other player input, and I think at this point we should not introduce more callbacks that don't use the sync system, to make this rewrite harder than it already is.

> In the long run, I imagined that those callbacks will be client side key stroke handlers and they will invoke sync system.

If the callback triggers a health increase, then how can the server verify that you actually are allowed to increase your health? Furthermore there is a bit of a synchronization issue, if another player places a cactus in your path, then you should take damage, even if your client didn't see it yet. That is only possible if the server evaluates the callback

---

I don't really get the DDOS concern, 1 packet per frame is already a given, and a few bytes more wouldn't make a difference there.

> UI opening

Should be synced, so that when e.g. another player breaks your workbench, then the client can just roll back and close the UI (instead of the server having to tell the client about it).

tillpp:
> all the animation state should be shared through an inventory component

We want to store animation Data in the model component.
(no idea what this entire conversation is about, i've got pinged here)

Argmaster:
Okay, that makes sense.
Not sure how to do that tho.

So, my (updated) thinking:
We sort of can't trust anything; so we could for example have sync command (lets temporarily call it `PlayerInput`)
We need to do `rayIntersect` on both sides, otherwise players can just give you arbitrary data. `MeshSelection` logic needs to become more side agnostic.
If we have ECS in place, I think we can rely on it to give us information about player orientation on both sides, so we don't need to send it. Again, with ECS we probably don't need to send inventory nor slot either, as we would rely on currently selected slot (?) if ECS knows that.
We would need to send what kind of input it is.
From there, *I* would dispatch different types of inputs into callbacks that implement sync transactions (possible?) so we can change them depending on items.
`PlayerInput` -> `onLeftClick/onRightClick/...?` callbacks still need to be aware which side they execute on. 
Using break/place terms doesn't make sense in all contexts, I don't think it makes sense to differentiate this way.
We need to move `BlockDropLocation` (basically drop spawning logic) to `itemdrop.zig` IMO, it would make more sense than current `UpdateBlock.BlockDropLocation`. Either way with ECS and access to chunk data we don't need to send it.
So, sync system ends up in the middle of everything, not at the end like now.

IntegratedQuantum:
Yeah, indeed we need to do the raycast on both sides, and we need to recalculate it in the sync command that interacts with things.

> If we have ECS in place, I think we can rely on it to give us information about player orientation on both sides, so we don't need to send it.

Currently, and I'm not sure if ECS changes this, the orientation is not properly synced through the sync system yet so it wouldn't be reliable.

> Using break/place terms doesn't make sense in all contexts, I don't think it makes sense to differentiate this way.

Yeah, I think it should be differentiated based on the key you press.

> Either way with ECS and access to chunk data we don't need to send it.

Yeah, indeed the BlockDropLocation is only sent because we don't have access to the raycast result on the server currently.

Argmaster:
> Currently, and I'm not sure if ECS changes this, the orientation is not properly synced through the sync system yet so it wouldn't be reliable.

Well it better start being, because its sort of game changing; it decides if you can hit someone, which direction you shoot an arrow, what part of body you get hit into. And we need it for raycast. It also changes orientation of player model, tho it depends how precise we want the entity intersection in the end, capsule doesn't care.

Wunka:
I don't quite understand what the result of the discussion was now. I have at least removed breakBlock as a callback and removed deltatime and passed Key.Modifiers. 

So I think the end question I have: is this now blocked by some rewrite?

IntegratedQuantum:
> is this now blocked by some rewrite?

No this is not blocked, just be aware of where we need to go in the future.

So please apply the remaining thing from https://github.com/PixelGuys/Cubyz/pull/2898#pullrequestreview-4205908114: Properly sync the callback execution with the server.

Wunka:
So from what I understand, some sync command like this:
```zig
const UseItem = struct {
	source: InventoryAndSlot,
	key: // some kind of enum to define which "key" / action  was pressed / done independent from user specific settings
	mod: Key.Modifier
	// information that should not be here but for now not available from serverside?
	selectedBlock: Block
```

Wunka:
And then in there the callback would be run from the item with the context struct from sync command?

Argmaster:
> So from what I understand, some sync command like this:
> ```zig
> const UseItem = struct {
> 	source: InventoryAndSlot,
> 	key: // some kind of enum to define which "key" / action  was pressed / done independent from user specific settings
> 	mod: Key.Modifier
> 	// information that should not be here but for now not available from serverside?
> 	selectedBlock: Block
> ```

Neither it has to be *use*, nor it has to be of an *item* (it can be an empty slot). Maybe just KeyPress?
Selected block has to at least be optional, since player may or may not be looking at one. 
Mind that this struct has to contain enough data to run both server and client side code.

Wunka:
> Neither it has to be _use_, nor it has to be of an _item_ (it can be an empty slot). Maybe just KeyPress? Selected block has to at least be optional, since player may or may not be looking at one. Mind that this struct has to contain enough data to run both server and client side code.

Yeah polishing it out is something different, just wanted a general idea out to find out if I misunderstood something completely

IntegratedQuantum:
> So from what I understand, some sync command like this:

For now I think this would be a good start, we will however need to make some changes in the future (serverside block detection, server-synced item slot selection, ...)

Wunka:
Okay I made a new sync method. I also moved the `onInteract` from `placeBlock` there. (I know it should probably be done in a follow up pr). I do want to add to the `onInteract` callback the `ctx` paremeter so it can also do server side stuff (like setting a spawn point / protecting places). So see this inclusion of `onInteract` as a proof of concept how I kinda invision this system.

Wunka:
Also I don't like the name `GameKeys` but I have literally no idea what to name it

IntegratedQuantum:
> Also I don't like the name GameKeys but I have literally no idea what to name it

~~KeyPress?~~ edit: nevermind

src/sync.zig
```diff
@@ -1771,6 +1773,63 @@ pub const Command = struct { // MARK: Command
...
+		key: main.game.GameKeys,
+
+		// TODO remove after information is thrugh other methods synced with server
+		lastDir: Vec3f,
+		selectedBlockPos: ?Vec3i = null,
+
+		pub fn run(self: KeyPress, ctx: Context) error{serverFailure}!void {
+			switch (self.key) {
+				.leftClick => {
+					_ = main.game.Player.inventory.getItem(self.source.slot).onLeftClick().run(.{
```
IntegratedQuantum, line 1789: Could you also start thinking about the corresponding right click callback?
I wonder if this switch case is the right abstraction, or if we should instead pass the key type to the callback (to allow e.g. a single callback to handle the selection wand behavior)

src/renderer.zig
```diff
@@ -986,14 +986,50 @@ pub const MeshSelection = struct { // MARK: MeshSelection
...
+					const block = mesh.chunk.getBlock(blockPos[0] - mesh.pos.wx, blockPos[1] - mesh.pos.wy, blockPos[2] - mesh.pos.wz);
+					if (!block.onInteract().isNoop()) {
+						main.sync.client.executeCommand(.{.keyPress = .{
+							.lastDir = lastDir,
+							.selectedBlockPos = selectedBlockPos,
+							.key = key,
+							.mod = mods,
+							.source = .{.inv = inventory.super, .slot = slot},
+						}});
+						return;
```
IntegratedQuantum, line 1017: This is breaking behavior, with this you can't place a block on a workbench anymore.

IntegratedQuantum (changes requested):
I think that we may need to start with more rewrites first (such that all onPress functions are synced), since the current implementation is breaking behavior since some things are inside the sync command, while others aren't.

src/sync.zig
Wunka: I thought about it too. 
We need a way to know that we can fallback to breaking of blocks / placing of blocks. 
So we need either a way to know that the callback will acutally do anything (harder if we join `onLeftClick` / `onRightClick`, or better said we would need a extra method for each callback which would return if the callback will actually do anything)
Or the breaking / placing of blocks gets integrated into their own callback. then we can go by `.handeld` / `.ignored` (now not possible because it is only known at the time of running the sync command

src/renderer.zig
Wunka: What is the exact breaking behavior here? you can still place things via shift

IntegratedQuantum: Oh right, I missed the mods.shift check above.

Either way, the problem that I see here is that the result of the interact callback is not evaluated, the intention here is that when the callback return `.ignored` the next action is executed (→ placing a block).

Argmaster:
Actually, what happens if player rebinds left click to not-a-left-click? Not even mentioning controller. `onLeftClick` makes way less sense than I thought.
So what is that really? It's not `onPlace`/`onBreak` because that makes no sense in case of selection wand, not even mentioning accessories.
`onPrimaryAction` `onSecondaryAction`?
We can't bind keys directly, so we need a way to create abstract "actions" in game (with names) and then they can be used to identify player activity in stable manner, while keys are rebound.

src/renderer.zig
Wunka: Yeah exactly thats why we either need the second function (mentioned in my other comment) or we first need to prepare break / place to also work in this system.

Wunka:
> Actually, what happens if player rebinds left click to not-a-left-click? Not even mentioning controller. `onLeftClick` makes way less sense than I thought. So what is that really? It's not `onPlace`/`onBreak` because that makes no sense in case of selection wand, not even mentioning accessories. `onPrimaryAction` `onSecondaryAction`? We can't bind keys directly, so we need a way to create abstract "actions" in game (with names) and then they can be used to identify player activity in stable manner, while keys are rebound.

Nice, the perfect words that didn't came to my mind. Will rename!

src/sync.zig
```diff
@@ -1771,6 +1773,63 @@ pub const Command = struct { // MARK: Command
...
+						.mod = self.mod,
+					});
+				},
+				.rightClick => {
+					if (ctx.side == .server) return;
+					const blockPos = self.selectedBlockPos orelse return;
+
+					const mesh = main.renderer.mesh_storage.getMesh(.initFromWorldPos(blockPos, 1)) orelse return;
+					const block = mesh.chunk.getBlock(blockPos[0] - mesh.pos.wx, blockPos[1] - mesh.pos.wy, blockPos[2] - mesh.pos.wz);
+					_ = block.onInteract().run(.{
```
IntegratedQuantum, line 1803: Currently this may be problematic as well, e.g. when it comes to opening a GUI, you will reopen the GUI every time the client needs to rollback and reapply changes. I think we should make GUI opening server-initiated (this would also make other issues easier to solve)

Argmaster:
> Nice, the perfect words that didn't came to my mind. Will rename!

It implies more than just renames. I don't have much experience with 3D engines, but after a [brief look at Godot](https://docs.godotengine.org/en/stable/getting_started/first_3d_game/02.player_input.html) (and similar systems in Unity and UE4/5) we should probably consider how they implemented it.

My understanding is that common pattern is to have `actions` that are ~~fixed~~ mapping `Key+Modifiers -> Action` (also involves distinguishing between press and release, to detect hold) and then the engine operates on `Actions` instead of specific keys. Quite a jump from current status quo I think?

Argmaster:
Not something we have to implement right now, but we should discuss that and have in mind eg. that in future we may want to eliminate modifiers from the signature.

tillpp:
Since there hadn't be a change in a month ( No shame on wunka, we all get bussy)
Would it be ok for you if someone else takes over this PR and continues it? ( cause I need this for my PVP PR )

(that someone might be me)

Wunka:
> Since there hadn't be a change in a month ( No shame on wunka, we all get bussy) Would it be ok for you if someone else takes over this PR and continues it? ( cause I need this for my PVP PR )
> 
> (that someone might be me)

My block is this:
> we first need to prepare break / place to also work in this system.

So if someone (who might be named tillpp) wants to do that. I am happy to give this pr up.

src/sync.zig
tillpp: Could you give me an example of an item where it wouldn't be that clear if rightClick would place the block, until the callback is called?

Wunka: > Could you give me an example of an item where it wouldn't be that clear if rightClick would place the block, until the callback is called?

Any Item which can interact with specific blocks. For example say we had a itemframe. before calling the callback: Do you know if the block is placed in front of the itemframe or put into it? We can't until we either get `.handeld` or `.ignored` from the callback.

tillpp: i see,

I assume putting the block placing as part of the command would lead to too much delay between pressing and seeing the block in the world ( because of ping )?

I propose the following: (which is actually similar/close/**same** to your idea)
the client+item evaluates first if the command would do anything (with a separate callback only called on the client), and then decide to place or send the command

what do you think?

Wunka: > I assume putting the block placing as part of the command would lead to too much delay between pressing and seeing the block in the world ( because of ping )?

Nope as we do client side prediction

> I propose the following: (which is actually similar/close/**same** to your idea) the client+item evaluates first if the command would do anything (with a separate callback only called on the client), and then decide to place or send the command

While this would be a solution it would mean we need "two" systems just because we didn't migrate another thing. 
Would probably need @IntegratedQuantum opinion but from my understanding it is the goal to when we get the callbacks to also "instantly" migrate everthing in one swoop

IntegratedQuantum: > I assume putting the block placing as part of the command would lead to too much delay between pressing and seeing the block in the world ( because of ping )?

Doing it like this is the end goal, though I suspect it's going to take more steps to get there.

> the client+item evaluates first if the command would do anything (with a separate callback only called on the client), and then decide to place or send the command

This is a workaround, and it duplicates logic (you need the same logic in the shared callback to check if the condition is still true). Adding this now would just mean more refactoring in the future to remove it again.

> but from my understanding it is the goal to when we get the callbacks to also "instantly" migrate everthing in one swoop

No, not necessarily, we should do small. There are multiple distinct problems here (server-initiated GUI opening, syncing player position and rotation, syncing (some) player key presses, adding new callbacks, callbacks shared between both sides #2489).
It is also not a requirement to migrate all of them at once, the resulting PR would be too big anyways.
For now it would be fine to only send the command when all the other client-side callbacks returned `.ignored`, then we can slowly migrate them.

src/game.zig
```diff
@@ -689,16 +688,16 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
-			Player.placeBlock(main.KeyBoard.key("placeBlock").modsOnPress);
+			Player.keyPress(.rightClick, deltaTime, main.KeyBoard.key("placeBlock").modsOnPress);
 		}
 	}
-	if (nextBlockBreakTime) |*breakTime| {
+	if (GameKeys.nextLeftClickTime) |*breakTime| {
 		if (breakTime.durationTo(time).nanoseconds >= 0 or !Player.isCreative()) {
 			breakTime.* = breakTime.addDuration(main.settings.updateRepeatSpeed);
-			Player.breakBlock(deltaTime);
+			Player.keyPress(.leftClick, deltaTime, main.KeyBoard.key("breakBlock").modsOnPress);
```
tillpp, line 700: since we replaced "next **Block**_Break_ Time" with "next **Left**_Click_ Time", we could consider to also rename "_break_**Block**" to something like "_click_**Left**"

src/game.zig
```diff
@@ -689,16 +688,16 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
+	if (GameKeys.nextRightClickTime) |*placeTime| {
 		if (placeTime.durationTo(time).nanoseconds >= 0) {
 			placeTime.* = placeTime.addDuration(main.settings.updateRepeatSpeed);
-			Player.placeBlock(main.KeyBoard.key("placeBlock").modsOnPress);
+			Player.keyPress(.rightClick, deltaTime, main.KeyBoard.key("placeBlock").modsOnPress);
 		}
 	}
-	if (nextBlockBreakTime) |*breakTime| {
+	if (GameKeys.nextLeftClickTime) |*breakTime| {
 		if (breakTime.durationTo(time).nanoseconds >= 0 or !Player.isCreative()) {
```
tillpp, line 698: we slow down the player in creative with breakTime since breaking is instantly.
This is ok, when we assume this code only handles breaking blocks, but i could image future tools where we would want the item to behave the same in creative and in survival.

Nothing that we have to fix now, but something we might want to think about in the future (in an issue)

src/sync.zig
```diff
@@ -1771,6 +1773,63 @@ pub const Command = struct { // MARK: Command
 			};
 		}
 	};
+
+	const KeyPress = struct { // MARK: KeyPress
+		source: InventoryAndSlot,
+		mod: main.Window.Key.Modifiers,
+		key: main.game.GameKeys,
+
+		// TODO remove after information is thrugh other methods synced with server
```
tillpp, line 1782: can you explain what you had in mind with this TODO?

src/sync.zig
tillpp, line 1789: maybe i missing something, but wouldn't this lead in multiplayer to cause the Host to trigger onLeftClick when a different player is actually doing it?

src/sync.zig
tillpp: ~~where can i find this rollback functionality?
The sync code is the code i am least familiar with, but i can't find how the block.onInteract() would be rolledback and executed again~~

found it

tillpp (commented):
I have a few Questions, such that a feature PR continuer (potentially me) can work on this.

src/sync.zig
Wunka: https://github.com/PixelGuys/Cubyz/pull/2898#issuecomment-4363522445
These two parameters below should in the future be already known by the server. So then it should not be passed via this sync command

src/sync.zig
Wunka: Could be. I can currently not test. But a simple if(side == .server) would fix it

src/sync.zig
tillpp: When doing my PVP PR i thought about this, and came to the conclusion, that it would be best to send it anyway, and check on the server if they match/are close enough.

Imagine trying to hit a mob, but on the server another Player has moved already in front of you, so you would hit your friend even though you intended and aimed at the mob.

By checking if they match up/are close enough this can be prevented

src/game.zig
IntegratedQuantum: The naming leftClick is not particularly good here since it's remappable.
I'd suggest to name it `break/attack` and `place/use` in the UI and in code.

src/sync.zig
IntegratedQuantum: On the other hand, if we do not check line-of-sight on the server then the player could hit through walls.
The common solution here is to simulate on the server what the client would have seen, but I think such a system is too complicated.
I think accidentally hitting a friend is fine and I think we can hide it to some degree behind animation time.

tillpp: > On the other hand, if we do not check line-of-sight on the server then the player could hit through walls.

I didnt say i didnt want to check the line of sight on the server.
I want to check them on both! And only do the action if they match up.
the client basicly sends position of block, type of block, the server checks line of sight, checks if they both aiming at the same position and same blocktype

PR #2899, opened 2026-04-12
Async those block updates (conquering the flicker biome addon)

### LINKED ISSUE #277: Placing blocks lags a lot in debug mode.

IntegratedQuantum:
This is done by converting the global block update queue into a per chunk block update queue.
The actual block updates are also done on the network thread, so they still appear right away, just not visually.
As for the visuals, I have not seen noticable delays when placing blocks in release modes, even with background noise from the flicker addon, in debug mode the story is different though, but we had the same amount of lag before, it's just not freezing the render thread anymore.

https://github.com/user-attachments/assets/8cfba26f-340c-46ee-aa4c-b650aefcdaf4

Should help with #2585
fixes #277 (the lag still exists, and causes a noticable delay, but it doesn't freeze the interface, so you can keep working which is what we care about)

Remaining issues:
- [x] There is some flickering in the light updates, I think we need to find a way to avoid uploading the mesh if the light was updated in the mean-time (how can we test this?)
- [x] verify that we are allowed to update blocks from the network thread without causing race conditions
- [x] actually test for correctness with some light blocks

### DIFF
--- a/src/renderer/chunk_meshing.zig
+++ b/src/renderer/chunk_meshing.zig
@@ -632,6 +632,8 @@ pub const ChunkMesh = struct { // MARK: ChunkMesh
 	lightListNeedsUpload: bool = false,
 	lightAllocation: graphics.SubAllocation = .{.start = 0, .len = 0},
 
+	blockUpdateQueue: main.utils.CircularBufferQueue(Vec3i) = undefined,
+
 	lastNeighborsSameLod: [6]?*const ChunkMesh = @splat(null),
 	lastNeighborsHigherLod: [6]?*const ChunkMesh = @splat(null),
 	isNeighborLod: [6]bool = @splat(false),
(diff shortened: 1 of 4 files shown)


Argmaster:
image

Achieved by `/replace cubyz:air` of large region followed by `/undo` and `/redo`, this appears randomly, I assume its lighting bug?
Breaking nearby block / placing light block updates it to correct state.

image

(ignore the blue dot, its a crosshair from Crosshair X)

IntegratedQuantum:
Good catch. Seems like it wasn't correctly updating the mesh light data in some race conditions.
The issue will still appear, but it will correct itself quickly enough.

PR #2900, opened 2026-04-13
ECS Part 13 | Character Selection GUI

tillpp:
https://github.com/user-attachments/assets/df2c837a-ca73-4287-9e41-7e6fde3bd7c7

TODO:
- [x] add entity model name above the frames
- [x] call the /avatar <entityModelId> in the background (kind of like players calls /kick when kicking a player)
- [x] have a scrollbar for when there are more then 5 models


tillpp:
The Button is programmer Art, i hope maybe one of our talented **Artists** might come up with a redesign.

I was originally going to implement this button into where the accessories are, but since they aren't ready yet, I will add that button to the inventory for now.

grafik

tillpp:
ready for review.
This is going to be my last opened PR before the 0.3.0 Release.
Completing the ECS PR series for the most Part. :partying_face:

Argmaster:
Things about the UI/UX
- nowhere in the UI we have rounded corners, so for consistency we shouldn't have them here.
- background IMO shouldn't be pitch black, could eg. be much darker version of window background texture?
- displayed models should be guaranteed to have some space in every axis, right now they have space horizontally, but they touch edges of display space vertically.
Only partially related: IMO it's impractical to guarantee consistent scale of models over ability to view the whole model. IMO it would be a good approach to get min and max and scale to fit vertically/horizontally.
- it would make sense to make windows have 1:2 / 2:3 proportions, since most models will be somewhere in the neighborhood of that proportion.
This comment was written purely based on provided screenshots

tillpp:
> nowhere in the UI we have rounded corners, so for consistency we shouldn't have them here.
> background IMO shouldn't be pitch black, could eg. be much darker version of window background texture?
> it would make sense to make windows have 1:2 / 2:3 proportions, since most models will be somewhere in the neighborhood of that proportion.


I am going to repeat what i told quantum in his DMs 3 months ago:
I am a programmer, I make the functionality, ugly but functional, and the Artist can change,fine tune it to its liking later, like changing the background image or the proportions of the frame.

~~(thought, i can already change the proportions)~~
Edit: no, i actually, i am going to let the artist figure out the right proportions later

> displayed models should be guaranteed to have some space in every axis, right now they have space horizontally, but they touch edges of display space vertically.

yeah i agree, i can add some vertical spacing.

> Only partially related: IMO it's impractical to guarantee consistent scale of models over ability to view the whole model. 

funny, i did this first, then people told me it would be better to have a consistent scale, to be able to see which model is bigger then what other model. (also note that most playermodels should have roughly the same height anyway ). I could change it back, but I wouldn't want people demanding me change it back and forth again. So let's do a poll?

> IMO it would be a good approach to get min and max and scale to fit vertically/horizontally.

You mean getting the dimensions of the model and scaling it down/up such that it fits in the EntityModelFrame, right?

Argmaster:
> funny, i did this first, then people told me it would be better to have a consistent scale, to be able to see which model is bigger then what other model. (also note that most playermodels should have roughly the same height anyway ). I could change it back, but I wouldn't want people demanding me change it back and forth again. So let's do a poll?

People are silly and rarely understand full extent of what they say. Yeah, unless Quantum makes a call I guess poll it is.
Wait until someone makes a cool dragon model and you can't even see it in the UI because camera is focused perfectly in between it's legs.

PR #2912, opened 2026-04-15
Cirrus trees

careeoki:
A sky island wood type.

Instead of `durable` modifier like other woods, cirrus has `light`. It also does enough damage to break nimbusite (which has 4 resistance,) so if you run out of materials on the sky islands or are doing some sky island challenge, you are still able to make some tools.
(planks are outdated here)
image
You can also see some vines and stalactites on the islands. I also removed daffodils to put more focus on the unique stuff. Daisies are still here cause they fit
image

- [x] vine bottom
- [x] vine item
- [x] texture final pass
- [x] build the trees


H41ogen:
i love this, but something just feels off about the planks texture, and i can't put a finger on it.

TyrrelMeter:
> i love this, but something just feels off about the planks texture, and i can't put a finger on it.

Might be a little to dark or to much contrast

du82:
I like it! I think we need lots more blocks and unique textures!

Wbjitscool:
Nice

Crepestrom:
could you give the biomes with the cirrius tree structure a tag with .cirrus
this would help with biome independant structures

Wbjitscool:
uh oh something must of broken the compiler it's showing an x

assets/cubyz/blocks/vine/cirrus.zig.zon
Argmaster, line 1: This file has to be renamed so the file name ends with `.zig.zon`

Argmaster (changes requested):

careeoki:
This is ready.
image

du82:
I really like these @careeoki

Wbjitscool:
Same

IntegratedQuantum (changes requested):
Overall looks really nice, and the idea of making leaf rings is is also good.
I also like the detail on the roots, that they curl upwards, as if trying to reach the soil again.
And it fits nicely at night with the pink light from the flowers.

But there is one thing that I think should be improved: variety
You already tried to fix it, by adding the leaning variants, and they are good, but the base variant looks too symmetrical.
I'd suggest, instead of placing the ring in one SBB, to split the ring into four sections, one for each side, this would allow you to have more variants and some breaks in the symmetry by having separate leaf and branch shapes. This could also allow for fun things like missing a segment.

With a little trickery, you can also make the leaf height consistent with different branch heights, here is a quick illustration of the idea:
Image

careeoki:
Done, they should have much more variation and be less symmetrical 

image

IntegratedQuantum (approved):
This looks great, and a lot more organic than the first attempt, thanks for putting in all the extra work!

PR #2914, opened 2026-04-16
ECS Patch 3 | moved entityModel reset to main thread

### LINKED ISSUE #2906: Consistent Segfault on World Exit

After one of the last few updates (something between ac54ee2f and e0dac499), I now get consistent segfaults when exiting a world with the following error message:
```sh
Disconnected
Disconnected
Segmentation fault at address 0xa38
???:?:?: 0x7f6c71150f69 in ??? (/usr/lib/libGLdispatch.so.0)
Unwind error at address `/usr/lib/libGLdispatch.so.0:0x7f6c71150f69` (unwind info unavailable), remaining frames may be incorrect
/home/richardb/src/Cubyz/src/entityModel.zig:291:15: 0x13e3737 in reset (Cubyz)
  model.deinit();
              ^
/home/richardb/src/Cubyz/src/server/world.zig:328:27: 0x13734c0 in deinit (Cubyz)
  main.assets.unloadAssets();
                          ^
/home/richardb/src/Cubyz/src/server/server.zig:681:14: 0x1320ebf in startFromExistingThread (Cubyz)
 defer deinit();
             ^
/home/richardb/src/Cubyz/src/server/server.zig:675:25: 0x14a284e in startFromNewThread (Cubyz)
 startFromExistingThread(name, port);
                        ^
???:?:?: 0x7f6c9aea7960 in ??? (/usr/lib/libc.so.6)
./debug_linux.sh: line 31: 13751 Aborted                    (core dumped) ./zig-out/bin/Cubyz
```

tillpp:
(maybe) fixes #2906

moves the entityModel reset to main thread.

### DIFF
--- a/src/assets.zig
+++ b/src/assets.zig
@@ -786,7 +786,6 @@ pub fn unloadAssets() void { // MARK: unloadAssets()
 	main.particles.ParticleManager.reset();
 	main.rotation.reset();
 	main.Tag.resetTags();
-	main.entityModel.reset();
 
 	// Remove paths from asset hot reloading:
 	var dir = main.files.cwd().openIterableDir("assets") catch |err| {
--- a/src/game.zig
+++ b/src/game.zig
@@ -670,6 +670,7 @@ pub const World = struct { // MARK: World
 		self.entityModelPalette.deinit();
 		self.manager.deinit();
 		main.server.stop();
+		main.entityModel.reset();
 
 		Player.super.deinit(.client);
 


Argmaster:
Hello kind sr, I am here to announce that with this change I am unable to reproduce that pesky seg fault :))))

Argmaster (approved):

IntegratedQuantum (changes requested):
Am I correct to assume that the problem is that it also calls assets.reset on the server-side, causing it call into the driver to delete the models?

Maybe we should make a separate function that unloads client assets (e.g. #1368 also belongs in this category, same with several other client side data structures) (doesn't need to be here, change request is only for answering my first question)

tillpp:
just realized that this Patch fixes one issue but introduces another issue.

it resets the assets client side, but never resets the assets server side.
Which would be a memory Leak for headless servers.

(for non-headless servers this shouldnt be a memory leak)

IntegratedQuantum:
> it resets the assets client side, but never resets the assets server side.
> Which would be a memory Leak for headless servers.

Actually not true. All other model data uses the worldArena, so no leaks happen.

IntegratedQuantum (approved):

tillpp:
but the entityModels variable doesnt get reset on headless servers when leaving a world and join another.
(theoretically that never happens i guess, because server owners will just restart before switching the world)
But for a clean design it should still be considered. (In case we ever think about world switching on headless servers or restart a sever without closing the application or something similar)

IntegratedQuantum:
headless server is not intended to be able to leave and rejoin a world.

PR #2922, opened 2026-04-17
Add `is_flipped_vertically` flag to image loading

Argmaster:
So, I usually say that flags are bad, but they are not *that* bad, to duplicate whole function body :)))


src/graphics.zig
```diff
@@ -2534,25 +2534,12 @@ pub const Image = struct { // MARK: Image
...
-		const data = stb_image.stbi_load(nullTerminatedPath.ptr, @ptrCast(&result.width), @ptrCast(&result.height), &channel, 4) orelse {
-			return error.FileNotFound;
-		};
-		main.stackAllocator.free(nullTerminatedPath);
-		result.imageData = allocator.dupe(Color, @as([*]Color, @ptrCast(data))[0 .. result.width*result.height]);
-		stb_image.stbi_image_free(data);
-		return result;
-	}
-	pub fn readUnflippedFromFile(allocator: NeverFailingAllocator, path: []const u8) !Image {
+	pub fn readFromFile(allocator: NeverFailingAllocator, path: []const u8, options: struct { is_flipped_vertically: bool = false }) !Image {
```
```suggestion
	pub fn readFromFile(allocator: NeverFailingAllocator, path: []const u8, options: struct { isFlippedVertically: bool = true }) !Image {
```

Also previous default was `true`, since OpenGL wants them flipped.

IntegratedQuantum (changes requested):

src/graphics.zig
Argmaster: The default was true, but logically it is not what you expect.

Argmaster: We can add alias `readFromFileOpenGl` or something that by default flips the image.

IntegratedQuantum: Maybe an enum would be better?

Argmaster:
```enum
const Orientation = enum {
	asIs,
	openGl,	
}
```

IntegratedQuantum: Yes

Argmaster: Updated. I came to a conclusion that this default was not helping and removed it. We are still come ahead by -13 lines of code.

IntegratedQuantum:
CI is still failing

Argmaster:
> CI is still failing

Now it works

IntegratedQuantum (approved):

PR #2927, opened 2026-04-18
More damage types

Crepestrom:
Adds These damage types

acid
explosion
poison
venom
drowning
cold


IntegratedQuantum:
I think for now we should only add the damage types we need. In the future we'll probably need to consider allowing addons to add and define their own damage type (It could just be a tag with a corresponding format string in the language once we have language support).

PR #2936, opened 2026-04-18
Axes and Branches

Crepestrom:
Buffs Axes to be back in line with other tools in terms of durability

But it also makes decaying branches drop at a 50% chance in order to balance it


careeoki:
Getting these errors:
```
[error]: Cannot use 'auto' in this context
[error]: Cannot use 'auto' in this context
[error]: Cannot use 'auto' in this context
[error]: Cannot use 'auto' in this context
[error]: Cannot use 'auto' in this context
[error]: Cannot use 'auto' in this context
[error]: Cannot use 'auto' in this context
[error]: Cannot use 'auto' in this context
```

When chopping the thin trees in `forest/thin_birch`, they seem to not drop any branches at all.

I do like this better than the durability nerf, and it also reduces the amount of branches clogging your inventory.
I think it could be a little more generous, like 60%, so that small trees made only of branches still give some wood. Sometimes they give you nothing.

PR #2937, opened 2026-04-18
fix particle collisions

RanPix:
there is still an issue if particles spawn inside the block, but i think that should be solved in a different pr


src/particles.zig
```diff
@@ -244,32 +244,24 @@ pub const ParticleSystem = struct {
...
-						box.max[2] - hitBox.min[2]
-					else
-						box.min[2] - hitBox.max[2];
+				const size = @as(f64, @floatCast(ParticleManager.types.items[particle.typ].size))*0.5;
+				const hitBox: game.collision.Box = .{.min = @splat(-size), .max = @splat(size)};
+
+				inline for (0..3) |j| {
+					const move = particleLocal.velAndRotationVel[j]*deltaTime;
+					if (main.game.collision.collides(.client, @enumFromInt(j), move, v3Pos, hitBox)) |box| {
+						const skin = 0.0001;
```
IntegratedQuantum: I did remove that from the other collision functions a while ago because it was causing different issues.
For particles we do however need them (unless we go fixed-point) due to accuracy loss when doing the relative-player coordinate thing.

IntegratedQuantum (changes requested):
Bad news: It still breaks sometimes, as you can see I spawned a ball on the left, but some particles phased through the rocks, and reappeared in small groups on the right:
Image

IntegratedQuantum:
I still observe disappearences when they are dropped from high altitudes, especially at high coordinates.

RanPix:
will probably have to wait until fixedpoints are added

PR #2940, opened 2026-04-19
Allow allocators in testing (take 2)

### LINKED ISSUE #1286: Allocators in testing

Argmaster:
This pull request default initializes `main.stackAllocator` and `main.globalAllocator` to testing allocator.

Resolves: #1286


src/utils/heap.zig
```diff
@@ -11,8 +12,12 @@ pub const testingAllocator = testingErrorHandlingAllocator.allocator();
 pub const allocators = struct { // MARK: allocators
 	pub var globalGpa = std.heap.DebugAllocator(.{.thread_safe = true}){};
 	pub var handledGpa = ErrorHandlingAllocator.init(globalGpa.allocator());
+
+	pub var testingGpa = if (builtin.is_test) ErrorHandlingAllocator.init(std.testing.allocator) else undefined;
```
IntegratedQuantum: Why not just use the one that's 5 lines above it?

IntegratedQuantum (changes requested):
Did you check if they are usable in testing?

src/utils/heap.zig
Argmaster: Hm right, well I didn't expect it there, so I completely ignored it.

src/utils/heap.zig
```diff
@@ -1,3 +1,4 @@
+const builtin = @import("builtin");
```
IntegratedQuantum: unintended change

src/main.zig
```diff
@@ -50,10 +50,10 @@ const file_monitor = utils.file_monitor;
 const Vec2f = vec.Vec2f;
 const Vec3d = vec.Vec3d;
 
-pub threadlocal var stackAllocator: heap.NeverFailingAllocator = undefined;
+pub threadlocal var stackAllocator: heap.NeverFailingAllocator = if (builtin.is_test) heap.testingAllocator else undefined;
 pub threadlocal var seed: u64 = undefined;
 threadlocal var stackAllocatorBase: heap.StackAllocator = undefined;
-pub const globalAllocator: heap.NeverFailingAllocator = heap.allocators.handledGpa.allocator();
+pub const globalAllocator: heap.NeverFailingAllocator = if (builtin.is_test) heap.allocators.testingGpa.allocator() else heap.testingAllocator;
```
```suggestion
pub const globalAllocator: heap.NeverFailingAllocator = if (builtin.is_test) heap.testingAllocator else heap.allocators.handledGpa.allocator();
```

IntegratedQuantum (changes requested):

Argmaster:
One issue with that is that there is no difference between deallocating something with `stackAllocator` and `globalAllocator` in testing, right?
Also, is the testing allocator reset every test or just at the very end?

IntegratedQuantum:
Yes, that is true, though that is technically also possible (to some degree) at runtime, since the stackAllocator uses the globalAllocator as a fallback. If it's trivial to fix, go for it, otherwise I'd say it's not worth the effort.

> Also, is the testing allocator reset every test or just at the very end?

It leak-checks every test.

Argmaster:
> Yes, that is true, though that is technically also possible (to some degree) at runtime, since the stackAllocator uses the globalAllocator as a fallback. If it's trivial to fix, go for it, otherwise I'd say it's not worth the effort.

Hm, then its not trivial to fix I think and its not really critical tbh. IMO out of scope for this PR.

IntegratedQuantum:
Alright, then I'm still waiting for an answer to my earlier question:
> Did you check if they are usable in testing?

Argmaster:
> Did you check if they are usable in testing?

Right, I didn't push it;
I added a small test which I think should be sufficient:

```zig
test "allocators" {
	const allocation1 = stackAllocator.create(u64);
	stackAllocator.destroy(allocation1);

	const allocation2 = globalAllocator.create(u64);
	globalAllocator.destroy(allocation2);
}
```

src/main.zig
```diff
@@ -668,3 +668,11 @@ test "abc" {
 	refAllDeclsRecursiveExceptCImports(@This());
 	_ = @import("zon.zig");
 }
+
+test "allocators" {
```
IntegratedQuantum: I'd prefer a more precise name here.
```suggestion
test "allocators are usable in tests" {
```

IntegratedQuantum (changes requested):

src/main.zig
Argmaster: Yessir

IntegratedQuantum (approved):

PR #2953, opened 2026-04-20
Add young oak trees to more biomes

careeoki:
Adds some more variation to biomes with oak trees.
image

Also replaces the simple trees in `bush_mountains`. You might be thinking, shouldn't the bush mountain have bushes? And you would be right. But those trees were definitely trees and not bushes.
Also also replaced the simple trees in the `thicket`, added .keepOriginalTerrain, and made them slightly more common. You literally never see this biome. I resisted making a bunch of changes to it because that should really be a seperate PR.


assets/cubyz/biomes/rocky_grassland.zig.zon
```diff
@@ -25,11 +25,17 @@
...
+			.chance = 0.014,
 			.type = .round,
 			.height = 6,
 			.height_variation = 3,
 		},
+		.{
+			.id = "cubyz:sbb",
+			.structure = "cubyz:tree/oak/young",
+			.placeMode = .degradable,
+			.chance = 0.02,
```
IntegratedQuantum: Is this a typo? At the top you remove 0.002, here you added 0.02.

IntegratedQuantum (changes requested):

assets/cubyz/biomes/rocky_grassland.zig.zon
careeoki: Yeah, a typo, fixed.

IntegratedQuantum (approved):

PR #2958, opened 2026-04-20
Make it possible to only select block if using the right tools

### LINKED ISSUE #2941: Make it possible to only select block if using the right tools
For example flowers. When using a sickle, it should select the flower. But if not using a sickle, you should select the block underneath it, while ignoring the hitbox of the flower. So that you can either break the block underneath or build inside of the flower while destroying it

Discussed in #2267 and #1558

DellieDelta:
Replaces the block field `.selectable` with `.selectionRule`

This to make `.selectionRule = .toolEffective` possible, which only selects the block if the used tool is effective on it :)

Fixes #2941

Progress towards #1558
Progress towards #627

### DIFF
--- a/assets/cubyz/blocks/lava.zig.zon
+++ b/assets/cubyz/blocks/lava.zig.zon
@@ -1,6 +1,7 @@
 .{
 	.tags = .{.fluid, .hot},
 	.drops = .{},
+	.selectionRule = .toolEffective,
 	.replaceable = true,
 	.degradable = true,
 	.transparent = true,
--- a/assets/cubyz/blocks/lava.zig.zon
+++ b/assets/cubyz/blocks/lava.zig.zon
@@ -1,7 +1,6 @@
 .{
 	.tags = .{.fluid, .hot},
 	.drops = .{},
-	.selectionRule = .none,
 	.replaceable = true,
 	.degradable = true,
 	.transparent = true,
--- a/assets/cubyz/blocks/lava.zig.zon
+++ b/assets/cubyz/blocks/lava.zig.zon
@@ -1,7 +1,7 @@
 .{
 	.tags = .{.fluid, .hot},
 	.drops = .{},
-	.selectable = false,
+	.selectionRule = .none,
 	.replaceable = true,
 	.degradable = true,
 	.transparent = true,
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -68,7 +68,7 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-const SelectionRule = enum { always, toolEffective, none };
+const SelectionRule = enum { always, toolEffective, never };
 
 var _transparent: [maxBlockCount]bool = undefined;
 var _collide: [maxBlockCount]bool = undefined;
@@ -539,7 +539,7 @@ pub const Block = packed struct(u32) { // MARK: Block
 		return switch (self.selectionRule()) {
 			.always => true,
 			.toolEffective => item == .proceduralItem and item.proceduralItem.isEffectiveOn(self),
-			.none => false,
+			.never => false,
 		};
 	}
 };
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -139,7 +139,7 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
 	_light[size] = zon.get(u32, "emittedLight", 0);
 	_absorption[size] = zon.get(u32, "absorbedLight", 0xffffff);
 	_degradable[size] = zon.get(bool, "degradable", false);
-	_selectionRule[size] = zon.get(SelectionRule, "selectionRule", SelectionRule.always);
+	_selectionRule[size] = zon.get(SelectionRule, "selectionRule", .always);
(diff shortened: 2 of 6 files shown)


src/renderer.zig
```diff
@@ -933,10 +933,24 @@ pub const MeshSelection = struct { // MARK: MeshSelection
 		while (total_tMax < closestDistance) {
 			const block = mesh_storage.getBlockFromRenderThread(voxelPos[0], voxelPos[1], voxelPos[2]) orelse break;
 			if (block.typ != 0) blk: {
-				const fluidPlaceable = item == .baseItem and item.baseItem.hasTag(.fluidPlaceable);
-				const holdingTargetedBlock = item == .baseItem and item.baseItem.block() == block.typ;
-				if (block.hasTag(.air) and !holdingTargetedBlock) break :blk;
-				if (block.hasTag(.fluid) and !fluidPlaceable and !holdingTargetedBlock) break :blk; // TODO: Buckets could select fluids
+				const isSelectable: bool = rules: {
+					const holdingTargetedBlock = item == .baseItem and item.baseItem.block() == block.typ;
+					if (holdingTargetedBlock) break :rules true;
```
Wunka:
```suggestion
					if (item == .baseItem and item.baseItem.block() == block.typ) break :rules true;
```
or you can put a `()` around the bool statement in the first line. but thats currently not really good readable (just my opinion)

DellieDelta: Good one! I agree

assets/cubyz/blocks/lava.zig.zon
```diff
@@ -1,7 +1,6 @@
 .{
 	.tags = .{.fluid, .hot},
 	.drops = .{},
-	.selectable = false,
```
IntegratedQuantum, line 4: Please set them to `toolEffective` for now

src/blocks.zig
```diff
@@ -525,6 +526,22 @@ pub const Block = packed struct(u32) { // MARK: Block
 	pub fn canBeChangedInto(self: Block, newBlock: Block, item: main.items.ItemStack, shouldDropSourceBlockOnSuccess: *bool) main.rotation.RotationMode.CanBeChangedInto {
 		return newBlock.mode().canBeChangedInto(self, newBlock, item, shouldDropSourceBlockOnSuccess);
 	}
+
+	pub fn isSelectableByItem(self: Block, item: Item) bool {
+		if (item == .baseItem and item.baseItem.block() == self.typ) return true;
+
+		if (self.hasTag(.air)) return false;
```
IntegratedQuantum: Could you remove this case by setting all .air blocks to `.selectionRule = .never`?

src/blocks.zig
```diff
@@ -68,6 +68,8 @@ pub const Ore = struct {
 	seed: u64,
 };
 
+const SelectionRule = enum { always, toolEffective, none };
```
IntegratedQuantum: Never is more informative here I think.
```suggestion
const SelectionRule = enum { always, toolEffective, never };
```

src/blocks.zig
```diff
@@ -137,9 +139,8 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
 	_light[size] = zon.get(u32, "emittedLight", 0);
 	_absorption[size] = zon.get(u32, "absorbedLight", 0xffffff);
 	_degradable[size] = zon.get(bool, "degradable", false);
-	_selectable[size] = zon.get(bool, "selectable", true);
+	_selectionRule[size] = zon.get(SelectionRule, "selectionRule", SelectionRule.always);
```
IntegratedQuantum: Enum literals don't need to specify the type when the result type is known.
```suggestion
	_selectionRule[size] = zon.get(SelectionRule, "selectionRule", .always);
```

src/blocks.zig
```diff
@@ -525,6 +526,22 @@ pub const Block = packed struct(u32) { // MARK: Block
 	pub fn canBeChangedInto(self: Block, newBlock: Block, item: main.items.ItemStack, shouldDropSourceBlockOnSuccess: *bool) main.rotation.RotationMode.CanBeChangedInto {
 		return newBlock.mode().canBeChangedInto(self, newBlock, item, shouldDropSourceBlockOnSuccess);
 	}
+
+	pub fn isSelectableByItem(self: Block, item: Item) bool {
+		if (item == .baseItem and item.baseItem.block() == self.typ) return true;
+
+		if (self.hasTag(.air)) return false;
+		if (self.hasTag(.fluid)) {
+			const fluidPlaceable = item == .baseItem and item.baseItem.hasTag(.fluidPlaceable);
```
IntegratedQuantum, line 534: I think this is also a case we should dehardcode in the future (doesn't need to happen in this PR, but maybe you could look into this?)

IntegratedQuantum (changes requested):

src/blocks.zig
DellieDelta: Sounds good! It'll need to happen anyway when adding buckets :)

IntegratedQuantum (approved):

DellieDelta:
Thanks for the review Quantum!

IntegratedQuantum:
Well thanks for the PR

PR #2961, opened 2026-04-20
Basalt mountains

careeoki:
A mountain biome with basalt and mahogany trees. It is also a parent biome for volcanoes.
image


ikabod-kee:
Put some boulders in the rockslide! Maybe also add flat bits of jungle!

Argmaster:
Okay, we have SDFs now, how about we add some large SDF spikes to it, so it's not just another type of mountains? Angled around 45deg, not perfectly vertical. No idea why would they exist, but imo they could look cool :3

Wbjitscool:
anything happening now

careeoki:
Will revisit this in the future when we can create more unique mountain shapes (surface SDFs)

PR #2964, opened 2026-04-21
Split calculateMotion out of physics.update

codemob-dev:
Progress towards #1728


tillpp:
I was actually thinking about moving the physics code into the an entityComponent+entitySystem, to allow things like mobs to also have physics (and to prevent people from cheating)

Would your changes, be in conflict with those ideas?

codemob-dev:
> I was actually thinking about moving the physics code into the an entityComponent+entitySystem, to allow things like mobs to also have physics (and to prevent people from cheating)

Quantum gave a good explanation for this in the issue: https://github.com/PixelGuys/Cubyz/issues/1728#issuecomment-3140250128

codemob-dev:
Whoops, misclick.

tillpp:
ah ok just wanted to make sure that is still true after your changes^^

src/physics.zig
```diff
@@ -37,32 +37,33 @@ pub fn calculateFriction(volumeProperties: *const collision.VolumeProperties, fr
 	}
 }
 
-pub fn update(deltaTime: f64, inputAcc: Vec3d, jumping: bool) void { // MARK: update()
+pub fn calculateMotion(deltaTime: f64, friction: *const FrictionState, volumeProperties: *const collision.VolumeProperties, density: f64, pos: Vec3d, velocity: *Vec3d, inputAcc: Vec3d, applyGravity: bool, jumpHeight: f64) Vec3d {
```
IntegratedQuantum: Instead of a bool `applyGravity`, I think passing a float `gravity` would be more useful (note that some calculations, like the friction coefficient, should still use the global gravity), e.g. we may want to have projectiles or particles or entities with custom gravity at some point.

Also pass by value instead of by const reference, unless you have a real benchmark that proves it is faster.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #2965, opened 2026-04-21
Darker mahogany leaves

careeoki:
Mahogany leaves are very close in color to lush grass, which can make it hard to tell the ground from the leaves in the new jungle.
image
So I made them darker to help with this, you can see how it is a lot clearer and less like one big green blob.
image
Here's a greyscale comparison to see the values better:
image

Overall I think this is a positive change, we now have a greater diversity of leaf colours. The new mahogany sits nicely between willow and oak.
image


ikabod-kee:
That's better!!

IntegratedQuantum:
I'm not sure about this, to me the jungle feels colder now, maybe we could add contrast without sacrificing the warm colors?

Also I think you may have overcorrected for the lack of shadows here. Once we get shadows, you will see a difference between the leaves and the ground, as the leaves will be more exposed to sunlight.

Argmaster (approved):

Argmaster:
master + no contrast

image

---
master + contrast 13%

image

---
dark-mahogany + no contrast

image

---
dark-mahogany + contrast 13%

image

Also I would like to point out that ones monitor has huge impact on perceived color and contrast, in a way that cannot be captured on a screenshot. 
Out of two monitors I have, one presents everything significantly darker than the other (NT 250 cd/m2 vs IPS 300/350 cd/m2 is the main difference in my case). On the darker one the `master` jungle is kind of a green blob, but on the brighter one it is not.

At first glance I thought its fine change, but tbh the jungle leaves end up being quite dark now, not sure if that's good, I don't think there is a significant color difference between trees and ground level plants in the jungle. Jungle is in fack kind of a green mess:

image

(I guess lets ignore the fact that there isn't much flat grass in real jungle...)

careeoki:
Alright, I'm going to close this for now, this is something to keep an eye on but let's wait until we have shadowmaps.

Issue #2972, opened 2026-04-23
Remove protocols.asynchronous option


IntegratedQuantum:
It is only used for chunks and lightmaps.
It seemed like a good solution back then, but after seeing that the number of protocols that use it is still just 2, I don't think it's worth it over giving each just their own Task.
For chunks it also makes sense to combine it with the mesh generation task, which would reduce the number of tasks needed, and might reduce energy usage, as more data remains in the cache.

PR #2973, opened 2026-04-23
Do the decompression and the chunk creation in one Task

IntegratedQuantum:
progress towards #2972

No measurable performance improvement was found

### DIFF
--- a/src/network/protocols.zig
+++ b/src/network/protocols.zig
@@ -319,17 +319,67 @@ pub const chunkRequest = struct { // MARK: chunkRequest
 
 pub const chunkTransmission = struct { // MARK: chunkTransmission
 	pub const id: u8 = 3;
-	pub const asynchronous = true;
-	fn clientReceive(_: *Connection, reader: *utils.BinaryReader) !void {
-		const pos = chunk.ChunkPosition{
-			.wx = try reader.readInt(i32),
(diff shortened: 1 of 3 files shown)


src/network/protocols.zig
```diff
@@ -319,17 +319,67 @@ pub const chunkRequest = struct { // MARK: chunkRequest
...
+			main.globalAllocator.destroy(self);
+		}
+	};
 	fn clientReceive(_: *Connection, reader: *utils.BinaryReader) !void {
-		const pos = chunk.ChunkPosition{
-			.wx = try reader.readInt(i32),
-			.wy = try reader.readInt(i32),
-			.wz = try reader.readInt(i32),
-			.voxelSize = try reader.readInt(u31),
+		const task = main.globalAllocator.create(MeshGenerationTask);
```
Wunka, line 372:
```suggestion
		const task = main.globalAllocator.create(MeshGenerationTask);
		errdefer main.globalAllocator.destroy(task);
```

Wunka (changes requested):
The Github UI suggests me as a reviewer so here is my review.

src/network/protocols.zig
IntegratedQuantum: good catch

PR #2974, opened 2026-04-23
Set value of DiscreteSlider when pressing on bar outside button

matjojo:
Same diff as #2881. I've pressed every slider I could find in the menus a few times, and note no weird behaviours anywhere.


IntegratedQuantum (approved):
Since these components are so similar, it may also make sense to investigate if we can merge them.

PR #2977, opened 2026-04-25
Trampoline

Crepestrom:
a bouncy trampoline

mean for the point when a player gets iron bars and starts to need more mobility tools to explore deeper
image

It has the same Bouncyness as a glimmergill cap


Wbjitscool:
cool would be useful in deep caves and other places

Argmaster:
But we already have those in form of bouncy mushrooms.

IntegratedQuantum:
I think it is a neat idea and would probably be good inside of an addon, but I don't think we should add it to the base game just yet. We already have glimmergill for this purpose, and while the trampoline looks nice and could be a good building block, I think we should first figure out if we really want to allow this bouncy-block-based vertical transportation (see #1731)

Issue #581, opened 2024-07-19
Experimental: Add downwards momentum when walking and sprinting


ikabod-kee:
Potentially helps you from flying off stairs when going down them.

Should only add downwards momentum when grounded.

PR #2978, opened 2026-04-25
Coyote friction prototype

IntegratedQuantum:
This is a possible solution for #1969 by giving you a short time window where you have more control after walking off a ledge.
In the situation mentioned in #1969 this gives you enough control to safely reach the ledge.

Currently this does not work properly due to having no distinction between ground friction and volume friction at the point the coyote friction is applied, this make the fall appear somewhat unnatural. This should be easier to fix after #1728. This is a blocker.

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -96,13 +96,16 @@ pub const Player = struct { // MARK: Player
 	pub var selectionPosition2: ?Vec3i = null;
 
 	pub var friction: physics.FrictionState = .{.current = 0, .mobile = 0};
+	pub var lastGroundFriction: physics.FrictionState = .{.current = 0, .mobile = 0};
 	pub var volumeProperties: physics.collision.VolumeProperties = .{.density = 0, .maxDensity = 0, .mobileFriction = 0, .terminalVelocity = 0};
 
 	pub var onGround: bool = false;
 	pub var jumpCooldown: f64 = 0;
 	pub var jumpCoyote: f64 = 0;
+	pub var frictionCoyote: f64 = 0;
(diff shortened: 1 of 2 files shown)


ikabod-kee:
It's neat, but it feels weird that there's no air control after jumping. Also the eye-velocity makes it seem floatier than it should be.

ikabod-kee:
Why not apply coyote friction to jumping? That actually might be enough.

IntegratedQuantum:
> Also the eye-velocity makes it seem floatier than it should be.

As explained in the post, that's because we don't have directional friction yet, so it also applies friction downwards, making it floaty when the fall starts. This has nothing to do with the eye velocity.

> Why not apply coyote friction to jumping? That actually might be enough.

I don't think we need this, unlike dropping off a ledge where you have less control when the drop happens, on jumping you do control when you hit the button, so you can already control the direction and speed before the jump happens.

ikabod-kee:
> I don't think we need this, unlike dropping off a ledge where you have less control when the drop happens, on jumping you do control when you hit the button, so you can already control the direction and speed before the jump happens.

Don't knock it till you try it :P

IntegratedQuantum:
What I'm saying is you already have the same amount of control, you just have to press the movement keys before you jump, not after.

ikabod-kee:
Would you be able to double the friction coyote time, but half the friction? I feel like that'd be a little more gooder.

ikabod-kee:
Actually, do you remember #581? I think in combination with this PR would make it feel more responsive (aka prevent the floatiness when walking off blocks)

Issue #2491, opened 2026-01-14
Make a callback that is checked before placing to see if a block can be placed.


IntegratedQuantum:
requires #2489

PR #2983, opened 2026-04-26
Allow blocks to define that they are only selectable when item can be placed on it

DellieDelta:
To remove the `.fluid` case in item selection, preparation for #627

Adds the block field `.placementTags` and the block selectionRule `.placeable`

It's a simple version for now (selection happens if a placementTag overlaps with a block tag), but it can be changed in the future to some kind of condition checking. Like replacing it with `.placementConstraints` (like tagOverlap, onTop, onBottom), which can work similarly to `.selectionRule`. To get closer to fixing #2491 

I'll also make it possible separately to define multiple selectionRules per block in #2987, to make both the selectionRules `.toolEffective` and `.placeable` possible on the same block, in preparation for adding buckets

### DIFF
--- a/assets/cubyz/blocks/duckweed.zig.zon
+++ b/assets/cubyz/blocks/duckweed.zig.zon
@@ -1,5 +1,6 @@
 .{
 	.tags = .{.cuttable},
+	.placementTags = .{.fluid},
 	.blockHealth = 0.1,
 	.drops = .{
 		.{.items = .{.auto}},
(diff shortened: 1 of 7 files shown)


DellieDelta:
Closing because I don't want to add the field .placementTags that'll only be temporary anyway, and because fitting this into the existing tag system leads to other problems. Sooo this can be revisited when there's some kind of better placement constraint checking

PR #2985, opened 2026-04-27
Link against musl when releasing Linux binaries

### LINKED ISSUE #2984: Statically link release artifacts
Currently, the builds released by the `Release` action are built against glibc, which is statically linked by default.  
This means that these binaries cannot be used on systems like *NixOS*, which do not support running dynamically-linked binaries.

My proposal is to build against `musl`, which is fully statically-linked, resulting in a binary that works on all systems (even systems that use glibc).

indium114:
Closes #2984 

Links against `musl` libc for Linux binaries, meaning that it can run on systems like NixOS that don't support dynamically-linked binaries.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,10 +21,10 @@ jobs:
         with:
           version: ${{ env.VERSION }}
 
-      - run: zig build -Dtarget=x86_64-linux-gnu.2.27 -Doptimize=ReleaseFast -Drelease=true -Dcpu=baseline -Dversion="$tag"
+      - run: zig build -Dtarget=x86_64-linux-musl -Doptimize=ReleaseFast -Drelease=true -Dcpu=baseline -Dversion="$tag"
       - run: tar -czf Linux-x86_64.tar.gz --transform 's,^,Cubyz/,' assets/cubyz launchConfig.zon -C zig-out/bin Cubyz
 
-      - run: zig build -Dtarget=aarch64-linux-gnu.2.27 -Doptimize=ReleaseFast -Drelease=true -Dcpu=baseline -Dversion="$tag"
+      - run: zig build -Dtarget=aarch64-linux-musl -Doptimize=ReleaseFast -Drelease=true -Dcpu=baseline -Dversion="$tag"
       - run: tar -czf Linux-aarch64.tar.gz --transform 's,^,Cubyz/,' assets/cubyz launchConfig.zon -C zig-out/bin Cubyz
 
       - run: zig build -Dtarget=x86_64-windows-gnu -Doptimize=ReleaseFast -Drelease=true -Dcpu=baseline -Dversion="$tag"


Argmaster:
I don't think it we can just "link against musl when releasing Linux binaries".
Building binary this way makes it impossible to run it on `Ubuntu 24.04` (with OS in state mostly out of the box, so musl is not installed)

```
 ..../debug_linux.sh -Dtarget=x86_64-linux-musl
Detecting Zig compiler...
Zig compiler is valid.
Building Zig Cubyz (-Dtarget=x86_64-linux-musl) from source. This may take a few minutes...
Cubyz successfully built!
Launching Cubyz.
Starting game with version 0.3.0-dev
GLFW Error(65544): X11: Failed to load Xlib
thread 23427 panic: Failed to initialize GLFW
/home/argmaster/dev/cubyz1/src/graphics/Window.zig:723:3: 0x1dc4777 in init (main.zig)
                @panic("Failed to initialize GLFW");
  ^
/home/argmaster/dev/cubyz1/src/main.zig:467:28: 0x1404b32 in main (main.zig)
        if (!headless) Window.init();
                           ^
/home/argmaster/dev/cubyz1/compiler/zig/lib/std/start.zig:699:88: 0x1404463 in callMain (std.zig)
                                                                                       ^
compiler/zig/lib/libc/musl/src/env/__libc_start_main.c:95:7: 0x242c319 in libc_start_main_stage2 (compiler/zig/lib/libc/musl/src/env/__libc_start_main.c)
        exit(main(argc, argv, envp));
      ^
Unwind error at address `/proc/self/exe:0x242c319` (unwind info unavailable), remaining frames may be incorrect
./debug_linux.sh: line 31: 23427 Aborted                 (core dumped) ./zig-out/bin/Cubyz
```

We would probably need to rebuild our deps with musl (?) but then how will eg. nvidia drivers behave? No idea how incompatibility looks here, but a tiny bit of reddit surfing suggested that it might be an issue.

indium114:
> I don't think it we can just "link against musl when releasing Linux binaries". Building binary this way makes it impossible to run it on `Ubuntu 24.04` (with OS in state mostly out of the box, so musl is not installed)
> 
> ```
>  ..../debug_linux.sh -Dtarget=x86_64-linux-musl
> Detecting Zig compiler...
> Zig compiler is valid.
> Building Zig Cubyz (-Dtarget=x86_64-linux-musl) from source. This may take a few minutes...
> Cubyz successfully built!
> Launching Cubyz.
> Starting game with version 0.3.0-dev
> GLFW Error(65544): X11: Failed to load Xlib
> thread 23427 panic: Failed to initialize GLFW
> /home/argmaster/dev/cubyz1/src/graphics/Window.zig:723:3: 0x1dc4777 in init (main.zig)
>                 @panic("Failed to initialize GLFW");
>   ^
> /home/argmaster/dev/cubyz1/src/main.zig:467:28: 0x1404b32 in main (main.zig)
>         if (!headless) Window.init();
>                            ^
> /home/argmaster/dev/cubyz1/compiler/zig/lib/std/start.zig:699:88: 0x1404463 in callMain (std.zig)
>     if (fn_info.params[0].type.? == std.process.Init.Minimal) return wrapMain(root.main(.{
>                                                                                        ^
> compiler/zig/lib/libc/musl/src/env/__libc_start_main.c:95:7: 0x242c319 in libc_start_main_stage2 (compiler/zig/lib/libc/musl/src/env/__libc_start_main.c)
>         exit(main(argc, argv, envp));
>       ^
> Unwind error at address `/proc/self/exe:0x242c319` (unwind info unavailable), remaining frames may be incorrect
> ./debug_linux.sh: line 31: 23427 Aborted                 (core dumped) ./zig-out/bin/Cubyz
> ```
> 
> We would probably need to rebuild our deps with musl (?) but then how will eg. nvidia drivers behave? No idea how incompatibility looks here, but a tiny bit of reddit surfing suggested that it might be an issue.

Tried building without any libc argument and with static linking in build.zig, same issue.

i'll close this until I can think of a better solution, I'm not really a Zig developer

PR #2987, opened 2026-04-27
Make it possible to define multiple selectionRules for a block

DellieDelta:
Related: #2983

To make both the selectionRules `.toolEffective` and `.placeable` possible on the same block, in preparation for adding buckets (#627)

Also rename it to selectionCapabilities, because it's clearer that any being true will lead to the block being selected

### DIFF
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -76,18 +76,9 @@ const SelectionCapability = enum {
 	pub fn loadSelectionCapabilitiesFromZon(_allocator: main.heap.NeverFailingAllocator, zon: main.ZonElement) []SelectionCapability {
 		var capabilities = main.List(SelectionCapability).initCapacity(_allocator, zon.toSlice().len);
 		for (zon.toSlice()) |capabilityZon| {
-			switch (capabilityZon) {
-				.string, .stringOwned => |string| {
-					if (std.meta.stringToEnum(SelectionCapability, string)) |capability| {
-						capabilities.appendAssumeCapacity(capability);
-					} else {
-						std.log.err("SelectionCapability '{s}' is invalid. Ignoring", .{string});
-					}
-				},
-				else => {
-					std.log.err("SelectionCapability element is not a string (type: {s}). Ignoring", .{@tagName(capabilityZon)});
-				},
-			}
+			if (capabilityZon.as(?SelectionCapability, null)) |capability| {
+				capabilities.appendAssumeCapacity(capability);
+			} else std.log.err("SelectionCapability is invalid. Ignoring", .{});
 		}
 		return capabilities.toOwnedSlice();
 	}
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -68,10 +68,8 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-const SelectionCapability = enum {
-	always,
+const SelectionCapability = enum(u8) {
 	toolEffective,
-	never,
 
 	pub fn loadSelectionCapabilitiesFromZon(_allocator: main.heap.NeverFailingAllocator, zon: main.ZonElement) []SelectionCapability {
 		var capabilities = main.List(SelectionCapability).initCapacity(_allocator, zon.toSlice().len);
@@ -85,9 +83,7 @@ const SelectionCapability = enum {
 
 	pub fn allowsItemSelection(self: SelectionCapability, item: Item, block: Block) bool {
 		return switch (self) {
-			.always => true,
 			.toolEffective => item == .proceduralItem and item.proceduralItem.isEffectiveOn(block),
-			.never => false,
 		};
 	}
 };
@@ -166,9 +162,6 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
 	if (zon.getChildOrNull("selectionCapabilities")) |capabilitiesZon| {
 		const capabilities = SelectionCapability.loadSelectionCapabilitiesFromZon(main.stackAllocator, capabilitiesZon);
 		defer main.stackAllocator.free(capabilities);
-		if (capabilities.len == 0) {
-			std.log.err("Field '.selectionCapabilities' is an empty array. This block can never be selected. Did you mean '.selectionCapabilities = .{{.never}}' instead?", .{});
-		}
 		selectionCapabilities = main.worldArena.allocator.dupe(SelectionCapability, capabilities) catch unreachable;
 	}
 	_selectionCapabilities[size] = selectionCapabilities;
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -68,7 +68,7 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-pub const SelectionCapability = enum {
+const SelectionCapability = enum {
 	always,
 	toolEffective,
 	never,
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -69,17 +69,36 @@ pub const Ore = struct {
 };
 
 pub const SelectionCapability = enum {
-    always,
-    toolEffective,
-    never,
+	always,
+	toolEffective,
+	never,
 
-    pub fn allowsItemSelection(self: SelectionCapability, item: Item, block: Block) bool {
-        return switch (self) {
-            .always => true,
-            .toolEffective => item == .proceduralItem and item.proceduralItem.isEffectiveOn(block),
-            .never => false,
-        };
-    }
+	pub fn loadSelectionCapabilitiesFromZon(_allocator: main.heap.NeverFailingAllocator, zon: main.ZonElement) []SelectionCapability {
+		var capabilities = main.List(SelectionCapability).initCapacity(_allocator, zon.toSlice().len);
+		for (zon.toSlice()) |capabilityZon| {
+			switch (capabilityZon) {
+				.string, .stringOwned => |string| {
+					if (std.meta.stringToEnum(SelectionCapability, string)) |capability| {
+						capabilities.appendAssumeCapacity(capability);
+					} else {
+						std.log.err("SelectionCapability '{s}' is invalid. Ignoring", .{string});
+					}
+				},
+				else => {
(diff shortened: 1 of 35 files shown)


src/blocks.zig
```diff
@@ -68,7 +68,38 @@ pub const Ore = struct {
...
+				.string, .stringOwned => |string| {
+					if (std.meta.stringToEnum(SelectionCapability, string)) |capability| {
+						capabilities.appendAssumeCapacity(capability);
+					} else {
+						std.log.err("SelectionCapability '{s}' is invalid. Ignoring", .{string});
+					}
+				},
+				else => {
+					std.log.err("SelectionCapability element is not a string (type: {s}). Ignoring", .{@tagName(capabilityZon)});
+				},
```
Wunka: you don't need this complicated logic here. Most of the work was already done for you. Look at the `as(` method of the zon. You can even do: `.as(?SelectionCapability, null) orelse {//Handle not right value...`

Wunka (changes requested):
Maybe never can just be the empty list? 
But you don't have to follow up on that.

DellieDelta:
> Maybe never can just be the empty list? But you don't have to follow up on that.

Good idea! Then the whole check can go. It's a valid state anyway

src/blocks.zig
```diff
@@ -139,7 +157,15 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
...
 	_degradable[size] = zon.get(bool, "degradable", false);
-	_selectionRule[size] = zon.get(SelectionRule, "selectionRule", .always);
+
+	var selectionCapabilities: ?[]SelectionCapability = null;
+	if (zon.getChildOrNull("selectionCapabilities")) |capabilitiesZon| {
+		const capabilities = SelectionCapability.loadSelectionCapabilitiesFromZon(main.stackAllocator, capabilitiesZon);
+		defer main.stackAllocator.free(capabilities);
+		selectionCapabilities = main.worldArena.allocator.dupe(SelectionCapability, capabilities) catch unreachable;
+	}
+	_selectionCapabilities[size] = selectionCapabilities;
```
Wunka: why are you duping here? just pass the `main.worldArena` for the allocator

Wunka (changes requested):

src/blocks.zig
DellieDelta: Ahh good one, I copied it from other places in `blocks.zig`. But looks like they can also use worldArena directly, so I'll make a PR for that

src/blocks.zig
```diff
@@ -68,7 +68,25 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-const SelectionRule = enum { always, toolEffective, never };
+const SelectionCapability = enum(u8) {
```
Wunka: you don't need a backing type here

```suggestion
const SelectionCapability = enum {
```

Wunka (changes requested):

src/blocks.zig
DellieDelta: Zig makes enums that have only one value 0 bytes, and the allocator gives an error when trying to make an array with 0 byte elements, that's why I have be explicit about it

Wunka: ah ok. then its ok

Wunka (approved):

src/blocks.zig
```diff
@@ -79,7 +97,7 @@ var _blockResistance: [maxBlockCount]f32 = undefined;
 
 /// Whether you can replace it with another block, mainly used for fluids/gases
 var _replaceable: [maxBlockCount]bool = undefined;
-var _selectionRule: [maxBlockCount]SelectionRule = undefined;
+var _selectionCapabilities: [maxBlockCount]?[]SelectionCapability = undefined;
```
IntegratedQuantum:
```suggestion
var _selectionCapabilities: [maxBlockCount]?[]const SelectionCapability = undefined;
```

src/blocks.zig
```diff
@@ -139,7 +157,13 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
 	_light[size] = zon.get(u32, "emittedLight", 0);
 	_absorption[size] = zon.get(u32, "absorbedLight", 0xffffff);
 	_degradable[size] = zon.get(bool, "degradable", false);
-	_selectionRule[size] = zon.get(SelectionRule, "selectionRule", .always);
+
+	var selectionCapabilities: ?[]SelectionCapability = null;
```
IntegratedQuantum: Instead of being nullable, I'd prefer a `.all` capability that is the default.
You could put that here as `&.{.all}` without needing to do any allocation.

src/blocks.zig
```diff
@@ -68,7 +68,25 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-const SelectionRule = enum { always, toolEffective, never };
+const SelectionCapability = enum(u8) {
+	toolEffective,
+
+	pub fn loadSelectionCapabilitiesFromZon(_allocator: main.heap.NeverFailingAllocator, zon: main.ZonElement) []SelectionCapability {
+		var capabilities = main.List(SelectionCapability).initCapacity(_allocator, zon.toSlice().len);
```
IntegratedQuantum: Can be unmanaged
```suggestion
		var capabilities = main.ListUnmanaged(SelectionCapability).initCapacity(_allocator, zon.toSlice().len);
```

src/blocks.zig
```diff
@@ -68,7 +68,25 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-const SelectionRule = enum { always, toolEffective, never };
+const SelectionCapability = enum(u8) {
+	toolEffective,
+
+	pub fn loadSelectionCapabilitiesFromZon(_allocator: main.heap.NeverFailingAllocator, zon: main.ZonElement) []SelectionCapability {
```
IntegratedQuantum: Functions that accept only arena allocators should call the parameter just arena
```suggestion
	pub fn loadSelectionCapabilitiesFromZon(arena: main.heap.NeverFailingAllocator, zon: main.ZonElement) []SelectionCapability {
```

src/blocks.zig
```diff
@@ -68,7 +68,25 @@ pub const Ore = struct {
...
+		var capabilities = main.List(SelectionCapability).initCapacity(_allocator, zon.toSlice().len);
+		for (zon.toSlice()) |capabilityZon| {
+			if (capabilityZon.as(?SelectionCapability, null)) |capability| {
+				capabilities.appendAssumeCapacity(capability);
+			} else std.log.err("SelectionCapability is invalid. Ignoring", .{});
+		}
+		return capabilities.toOwnedSlice();
```
IntegratedQuantum: Since this is using an arena, toOwnedSlice is unnecessary
```suggestion
		return capabilities.items;
```

IntegratedQuantum (changes requested):

src/blocks.zig
DellieDelta: Is it to make it clearer that the default case is `.all`? Because I personally prefer only making it `.all` when unspecified. Otherwise you can also add `.all` to the config, which is redundant

IntegratedQuantum: Right now it is not immediately obvious that `null` means everything selects it. Having a `all` SelectionCapability would avoid the use of optional.

DellieDelta: What do you think of this?

```
const alwaysSelectable = null;
var selectionCapabilities: ?[]SelectionCapability = alwaysSelectable;
```

IntegratedQuantum: In my opinion the problem is also at the callsite, where you do not have this extra constant.
This approach would be fine though, if you had a struct with a struct-local constant and a load and a check function.

DellieDelta: Sounds good! I'll see if I can get something that still looks clean, otherwise I'll just go with the `.any`

src/blocks.zig
```diff
@@ -68,7 +68,46 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-const SelectionRule = enum { always, toolEffective, never };
+const SelectionCapabilities = struct {
+	capabilities: ?[]const SelectionCapability,
+
+	pub const alwaysSelectable = SelectionCapabilities{.capabilities = null};
```
IntegratedQuantum: Construction with T{} is likely going to be removed https://github.com/ziglang/zig/issues/5038, so avoid it if it isn't necessary
```suggestion
	pub const alwaysSelectable: SelectionCapabilities = .{.capabilities = null};
```

src/blocks.zig
```diff
@@ -139,7 +183,13 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
 	_light[size] = zon.get(u32, "emittedLight", 0);
 	_absorption[size] = zon.get(u32, "absorbedLight", 0xffffff);
 	_degradable[size] = zon.get(bool, "degradable", false);
-	_selectionRule[size] = zon.get(SelectionRule, "selectionRule", .always);
+
+	if (zon.getChildOrNull("selectionCapabilities")) |capabilitiesZon| {
+		_selectionCapabilities[size] = SelectionCapabilities.loadFromZon(main.worldArena, capabilitiesZon);
+	} else {
+		_selectionCapabilities[size] = SelectionCapabilities.alwaysSelectable;
```
IntegratedQuantum: decl literals can be used here
```suggestion
		_selectionCapabilities[size] = .loadFromZon(main.worldArena, capabilitiesZon);
	} else {
		_selectionCapabilities[size] = .alwaysSelectable;
```

src/blocks.zig
```diff
@@ -68,7 +68,51 @@ pub const Ore = struct {
...
+
+		const capabilities = self.capabilities orelse return true;
+		for (capabilities) |capability| {
+			if (capability.allowsItemSelection(item, block)) return true;
+		}
+		return false;
+	}
+};
+
+const SelectionCapability = enum(u8) {
```
IntegratedQuantum: Since this is only used inside of SelectionCapabilities, I think it should be declared inside there, which would also allow reducing the redundant information in the struct name.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #2988, opened 2026-04-27
Headless server: bootstrap world if missing and handle SIGTERM/SIGINT

teemow:
Make `headlessServer=true` actually self-sufficient: today the binary panics if the world named in `autoEnterWorld` doesn't exist on disk, because world creation only happens through the GUI's `tryCreateWorld()`. That makes it impossible to ship the server in an immutable container without manually pre-seeding `saves/<name>/`.

This adds three optional `launchConfig` fields and a small bootstrap branch in `main()`:

- `createIfMissing` (default `false`) — when `true` *and* `headlessServer` is set, the world named in `autoEnterWorld` is created via `tryCreateWorld()` if it doesn't already exist.
- `worldSeed: ?u64` — optional seed for the auto-created world; random if unset.
- `worldPreset: []const u8` — preset name, defaults to `cubyz:default`.

Existing saves are never touched (the new code path is gated on `files.cubyzDir().hasDir("saves/<name>")` returning `false`). The `launchConfig.zon` template generated by `build.zig` documents the new keys.

The same change also installs POSIX `SIGTERM`/`SIGINT` handlers (only when running headless, only on POSIX) that call `server.stop()`, so orchestrators (systemd, Kubernetes, ...) can roll the process cleanly and let chunks flush during the configured grace period. Desktop client behaviour is unchanged.

Motivated by packaging Cubyz as a containerised dedicated server; happy to drop or rework parts of the API if you'd prefer a different shape.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -157,6 +157,9 @@ fn createLaunchConfig(b: *std.Build) !void {
 			\\    .autoEnterWorld = "",
 			\\    .headlessServer = false,
 			\\    // .preferredAuthenticationAlgorithm = .ed25519, // Uncomment and change this if you own a server in an outdated game version where the default algorithm got compromised.
+			\\    // .createIfMissing = false, // When true (and headlessServer is true), the world named in autoEnterWorld is created if it doesn't exist yet.
+			\\    // .worldSeed = 0, // Seed used when auto-creating a world. If unset, a random seed is generated.
+			\\    // .worldPreset = "cubyz:default", // World preset used when auto-creating a world.
 			\\}
 		;
 		try std.Io.Dir.cwd().writeFile(io.io(), .{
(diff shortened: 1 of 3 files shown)


IntegratedQuantum (changes requested):
First up, this PR does two completely separate and unrelated things, in the future please make one PR for each thing you do, [see also](https://github.com/PixelGuys/Cubyz/blob/master/docs/CONTRIBUTING.md#dont-put-multiple-changes-into-one-pull-request)

Secondly please search first if a PR addressing one of your concerns already exists, otherwise you are doing duplicate work, @BoySanic already worked on world creation in headless servers: #2251

As for the signal handler: I think this would also make a lot of sense for normal clients as well. Also did you check if Windows has something similar for this, and if there is a more general abstraction in the Zig standard library?

PR #2990, opened 2026-04-28
Make it possible to specify that a block should be destroyed/dropped when replaced by another block

DellieDelta:
Progress towards #1558

It's a little bit of a workaround because we don't have something like BreakBlock or ReplaceBlock yet instead of UpdateBlock, but it should be easy enough to pull apart when the time comes

Most relevant changes are in the first commit :)

### DIFF
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -70,6 +70,21 @@ pub const Ore = struct {
 
 const SelectionRule = enum { always, toolEffective, never };
 
+/// How this block behaves when trying to place a block in the same position
+const Replaceability = enum {
+	none,
+	destroy,
+	drop,
+
+	pub fn isReplaceable(self: Replaceability) bool {
+		return self != .none;
+	}
+
+	pub fn dropsOnReplace(self: Replaceability) bool {
+		return self == .drop;
+	}
+};
+
 var _transparent: [maxBlockCount]bool = undefined;
 var _collide: [maxBlockCount]bool = undefined;
 var _id: [maxBlockCount][]u8 = undefined;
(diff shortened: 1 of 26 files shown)


src/blocks.zig
```diff
@@ -70,15 +70,29 @@ pub const Ore = struct {
 
 const SelectionRule = enum { always, toolEffective, never };
 
+/// How this block behaves when trying to place a block in the same position
+const Replaceability = enum {
+	none,
+	destroy,
+	drop,
```
Argmaster, line 77: We might as well

```suggestion
const Replaceability = enum {
	no,
    yes,
	yesWithDrop,
```

src/blocks.zig
```diff
@@ -140,7 +154,7 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
 	_absorption[size] = zon.get(u32, "absorbedLight", 0xffffff);
 	_degradable[size] = zon.get(bool, "degradable", false);
 	_selectionRule[size] = zon.get(SelectionRule, "selectionRule", .always);
-	_replaceable[size] = zon.get(bool, "replaceable", false);
+	_replaceability[size] = zon.get(Replaceability, "replaceability", .none);
```
Argmaster, line 157: Its a property, it should be adjective, not a noun

src/blocks.zig
DellieDelta: In my opinion the original names are clearer and more extensible. For example if you want to add a fourth mode in the future that does something very different

src/blocks.zig
DellieDelta: I see it more like a mode of replacement, then a noun makes more sense I think

src/blocks.zig
Argmaster: 1. Future proofing doesn't work. A simple example of extension would be to add support for different drop than usually is dropped when block is broken. That would require you to use union instead of enum, a major refactor either way.
2. Similar pattern was already established in `rotation.RotationMode.CanBeChangedInto`

Argmaster:
So, overall I am not sure if this is flexible enough. What if instead we would support multiple `onUpdate` callbacks? 
We will have to go somewhat this route anyway.

Also I see either `vine_decay` (by me) or other callbacks violate naming convention... Sorry for that

src/blocks.zig
DellieDelta: Yeahhh, you're right. Next to yes or no I don't see it getting a "maybe" loll, unless we go to a more complex structure anyway. That makes it more like a property with subproperties kind of, but with enums you can just make it flat

DellieDelta:
Interestingg, you'd almost need like an onReplace callback where you can say that it should drop an item, or at least some way to do that within onUpdate. It might be a bit overkill for now, but on the other hand you don't want to introduce a configuration like yesWithDrop only to remove it later again anyway

Issue #2572, opened 2026-02-19
Add a flake.nix


SamsonLib:
Add a flake.nix for easy dependency mangagement on linux (nixos). I propose this one:

```
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    flake-utils.url = "github:numtide/flake-utils";
  };

  outputs =
    {
      self,
      nixpkgs,
      flake-utils,
    }:
    flake-utils.lib.eachDefaultSystem (
      system:
      let
        pkgs = import nixpkgs { inherit system; };
        deps = with pkgs; [
          zig
          bash
          libx11
          libxrandr
          libxi
          libxinerama
          libxcursor
          mesa
          libGL
          glfw
        ];
      in
      {
        devShells.default = pkgs.mkShell {
          name = "Cubyz";
          packages = deps;
          shellHook = ''
            export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:${builtins.toString (pkgs.lib.makeLibraryPath deps)}";
          '';
        };
      }
    );
}
```

SamsonLib:
For this to work the shebangs need to be changed from
```#!/bin/bash``` to 
```#!/usr/bin/env bash```

IntegratedQuantum:
Not sure how this works, but the dependencies look wrong. The game statically links a self-built glfw, the game doesn't work with system installed zig.
Also note the existing discussions about nix https://github.com/PixelGuys/Cubyz/pull/1967#issuecomment-3481904804

BoySanic:
https://github.com/NixOS/nixpkgs/pull/455683

PR #2991, opened 2026-04-29
Add Nix flake

tenyoru:
Adds a devShell with X11/Vulkan runtime libs so the game runs on NixOS.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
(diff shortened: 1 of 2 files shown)


tenyoru:
Should I update the PR to match the code style?

Wunka:
Please look at the disscusions from before: https://github.com/PixelGuys/Cubyz/issues/2572 and https://github.com/PixelGuys/Cubyz/pull/1967

tenyoru:
Oops, I didn’t check the other PRs. Thanks for the response!

tenyoru:
Added zig and zls as suggested in this comment: https://github.com/PixelGuys/Cubyz/pull/1967#issuecomment-3394386583

Argmaster:
The finally call is for Quantum to make, but:
- we don't have any active contributors that use NixOS, so unless you are going to maintain it (sorry, but we can't assume you will) no one will, so the code will degrade and cause maintenance issues
- as noted in this [comment](https://github.com/NixOS/nixpkgs/pull/455683) it is not strictly necessary to include it in the main repository
- there is already community initiative to do it from nix side https://github.com/NixOS/nixpkgs/pull/455683 imo it would be better to invest into that

IntegratedQuantum:
Generally I don't want to have any artifacts in the code that are specific to your setup, that includes editor config, and also OS-specific configurations. Especially when there seem to be other solutions for it.

tenyoru:
> * we don't have any active contributors that use NixOS, so unless you are going to maintain it (sorry, but we can't assume you will) no one will, so the code will degrade and cause maintenance issues

This is only relevant during development. If a Nix/NixOS user runs into a problem, they can adjust and commit the files when necessary. There’s no need for active maintainers for this, especially since it’s only two files.

If someone wants to contribute to the project, these files simply make it more convenient, so they don’t have to create their own setup from scratch.

PR #2994, opened 2026-04-30
Side agnostic physics

codemob-dev:
Adds a helper function `main.game.getBlockWithSide(comptime side: main.sync.Side, x: i32, y: i32, z: i32) ?Block` and use it to allow the physics functions to be called on the server side as well. At the moment this isn't particularly useful considering the physics functions themselves haven't been entirely generalized yet, but it is a prerequisite for #1663.


IntegratedQuantum (changes requested):
Please fix conflicts with the other physics PR, also in a bit there will be more conflicts due to #3000

codemob-dev:
> Please fix conflicts with the other physics PR, also in a bit there will be more conflicts due to #3000

Done!

IntegratedQuantum (approved):

PR #2995, opened 2026-04-30
Add `ListUnmanaged.print`

Argmaster:
This pull request adds `print` method to `ListUnmanaged` struct.
To avoid copying a leftover TODO from `print` function in `List`, it also makes an attempt on addressing it:

https://github.com/Argmaster/Cubyz/blob/ListUnmanaged.print/src/utils/list.zig#L213

> // TODO: Is there no easier way to make this without an extra copy?

We might as well just leave it for different time tho

Resolves: https://github.com/PixelGuys/Cubyz/pull/1425#discussion_r3132400987

### DIFF
--- a/src/utils/list.zig
+++ b/src/utils/list.zig
@@ -435,7 +435,8 @@ test "ListUnmanaged.print single call" {
 
 	list.print(main.stackAllocator, "foo {}", .{34});
 
-	try std.testing.expectEqualStrings(list.items, "foo 34");
+	try std.testing.expectEqualStrings("foo 34", list.items);
+	try std.testing.expect(list.items.len <= list.capacity);
 }
 
 test "ListUnmanaged.print with a string" {
@@ -444,7 +445,8 @@ test "ListUnmanaged.print with a string" {
 
 	list.print(main.stackAllocator, "foo {s}", .{"bar spam BUZZ"});
 
-	try std.testing.expectEqualStrings(list.items, "foo bar spam BUZZ");
+	try std.testing.expectEqualStrings("foo bar spam BUZZ", list.items);
+	try std.testing.expect(list.items.len <= list.capacity);
 }
 
 test "ListUnmanaged.print multiple writes" {
@@ -467,7 +469,8 @@ test "ListUnmanaged.print multiple writes" {
 		\\
 	;
 
-	try std.testing.expectEqualStrings(list.items, expected);
+	try std.testing.expectEqualStrings(expected, list.items);
+	try std.testing.expect(list.items.len <= list.capacity);
 }
 
 /// Holds multiple arrays sequentially in memory.
--- a/src/utils/list.zig
+++ b/src/utils/list.zig
@@ -421,12 +421,28 @@ pub fn ListUnmanaged(comptime T: type) type {
 	};
 }
 
-test "ListUnmanaged.print single call" {
+test "ListUnmanaged.print single call, buffer not preserved" {
 	var list: ListUnmanaged(u8) = .{};
+	const oldAddress = list.items.ptr;
+	defer list.deinit(main.stackAllocator);
+
+	list.print(main.stackAllocator, "foo {d:.1}", .{34});
+	const newAddress = list.items.ptr;
+
+	try std.testing.expect(oldAddress != newAddress);
+	try std.testing.expectEqualStrings("foo 34", list.items);
+	try std.testing.expect(list.items.len <= list.capacity);
+}
+
+test "ListUnmanaged.print initCapacity, buffer preserved" {
+	var list: ListUnmanaged(u8) = .initCapacity(main.stackAllocator, 6);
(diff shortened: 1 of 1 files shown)


src/utils/list.zig
```diff
@@ -406,9 +406,73 @@ pub fn ListUnmanaged(comptime T: type) type {
...
+		pub fn print(self: *@This(), allocator: NeverFailingAllocator, comptime fmt: []const u8, args: anytype) void {
+			// It seems that concepts of capacity and current size are swapped in Writer compared to our list.
+			// The `writer.buffer` (a slice) contains what we call capacity as its `len`, while the current
+			// use of the buffer is stored by `writer.end` attribute.
+			var buffer = self.items;
+			buffer.len = self.capacity;
+
+			// Unfortinately there is no constructor that would allow us to use partially filled buffer.
+			// We have to just create a buffer from slice and move the end pointer afterwards.
+			var writer = std.Io.Writer.Allocating.initOwnedSlice(allocator.allocator, buffer);
```
IntegratedQuantum: Would it help to use `fromArrayList`/`toArrayList`? Then we'd only need to convert our list to zig's list, which is less cumbersome.

IntegratedQuantum (changes requested):

src/utils/list.zig
Argmaster: Yeah, I guess then we can ignore the internals that needed so much descriptions. I completely ignored that since it was not starting with `init`.

Argmaster: Its so annoying they clean up the list and we cannot just pass a temporary >.<

Argmaster: Done.

src/utils/list.zig
```diff
@@ -406,9 +406,65 @@ pub fn ListUnmanaged(comptime T: type) type {
...
+
+test "ListUnmanaged.print multiple writes" {
+	var list: ListUnmanaged(u8) = .{};
+	defer list.deinit(main.stackAllocator);
+
+	// The tricky part of the implementation is to correctly reassign buffer bounds, so every time
+	// we use the list as print destination, we want to make sure it retains normal list behavior
+	// by inserting a single element.
+	list.append(main.stackAllocator, '\n');
+	list.print(main.stackAllocator, "BarFooSpam {}", .{main.vec.Vec2d{0.3, 0.4}});
```
IntegratedQuantum: The way how more complex objects are printed is possibly fragile, and I don't think we should rely on it in tests.
Plus the dependence on Vec2d (which is going to change once Zig adds glsl-style vectors) is unnecessary.

IntegratedQuantum (changes requested):

src/utils/list.zig
Argmaster: Ok, would a array be fine? I would want something more than just a single value tbh

IntegratedQuantum: Why though, what do you intend to test with a composite type?

Argmaster: Ok, yeah I guess it doesn't make that much sense. I was somewhat aiming to test if realloc works and if weirder things can still be printed properly.

What about I replace it with a test that checks if `.initCapactiy` preserves the `items` pointer?
I would also test if `items` pointer changes when not using `.initCapacity`.

Argmaster:
```diff
@@ -421,12 +421,28 @@ pub fn ListUnmanaged(comptime T: type) type {
        };
 }

+test "ListUnmanaged.print buffer preserved" {
+       var list: ListUnmanaged(u8) = .initCapacity(main.stackAllocator, 20);
+       const oldAddress = list.items.ptr;
+       defer list.deinit(main.stackAllocator);
+
+       list.print(main.stackAllocator, "foo {}", .{34});
+       const newAddress = list.items.ptr;
+
+       try std.testing.expect(oldAddress == newAddress);
+       try std.testing.expectEqualStrings("foo 34", list.items);
+       try std.testing.expect(list.items.len <= list.capacity);
+}
+
 test "ListUnmanaged.print single call" {
        var list: ListUnmanaged(u8) = .{};
+       const oldAddress = list.items.ptr;
        defer list.deinit(main.stackAllocator);

-       list.print(main.stackAllocator, "foo {}", .{34});
+       list.print(main.stackAllocator, "foo {d:.1}", .{34});
+       const newAddress = list.items.ptr;

+       try std.testing.expect(oldAddress != newAddress);
        try std.testing.expectEqualStrings("foo 34", list.items);
        try std.testing.expect(list.items.len <= list.capacity);
 }
```

Argmaster:
> /home/runner/work/Cubyz/Cubyz/build.zig.zon:41:11: error: bad HTTP response code: '502 Bad Gateway'
   .url = "https://github.com/PixelGuys/Cubyz-Libs/releases/download/11/cubyz_deps_x86_64-windows-gnu.tar.gz",
   
   <3

IntegratedQuantum (approved):

PR #3004, opened 2026-05-01
Rename callbacks to use snake_case naming convention

### LINKED ISSUE #2993: `vine_decay.zig` violates naming convention
Image

Or other callbacks do. Either way it has to be fixed

Argmaster:
Resolves: #2993


IntegratedQuantum:
Hmm, I'm not 100% sure here.
Technically they are files as a struct, which would mean they'd need to be `VineDecay`, but I really don't like this at all, all to be modded IDs should be lowercase.

On the other hand making an exception here, would make it more difficult to automatically check the formatting later.
Should we maybe remove the file as struct behavior from these files? It is pretty convenient, but maybe it wouldn't be too bad to drop it.

Argmaster:
I don't care and adding a path based behavior change is trivial (as long as the checks are done in some convenient language like Python)
This is pretty much an artificial problem.

IntegratedQuantum (approved):

Issue #2603, opened 2026-02-25
Tool handle progression


BoySanic:
### Description

It was pointed out in the discord that if you can get some glimmergill, its durable parameter is greater than any wood type on the surface.

I think there should be more progression beyond glimmergill. Maybe some more plant materials either below the surface or in the sky islands or in rare biomes on the surface that can be used to create more durable tools than either wood or glimmergill allow for currently.

IntegratedQuantum:
Yes, I would like to have some better wood progressions. Glimmergills was really just a decision of "Oh, what kind of stats should I give mushrooms?"

calebstenberg:
Should wood progression also include some woods that are harder to break?
Maybe we separate hard and soft wood for some slight variation in block health and material strength, and also have some more special wood types that you aren't meant to break with your hand and have some more varied effects?

PR #3005, opened 2026-05-02
Procedural Tools | Slots can have Tags now

Crepestrom:
Defined along with a procedural tool you can define which tags a slot has

you can define a tool to have multiple tags in one slot
you define which slots each tag has in it

image

also adds a new modifier restriction (currently only used for branches)
on_top_of
if the thing is on top of a slot with a tag .head branches will have decreased durability

new modifier
image

modifier being applied

image


Crepestrom:
related issues #2603 #89

src/items.zig
```diff
@@ -1287,6 +1311,15 @@ pub fn registerProceduralItem(assetFolder: []const u8, id: []const u8, zon: ZonE
...
 		slotInfos[i].optional = zonDisabled.as(usize, 0) != 0;
 	}
+	var tagfields: main.List(TagField) = .init(main.worldArena);
+	for (zon.getChild("tagfields").toSlice()) |paramZon| {
+		const tagFieldVal = tagfields.addOne();
+		const matrixZon = paramZon.getChild("matrix");
+		for (0..25) |i| {
+			tagFieldVal.hasTagMatrix[i] = matrixZon.getAtIndex(bool, i, false);
+		}
+		tagFieldVal.tagType = main.Tag.find(paramZon.get([]const u8, "tag", "not specified"));
```
Wunka: this should probably at least  log a warning or even an error. if someone specifies a tagfield matrix it should also have a tag.

src/proceduralItem/modifiers/restrictions/encased.zig
```diff
@@ -15,7 +15,7 @@ pub fn satisfied(self: *const Encased, proceduralItem: *const ProceduralItem, x:
 	var count: usize = 0;
 	for ([_]i32{-1, 0, 1}) |dx| {
 		for ([_]i32{-1, 0, 1}) |dy| {
-			if ((proceduralItem.getItemAt(x + dx, y + dy) orelse continue).hasTag(self.tag)) count += 1;
+			if (proceduralItem.checkForTagAt(x + dx, y + dy, self.tag) orelse continue) count += 1;
```
Wunka: this changes the logic of encased. I think this was an unintended change

src/items.zig
```diff
@@ -839,6 +848,21 @@ pub const ProceduralItem = struct { // MARK: ProceduralItem
 		return self.craftingGrid[@intCast(x + y*5)];
 	}
 
+	pub fn checkForTagAt(self: *const ProceduralItem, x: i32, y: i32, checkedTag: main.Tag) ?bool {
```
Wunka: why does this return an `?bool` ? why not just a `bool`? if its outside the x, y range it should be `false`

src/items.zig
```diff
@@ -839,6 +848,21 @@ pub const ProceduralItem = struct { // MARK: ProceduralItem
...
+		if (x < 0 or x >= 5) return null;
+		if (y < 0 or y >= 5) return null;
+
+		for (self.type.tagfields()) |specificTagField| {
+			if (specificTagField.hasTagMatrix[@intCast(x + y*5)] == true) {
+				if (specificTagField.tagType == checkedTag) {
+					return true;
+				}
+			}
```
Wunka:
```suggestion
			if (specificTagField.tagType != checkedTag) continue;
			return specificTagField.hasTagMatrix[@intCast(x + y*5)];
```
while not changing much, early returns are mostly better.

src/items.zig
```diff
@@ -839,6 +848,21 @@ pub const ProceduralItem = struct { // MARK: ProceduralItem
...
+		}
+		if ((self.getItemAt(x, y) orelse return false).hasTag(checkedTag)) return true;
+		return false;
```
Wunka, line 870: mhhh. this makes the the encases logic make sense. But I would say this change in general doens't make sense. 
I would say rename the function to something like `checkForSlotTagAt` to show that it only cares about the slot itself and not the item. 

If you still think this needs to be changed please do it in a seperate pr (I don't see how your other changes are depended on it) your pr is already pretty big

Wunka (changes requested):
As you have not linked it: https://discord.com/channels/443805812390100992/1103565202093506620/1499921286820921495
(where the idea came from)

this is an intersting change, but could not be the only way to do it...
Please create a corresponding issue with an explanation what the use case is.

Argmaster:
Heh we are coming back to where we came from.
"The Tags, or There and Back Again"

src/proceduralItem/modifiers/restrictions/encased.zig
Crepestrom: intended change

src/items.zig
Crepestrom: it cares about both the slot and item

src/items.zig
Crepestrom: i included a error now

IntegratedQuantum:
Is there no better way to specify these? A boolean matrix seems kind of clunky and error prone.

Crepestrom:
sure i transitioned it into the format the optional/disabled has

IntegratedQuantum:
Now that I think about it, optional and disable are also kind of tags, maybe it should all be defined in the same way?

Could you also go with the more compact `.xyz = .{matrix}` instead of `.{.name = .xyz, .matrix = .{matrix}}`

Crepestrom:
i will see if i can do that

Crepestrom:
though optional and disable being seperate can be fine since those define how the workbench is layed out while the tags only interact with properties

IntegratedQuantum:
Oh I absolutely want materials that can only be placed on fields with specific tags, which is pretty similar to what optional does.

Crepestrom:
i wont implement materials only being placed on certain tags yet but i see what you mean

Crepestrom:
i was not able to compress the form of the ZonObjects as you specified
and optional and disabled are outside the scope of this pr

Crepestrom:
if i were to modify disabled and optinal i would have to make the pr much bigger touching the workbench and inventory

IntegratedQuantum:
> i was not able to compress the form of the ZonObjects as you specified

It should be easy to do, you just need to iterate over the internal hashmap

src/proceduralItem/modifiers/restrictions/on_top_of.zig
```diff
@@ -0,0 +1,34 @@
...
+	result.* = .{
+		.tag = main.Tag.find(zon.get(?[]const u8, "tag", null) orelse blk: {
+			std.log.err("Missing tag field for encased restriction.", .{});
+			break :blk "not specified";
+		}),
+	};
+	return result;
+}
+
+pub fn printTooltip(self: *const OnTopOf, outString: *main.List(u8)) void {
```
Wunka:
```suggestion
pub fn printTooltip(self: *const OnTopOf, outString: *main.ListManaged(u8)) void {
```
we switched the default List in #3129 and #3152 . Thats where your error is coming from

Wunka (changes requested):

src/proceduralItem/modifiers/restrictions/on_top_of.zig
```diff
@@ -0,0 +1,34 @@
...
+		.tag = main.Tag.find(zon.get([]const u8, "tag") orelse blk: {
+			std.log.err("Missing tag field for encased restriction.", .{});
+			break :blk "not specified";
+		}),
+	};
+	return result;
+}
+
+pub fn printTooltip(self: *const OnTopOf, outString: *main.ListManaged(u8)) void {
+	outString.print("on top of .{s}", .{self.tag.getName()});
```
IntegratedQuantum, line 33: on tops of? I didn't know you can put multiple items into one slot.
Please chooses a more descriptive name here.

src/proceduralItem/modifiers/restrictions/on_top_of.zig
```diff
@@ -0,0 +1,34 @@
...
+	isOnTopOfTag = proceduralItem.checkForTagAt(x, y, self.tag);
+
+	return isOnTopOfTag;
+}
+
+pub fn loadFromZon(allocator: NeverFailingAllocator, zon: ZonElement) *const OnTopOf {
+	const result = allocator.create(OnTopOf);
+	result.* = .{
+		.tag = main.Tag.find(zon.get([]const u8, "tag") orelse blk: {
+			std.log.err("Missing tag field for encased restriction.", .{});
```
IntegratedQuantum: This is not the encased restriction

assets/cubyz/blocks/branch/birch.zig.zon
```diff
@@ -28,6 +28,14 @@
 					.id = "durable",
 					.strength = 0.45,
 				},
+				.{
+					.id = "fragile",
+					.strength = 0.1,
+					.restriction = .{
+						.id = .on_top_of,
+						.tag = .head,
```
IntegratedQuantum: I like the idea, but in my opinion it would be better to go the opposite route: Make the durable only available when in a handle slot.

src/items.zig
```diff
@@ -849,6 +858,18 @@ pub const ProceduralItem = struct { // MARK: ProceduralItem
 		return self.craftingGrid[@intCast(x + y*5)];
 	}
 
+	pub fn checkForTagAt(self: *const ProceduralItem, x: i32, y: i32, checkedTag: main.Tag) bool {
```
```suggestion
	pub fn hasMaterialOrSlotTagAt(self: *const ProceduralItem, x: i32, y: i32, checkedTag: main.Tag) bool {
```

src/items.zig
```diff
@@ -634,6 +642,7 @@ pub const ProceduralItemType = struct { // MARK: ProceduralItemType
 	tags: []main.Tag,
 	properties: []PropertyMatrix,
 	slotInfos: [25]SlotInfo,
```
IntegratedQuantum, line 644: Please remove this, they should now be accessed through tags

src/proceduralItem/modifiers/restrictions/encased.zig
```diff
@@ -15,7 +15,7 @@ pub fn satisfied(self: *const Encased, proceduralItem: *const ProceduralItem, x:
 	var count: usize = 0;
 	for ([_]i32{-1, 0, 1}) |dx| {
 		for ([_]i32{-1, 0, 1}) |dy| {
-			if ((proceduralItem.getItemAt(x + dx, y + dy) orelse continue).hasTag(self.tag)) count += 1;
+			if (proceduralItem.checkForTagAt(x + dx, y + dy, self.tag)) count += 1;
```
IntegratedQuantum: Since you combined them, I'd suggest a naming convention that distinguishes slot tags from item tags.
e.g. we did this for cave layers which have to end with `_layer`.

We could do the same thing by mandating that `slot` tags end with `_slot`.
Then it would be much easier for the player to see that something is a slot tag vs an item tag.

src/items.zig
```diff
@@ -634,6 +642,7 @@ pub const ProceduralItemType = struct { // MARK: ProceduralItemType
...
+	tagfields: []TagField,
```
```suggestion
	slotTags: []TagField,
```

src/items.zig
```diff
@@ -1318,19 +1339,30 @@ fn loadPixelSources(assetFolder: []const u8, id: []const u8, layerPostfix: []con
...
-		if (i >= 25) {
-			std.log.err("disabled array of {s} has too many entries", .{id});
-			break;
+	var tagfields: main.List(TagField) = .empty;
+	{
+		defer tagfields.deinit(main.stackAllocator);
+		var it = zon.object.iterator();
+		while (it.next()) |entry| {
+			if (std.mem.eql(u8, entry.key_ptr.*, "optional")) continue;
+			if (std.mem.eql(u8, entry.key_ptr.*, "disabled")) continue;
```
IntegratedQuantum, line 1358: I think there is no reason not to make this information accessible to addons.

IntegratedQuantum (changes requested):

src/items.zig
Crepestrom: could you provide more information
im not sure why it wouldnt be accessible by addons

src/proceduralItem/modifiers/restrictions/encased.zig
Crepestrom: added

src/items.zig
Crepestrom: renamed

assets/cubyz/blocks/branch/birch.zig.zon
Crepestrom: swapped

src/proceduralItem/modifiers/restrictions/on_top_of.zig
Crepestrom: oops fixed

Crepestrom:
tagFeilds have been renamed to slotTags

src/items.zig
IntegratedQuantum: I mean you should treat these like normal slot tags, so they can be accessed by addons (e.g. to add effects if you are in an optional slot, or next to `x` disabled slots)

Crepestrom: sure i dont mind that then

PR #3006, opened 2026-05-02
Require sickles for selecting and collecting foliage (only config changes)

### LINKED ISSUE #1558: Require sickles for collecting leaves, grass plants, etc

### LINKED ISSUE #2384: Infinite Apples [Bug]

DellieDelta:
Fixes #1558
Fixes #2384

groups

There are three types of foliage, which are handled differently. From left to right:

- Wood-like foliage (cactus, mushroom blocks, and their branches): 
    - Left unchanged, always allow selection and drops
- Foliage with a hitbox (leaves, lilypads, moss): 
   - Always allow selection, but only drop when using a sickle
- Replaceable foliage that you can walk through:
    - Only allow selection and drops when holding a sickle

Commits are split based on these foliage types

### DIFF
--- a/assets/cubyz/blocks/bluebells.zig.zon
+++ b/assets/cubyz/blocks/bluebells.zig.zon
@@ -2,8 +2,10 @@
 	.tags = .{.cuttable},
 	.blockHealth = 0.2,
 	.drops = .{
-		.{.items = .{.auto}},
+		.{.items = .{.auto}, .allowedToolTags = .{.cuttable}},
 	},
+	.selectionRule = .toolEffective,
+	.replaceable = true,
 	.degradable = true,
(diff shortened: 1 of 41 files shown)


IntegratedQuantum (approved):
Thank you, I noticed on more issue #3008, other than that it looks good.

Issue #2921, opened 2026-04-17
Tree Biome Independant Structures


Crepestrom:
This is just to organize the different Biome Independant Structure table Prs

Tags:
.oak #2788 
.birch #2880 
.pine #2882 
.mahogany #2915 
.willow #2916 
.baobab #2917 
.palm #2919 
.cactus #2920

IntegratedQuantum:
All PRs were merged

Issue #1364, opened 2025-04-27
Snow layers


careeoki:
**basic implementation:** 
- Has a rotation mode that allows for varying heights of snow (#1255 ?) Allows for snow on top of leaves and such.

**future mechanics:**
- Does not have collision, though walking through it slows you down (similar to #906). The higher the snow pile is, the more it will slow you down. The first layer has no effect on movement. _(a material could provide a snowshoe modifier for boots/accessories?)_
-  It's a #805? _(different heights falling ontop of eachother will combine together?)_
- This collects on the ground while it's snowing (#1046) (not infinitely :P). It's able to be melted by heat sources.

ikabod-kee:
How about making "snow" into a sort of frost layer instead? It can be applied to most blocks and make it quite white and frosty.

ikabod-kee:
Or, maybe frost can even have 2 stages. One that makes the block look frosty, then the second stage that gives the block the mere appearance of snow.

ikabod-kee:
<img width="391" height="581" alt="Image" src="https://github.com/user-attachments/assets/dfac1f71-056d-4f6d-a30d-e19c4549441a" />

Here's kind of a doodoo example.

ikabod-kee:
Then again, we can always try out a shader.

careeoki:
Snow != frost though. I want to be able to have fluffy snow piling up on leaves. A frost ore could be a seperate block, maybe

![Image](https://github.com/user-attachments/assets/197243e6-3273-4211-8039-df0d5fe90053)

ikabod-kee:
Ooh, you know what? We could have it be an ore pile combo.

ikabod-kee:
<img width="64" height="320" alt="Image" src="https://github.com/user-attachments/assets/d9b306b1-7b62-43b9-bce2-2a44318ca52c" />

ikabod-kee:
Okay, actually. I can easily see frost being an ore, and snow piles just being a pile XD nevermind

PR #3010, opened 2026-05-02
Revamped Tall Mountains

TyrrelMeter:
I've been thinking about current cubyz mountain generation, and realized that we can do better. With better generation settings I've created an even better tall mountain, with more trees to make it more realistic. I'm not completely happy or settled on the design yet, but I can say for sure it is an improvement. There are also some generation bugs with the mountains I'll need to fix.

Cubyz 0 3 0-dev 5_2_2026 6_47_39 PM
Cubyz 0 3 0-dev 5_2_2026 6_46_45 PM
Cubyz 0 3 0-dev 5_2_2026 6_46_57 PM
Cubyz 0 3 0-dev 5_2_2026 6_46_53 PM

### DIFF
--- a/src/server/terrain/sdf_models/cluster.zig
+++ b/src/server/terrain/sdf_models/cluster.zig
@@ -31,8 +31,8 @@ pub fn initAndGetExtend(zon: ZonElement) sdf.SdfModel.InitResult {
 	defer list.deinit();
 
 	var maxExtend: vec.Boxi = .{
-		.min = @splat(-1e9),
-		.max = @splat(1e9),
+		.min = @splat(1e9),
+		.max = @splat(-1e9),
 	};
 
 	for (zon.getChild("children").toSlice()) |child| {
@@ -68,8 +68,9 @@ pub fn instantiate(self: *@This(), arena: NeverFailingAllocator, seed: *u64) Sdf
 	var maxPos: Vec3i = @splat(-1e9);
 	for (self.children, instance.children) |entry, *result| {
 		result.* = entry.model.instantiate(arena, seed);
-		result.minBounds +%= @trunc(entry.positionOffset + entry.randomOffset*main.random.nextFloatVectorSigned(3, seed));
-		result.maxBounds +%= @trunc(entry.positionOffset + entry.randomOffset*main.random.nextFloatVectorSigned(3, seed));
+		const offset = entry.positionOffset + entry.randomOffset*main.random.nextFloatVectorSigned(3, seed);
+		result.minBounds +%= @trunc(offset);
+		result.maxBounds +%= @trunc(offset);
 		minPos = @min(minPos, result.minBounds);
 		maxPos = @max(maxPos, result.maxBounds);
 	}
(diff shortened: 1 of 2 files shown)


TyrrelMeter:
Hey, I know that this biome is really cool, but can you please move it into a cave? It would look much better there.

careeoki (changes requested):
This is undeniably very cool, and is a lot more detailed than what we have now.  But 

- Can you please split the cold mountain into a seperate PR? Yeah technically this is only 2 biomes, but there's a lot of individual subbiomes to look at here. (i also really don't like how snow biomes with trees look without #1364 , they just look weird with snow on the ground and nothing on the trees.)
- I think it's a little too tall. I would aim for 1000 blocks, for the sanity of anyone wanting to climb this. I like the amount of slopes there are, so see if you are able to just squish them closer vertically.
- There are a lot of big flat walls like this. Ideally there should be a smoother slope upwards. Again, since you added more slopes, you might be able to make it nicer by lowering the base biome.
image
image

Argmaster:
It would be cool if we could force what biomes are next to it or by some other way control how steep the climb is, as it gets pretty ridiculous... in some cases. Here same mountain from two different sides:

image

image

TyrrelMeter:
I updated the mountains base to be at 128 for this and the cold mountain PR but forgot to put a comment in to notify people about it.

assets/cubyz/biomes/tall_mountain/base.zig.zon
```diff
@@ -1,26 +1,93 @@
...
+	.maxSubBiomeCount = 1,
+	.stoneBlock = "cubyz:slate/base",
 	.music = "cubyz:out_of_breath",
-
+	.validPlayerSpawn = false,
+	.ground_structure = .{
+		"cubyz:grass",
+		"4 to 6 cubyz:soil",
+	},
+	.structures = .{
```
careeoki: I think you could reduce the amount of trees here. All of these big trees make it hard to navigate up the mountain (not in a fun way)

careeoki (changes requested):
Biome-wise it looks good, except one minor thing.
But something is happening here and creating these completely flat mountain sides.
Image
Image
@IntegratedQuantum you're better at identifying these technical problems than I am

assets/cubyz/biomes/tall_mountain/base.zig.zon
```diff
@@ -1,26 +1,93 @@
...
-	.mountains = 150,
-
+	.smoothBeaches = true,
+	.radius = 370,
+	.mountains = 170,
 	.chance = 0,
-
+	.maxSubBiomeCount = 1,
+	.stoneBlock = "cubyz:slate/base",
 	.music = "cubyz:out_of_breath",
```
careeoki: You should update this to the new music format (see #3044)

careeoki (changes requested):
Now that there are trees here, you should add the tags for them (.oak, .birch, .pine) (see #2921)

careeoki:
I think you could take advantage of the young oak SBBs here, especially since the mountains parent biome uses them.

TyrrelMeter:
ok I have no idea what I'm doing with github and just messed a bunch of stuff up because I don't know what I'm doing

TyrrelMeter:
#3098 is the new PR since I accidentally closed this one

IntegratedQuantum:
> you're better at identifying these technical problems than I am

It's a bug in the code, only happens at the border of the map fragments, I'll make an issue, no need to worry about it here.

TyrrelMeter:
thanks quantum

PR #3015, opened 2026-05-02
Cubyz outline

Crepestrom:
image

image

image

adds a outline shader to cubyz


du82:
I do not like this, it looks cartoonish and not like Minecraft

Crepestrom:
it doesnt need to look like minecraft

IntegratedQuantum:
I personally do not like outline shaders. Furthermore this implementation seems to have some implementation issues (look at a pond, also some of the lines appear at an offset in one of the images, the shape of lines also looks weird at the corners).
And it also makes LOD transitions much more visible, which is something we don't want.

Beyond that I also don't want any significant shader development before the vulkan migration is done.

PR #3019, opened 2026-05-03
Clear the player job queue before checking the reference count

IntegratedQuantum:
there may still be tasks in it

this fixes #2858


Wunka (approved):
YES FINNALY! 🎉  🎉

Issue #2818, opened 2026-04-04
Add a better way to display tool restrictions


IntegratedQuantum:
Current description is hard to understand, e.g. encased means that it's a direct or diagonal, but it's just as valid to assume that it only allows direct neighbors.

I think it would be good if we could color the relevant slots in the crafting grids when you hover over the item. The description should then display the same color.

Another option would be to have a little graphic in the tooltip.

Argmaster:
> I think it would be good if we could color the relevant slots in the crafting grids when you hover over the item. The description should then display the same color.

I am not sure about exact interpretation, but what I think a mature (not easy to implement, but a good UX) solution would be to:
When player holds an item in UI (has it in hand slot) and hovers it over crafting grid slot (don't even ask how would that work) to add green and red (something) to the slots nearby that contribute positively/neutrally to the modifier activation if the item was placed in hovered slot. Then, additional (something) would show up on the item attached to the mouse when the modifier is actually activated.
Crazy complicated to implement, very obvious when user intersects with it.
Something could be background color, outline, or tick/cross in corner, although extra images would create noise. Different information should be presented though different means imo.

Another thing to do for same of different purpose would be to display secondary tooltip when hovering over slots in crafting inventory.

Another thing we could do is to light up (colorful background) items in player inventory that interact with each other, although that is a little bit noisy since there will be a lot of interactions and player will not be interested in most of them.
A more subtle way would be to display small circles (3x3 pixels) in top left corner for every modifier that would interact with currently held item

IntegratedQuantum:
> When player holds an item in UI (has it in hand slot) and hovers it over crafting grid slot (don't even ask how would that work) to add green and red (something) to the slots nearby that contribute positively/neutrally to the modifier activation

The difficult part is that there may be multiple modifiers with possibly different restrictions

PR #3023, opened 2026-05-03
Colorable Item Slots

Crepestrom:
Item slots can now be recolored

image
image


ikabod-kee:
Ooh, what if custom icons for them next?

DellieDelta:
Also related: #2818

Argmaster:
Okay, I sort of think that static non-gray colors are not a great idea. Support for coloring slots on demand is great (functionally, haven't looked at the code yet), but hard coding colors (other than gray) in zon files not so much. It ends up noisy, too much implicit information; what do the colors even mean?. Additionally it poses a accessibility issue for color blind people. 

I guess if we were to use gray scale it would make some sense to mark same types of slots, but then required slots would need something extra to emphasize their function. An asterisk or something.

Argmaster:
Actually, there is even more issues, the final look of workbench inventory is going to be much different; it's going to have blue background (as of now that's the plan)

Wunka:
I think it would be better to combine it with #3005. so you can set colors on tags. that way we could also move the `!optional = darker color` logic there..

IntegratedQuantum:
A different option would be to just add small icons to the slot, together with a map legend explaining the symbols.

ikabod-kee:
A legend is not really needed. You can just have it show its stats on being hovered.

IntegratedQuantum:
Well a legend has the advantage that you don't need to hover over all the slots individually to understand their meaning.

ikabod-kee:
using a legend doesn't sound very modular.

Crepestrom:
image

Updated to modern version
colors have been neutralized

PR #3028, opened 2026-05-03
Make selection wand select foliage

### LINKED ISSUE #3025: The selection wand cannot select plants

Which is preventing me from selecting a blueprint like this:
Image

DellieDelta:
Fixes #3025

### DIFF
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -528,7 +528,12 @@ pub const Block = packed struct(u32) { // MARK: Block
 	}
 
 	pub fn isSelectableByItem(self: Block, item: Item) bool {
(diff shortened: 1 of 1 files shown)


Wunka:
I think this is ok as a short time fix. but in long term I think this should be handeld differently. 
Maybe through a callback? #2898 

I atleast think that we should aim for 0 mentioning of any kind of specific item, block etc in the code.
And with how things currently are going I think this is not only my opinion.

DellieDelta:
I agree, at some point we probs need some general way to say that an item should always be able to select blocks

IntegratedQuantum (approved):
I do not like the hardcoding fix, please try to work towards a proper fix for this soon.
I'm only merging this, since it directly affected carrie.

PR #3032, opened 2026-05-04
Tiny cleanup: use worldArena directly in blocks.zig

DellieDelta:


IntegratedQuantum (approved):
Note that in general arenas are pretty bad with resizing data structure, every resize leaks the previous memory and thus you end up using more memory than you actually need.
But luckily here this is not relevant since it's created with the right capacity (ignoring the error case)

DellieDelta:
Good to know!

Issue #3031, opened 2026-05-04
3x3x1 panel mining


Crepestrom:
### Description

if we were to implement mining with a 3x3 mining area what should it be
should it be a tool type
a tool modifier
or an accessory or armor

Crepestrom:
i ask because i have the code for the mining part but not for what should be needed to trigger it

careeoki:
basic concept for making mining patterns
<img width="1301" height="513" alt="Image" src="https://github.com/user-attachments/assets/0ce5c99f-a587-4a5f-b7f8-8078336b1ce9" />
With this you could design any 3x3 shape and mine with that. It would be cooler to do this when crafting the tool itself but not every tool has a 3x3 square of slots

Argmaster:
> basic concept for making mining patterns
Image

The issue is that we don't really have slots in tools that are exclusively for side effects, so if we were to use flowers, they will have low stats and thus the tool stats will immediately tank.
Could we maybe change this?
And significantly drop the impact of some slots on the tool stats? Honestly, maybe even to 0.
Another option is to conditionally exclude items from stat math.
We could enable modifiers to do that?

Argmaster:
Oh I think I ignored the core of your suggestion, sorry Carrie.
Anyway I recall talking with Quantum about using flowers to select multiblock pattern, that is what I refered to mostly.

IntegratedQuantum:
It could also be an accessory that you equip, so it applies to placement as well

Crepestrom:
It would probably be good that if you were crouching it only mined your selected block and not the extra area

Issue #1678, opened 2025-07-17
Attack The Block! (ATB)


ikabod-kee:
<img width="606" height="552" alt="Image" src="https://github.com/user-attachments/assets/0a086b32-f1a4-4fcb-bd41-1d971861d634" />

In case undertime doesn't work out for you guys.

ikabod-kee:
In case you're wondering what the threshold would be. You'd have to aim at the edge pixels to perform the damage-split.
<img width="1858" height="1057" alt="Image" src="https://github.com/user-attachments/assets/9c575527-1cd2-458c-9cd0-f3beef16e7a0" />
This *wouldn't* proc the damage split.

<img width="1858" height="1057" alt="Image" src="https://github.com/user-attachments/assets/adad981d-d7dc-40e3-b6ed-47cee90f62ab" />
This *would.*

ikabod-kee:
Note that this requires storing damage for more than one block (as pointed out by ARG)

ikabod-kee:
A more descript version of this: When the player swings their tool (or even fist) it casts a ray and places a 1.5x1.5x1.5 pixel hitbox on the block. The damage will be distributed evenly across all blocks that are touched by this hitbox. If one of the blocks is going to be destroyed by this action, then the leftover damage is automatically distributed to other blocks in this hitbox.

Why this system would work:
- The player now has more control over their swings compared to undertime or overdamage methods
- Players can mine tunnels with ease by aiming at a block intersection.
- Way more satisfying.

Concerns you may have and why they aren't much of an issue:
- "Won't the player accidentally perform this action?"
If the player is aware of this action, which they will probably find on their own within the first hour of playing, then they will try aiming at the center of the block more often. Also, most players aim for the center of blocks while mining anyways.
- "Ikabod I'm scared."
It's okay.

ikabod-kee:
@Argmaster Could you add your suggestion of how this should work here?

ikabod-kee:
For scythes, maybe they could work like this?

<img width="464" height="442" alt="Image" src="https://github.com/user-attachments/assets/042292a9-cf27-408f-a023-102369040d2c" />

ikabod-kee:
<img width="463" height="440" alt="Image" src="https://github.com/user-attachments/assets/e0f5d4e6-4d5a-4553-bb91-6318f1e67122" />

Don't like that? We got you covered.

ikabod-kee:
[...]

PR #3035, opened 2026-05-05
3x3 block breaking

Crepestrom:
sulfer has a new property that allows 3x3 breaking

its not perfect and im not sure what the balancing on it should be

#3031


Wbjitscool:
this would be a nice feature to have

Argmaster:
1. Under no circumstances we should make the relation between modifiers and breakign logic bidirectional. Breaking logic calls into modifiers, not other esy around, because the latter cannot be extended.
2. There are upcoming changes to breaking logic https://github.com/PixelGuys/Cubyz/pull/2898
That make current implementation obsolete in one way or another.
3. We might want more control over the shape than just 3x3

Wunka (changes requested):
I too agree with Argmaster.
This should be handeld differently.
I think it could be maybe interesting to instead of touching the break block logic, to touch the select logic. 

Atleast you need to go through a function that does more than return a bool.

I would throw the idea out to maybe make a `.changeSelectedBlock` callback which then also makes it so (atleast for your modifier) that the other surrounding blocks are also selected. But for that first there needs to be a rewrite to even allow selecting multiple blocks.

Wbjitscool:
Yeah

m-m-0-m:
An alternative would be to implement #1678 and make the picaxe `size` larger than the block size.
Then, if you were to place your cursor in the center of a block, you could break a 3x3x1 chunk all at once.

Wunka:
> An alternative would be to implement #1678 and make the picaxe `size` larger than the block size. Then, if you were to place your cursor in the center of a blck, you could break a 3x3x1 chunk.

yeah. that would still mean we need the rewrite for selecting multiple blocks

IntegratedQuantum:
I would prefer to wait on this until we have onUse/onBreak callbacks and these actions are properly synced with the server.

Also your implementation is insufficient, as it hardcodes additional restrictions into the shape (it must be a box, damage is distributed equally), instead of exposing the shape to the outside, the hit callback should instead handle the shape locally and call into the block breaking code instead.

PR #3036, opened 2026-05-05
Move stepping related eye movement calculations into calculateEyeMovement

codemob-dev:
Progress towards https://github.com/PixelGuys/Cubyz/issues/1728 and https://github.com/PixelGuys/Cubyz/issues/1663


IntegratedQuantum (changes requested):
Would it be possible to combine the two eye update methods? I think we could afford to do calculateEyeMovement after checking the stepping height.

codemob-dev:
> Would it be possible to combine the two eye update methods? I think we could afford to do calculateEyeMovement after checking the stepping height.

Should be. I'll test to see how well it works.

codemob-dev:
Seems to work fine!

IntegratedQuantum (approved):

PR #3037, opened 2026-05-05
Add a general way to ignore the selection rule for blocks, and let selection wand use it

DellieDelta:
In essence it's an override (blocks define selectability, but items can choose to override it), but for now the only use case is ignoring it completely, so I made it a simple boolean


IntegratedQuantum:
Maybe we need to use generic callbacks for selection (see comment in #3040), adding a new property every time we discover a new edge case doesn't scale well, and limits mods too.

DellieDelta:
Good idea, callbacks are perfect for overrides like this

PR #2659, opened 2026-03-04
Flowers folder

Crepestrom:
Moves all flowers to a flower folder
-bluebells
-cactus flower
-castilleja
-daffodil
-daisies
-dandelions
-hibiscus
-marigold
-trumpet_lily
-vetch

PR #3038, opened 2026-05-05
Add `flower` tag to all flowers

Argmaster:
When thinking about https://github.com/PixelGuys/Cubyz/issues/3031#issuecomment-4381945498 I noticed there is no easy way to find all the flowers and since mushrooms got their tag, why flowers can't have one?


Argmaster:
Mutually exclusive with #2659 (or is it?)

ikabod-kee:
This is very late, but I don't see why this shouldn't be merged.

assets/cubyz/blocks/osier.zig.zon
IntegratedQuantum, line 1: I wouldn't categorize it as a flower

IntegratedQuantum (changes requested):
hibiscus is missing

Argmaster:
I have finally updated this one.

IntegratedQuantum (approved):

Issue #2930, opened 2026-04-18
Replace `@cImport`, it was deprecated upstream


IntegratedQuantum:

AfmanS:
Should this be done in the same manner as https://github.com/PixelGuys/Cubyz/pull/2058?

IntegratedQuantum:
No, I think all includes should be added to a single .c files with comments explaining what they are for.

Issue #2950, opened 2026-04-20
Standardized imports


codemob-dev:
We use two different import styles in different places without much consistency. (`const foo = @import("foo.zig");` vs `const foo = main.foo;`) We should choose one and use it everywhere.

Eg. `itemdrop.zig` mainly uses the first style, while `blueprint.zig` mainly uses the second.
The first few imports of `itemdrop.zig`:
```zig
const std = @import("std");

const blocks = @import("blocks.zig");
const chunk_zig = @import("chunk.zig");
const ServerChunk = chunk_zig.ServerChunk;
const game = @import("game.zig");
const World = game.World;
const ServerWorld = main.server.ServerWorld;
const graphics = @import("graphics.zig");
const c = graphics.c;
const items = @import("items.zig");
const ItemStack = items.ItemStack;
const ZonElement = @import("zon.zig").ZonElement;
const main = @import("main");
const physics = main.game.physics;
const random = @import("random.zig");
const settings = @import("settings.zig");
const utils = @import("utils.zig");
const vec = @import("vec.zig");
```

The first few imports of `blueprint.zig`:
```zig
const std = @import("std");

const main = @import("main");
const Compression = main.utils.Compression;
const ZonElement = @import("zon.zig").ZonElement;
const vec = main.vec;
const Vec3i = vec.Vec3i;

const Array3D = main.utils.Array3D;
const Block = main.blocks.Block;
const NeverFailingAllocator = main.heap.NeverFailingAllocator;
const User = main.server.User;
const ServerChunk = main.chunk.ServerChunk;
const Degrees = main.rotation.Degrees;
const Tag = main.Tag;
```

Note that `itemdrop.zig` uses `const vec = @import("vec.zig");` while `blueprint.zig` uses `const vec = main.vec;`

I am personally in favor of the second option because it makes it easier to respect `pub const ... = @import(...)` statements. We could probably enforce this by making the CI fail if there are any `@import` statements that aren't public.

Wunka:
I also like the second option more. Enforcing it is sadly not possible as sometimes using the second options leets to recursive imports. Also sometimes you need to import things which are not in the import "tree" (see for example https://github.com/PixelGuys/Cubyz/pull/2585#issuecomment-3959005357). I would atleast say we should put this prefernce in the contributing guidelines

Argmaster:
There is a semantic difference between those two forms. If I remember correctly`@import` is relative to current file, so referencing file in a different parent directory is at least awkward.
[...]

PR #3039, opened 2026-05-05
Add Translate C to build.zig, remove @cImport from graphics.zig

AfmanS:
Progress towards #2930.
graphics.zig seemed the most complex so it's a good first file to discuss the move.
Mostly followed example from https://ziglang.org/download/0.16.0/release-notes.html#cImport-Moving-to-Build-System.
All C import consts in graphics.zig were renamed to c. 
I don't have the knowledge to write better comments on the c.h file.
Builds, passes tests and plays fine on Windows x86. I have no way to test that that __MACH__ preproc actually works on macOS.


src/graphics.zig
```diff
@@ -36,25 +23,6 @@ pub const ComputePipeline = pipelines.ComputePipeline;
 pub const Pipeline = pipelines.Pipeline;
 pub const vulkan = @import("graphics/vulkan.zig");
 
-pub const c = @cImport({
-	@cInclude("glad/gl.h");
-
-	// NOTE(blackedout): glad is currently not used on macOS, so use Vulkan header from the Vulkan-Headers repository instead
```
IntegratedQuantum, line 42: Please keep the comment

src/c.h
```diff
@@ -0,0 +1,24 @@
+// graphics.zig
```
IntegratedQuantum: Please add comments for what they are used for, not what file they are from.

src/graphics.zig
```diff
@@ -3,20 +3,7 @@
 const std = @import("std");
 const builtin = @import("builtin");
 
-pub const hbft = @cImport({
-	@cDefine("_BITS_STDIO2_H", ""); // TODO: Zig fails to include this header file
```
IntegratedQuantum, line 7: Oh, does this work now? Have you tried compiling this for linux with ReleaseSafe?

IntegratedQuantum (changes requested):

src/graphics.zig
AfmanS: done

src/c.h
AfmanS: I took a stab at it, but I'm not sure

src/graphics.zig
AfmanS: I hadn't tried compiling with ReleaseSafe at all, tried it on Windows and it fails. But it's not because of my changes, even on the master branch I get 10 of these errors when trying ReleaseSafe: 
```
install
└─ install Cubyz
   └─ compile exe Cubyz ReleaseSafe native 10 errors
.zig-cache\o\2a91e3a4144812e52f5ea6ae934a5d6c\cimport.zig:978:15: error: unused local constant
        const extern_local_wcscat_s = struct {
              ^~~~~~~~~~~~~~~~~~~~~
.zig-cache\o\2a91e3a4144812e52f5ea6ae934a5d6c\cimport.zig:1002:15: error: unused local constant
        const extern_local_wcscpy_s = struct {
```
Seems related to: https://codeberg.org/ziglang/translate-c/issues/327

AfmanS: I should say that without this line, the `#include`s that needed that `@cDefine` do fail, so it might be doing the job:
https://github.com/PixelGuys/Cubyz/pull/3039/changes#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR56

IntegratedQuantum: Yes, this is known this behavior is tracked in #2336, but for Linux we have a workaround for this issue (which is what this define is doing).
If you are on windows you can also cross-compile to linux with `-Dtarget=x86_64-linux-gnu` to check if it still works with your changes.

AfmanS: I can confirm that building with `-Doptimize=ReleaseSafe -Dtarget=x86_64-linux-gnu` works just fine

src/graphics.zig
```diff
@@ -3,20 +3,7 @@
...
-	@cInclude("freetype/ftcolor.h");
-	@cInclude("freetype/ftlcdfil.h");
-	@cInclude("freetype/ftsizes.h");
-	@cInclude("freetype/ftstroke.h");
-	@cInclude("freetype/fttrigon.h");
-	@cInclude("freetype/ftsynth.h");
-	@cInclude("hb.h");
-	@cInclude("hb-ft.h");
-});
+pub const c = @import("c");
```
IntegratedQuantum: Please make this private, all other files that used graphics.c, should now use the module with `@import("c")`

src/c.h
```diff
@@ -0,0 +1,27 @@
...
+#include <freetype/ftcolor.h>
+#include <freetype/ftlcdfil.h>
+#include <freetype/ftsizes.h>
+#include <freetype/ftstroke.h>
+#include <freetype/fttrigon.h>
+#include <freetype/ftsynth.h>
+#include <hb.h>
+#include <hb-ft.h>
+
+// used for rendering graphics and windows
```
IntegratedQuantum: windows is ambiguous which makes this confusing, I'd suggest windowing instead
```suggestion
// used for rendering, windowing and inputs
```

src/c.h
```diff
@@ -0,0 +1,27 @@
+// used for text rendering
```
IntegratedQuantum:
```suggestion
// used for text rendering and layouting
```

src/c.h
```diff
@@ -0,0 +1,27 @@
...
+// NOTE(blackedout): glad is currently not used on macOS, so use Vulkan header from the Vulkan-Headers repository instead
+#ifdef __MACH__
+#include <vulkan/vulkan.h>
+#include <vulkan/vulkan_beta.h>
+#else
+#include <glad/vulkan.h>
+#endif
+#include <GLFW/glfw3.h>
+
+// used for exporting and reading images to/from files
```
IntegratedQuantum:
```suggestion
// used for reading and writing image files
```

IntegratedQuantum (changes requested):

src/client/Entity.zig
```diff
@@ -4,7 +4,7 @@ const main = @import("main");
 const chunk = main.chunk;
 const game = main.game;
 const graphics = main.graphics;
-const c = graphics.c;
+const c = @import("c");
```
IntegratedQuantum: Could you please reorder them to be in a new block? Each module import and subimports should be in its own block.

src/gui/windows/performance_graph.zig
```diff
@@ -59,6 +59,7 @@ pub fn deinit() void {
 }
 
 pub fn render() void {
+	const c = @import("c");
```
IntegratedQuantum: All module importants should be done at the top of the file, please apply this to the other files as well.

IntegratedQuantum (changes requested):

src/client/Entity.zig
AfmanS: I tried changing just Entity.zig. Is this how it should be?

IntegratedQuantum: yes

AfmanS: done

src/gui/windows/performance_graph.zig
AfmanS: done

src/renderer.zig
```diff
@@ -6,18 +6,23 @@ const chunk = @import("chunk.zig");
...
+
 const main = @import("main");
+const gpu_performance_measuring = main.gui.windowlist.gpu_performance_measuring;
+const crosshair = main.gui.windowlist.crosshair;
 const Window = main.Window;
+
 const models = @import("models.zig");
 const network = @import("network.zig");
 const settings = @import("settings.zig");
+
```
IntegratedQuantum: This is actually incorrect, these are not real imports, they are legacy code (see #2950) and will be converted to `main.` at some point, so please leave them like they were and just put c at the bottom.

src/graphics.zig
```diff
@@ -3,20 +3,7 @@
...
-	@cInclude("freetype/ftcolor.h");
-	@cInclude("freetype/ftlcdfil.h");
-	@cInclude("freetype/ftsizes.h");
-	@cInclude("freetype/ftstroke.h");
-	@cInclude("freetype/fttrigon.h");
-	@cInclude("freetype/ftsynth.h");
-	@cInclude("hb.h");
-	@cInclude("hb-ft.h");
-});
+const c = @import("c");
```
IntegratedQuantum: Could you move the c at the bottom, always below main?

IntegratedQuantum (changes requested):

src/graphics.zig
AfmanS: Done

src/renderer.zig
AfmanS: Is this acceptable?

IntegratedQuantum (approved):

PR #3040, opened 2026-05-05
Allow blocks to define under what conditions they can be placed, for now only for selection

DellieDelta:
Lil demo: https://www.youtube.com/shorts/1qEEHvCX0kc

Related to #2491

Continuation of #2983

For now it's only enforced client-side through selection. But my idea for the future is that the placement constraints will get checked before placement by default through the same function when no `onPlace` callback is defined

And when you do define the callback, that you have to decide in there what to do with the placement constraint check result

Introduces a new concept for blocks, called `support`. Support is about the way other blocks can be placed against it. Soil can always be placed against, but water can only be placed against by lily pads and duckweed through defining the placement constraints

Gets rid of the `.fluidPlaceable` tag, and the case for it in selection

My plan is to make another PR later for also checking whether the neighbor is replaceable or not, so you can disable selection when underwater and holding a lily pad for example

This also opens the door for removing the same type of block check in selection (for vines, water, air, fog). You can add a new placement constraint like `allowedBlocks = .{.auto}`, something like that

### DIFF
--- a/assets/cubyz/blocks/lily_pad.zig.zon
+++ b/assets/cubyz/blocks/lily_pad.zig.zon
@@ -4,6 +4,9 @@
 	.drops = .{
 		.{.items = .{.auto}, .allowedToolTags = .{.cuttable}},
 	},
+	.placementConstraints = .{
+		.{.allowedTags = .{.fluid}, .allowedFaces = .{.top}},
+	},
 	.degradable = true,
 	.alwaysViewThrough = true,
 	.absorbedLight = 0x121012,
@@ -12,7 +15,6 @@
 	.texture = "cubyz:lily_pad",
 	.item = .{
 		.texture = "lily_pad.png",
-		.tags = .{.fluidPlaceable},
 	},
 	.lodReplacement = "cubyz:air",
 }
--- a/assets/cubyz/blocks/duckweed.zig.zon
+++ b/assets/cubyz/blocks/duckweed.zig.zon
@@ -4,6 +4,9 @@
 	.drops = .{
 		.{.items = .{.auto}, .allowedToolTags = .{.cuttable}},
 	},
+	.placementConstraints = .{
+		.{.allowedTags = .{.fluid}, .allowedFaces = .{.top}},
+	},
 	.selectionCapabilities = .{.toolEffective},
 	.replaceable = true,
 	.degradable = true,
@@ -21,7 +24,6 @@
 	.texture3 = "cubyz:duckweed/3",
 	.item = .{
 		.texture = "duckweed.png",
-		.tags = .{.fluidPlaceable},
 	},
 	.lodReplacement = "cubyz:air",
 }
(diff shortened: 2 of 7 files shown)


assets/cubyz/blocks/lily_pad.zig.zon
```diff
@@ -4,6 +4,9 @@
 	.drops = .{
 		.{.items = .{.auto}, .allowedToolTags = .{.cuttable}},
 	},
+	.placementConstraints = .{
+		.{.allowedTags = .{.fluid}, .allowedFaces = .{.top}},
```
IntegratedQuantum, line 8: lilypads should remain placeable on normal terrain (outside of water) in my opinion.

IntegratedQuantum (changes requested):
I like the idea of only selecting things if you can also place them in that orientation.

However I have two cirticisms of your current implementation:

In my opinion, for blocks without rotation, I should be able to place it independently of where I am looking at.
e.g. in this situation, when I right click, I expect a lilypad to be placed on the water, but when I can only select the top side of blocks, then this is not possible:

Image

Secondly, I think this implementation is insufficient, you will have a difficult time implementing #2491 with just tags and looking at only the selected neighbor block.

Instead it should be a generic callback (see callbacks.zig).

DellieDelta:
Ahhh, I had no idea you could define custom fields in callbacks, that's nice! So I'm imagining something like this then:

```
.onPlace .{
	.type = .check_constraints,
	.constraints = .{
		.{.allowedTags = .{.fluid}, .allowedFaces = .{.top}},
	},
},
```

And then selection could be something like this in pseudocode:

```
fn isSelectable(blockToPlace, targetBlock) bool {
	if (blockToPlace has onPlace callback) {
		return blockToPlace.onPlace.run() == .handled
	}

	if (targetBlock.support == .no) return false
	return targetBlock.neighbor.replaceable
}
```

But before I start working on this, would you be fine with already having an `onPlace` callback that isn't used yet for placement, but only for the `.placeable` case in selection? I saw that the actual placement checking can only be implemented after the world interface refactor, but I think selection is not part of that

IntegratedQuantum:
`.constraints = .{`
I don't like this, since it is unclear whether the list is `or` or `and`. For material modifiers we have composable `and` `or` and `not` restrictions, maybe the same system would be better here as well
```
.canSelect = .{
	.type = "cubyz:or"
	.children = .{
		.{
			.type = "cubyz:check_neighbor_block_tags",
			.neighbors = .{.top},
			.requiredTag = .fluid,
		},
		.{
			.type = "cubyz:check_neighbor_block_tags",
			.neighbors = .{.top},
			.requiredTag = .solid,
		},
	},
},
```

> 	if (blockToPlace has onPlace callback) {

I would suggest to always have the callback, and handle the other logic inside the default callback.

> I saw that the actual placement checking can only be implemented after the world interface refactor

You'll need the world interface to get the neighbor blocks

DellieDelta:
I think it's best to close this for now then, and to pick it back up again when the code is ready for placement checking. I like the idea of making the restrictions composable though!

PR #3042, opened 2026-05-06
Split vertical collision functions out of physics.update

codemob-dev:
Progress towards #1728 and #1663


src/physics.zig
```diff
@@ -553,49 +553,61 @@ pub fn calculateWallCollision(comptime side: main.sync.Side, motion: *Vec3d, pos
...
-			} else {
-				velocityChange = Player.super.vel[2];
-				Player.super.vel[2] = 0;
-			}
-			const damage: f32 = @floatCast(@round(@max((velocityChange*velocityChange)/(2*baseGravity) - 7, 0))/2);
+		// Always unstuck upwards for now
+		while (collision.collides(side, .z, 0, pos.*, hitBox)) |_| {
+			pos[2] += 1;
+		}
+		return velocityChange;
```
IntegratedQuantum: Could you not compute the velocity change on the callsite? It's literally just `@abs(@abs(preVel[2]) - @abs(vel[2]))`

IntegratedQuantum (changes requested):

src/physics.zig
codemob-dev: Done

IntegratedQuantum (approved):

PR #3045, opened 2026-05-06
Calculate star opacity with ambientLight instead of dayCycle

AfmanS:
In preparation for an asymmetric day-night cycle (#1211), I think it makes more sense, is simpler and faster for star opacity to be based on ambient light. This way it's not coupled to the very specific transitions between times of day. I tested and it looks fine.
I checked and the `if (starOpacity != 0) {` block is not entered during the day and at night starOpacity is 1. I used `std.debug.print("{d}\n", .{starOpacity });` (in case {d} does some FP rounding magic I should know).


IntegratedQuantum (changes requested):
I don't like that this creates a dependence between the brightness of the blocks and the brightness of the stars (and you already feel the problems of it here, with the magic constant 0.9).
I think it would be better to have some struct that stores the time of day and has functions to query the brightness as well as other properties.

AfmanS:
> I think it would be better to have some struct that stores the time of day and has functions to query the brightness as well as other properties.

You mean something like removing dayNightLightFactor from World and placing it in another struct dedicated to day length, day cycle transitions and star opacity? https://github.com/PixelGuys/Cubyz/blob/604ca82c84d4988c89cc9aba12222c405d6d92eb/src/game.zig#L375

IntegratedQuantum:
Yeah, basically all functions that relate to the day-night cycle should be in one place.

AfmanS:
Quite some changes... The change to the star rotation makes it realistic, with a full rotation per day. But it does look faster.

src/gui/windows/debug.zig
```diff
@@ -66,6 +66,8 @@ pub fn render() void {
 		y += 8;
 		draw.print("Game Time: {}", .{main.game.world.?.gameTime.load(.monotonic)}, 0, y, 8, .left);
 		y += 8;
+		draw.print("Day Time: {}", .{main.game.dayCycle.?.dayTime}, 0, y, 8, .left);
```
IntegratedQuantum: Could you put this in the previous line? The debug menu has already too many lines

src/gui/windows/multiplayer.zig
```diff
@@ -57,6 +57,7 @@ fn join() void {
 	if (connection) |_connection| {
 		_connection.world = &main.game.testWorld;
 		main.game.world = &main.game.testWorld;
+		main.game.dayCycle = .{};
```
IntegratedQuantum: I think this should be in the world and inited together with it.

src/server/world.zig
```diff
@@ -1319,3 +1317,11 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		self.mutex.unlock();
 	}
 };
+
+pub const ServerDayCycle = struct { // MARK: DayCycle
```
IntegratedQuantum: Please put the relevant constants into game.DayCycle, this constant does not belong here.
Constants in two places are unmaintainable.

IntegratedQuantum (changes requested):

src/game.zig
```diff
@@ -415,26 +379,119 @@ pub const World = struct { // MARK: World
...
-			fog.skyColor = biomeFog.skyColor*skyColorFactor;
-			fog.density = biomeFog.density;
-			fog.fogLower = biomeFog.fogLower;
-			fog.fogHigher = biomeFog.fogHigher;
-		}
 		network.protocols.playerPosition.send(self.conn, Player.getPosBlocking(), Player.getVelBlocking(), @intCast(newTime & 65535));
+		self.dayCycle.update(deltaTime);
 	}
+
+	pub const DayCycle = struct { // MARK: DayCycle
```
IntegratedQuantum: I'm not too happy with the name.
Maybe DayTime would fit better?

src/gui/windows/debug.zig
AfmanS: done

src/gui/windows/multiplayer.zig
AfmanS: done

src/server/world.zig
AfmanS: done

src/game.zig
```diff
@@ -415,26 +379,119 @@ pub const World = struct { // MARK: World
...
+		const dayCycleLength: u63 = 12000; // Length of one in-game day in 100ms. Midnight is at DAY_CYCLE/2. Sunrise and sunset each take about 1/16 of the day. Currently set to 20 minutes
+		const minimumAmbientLight: f32 = 0.1;
+
+		biomeFog: Fog = Fog{.skyColor = .{0.8, 0.8, 1}, .fogColor = .{0.8, 0.8, 1}, .density = 1.0/15.0/128.0, .fogLower = 100, .fogHigher = 1000},
+		fog: Fog = Fog{.skyColor = .{0.8, 0.8, 1}, .fogColor = .{0.8, 0.8, 1}, .density = 1.0/15.0/128.0, .fogLower = 100, .fogHigher = 1000},
+		ambientLight: f32 = 0,
+		dayTime: i64 = 0,
+
+		pub fn getNightStart() u63 {
```
IntegratedQuantum: Could be a constant.

Also, following the previous code, I think this should be midnight.

src/game.zig
```diff
@@ -415,26 +379,119 @@ pub const World = struct { // MARK: World
...
-			fog.density = biomeFog.density;
-			fog.fogLower = biomeFog.fogLower;
-			fog.fogHigher = biomeFog.fogHigher;
-		}
 		network.protocols.playerPosition.send(self.conn, Player.getPosBlocking(), Player.getVelBlocking(), @intCast(newTime & 65535));
+		self.dayCycle.update(deltaTime);
 	}
+
+	pub const DayCycle = struct { // MARK: DayCycle
+		const dayCycleLength: u63 = 12000; // Length of one in-game day in 100ms. Midnight is at DAY_CYCLE/2. Sunrise and sunset each take about 1/16 of the day. Currently set to 20 minutes
```
IntegratedQuantum: Doesn't need the explicit type.
```suggestion
		const dayCycleLength = 12000; // Length of one in-game day in 100ms. Midnight is at DAY_CYCLE/2. Sunrise and sunset each take about 1/16 of the day. Currently set to 20 minutes
```

src/game.zig
```diff
@@ -415,26 +379,119 @@ pub const World = struct { // MARK: World
...
+			const biome = world.?.playerBiome.load(.monotonic);
+
+			const t = 1 - @as(f32, @floatCast(@exp(-2*deltaTime)));
+
+			self.biomeFog.fogColor += (biome.fogColor - self.biomeFog.fogColor)*@as(Vec3f, @splat(t));
+			self.biomeFog.skyColor += (biome.skyColor - self.biomeFog.skyColor)*@as(Vec3f, @splat(t));
+			self.biomeFog.density += (biome.fogDensity - self.biomeFog.density)*t;
+			self.biomeFog.fogLower += (biome.fogLower - self.biomeFog.fogLower)*t;
+			self.biomeFog.fogHigher += (biome.fogHigher - self.biomeFog.fogHigher)*t;
+			// Ambient light:
```
IntegratedQuantum: the comment was useless before

src/game.zig
```diff
@@ -415,26 +379,119 @@ pub const World = struct { // MARK: World
...
+			{
```
IntegratedQuantum: the scope can be removed now that it's in its own function

IntegratedQuantum (changes requested):

src/game.zig
AfmanS: I'd argue it's more useful for it to be the start of each day phase, someone who sets night probably wants it longer rather than shorter, name for day

src/game.zig
AfmanS: done

src/game.zig
AfmanS: done

src/game.zig
AfmanS: done

src/game.zig
AfmanS: done

src/game.zig
IntegratedQuantum: Then please introduce a `dayStart` constant that is used in the other case to make the `/time` command consistant.

(and also adjust the nightStart constant to actually be at night start)

IntegratedQuantum (changes requested):
Please solve the night start/midnight inconsistency.

src/game.zig
AfmanS: done

IntegratedQuantum (approved):

PR #3047, opened 2026-05-07
Update encased.zig

Crepestrom:
makes encased give a warning if you do not put in a correct tag
"Could not find the tag for Encased restriction"

### DIFF
--- a/src/proceduralItem/modifiers/restrictions/encased.zig
+++ b/src/proceduralItem/modifiers/restrictions/encased.zig
@@ -27,7 +27,7 @@ pub fn loadFromZon(allocator: NeverFailingAllocator, zon: ZonElement) *const Enc
 		.tag = main.Tag.find(zon.get(?[]const u8, "tag", null) orelse blk: {
 			std.log.err("Could not find the tag for Encased restriction", .{});
 			break :blk "not specified";
-			}),
+		}),
 		.amount = zon.get(usize, "amount", 8),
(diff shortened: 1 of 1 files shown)


src/proceduralItem/modifiers/restrictions/encased.zig
```diff
@@ -24,7 +24,10 @@ pub fn satisfied(self: *const Encased, proceduralItem: *const ProceduralItem, x:
 pub fn loadFromZon(allocator: NeverFailingAllocator, zon: ZonElement) *const Encased {
 	const result = allocator.create(Encased);
 	result.* = .{
-		.tag = main.Tag.find(zon.get([]const u8, "tag", "not specified")),
+		.tag = main.Tag.find(zon.get(?[]const u8, "tag", null) orelse blk: {
+			std.log.err("Could not find the tag for Encased restriction", .{});
```
Argmaster: Your current message is ambiguous, may be interpreted as "there is a tag field, but the value is not recognized" or "there is no tag field and there should be one". I would suggest something akin to:

```suggestion
			std.log.err("Missing tag field for encased restriction.", .{});
```

Long term I would consider propagating some item information, like ID to here so we can have a more descriptive message. Out of scope for now.

Argmaster (changes requested):

IntegratedQuantum (approved):

PR #3048, opened 2026-05-07
Add a workbench to the rare structure pool

ikabod-kee:
## That's it. That's the only change.

image 
Pictured: Naturally Spawning Workbench


Wbjitscool:
lol

IntegratedQuantum:
I'm not sure if we should have this. The workbench is already placed deliberately in a few select biomes (→stone forest, the root, spawn), and I think having it everywhere takes away from this.

ikabod-kee:
You do make a good point, that does seem pretty deliberate.

PR #3049, opened 2026-05-07
Switch to Io.net (2nd try)

Wunka:
Sadly I cannot reopen #2466 so I need to open a 2nd pr

What has changes since #2466? 

- Switched from the bad implementation to avoid the windows error to the "babysitter" version

How does performance look?

- Its a little bit worse than current implementation as we need the "babysitter" thread. 
- Joining  headless server (on same machine, headless server ran on current implementation) with 1.5 GB of assets: the times where ~1m and 25s for current implementation and 1m 37s for this branch. so about 12 seconds slower. 
- Improving this would probably be complicated, so I suggest to take that hit until `receiveTimeout` is updated in zig to work on windows

How will this continue?

- I plan on splitting this pr up into these parts:

1. update #2259 to use `Io.net.IpAddress` as the inner type (either by replacing that pr or making a follow up pr when it gets merged)
2. add the "babysitter" method (maybe make a generlized version for utils?)
3. the actual switch (would probably be just this pr)


IntegratedQuantum:
Would it be possible to still use the faster method on Linux, together with a TODO and an issue?
Also could you please measure latency changes and impact on CPU usage as well? If it continuously uses up to 3 threads, it might be pretty bad, but if 2 are mostly idle, then it's not a big deal.

Wunka:
> Would it be possible to still use the faster method on Linux, together with a TODO and an issue? Also could you please measure latency changes and impact on CPU usage as well? If it continuously uses up to 3 threads, it might be pretty bad, but if 2 are mostly idle, then it's not a big deal.

Added that it uses `receiveTimeout` when not on windows. 
It seems to use more than just 2/3 threads
Example: CPU usage on threads filterd by network threads with the commit before I diverge:
image
And after my changes:
image

(on both I joined a world after 30s)

when it comes to latency. Here is when using debug mode: (all players on same machine)
image

and here on in release safe:

image

IntegratedQuantum:
Why does it show so many threads in the second one?

Wunka:
> Why does it show so many threads in the second one?

That is something I would also like to know. Maybe it is to fast and cannot reuse an old one directly? (I have no idea if that makes even sense)

Wunka:
ok.... I sadly need to confirm that I made the RTT screenshot with the newest version of my pr without realizing. So it had the "when no windows use the correct method" implementation.... 
Here is the correct value for release safe compiled joined user:
image
.... not good

Wunka:
I think it might be better to just try making a wrapper around `std.Io.net.Socker` to fallback to `winsock2` when on windows. This "babysitter" thingy seems to not be good enough

Wunka:
Or what do you think @IntegratedQuantum ? We would atleast get the as it seems performance increase on linux.

IntegratedQuantum:
> We would atleast get the as it seems performance increase on linux.

I'd doubt those numbers actually. They also use poll and the only difference is that they use recvmsg instead of recvfrom, and from what I can tell recvfrom is less generic and thus probably as fast or faster. But overall the IO implementation has probably more overhead.

The main motivation to switch to it, is so we don't have to maintain the netcode, but if we'd need to maintain Windows anyways which was also the bigger pain to work with, then I see no point.

Wunka:
Then I will let this go for now until (hopefully) the next zig release. But I will still do the IpAddress rewrite because I still think this would help for #859 (and / or for the later Io.net rewrite). I have already a working version on top of #2259

PR #3056, opened 2026-05-08
use `std.Io.net.IpAddress` instead of making our own

Wunka:
This can be seen as a follow up or replacement for #2259 
It is also a preparation for #859 and #2338

The last two commits switched from `Ip4Address` to the generlized `IpAddress` so that can be reverted if wanted


Wunka:
@codemob-dev I have taken the liberty to switch the inner type. I don't know if there any requirements for the `SocketAddress` to work good with #2180 so if you could take a short look?

IntegratedQuantum:
I think this should be done after #2259, there are some changes in here that are breaking (→changing the location of the question mark) and should be discussed.

Wunka:
Ok. I will then rebase when #2259 is merged. Just for documentation (not to already start the discussion) the change of the question mark was done so I don't need to implement the format function myself again just to put the ? before the port. That's the only reason

src/network.zig
```diff
@@ -115,10 +116,11 @@ const Socket = struct {
...
 	}
 
-	fn send(self: Socket, data: []const u8, destination: Address) void {
-		const addr = posix.sockaddr.in{
-			.port = @byteSwap(destination.port),
-			.addr = destination.ip,
+	fn send(self: Socket, data: []const u8, destination: SocketAddress) void {
+		const addr = blk: {
+			var posixAddr: std.Io.Threaded.PosixAddress = undefined;
+			_ = std.Io.Threaded.addressToPosix(&destination.address, &posixAddr);
```
IntegratedQuantum: The length (return value) should probably be passed into the other functions instead of using `@sizeOf`

src/network.zig
```diff
@@ -237,24 +238,64 @@ pub fn deinit() void {
...
+	fn parseInner(string: []const u8, defaultPort: ?u16) !struct { ip: []const u8, isSymmetricNAT: bool, port: u16 } {
+		var parts = std.mem.splitScalar(u8, string, ':');
+		const ip = parts.first();
+		var portString = parts.rest();
+		var isSymmetricNAT = false;
+		if (portString.len > 0 and portString[portString.len - 1] == '?') {
+			isSymmetricNAT = true;
+			portString = portString[0 .. portString.len - 1];
+		}
+		const port = try (std.fmt.parseUnsigned(u16, portString, 10) catch |err| (defaultPort orelse err));
```
IntegratedQuantum: Instead of ignoring the error when a defaultPort is present, this should only use the default port when the portString.len == 0

IntegratedQuantum (changes requested):

src/network.zig
Wunka: both done

IntegratedQuantum (changes requested):
Since the network code is quite attractive to windows issues and you are using new standard library functions, please also find someone to test this on Windows

Wunka:
This was tested on the last playsession by a windows user, no problems were reported

IntegratedQuantum (approved):

PR #3057, opened 2026-05-08
Craftable meteorite block

### LINKED ISSUE #3016: Meteorite blocks are uncraftable

they exist, you can't craft them

codemob-dev:
Fixes #3016

### DIFF
--- a/assets/cubyz/recipes/special_recipes.zig.zon
+++ b/assets/cubyz/recipes/special_recipes.zig.zon
@@ -48,6 +48,10 @@
 		.output = "cubyz:iron_block",
 	},
 	.{
+		.inputs = .{"4 cubyz:meteorite_ingot"},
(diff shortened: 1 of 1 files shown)


careeoki (changes requested):
It should also be `.reversible` so you can get your ingots back.

codemob-dev:
> It should also be `.reversible` so you can get your ingots back.

Looking at it, most of the ingot→block recipes aren't reversible, I might want to fix that in another PR.

careeoki:
They are reversible, but it's in a separate recipe rather than using the bool.

codemob-dev:
> They are reversible, but it's in a separate recipe rather than using the bool.

Yeah, just noticed that.

IntegratedQuantum (approved):

PR #3060, opened 2026-05-08
Make SelectionCapabilities a bitmask

DellieDelta:


src/blocks.zig
```diff
@@ -70,49 +70,50 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-const SelectionCapabilities = struct {
-	capabilities: ?[]const Capability,
+const SelectionCapabilities = packed struct(u3) {
+	never: bool = false,
+	always: bool = false,
```
IntegratedQuantum: I think it's a bit odd to have it like this. Maybe a tagged union would make more sense?
```zig
union(enum) {
	never: void,
	always: void,
	custom: packed struct(u1) {
		toolEffective: bool,
	},
}
```

IntegratedQuantum (changes requested):

src/blocks.zig
```diff
@@ -70,49 +70,52 @@ pub const Ore = struct {
 	seed: u64,
 };
 
-const SelectionCapabilities = struct {
-	capabilities: ?[]const Capability,
+const SelectionCapabilities = union(enum) {
+	const backingType = u1;
```
IntegratedQuantum: types should start with capital letter
```suggestion
	const BackingType = u1;
```

src/blocks.zig
```diff
@@ -70,49 +70,52 @@ pub const Ore = struct {
...
+	},
 
-		pub fn allowsSelectionByItem(self: Capability, block: Block, item: Item) bool {
-			return switch (self) {
-				.toolEffective => item == .proceduralItem and item.proceduralItem.isEffectiveOn(block),
-			};
-		}
-	};
+	pub const neverSelectable: SelectionCapabilities = .{.never = {}};
+	pub const alwaysSelectable: SelectionCapabilities = .{.always = {}};
```
IntegratedQuantum: These are redundant, since you can just use `.never`, coercion from enum literals to void union fields is possible.

src/blocks.zig
```diff
@@ -70,49 +70,52 @@ pub const Ore = struct {
...
 	pub fn allowsSelectionByItem(self: SelectionCapabilities, block: Block, item: Item) bool {
-		if (item == .baseItem) {
-			const base = item.baseItem;
-			if (base.block() == block.typ or std.mem.eql(u8, base.id(), "cubyz:selection_wand")) {
-				return true;
-			}
-		}
+		if (self == .never) return false;
+		if (self == .always) return true;
+		return switch (item) {
```
IntegratedQuantum: Should be hidden behind `if (self.custom.toolEffective) {`

IntegratedQuantum (changes requested):

src/blocks.zig
```diff
@@ -70,49 +70,52 @@ pub const Ore = struct {
...
+	const BackingType = u1;
 
-	const Capability = enum(u8) {
-		toolEffective,
+	always: void,
+	custom: packed struct(BackingType) {
+		toolEffective: bool = false,
+	},
 
-		pub fn allowsSelectionByItem(self: Capability, block: Block, item: Item) bool {
```
IntegratedQuantum, line 79: I think it would be nicer to still have this function on the `custom` struct with an if for each condition, so we have a clearer distinction between the features, and the stuff that's currently hardcoded.

IntegratedQuantum (changes requested):
Just to be sure: Did you test some of the edge cases after the new refactor?

src/blocks.zig
DellieDelta: Good one!

DellieDelta:
> Just to be sure: Did you test some of the edge cases after the new refactor?

Yess, I do my best to test it so that it doesn't have to be retested in-game anymore

src/blocks.zig
IntegratedQuantum: The way I imagined it was just a single line, with all the other stuff that isn't relevant for the check, on the outside
```
pub fn allowsSelectionByItem(self: @This(), block: Block, item: Item) bool {
    if (self.toolEffective) {
        if (...) return true;
    }
    // Then you could easily put future checks here guarded by their own ifs
    return false;
}
```

DellieDelta: Ahh okay, I misunderstood

src/blocks.zig
```diff
@@ -70,49 +70,56 @@ pub const Ore = struct {
...
-			return switch (self) {
-				.toolEffective => item == .proceduralItem and item.proceduralItem.isEffectiveOn(block),
-			};
+		pub fn allowsSelectionByItem(self: @This(), block: Block, item: Item) bool {
+			if (@as(BackingType, @bitCast(self)) == 0) return false;
+
+			if (self.toolEffective) {
+				if (item == .proceduralItem) {
+					return item.proceduralItem.isEffectiveOn(block);
```
IntegratedQuantum: Should be structured like this, so it does not skip checks below it.
```suggestion
				if (item == .proceduralItem and  item.proceduralItem.isEffectiveOn(block)) {
					return true;
```

src/blocks.zig
```diff
@@ -70,49 +70,56 @@ pub const Ore = struct {
...
+	always: void,
+	custom: packed struct(BackingType) {
+		toolEffective: bool = false,
 
-		pub fn allowsSelectionByItem(self: Capability, block: Block, item: Item) bool {
-			return switch (self) {
-				.toolEffective => item == .proceduralItem and item.proceduralItem.isEffectiveOn(block),
-			};
+		pub fn allowsSelectionByItem(self: @This(), block: Block, item: Item) bool {
+			if (@as(BackingType, @bitCast(self)) == 0) return false;
```
IntegratedQuantum: By the way I you should be able to do `self == .{}` or `self == @This(){}`, since it's default-initialized to false.
Then you don't need the BackingType variable

src/blocks.zig
```diff
@@ -70,49 +70,56 @@ pub const Ore = struct {
...
+
+			return false;
 		}
-	};
+	},
 
-	pub const alwaysSelectable: SelectionCapabilities = .{.capabilities = null};
+	const Capability = enum {
+		toolEffective,
+	};
```
IntegratedQuantum: By the way there is also `std.meta.FieldEnum(comptime T: type)`, you don't need to use it here yet, but just so you know this exists.

IntegratedQuantum (changes requested):

src/blocks.zig
DellieDelta: That's very handy, I'll use it here :)

IntegratedQuantum (approved):

PR #3061, opened 2026-05-09
[Accessories Part I] Slots!

codemob-dev:
The icons and slots are just placeholders for testing until @careeoki makes some better textures. Currently, any item can be placed in the slots for testing purposes but can change that to allow no items if needed.

Screenshot_2026-05-08_20-32-06


Wbjitscool:
there's check issues it needs some fixes according to github

Wunka:
Oh no. Now that even such a pr is out I finnally need to go back to the callbacks (no we have two accessorie pr series)

Argmaster:
Can we have the items in columns instead of rows? Inventory already takes a lot of horizontal space.
Alternatively we could make it 1 row so it can be placed above inventory.
General point is to design an optimal placement for that window.

Wbjitscool:
it's still having an issue with 1 check failing

codemob-dev:
> Can we have the items in columns instead of rows? Inventory already takes a lot of horizontal space.

Like this?
Screenshot_2026-05-09_08-39-48

Wbjitscool:
yeah

Argmaster:
Definitely suboptimal layout for presentation, but yeah I guess thats what I described.
Ok, an what about horizontal layout? 

```
<ring> <ring> <ring> <ring> <ring> <ring> <empty> <necklace> <necklace>
```

codemob-dev:
> Definitely suboptimal layout for presentation, but yeah I guess thats what I described.
Ok, an what about horizontal layout?

I feel like this would scale really badly once we add armor and other slots

Argmaster:
> I feel like this would scale really badly once we add armor and other slots

Why? Just add a row above or below. Well, unless you want to have character presentation.
In that case, I would say

|        |       |      |          |        |            |
|--------|-------|------|----------|--------|------------|
| player | model | here | helm     | ring 1 | necklace 1 |
|        |       |      | cuirass  | ring 2 | necklace 2 |
|        |       |      | gauntlet | ring 3 | other 1    |
| player | model | here | cuisses  | ring 4 | other 2    |
| x      | y     | z    | greaves  | ring 5 | other 3    |


Which is almost what you made...
But I would ask for placeholders then, so we have the general idea there.
Preferably as inactive slots.
Also so we can reality check my stupid ideas

codemob-dev:
> Why? Just add a row above or below. Well, unless you want to have character presentation.
In that case, I would say

How would you suggest we specify the slot order in the zon?

careeoki:
This is how I would think to lay it out. @IntegratedQuantum 's input here would be helpful, like specifically how many slots do we want, what kind of labeled slots do we want
image

Wbjitscool:
> This is how I would think to lay it out. @IntegratedQuantum 's input here would be helpful, like specifically how many slots do we want, what kind of labeled slots do we want image

that looks better

IntegratedQuantum:
> This is how I would think to lay it out. @IntegratedQuantum 's input here would be helpful, like specifically how many slots do we want, what kind of labeled slots do we want

I think addons should be able to add custom accessory slots, so we should have a GUI that is expandable. I would suggest maybe putting the main body slots on the left and have a scrollable list on the right (maybe 2 slots wide, following your desing) with all the other slots.

> what kind of labeled slots do we want

For now I would suggest to keep the list small (mabye 1 necklace slot for the healing accessories and 2 ring slots) and expand it when we need to.

codemob-dev:
> I think addons should be able to add custom accessory slots, so we should have a GUI that is expandable. I would suggest maybe putting the main body slots on the left and have a scrollable list on the right (maybe 2 slots wide, following your desing) with all the other slots.

Would the armor slots be hardcoded then? If not, how would we specify their location in the zon?

IntegratedQuantum:
You'll probably need some sort of priority hint for the order, you can just make the first accessories appear in the four armor slots

codemob-dev:
> I think addons should be able to add custom accessory slots, so we should have a GUI that is expandable. I would suggest maybe putting the main body slots on the left and have a scrollable list on the right (maybe 2 slots wide, following your desing) with all the other slots.

Not exactly what you asked, but this is what I came up with:
image

The window currently expands infinitely to the left as you add more slots, and you can specify an index to try to order the slots by.

Once I get textures from @careeoki and get armor slots it will look a lot better.

careeoki:
[Alright, here you go, as requested](https://github.com/codemob-dev/Cubyz/pull/1)

image

Next you should do this: 

> You'll probably need some sort of priority hint for the order, you can just make the first accessories appear in the four armor slots

codemob-dev:
Got it working!

image

assets/cubyz/accessory_slots/feet.zig.zon
```diff
@@ -0,0 +1,5 @@
+.{
+	.texture = "feet.png",
```
IntegratedQuantum, line 2: Please also give them some tags

src/entityComponent/accessories.zig
```diff
@@ -0,0 +1,93 @@
...
+
+	pub fn getAccessories(entityId: u32) ?*Inventory.ClientInventory {
+		return &(components.get(@enumFromInt(entityId)) orelse return null).accessories;
+	}
+
+	pub fn load(entityId: u32, reader: *utils.BinaryReader, version: u32) main.entity.EntityComponentLoadError!void {
+		if (version != entityComponentVersion) return error.InvalidComponentVersion;
+		const accessories = &components.add(main.globalAllocator, @enumFromInt(entityId)).accessories;
+		accessories.* = .init(main.globalAllocator, items.accessory_slots.getTotalSlotCount(), .serverShared, .{.playerAccessories = entityId}, .{});
+		accessories.super.fromBytes(reader);
```
IntegratedQuantum, line 44: Not necessary, init will already load it using the sync system.

IntegratedQuantum (changes requested):
A few things to consider:
- you added migrations, but you are not using them
- what about accessory slots that aren't unlocked by default or for every player/player type?

Argmaster:
What if we want a material that adds more accessory slots 🙃

PR #3062, opened 2026-05-09
Parachute

Crepestrom:
image

It slows your fall


Crepestrom:
about 2/3rds speed as walking

Crepestrom:
also adds physicsproperty to baseitem incase we want other items to affect physics

the friction and air movementspeed are split off into functions to work with the coming accesories pr to hook into

Wbjitscool:
nice

Argmaster:
This is very much something that should be done through attaching ECS components and using proper item callbacks. We can't indefinitely add new fields to items/blocks, it doesn't scale.

As for the idea itself, it needs polishing and balancing.
Maybe flight time limit, possibly by consuming energy or durability during flight, so you can't just fly from the top of the world down to the void.
Maybe some kind of penalty of pulling out parachute with max velocity already reached should be added, eg. burst of energy use at the start, burst of damage to item, so you can't just fall 40k blocks and pull the parachute at the end at **no cost**.
Another way to balance it is to make it very slow, so people have to time it as close to the earth to avoid enjoying 5 minutes of falling 100 blocks, to involve skill.
It should be procedurally crafted, so we can have a mix of mentioned tools depending on materials used.
It should fill different niche than glider, with glider enabling fast horizontal travel without slowing you down too much, so you will still take damage if you fly straight down/into a wall. Parachute would allow you to slow down at high speeds.

At the same time I can imagine others may just say "no we have glider planned"

Crepestrom:
i just made this to be similar to the chisel where we have a test version now and we can add procedural crafting later

PR #3066, opened 2026-05-09
Remove all uses of @cImport

### LINKED ISSUE #2930: Replace `@cImport`, it was deprecated upstream

AfmanS:
Closes #2930 
Condition.zig will be removed when we increase the version so it may not matter in that case


src/utils/Condition.zig
```diff
@@ -144,9 +144,7 @@ const SingleThreadedImpl = struct {
 const WindowsImpl = struct {
 	condition: c.CONDITION_VARIABLE = .{},
 
-	const c = @cImport({
-		@cInclude("windows.h");
-	});
+	const c = @import("c");
```
IntegratedQuantum: Please keep this as is, this file is scheduled to be remove during the next update anyways

src/c.h
```diff
@@ -25,3 +25,39 @@
 // used for reading and writing image files
 #include <stb/stb_image.h>
 #include <stb/stb_image_write.h>
+
+// used for transmitting and loading models and textures
```
IntegratedQuantum: We only use it for model loading
```suggestion
// used for loading models
```

src/c.h
```diff
@@ -25,3 +25,39 @@
...
+#include <cgltf.h>
+
+// used for virtual memory management on Windows
+#ifdef _WIN32
+#include <memoryapi.h>
+#endif
+
+// used for SSL and TLS
```
IntegratedQuantum: SSL is deprecated since 2015, and we don't use it
```suggestion
// used for TLS
```

src/c.h
```diff
@@ -25,3 +25,39 @@
...
+#include <unistd.h>
+#include <errno.h>
+#endif
+
+// used for audio
+#include <miniaudio.h>
+#define STB_VORBIS_HEADER_ONLY
+#include <stb/stb_vorbis.h>
+
+// used to interact with shaders and SPIR-V
```
IntegratedQuantum:
```suggestion
// used to compile shaders to SPIR-V
```

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #3078, opened 2026-05-11
Add an indicator for if you are on the ground or not to the crosshair

### LINKED ISSUE #2499: Indicator for being on a block
I think one of the reasons for the dislike of low air mobility in games (compared to real life), is because a crucial information is missing.
In real life we can see our feet in our peripheral vision, this allows our brain (unconsciously) to sync up our feet with a target line or ledge on the ground. To confirm this I did an experiement: I blocked my downwards peripherals with my hands, and it made me much less accurate at hitting and jumping from a line on the ground.

So to solve this I propose an indicator for whether or not the player is currently on the ground. Together with coyote time and coyote frictions (see https://github.com/PixelGuys/Cubyz/issues/1969#issuecomment-3403227249), this should be enough to give the player some time to react when stepping of a block.
As for how that would look, I propose just a small horizontal bar below the crosshair that gets toggled on or off.

codemob-dev:
Fixes #2499 
image
image


IntegratedQuantum:
It would be nice if some people could try this, I also had a prototype of this when I worked on more air control, and it wasn't actually as intuitive or as useful as I hoped it would be.

DellieDelta:
For me it doesn't add much compared to not having anything

careeoki (changes requested):
Yeah, I'm not really sure if this helps much at all. Mostly it feels distracting because it keeps flickering on and off while jumping and climbing around.
Also it being on the crosshair makes it feel like it would have something to do with selecting/placing blocks, when it doesn't.

IntegratedQuantum:
I also played around with a few variants, where it's a bar that gets larger depending on how much area you are on, but it also didn't feel intuitive. So I will close this and the issue for now.

PR #3081, opened 2026-05-13
More user-friendly argument parsing

H41ogen:
Use tokenizeScalar instead of splitScalar in argparse.zig. This allows for trailing and duplicate spaces.

### DIFF
--- a/src/argparse.zig
+++ b/src/argparse.zig
@@ -320,10 +320,10 @@ test "x or xy negative empty" {
 	try std.testing.expectEqualStrings(
 		\\---
 		\\x
-		\\Expected a number for <x>, found ""
+		\\Missing argument at position <x>
 		\\---
 		\\xy
-		\\Expected a number for <x>, found ""
+		\\Missing argument at position <x>
 		\\---
 	, errors.items);
 	try std.testing.expectError(error.ParseError, resultOrError);
--- a/src/argparse.zig
+++ b/src/argparse.zig
@@ -64,12 +64,12 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 			.parse => error{ParseError}!T,
 		} {
 			var result: T = undefined;
-			var split = std.mem.splitScalar(u8, args, ' ');
+			var tokens = std.mem.tokenizeScalar(u8, args, ' ');
 
 			var tempErrorMessage: ListUnmanaged(u8) = .{};
 			defer tempErrorMessage.deinit(main.stackAllocator);
 
-			var nextArgument: ?[]const u8 = split.next();
+			var nextArgument: ?[]const u8 = tokens.next();
 
 			inline for (s.fields) |field| {
 				const value = resolveArgument(field.type, allocator, field.name[0..], nextArgument, &tempErrorMessage);
@@ -85,7 +85,7 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 				} else {
 					@field(result, field.name) = value catch unreachable;
 					tempErrorMessage.clearRetainingCapacity();
-					nextArgument = split.next();
+					nextArgument = tokens.next();
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


codemob-dev:
We don't want trailing spaces though.

H41ogen:
It's a QOL thing. I've seen people multiple times paste a command and had a trailing space, or accidentally putting a trailing space, and not realizing why the command didn't work because the error message is unclear

codemob-dev:
> It's a QOL thing. I've seen people multiple times paste a command and had a trailing space, or accidentally putting a trailing space, and not realizing why the command didn't work because the error message is unclear

True, yeah.

H41ogen:
speaking of that, i completely forgot about trailing newlines; those happen even more often than trailing spaces when pasting commands

Wunka:
We could also trim the input. But I agree we probably should also allow multiple spaces to seperate arguments

Wunka:
But from the test results it seems your code changes the logic a little

H41ogen:
> But from the test results it seems your code changes the logic a little

yeah, i'm trying to figure out exactly what's going on with that

H41ogen:
> > But from the test results it seems your code changes the logic a little
> 
> yeah, i'm trying to figure out exactly what's going on with that

It's because now parsing an empty string returns 0 arguments instead of 1 argument == to `""`. Unless I'm wrong, this is expected behavior.

Wunka:
> > > But from the test results it seems your code changes the logic a little
> > 
> > 
> > yeah, i'm trying to figure out exactly what's going on with that
> 
> It's because now parsing an empty string returns 0 arguments instead of 1 argument == to `""`. Unless I'm wrong, this is expected behavior.

Oh my god. I hated this behavior. thanks this is now my favorite pr.

H41ogen:
dunno if i want to include stripping things like newlines in this pr, since I don't know what would be the most optimal way to do that

Wunka (approved):

src/argparse.zig
Argmaster: If you look at the original PR, you will see that the reason for this double check is the behavior of splitScalar which for an empty string returns an empty string first time next() is used. Is it still the case with tokenizeScalar?

Argmaster:
So, we will still need to replace tokenizeScalar with custom split iterator, since we need to support quotes and parentheses, but I don't see an issue with adding this small improvement before improved split Iterator.

src/argparse.zig
```diff
@@ -320,10 +320,10 @@ test "x or xy negative empty" {
 	try std.testing.expectEqualStrings(
 		\\---
 		\\x
-		\\Expected a number for <x>, found ""
+		\\Missing argument at position <x>
```
Argmaster, line 323: That's weird, why did that change?
Did this fail in CI before?

Wunka: Yes. see https://github.com/PixelGuys/Cubyz/pull/3081#issuecomment-4444833822
the `splitIterator` gives a `""` as the first `next` when given `""` as argument. while the tokenizer gives a `null` value (better in my opinion)

Argmaster: ah I see, yeah I see that now.

Argmaster: So the https://github.com/PixelGuys/Cubyz/pull/3081#discussion_r3240169055 should be changed no?

src/argparse.zig
Wunka: true. with the tokenizer `""` cannot happen anymore so that case should be removed

IntegratedQuantum (changes requested):
Please apply changes suggested by Argmaster in https://github.com/PixelGuys/Cubyz/pull/3081/changes#r3240169055

H41ogen:
> Please apply changes suggested by Argmaster in https://github.com/PixelGuys/Cubyz/pull/3081/changes#r3240169055

sorry, i got sick twice and lost one of my jobs. it should be good now

IntegratedQuantum (approved):

PR #3082, opened 2026-05-13
Replace the only use of ** for array initialization with @splat

H41ogen:
Initialization of arrays with ** is getting removed in Zig 0.17, and there is only one remaining use of it in Cubyz's codebase.


IntegratedQuantum:
This file will be removed in the next Zig update

PR #3084, opened 2026-05-16
disallow shifclick on creative window when not creative

Wunka:
While testing something different, I saw that I could shift-click take items from the creative window even though I was not in creative mode and had cheats not enabled. 
This PR fixes this.

### DIFF
--- a/src/sync.zig
+++ b/src/sync.zig
@@ -1173,6 +1173,8 @@ pub const Command = struct { // MARK: Command
 		}
 
 		fn run(self: FillAnyFromCreative, ctx: Context) error{serverFailure}!void {
+			if (ctx.side == .server and ctx.user != null and ctx.gamemode != .creative) return;
+			if (ctx.side == .client and ctx.gamemode != .creative) return;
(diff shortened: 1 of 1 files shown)


TyrrelMeter:
Good thing you spotted that, that seems like a pretty big bug.

IntegratedQuantum (changes requested):
Couldn't this just be a single `if (ctx.gamemode != .creative) return`?

Wunka:
> Couldn't this just be a single `if (ctx.gamemode != .creative) return`?

True. I copied it just over from the other creative sync command. but that one had reasons which here don't exist to be so complicated.

Changed!

IntegratedQuantum:
Could you also simplify the other use-case?

Wunka:
> Could you also simplify the other use-case?

that one has a reason to be so complicated. it is called from the server with `user = null` when a player collects a drop

IntegratedQuantum:
Then it should be called with gamemode = .creative

Wunka:
it is.... sorry I should have just tested it out. Changed!

IntegratedQuantum (approved):

PR #3098, opened 2026-05-21
Tall mountain revamp

TyrrelMeter:
Sorry I had to resubmit this pull request, I'm not good with the github stuff so I made some mistakes.
I've been thinking about current cubyz mountain generation, and realized that we can do better. With better generation settings I've created an even better tall mountain, with more trees to make it more realistic. I'm not completely happy or settled on the design yet, but I can say for sure it is an improvement. There are also some generation bugs with the mountains I'll need to fix.

Cubyz 0 3 0-dev 5_2_2026 6_47_39 PM
Cubyz 0 3 0-dev 5_2_2026 6_46_45 PM
Cubyz 0 3 0-dev 5_2_2026 6_46_57 PM
Cubyz 0 3 0-dev 5_2_2026 6_46_53 PM


TyrrelMeter:
@careeoki If you could review this(and the cold mountain PR) within the next week so I can work on it when I get back from vacation that would be great.

assets/cubyz/biomes/tall_mountain/slope2.zig.zon
```diff
@@ -1,28 +1,81 @@
...
+		},
+		.{
+			.id = "cubyz:sbb",
+			.structure = "cubyz:tree/birch/silver/2",
+			.placeMode = .degradable,
+			.chance = 0.015,
+		},
+		.{
+			.id = "cubyz:simple_vegetation",
+			.block = "cubyz:grass_vegetation",
```
careeoki: Is the mismatched grass plants here intentional? Feels like it isn't since there's no cold grass plants. This applies to some of the other layers as well.

Image

careeoki (changes requested):
Just one thing I noticed, it seems good otherwise.

assets/cubyz/biomes/tall_mountain/slope2.zig.zon
TyrrelMeter: No it was not. Thanks for pointing that out, it is fixed now.

ikabod-kee:
I think if you used some SDFs to carve into the mountain or extend off of it, you could create some interesting flat "resting points" for the player and mobs. I do also think the base of the mountain could be a little wider. Also for a mountain, it is missing a lot of boulders and areas that don't have soil.

TyrrelMeter:
do we have the ability to generate surface biome SDFs yet?

careeoki (commented):
It seems fine to me. 
@ikabod-kee please re-review as you requested changes.

ikabod-kee:
Sorry for the delay! I'll get on it! o7

ikabod-kee (changes requested):
This is definitely better, though I will suggest making the mountains different heights, as all mountains I tested end at around the same height. It would also be nice to have multiple mountains that don't exactly have a peak.

TyrrelMeter:
I've made it so that the mountains can generate at different heights by changing it so that the mountain can stop at a slope 1 or 2 below the summit

ikabod-kee:
You forgot to push the commit

TyrrelMeter:
Wait, it worked once but now it doesn't give me a bit

TyrrelMeter:
Well that doesn't affect it on my end

TyrrelMeter:
alright it should be good now, turns out it just wasn't a low enough chance

ikabod-kee:
image

1. Mountains ending at slope6 or slope5 look quite botched, as if they're not finished. I suggest making a seperate sub-biome that replaces slope6/slope5, giving it a proper "cap."
2. The heights of the mountains still are quite similar. In the files, it seems that each sub-biome has the same min and max heights. It'd be nice to have a greater range of variation.
3. Instead of having a set radius, it'd be nice to see a range of radius from this biome.

TyrrelMeter:
(it shouldn't end at slope 5)

ikabod-kee:
Why not? Why not end at 4 or 3? Even 2?

TyrrelMeter:
Well its not much of a tall mountain if it ends too close to the ground

TyrrelMeter:
If it ends below the snow line then I think it should be a separate biome

TyrrelMeter:
Also tried making the radius a range but it breaks the whole generation by turning the mountain into a brick. Sorry. I've updated the height range, trying 3 separate summits, one for each of the top 3 slopes.

ikabod-kee (approved):
approved

TyrrelMeter:
There have been Compilation check failures and I've looked through them but I don't know what it wants me to change

IntegratedQuantum:
It wants you to use tabs instead of spaces for indentation, you can configure this in your editor.

assets/cubyz/biomes/tall_mountain/slope7.zig.zon
```diff
@@ -0,0 +1,49 @@
...
+		.mountain,
+		.pine,
+	},
+	.minHeight = 898,
+	.maxHeight = 898,
+	.smoothBeaches = true,
+	.radius = 80,
+	.mountains = 60,
+	.chance = 0,
+	.minSubBiomeCount = 0,
```
IntegratedQuantum: This variable doesn't exist

IntegratedQuantum:
Have you tried different slopes? I feel like the mountains are a bit too pointy.

TyrrelMeter:
I have tried like 10+ different slopes heights widths and mountainous to maximize look. I can tweak it further but without making the biome shorter or wider there isn't to much I can do. I've made summit three less this because I realized it was to thin.

IntegratedQuantum:
I don't just mean the summit, I mean the whole thing.

TyrrelMeter:
I've made it so that the mountains *should* look better now, but you should be the judge of that @IntegratedQuantum

IntegratedQuantum:
I just noticed I looked at an outdated version, and of course it looks nicer now.

assets/cubyz/biomes/tall_mountain/base.zig.zon
```diff
@@ -1,26 +1,104 @@
 .{
-	.properties = .{.mountain},
-	.radius = 320,
+	.properties = .{
+		.mountain,
+		.oak,
+		.birch,
+	},
 	.minHeight = 128,
 	.maxHeight = 128,
```
IntegratedQuantum, line 10: The base should not be underground:
Image

To avoid this you could use the `keepOriginalTerrain` trick to add to the base terrain instead of overwriting it with a fixed value, this is e.g. used in the stone rock subbiome.

here you could for example set the height of all biomes to `1000` and use `keepOriginalTerrain` as a percentage of how close it is to the top, e.g. here you could use `.keepOriginalTerrain=0.9` to make it 10% of the way to the top.

IntegratedQuantum (changes requested):

assets/cubyz/biomes/tall_mountain/slope7.zig.zon
TyrrelMeter: I don't even remember how that got there

assets/cubyz/biomes/tall_mountain/base.zig.zon
TyrrelMeter: Should be fixed now.

IntegratedQuantum:
CI failed

TyrrelMeter:
No it didn't @IntegratedQuantum

TyrrelMeter:
Oh please quantum reviewlop my PR

IntegratedQuantum (changes requested):
Please add tags to ensure that rare structures can spawn,
All snowy areas should have `.tags = .{.snowy},`, all areas with trees should have tags resembling their tree types.

TyrrelMeter:
Alright I've updated the tags.

IntegratedQuantum:
Please update all your cobbles and stones, see #3254

assets/cubyz/biomes/tall_mountain/base.zig.zon
```diff
@@ -1,26 +1,105 @@
 .{
-	.properties = .{.mountain},
-	.radius = 320,
+	.properties = .{
+		.mountain,
+		.oak,
+		.birch,
```
IntegratedQuantum: These are climate properties, tags need to be separate, in a `tags` field.

IntegratedQuantum (changes requested):

assets/cubyz/biomes/tall_mountain/base.zig.zon
TyrrelMeter: Everything should be in order now.

IntegratedQuantum: No it isn't. Please remove the duplicate entries. climate properties should only be mentioned in the properties field, tags should only be mentioned in the tags field.

TyrrelMeter: So I removed tree types + snowy from properties and mountains from tags, is that good?

IntegratedQuantum (approved):

IntegratedQuantum:
@TyrrelMeter this PR caused an error.
Every time you remove a biome (`cubyz:tall_mountain/peak` in this case) you must add a migration for it. Otherwise it will be missing in existing worlds, printing an error.

TyrrelMeter:
Oh, sorry, I haven't done stuff with upgrading older biomes before. Is there something I need to do to fix this?

IntegratedQuantum:
I'll take care of it, but please be more careful next time.
I'll also try to finally get some automated testing for this done.

TyrrelMeter:
I will be more careful.

Issue #3073, opened 2026-05-11
Update commands and commands helper to the new parser pattern


Wunka:
With #1425 we got an argument parser for commands. All commands should be migrated to it and new ones should use it.
For commands that need a parser feature which currently doens't exist, first the parser should be updated

Command helper functions/structs that need to be integrated:

- [x] `parseCoordinates` #3103 
- [x] `Target` #3103

General Commands that need to be updated:

- [x] `clear` #3105 
- [x] `gamemode` #3106 
- [x] `help` #3117
- [x] `invite` #3113
- [x] `kick` #3108 
- [x] `kill` #3107
- [x] `particles` #3390 
- [x] `spawn` #3110
- [x] `tickspeed` #3114
- [x] `tp` #3104 

Permission Command that needs update:
- [x] `perm` #3112 

World edit commands that need update:
- [x] `undo` #3327 
- [x] `redo` #3328 
- [x] `pos1` #3323 
- [x] `pos2` #3324 
- [x] `deselect` #3160 
- [x] `copy` #3145 
- [x] `paste` #3164 
- [x] `blueprint` #3138 
- [x] `rotate` #3386 
- [x] `set` #3387 
- [x] `mask` #3161 
- [x] `replace` #3388 
- [x] `toggledecay` #3389 

things that need removal after certain migrations are done:
- [x] `parseAxis` used by:  `particles`, `spawn`, `tp`
- [x] `parseCoordinates` used by: `particles`, `spawn`, `tp`
- [x] `parsePlayerIndexAndIncreaseRefCount` used by: `kick`, `spawn`, `perm`, `tp`

things that need to be changed after migration is done:
- [x] instead of giving `args: []const u8` as parameter, commands should directly get a paramter from type `Args`

Wunka:
as I would like to work on #3223. @Argmaster are you going to continue working on the wordedit commands soon or can I take over?

Argmaster:
> as I would like to work on #3223. @Argmaster are you going to continue working on the wordedit commands soon or can I take over?

If you want to, please do!
Currently not really capable of doing that myself due to weather.

PR #3103, opened 2026-05-22
add Axis and TargetArgs Helper struct which are conform to the argsparser

Wunka:
Progress towards #3073
Example of this used can be seen in #3104

### DIFF
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -75,6 +75,20 @@ pub const Axis = struct {
 	}
 };
 
+pub const Coordinates = struct {
+	x: Axis,
+	y: Axis,
+	z: ?Axis,
+
+	pub fn toVec(self: Coordinates, source: *User) main.vec.Vec3d {
+		return .{
+			self.x.toValue(source.player().pos[0]),
+			self.y.toValue(source.player().pos[1]),
+			if (self.z) |z| z.toValue(source.player().pos[2]) else source.player().pos[2],
+		};
+	}
+};
+
 fn parseAxis(arg: []const u8, playerPos: f64, source: *User) !f64 {
 	const hasTilde = if (arg.len == 0) false else arg[0] == '~';
 	const numberSlice = if (hasTilde) arg[1..] else arg;
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -49,11 +49,11 @@ pub fn execute(msg: []const u8, source: *User) void {
 	}
 }
 
-pub const Axis = struct {
+pub const Coordinate = struct {
 	hasTilde: bool,
 	value: f64,
 
-	pub fn parse(allocator: NeverFailingAllocator, _: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!Axis {
+	pub fn parse(allocator: NeverFailingAllocator, _: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!Coordinate {
 		const hasTilde = arg[0] == '~';
 		const numberSlice = if (hasTilde) arg[1..] else arg;
 		if (hasTilde and numberSlice.len == 0) return .{.hasTilde = true, .value = 0};
@@ -70,7 +70,7 @@ pub const Axis = struct {
 		};
 	}
 
-	pub fn toValue(self: Axis, playerPos: f64) f64 {
+	pub fn toValue(self: Coordinate, playerPos: f64) f64 {
 		return std.math.clamp(if (self.hasTilde) playerPos + self.value else self.value, -1e9, 1e9); // TODO: Remove clamp after #310 is implemented
 	}
 };
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -59,12 +59,12 @@ pub const Coordinate = union(enum) {
 		if (isRelative and numberSlice.len == 0) return .{.relative = 0};
 		if (isRelative) {
 			return .{.relative = std.fmt.parseFloat(f64, numberSlice) catch {
-				errorMessage.print(allocator, "<{s}>: Expected number, found \"{s}\"", .{name, numberSlice});
+				errorMessage.print(allocator, "Expected number for <{s}>, found \"{s}\"", .{name, numberSlice});
 				return error.ParseError;
 			}};
 		}
 		return .{.absolute = std.fmt.parseFloat(f64, numberSlice) catch {
-			errorMessage.print(allocator, "<{s}>: Expected number or \"~\", found \"{s}\"", .{name, arg});
+			errorMessage.print(allocator, "Expected number or \"~\" for <{s}>, found \"{s}\"", .{name, arg});
 			return error.ParseError;
 		}};
 	}
@@ -173,11 +173,11 @@ pub const PlayerIndex = struct {
 
 	pub fn parse(allocator: NeverFailingAllocator, name: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!PlayerIndex {
 		if (!std.ascii.startsWithIgnoreCase(arg, "@")) {
-			errorMessage.print(allocator, "<{s}>: Player index specifiers always start with @, found \"{s}\"", .{name, arg});
+			errorMessage.print(allocator, "Expected to start with @ for <{s}>, found \"{s}\"", .{name, arg});
 			return error.ParseError;
 		}
 		return .{.index = std.fmt.parseInt(usize, arg[1..], 10) catch {
-			errorMessage.print(allocator, "<{s}>: Player index must be an integer, found \"{s}\"", .{name, arg[1..]});
+			errorMessage.print(allocator, "Expected and integer after @ for <{s}>, found \"{s}\"", .{name, arg[1..]});
 			return error.ParseError;
 		}};
 	}
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -49,32 +49,37 @@ pub fn execute(msg: []const u8, source: *User) void {
 	}
 }
 
-pub const Coordinate = struct {
-	isRelative: bool, // Relative coordinates are indicated by leading `~`.
-	value: f64, // Absolute value for absolute coordinates, offset for relative coordinates.
+pub const Coordinate = union(enum) {
+	relative: f64, // Relative coordinates are indicated by leading `~`.
+	absolute: f64,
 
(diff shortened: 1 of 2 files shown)


src/server/command.zig
```diff
@@ -117,3 +145,36 @@ pub const Target = struct {
 		if (self.increasedRefCount) self.user.decreaseRefCount();
 	}
 };
+
+pub const TargetArg = struct {
+	index: ?usize,
+
+	pub fn parse(allocator: NeverFailingAllocator, _: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!TargetArg {
+		if (arg.len == 0) return .{.index = null};
```
Argmaster: This should be a `return error.ParseError;` and honestly you should never receive an empty argument I believe, so may even be an assert material.

Wunka: :/ I hoped I could so that this would be easier. I will remove it then

src/server/command.zig
```diff
@@ -117,3 +145,36 @@ pub const Target = struct {
 		if (self.increasedRefCount) self.user.decreaseRefCount();
 	}
 };
+
+pub const TargetArg = struct {
```
Argmaster: `PlayerIndex` IMO would be more fitting?

src/server/command.zig
Argmaster: What would be easier this way? Parser should already report that it doesn't have enough arguments on its own.

Argmaster: And you can also use optionals if arguments are optional (unless we stripped that feature?)

Wunka: no, my parsing towords Taget from TargetArgs is easier. but I have found another easy way

Wunka: yes optionals are parsable

Argmaster: What's wrong with just making `TargetArg` optional and making `toTarget` a free function?

Wunka: I now switched to a `fromPlayerIndex` inside `Target`

src/server/command.zig
```diff
@@ -47,6 +49,32 @@ pub fn execute(msg: []const u8, source: *User) void {
 	}
 }
 
+pub const Axis = struct {
+	hasTilde: bool,
+	value: f64,
+
+	pub fn parse(allocator: NeverFailingAllocator, _: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!Axis {
+		const hasTilde = if (arg.len == 0) false else arg[0] == '~';
```
Argmaster: arg should never be empty; you should be able to assert this.

src/server/command.zig
```diff
@@ -47,6 +49,32 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+		const numberSlice = if (hasTilde) arg[1..] else arg;
+		if (hasTilde and numberSlice.len == 0) return .{.hasTilde = true, .value = 0};
+		return .{
+			.hasTilde = hasTilde,
+			.value = std.fmt.parseFloat(f64, numberSlice) catch {
+				if (hasTilde) {
+					errorMessage.print(allocator, "#ff0000Expected number, found \"{s}\"", .{numberSlice});
+				} else {
+					errorMessage.print(allocator, "#ff0000Expected number or \"~\", found \"{s}\"", .{arg});
```
Argmaster: You should specify the name (you marked it `_` in parameter list) in error message, as they can get ambiguous.

src/server/command.zig
```diff
@@ -47,6 +49,32 @@ pub fn execute(msg: []const u8, source: *User) void {
 	}
 }
 
+pub const Axis = struct {
```
Argmaster: This is not an axis, this is a `Coordinate`. A little bit do-everything-coordinate since it can be both relative and absolute, but a coordinate.

src/server/command.zig
```diff
@@ -47,6 +49,32 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+	hasTilde: bool,
```
Argmaster: Meaningless for a reader; implementation detail.

```suggestion
	isRelative: bool, // Relative coordinates are indicated by leading `~`.
```

src/server/command.zig
Wunka: is this not always shown? 
for me I always get all options shown and why each could not be parsed

src/server/command.zig
```diff
@@ -47,6 +49,32 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+	value: f64,
```
Argmaster:
```suggestion
	value: f64,  // Absolute value for absolute coordinates, offset for relative coordinates.
```
In general I dislike how this squishes relative and absolute coordinates into one construct. I get that its convenient, but it suddenly means we need relatively large quantities of comments to preserve basic knowledge.
We could have a
```
struct Coordinate {
    value: union(enum) {
        absolute: Absolute,
        relative: Relative,
    }
    struct Absolute {value: f32},
    struct Relative {offset: f32},
}
```

To type encode that, or maybe it would be fine to just 

```
    Coordinate union(enum) {
        absolute: Absolute,
        relative: Relative,
    }
    struct Absolute {value: f32},
    struct Relative {offset: f32},
```
I am not entirely sure how Parser will behave with something like this, I think it may be fine, even tho unions are treated in quite unique way.

That is more code tho, so I am not sure. Maybe a compromise on names and few comments is fine.

src/server/command.zig
Wunka: this and the other comment (your hasTilde vs isRelative). are becuase I use the same names as the other Axis/ Coordinate struct have

Argmaster: > other Axis/ Coordinate struct have

Where? Which one?

Wunka: ok they are no structs. but the names come from the `parseAxis` and `parseCoordinates` functions which can be removed when everything is migrated to the new system

src/server/command.zig
Wunka: I would let this question open. the parser would have no problem as I am nonetheless overwriting the parser function

src/server/command.zig
Argmaster: I see, but my opinion still holds. Legacy names are what they are, we can always improve on them.
Obv @IntegratedQuantum may disagree.

src/server/command.zig
Argmaster: Can you paste error message from this command? Possibly with mistake for y or z?

Wunka: image

Argmaster: Ok, it might be fine, since content points you to which of X Y Z are wrong. 
That tilda looks weird in quotes tho, I would leave it without (in the xyz case where you have reliable control over it)

src/server/command.zig
IntegratedQuantum: I agree. Better names are always appreciated

Wunka: renaming is done

src/server/command.zig
```diff
@@ -47,6 +49,32 @@ pub fn execute(msg: []const u8, source: *User) void {
 	}
 }
 
+pub const Coordinate = struct {
+	isRelative: bool, // Relative coordinates are indicated by leading `~`.
```
IntegratedQuantum: I would prefer an `enum {relative, absolute}` or maybe just use a union for the whole thing `absolute: f64, relative: f64`

src/server/command.zig
```diff
@@ -47,6 +49,32 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+					errorMessage.print(allocator, "#ff0000Expected number, found \"{s}\"", .{numberSlice});
+				} else {
+					errorMessage.print(allocator, "#ff0000Expected number or \"~\", found \"{s}\"", .{arg});
+				}
+				return error.ParseError;
+			},
+		};
+	}
+
+	pub fn toValue(self: Coordinate, playerPos: f64) f64 {
```
IntegratedQuantum: I would prefer a global function
```
pub fn resolveCoordinates(x: Coordinate, y: Coordinate, z: Coordinate, player: *User) Vec3d {}
```
This would make it less cumbersome to use.

src/server/command.zig
```diff
@@ -47,6 +49,32 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+		return std.math.clamp(if (self.isRelative) playerPos + self.value else self.value, -1e9, 1e9); // TODO: Remove clamp after #310 is implemented
+	}
+};
+
 fn parseAxis(arg: []const u8, playerPos: f64, source: *User) !f64 {
```
IntegratedQuantum, line 83: Please also add a TODO to the old parsing functions and mention them in the migration issue.

IntegratedQuantum (changes requested):

src/server/command.zig
IntegratedQuantum: I think we should add this information. Figuring it out from the content alone may be difficult when you have a longer command with more arguments.

Wunka: is this good? 
image

src/server/command.zig
Wunka: made it a union. @Argmaster I had to add union support.

src/server/command.zig

src/server/command.zig
Wunka: done

src/server/command.zig
Argmaster: Please follow same pattern as built in parsers for float, int and enum: `"Expected a {whatever} for <{s}>, found \"{s}\""`
Consistency for the win!

Wunka: New version.
image

src/server/command.zig
Argmaster: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

What did I tell you about TODOs.
Make a god damned ISSUE!

Wunka: (I did mention it in the migration issue....)

Argmaster: Ok, then its fine.

IntegratedQuantum: > What did I tell you about TODOs

You should always do both. TODOs are useful to find all the places mentioned in the issue, and they remain as a marker when inevitably one of the places is forgotten.

IntegratedQuantum (approved):

PR #3104, opened 2026-05-22
Tp command update

Wunka:
Depends on #3103
Progess for #3073

### DIFF
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -75,6 +75,20 @@ pub const Axis = struct {
 	}
 };
 
+pub const Coordinates = struct {
+	x: Axis,
+	y: Axis,
+	z: ?Axis,
+
+	pub fn toVec(self: Coordinates, source: *User) main.vec.Vec3d {
+		return .{
+			self.x.toValue(source.player().pos[0]),
+			self.y.toValue(source.player().pos[1]),
+			if (self.z) |z| z.toValue(source.player().pos[2]) else source.player().pos[2],
+		};
+	}
+};
+
 fn parseAxis(arg: []const u8, playerPos: f64, source: *User) !f64 {
 	const hasTilde = if (arg.len == 0) false else arg[0] == '~';
 	const numberSlice = if (hasTilde) arg[1..] else arg;
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -49,11 +49,11 @@ pub fn execute(msg: []const u8, source: *User) void {
 	}
 }
 
-pub const Axis = struct {
+pub const Coordinate = struct {
 	hasTilde: bool,
 	value: f64,
 
-	pub fn parse(allocator: NeverFailingAllocator, _: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!Axis {
+	pub fn parse(allocator: NeverFailingAllocator, _: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!Coordinate {
 		const hasTilde = arg[0] == '~';
 		const numberSlice = if (hasTilde) arg[1..] else arg;
 		if (hasTilde and numberSlice.len == 0) return .{.hasTilde = true, .value = 0};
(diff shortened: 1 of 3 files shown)


src/server/command/tp.zig
```diff
@@ -11,76 +13,100 @@ pub const usage =
...
-	if (std.mem.containsAtLeast(u8, args, 1, ":")) {
-		const biome = main.server.terrain.biomes.getByIdOptional(args) orelse {
-			source.sendMessage("#ff0000Couldn't find biome with id \"{s}\"", .{args});
-			return;
-		};
-		if (biome.isCave) {
-			source.sendMessage("#ff0000Teleport to biome is only available for surface biomes.", .{});
-			return;
+const Args = union(enum) {
+	@"/tp <biome>": struct { biome: struct {
```
Argmaster: I would argue against inlining struct definitions in argument definition. System was designed to be reusable and it should be used as such. Please extract it next to other reusable arg structs.

Wunka: done

src/server/command/tp.zig
```diff
@@ -11,76 +11,92 @@ pub const usage =
...
-				.dirNegY => wy -%= mapSize,
-				.dirPosY => wy +%= mapSize,
-				else => unreachable,
+	var errorMessage: main.ListUnmanaged(u8) = .{};
+	defer errorMessage.deinit(main.stackAllocator);
+
+	const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
+		source.sendMessage("#ff0000{s}", .{errorMessage.items});
+		return;
+	};
```
Wunka, line 33: @Argmaster  I don't like that I would need to write this every time. should we maybe make it so that a command can either have `pub fn execute(args: []const u8, source: *User) void {` or `pub fn execute(args: Args, source: *User) void {` ?

Argmaster: Yes, that is roughly the direction. We can't just do that now tho, since we have to maintain compatibility with unmigrated commands.

Argmaster: I mean, the direction is to have only `pub fn execute(args: Args, source: *User) void {`

Argmaster: This way we can use `ArgParser` to parse, autocomplete and generate usage for us. This is usual scope of argument parser library, at least based on Python's `argparse` and `click` (both have very different interface) which I have experience with.

Wunka: mhhh I can't find how the main function in zig works with the different argument possibilities, but probably some weird comptime magic. I will then wait until all are migrated.

src/server/command.zig
```diff
@@ -182,3 +182,14 @@ pub const PlayerIndex = struct {
 		}};
 	}
 };
+
+pub const Biome = struct {
+	biome: *const main.server.terrain.biomes.Biome,
```
IntegratedQuantum, line 187: Could we not add the parse function directly to the `Biome` struct? Skipping the pointer in the generic code shouldn't be too difficult.

IntegratedQuantum (changes requested):

src/server/command.zig
Wunka: I am ok with both options. I for now wanted parse functions in the command section. thats why I didn't touch the biome struct itself

Argmaster: You will have to implement `parse` `autocomplete` and possibly any other great ideas on the other side of the world inside `main.server.terrain.biomes.Biome`. 
IMO this unnecessarily complicates `Biome` struct with features used in completely different part of code.
That means coupling between systems and scattered implementation of similar features.
Not even mentioning that this unnecessarily complicates argparse, since ownership rules for raw pointers are ambiguous. As of now every object returned by argparse is owned by the caller, while in this case I think It is not, you are not allowed to dealloc whats behind the pointer?
Also its inconsistent with other 99% of argument parsing code which uses proxy objects

IntegratedQuantum: Yeah, that makes sense. Could we at least give it a different name then? I think it could be confusing to have two structs named Biome when both of them are distinctly different types, and once ZLS gets auto-importing you could mistakenly import the wrong one.

Wunka: `BiomeArg` ? and then in general just do the suffix Arg when we need to "copy" a struct?

Argmaster: I was thinking BiomeId, since that's what you have to type in the command

Wunka: renamed to BiomeId

IntegratedQuantum (approved):

PR #3105, opened 2026-05-22
clear command update

Wunka:
Progress towards #3073

### DIFF
--- a/src/server/command/clear.zig
+++ b/src/server/command/clear.zig
@@ -6,23 +6,23 @@ const User = main.server.User;
 pub const description = "Clears your inventory/chat";
 pub const usage = "/clear <inventory/chat>";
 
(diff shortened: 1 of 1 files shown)


IntegratedQuantum (approved):
I'm surprised that it's still the same number of lines. Though I guess that will go away once we do the parsing on the outside, and automate the usage string generation.

Argmaster:
> I'm surprised that it's still the same number of lines.

Its much less dense code, half of that are curly braces and empty lines now.
As Wunka recognized the fact we have to manually set up parsing is wasting quite some precious lines, we will get rid of extra 6 lines or so by moving that outside of function and passing `Args` instead of `[]const u8` 

---

Heh sorry for repeating your words 😅

PR #3106, opened 2026-05-22
gamemode command update

Wunka:

### DIFF
--- a/src/server/command/gamemode.zig
+++ b/src/server/command/gamemode.zig
@@ -13,8 +13,7 @@ pub const usage =
 ;
 
 const Args = union(enum) {
-	@"/gamemode <playerIndex> <mode>": struct { playerIndex: ?command.PlayerIndex, mode: main.game.Gamemode },
-	@"/gamemode <playerIndex>": struct { playerIndex: ?command.PlayerIndex },
+	@"/gamemode <playerIndex> <mode>": struct { playerIndex: ?command.PlayerIndex, mode: ?main.game.Gamemode },
 };
 
 const ArgParser = main.argparse.Parser(Args, .{.commandName = "/gamemode"});
@@ -33,12 +32,11 @@ pub fn execute(args: []const u8, source: *User) void {
 			const target = command.Target.fromPlayerIndex(params.playerIndex, source) catch return;
 			defer target.deinit();
 
-			main.sync.setGamemode(target.user, params.mode);
-		},
-		.@"/gamemode <playerIndex>" => |params| {
-			const target = command.Target.fromPlayerIndex(params.playerIndex, source) catch return;
-			defer target.deinit();
-			source.sendMessage("#ffff00{s}", .{@tagName(target.user.gamemode.load(.monotonic))});
+			if (params.mode) |mode| {
+				main.sync.setGamemode(target.user, mode);
+			} else {
+				source.sendMessage("#ffff00{s}", .{@tagName(target.user.gamemode.load(.monotonic))});
+			}
 		},
 	}
 }
--- a/src/server/command/gamemode.zig
+++ b/src/server/command/gamemode.zig
@@ -1,13 +1,17 @@
 const std = @import("std");
 
 const main = @import("main");
+const command = main.server.command;
 const User = main.server.User;
 
 pub const description = "Get or set your gamemode.";
-pub const usage = "/gamemode\n/gamemode <survival/creative>";
+pub const usage =
+	\\/gamemode <survival/creative>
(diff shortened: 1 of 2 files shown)


src/server/command/gamemode.zig
```diff
@@ -1,22 +1,31 @@
...
 const main = @import("main");
+const command = main.server.command;
 const User = main.server.User;
 
-pub const description = "Get or set your gamemode.";
-pub const usage = "/gamemode\n/gamemode <survival/creative>";
+pub const description = "Get or set the players gamemode.";
+pub const usage =
+	\\/gamemode <survival/creative>
+	\\/gamemode @playerIndex <survival/creative>
```
IntegratedQuantum, line 10: Previous implementation also had `/gamemode` to query the gamemode, especially now that you have added the player index this would be helpful.

IntegratedQuantum (changes requested):

src/server/command/gamemode.zig
Wunka: I brought it back. I somehow thought it couldn't do it

src/server/command/gamemode.zig
```diff
@@ -1,22 +1,44 @@
...
+	\\/gamemode
+	\\/gamemode @playerIndex
+;
+
+const Args = union(enum) {
+	@"/gamemode <playerIndex> <mode>": struct { playerIndex: ?command.PlayerIndex, mode: main.game.Gamemode },
+	@"/gamemode <playerIndex>": struct { playerIndex: ?command.PlayerIndex },
```
IntegratedQuantum: It could also be implemented with the second argument being optional. Not sure if that's better though.

I did however notice that the error message gets longer when there are more variants. @Argmaster I would like to hear your opinion here, should we try to minimize the number of variants?

Argmaster: Yeah I don't think we need two variants here, variants should be used when there are two different use cases (sub commands), not when there is literally an optional trailing argument. That makes it also faster, although performance is not significant in this case, since the parsing resets after failed branch. With current approach we would parse player index twice.

Wunka: changed. but I do have to say. optional arguments can give weird error messages e.g:
image

Argmaster: Uh is it because we don't distinguish ~~no~~ extra argument from argument that doesn't match the argument type restrictions?

Wunka: I have to sadly say, I don't understand your sentence. but from my understanding it is because the parser parses `/gamemode survival @0` as `/gamemode null survival @0`. as survival is not valid for playerindex and so playerIndex gets parsed as `null`

Argmaster: Does player index notice that survival is not a valid player index?

Wunka: yes as playerindexes always start with `@` and then only have a number

Argmaster: I see, so `survival` fails to match player index requirements and is then propagated and accepted by mode field, but then we have another argument pending that doesn't have anywhere to go. 
Firstly we should check max expected number of arguments against length of provided arguments, to make the message at least less confusing -> even generic "invalid argument" would be better. 
Secondly it's hard to provide good message here... We would have to speculatively eval everything after failed optional, but I'm not sure if that is viable and how much better that would make the messages. If we do that, we could at least report first failed optional as failure point. By speculatively I mean probably a recursive call? So we can error out if we encounter too many arguments at the end and there are no other branches. It will always end up a little bit weird, since input can at that point be pretty much anything.

IntegratedQuantum: How about we display how the command got matched?
```
/gamemode <optional playerIndex> <optional mode>
matched as
/gamemode <skipped> survival
has leftover arguments '@0'
```

src/server/command/gamemode.zig
```diff
@@ -1,22 +1,42 @@
 const std = @import("std");
 
 const main = @import("main");
+const command = main.server.command;
 const User = main.server.User;
 
-pub const description = "Get or set your gamemode.";
-pub const usage = "/gamemode\n/gamemode <survival/creative>";
+pub const description = "Get or set the players gamemode.";
```
IntegratedQuantum: grammar, also the → a since you are not referring to a specific player.
```suggestion
pub const description = "Get or set a player's gamemode.";
```

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #3107, opened 2026-05-22
Kill command update

Wunka:

### DIFF
--- a/src/server/command/kill.zig
+++ b/src/server/command/kill.zig
@@ -5,7 +5,10 @@ const command = main.server.command;
 const User = main.server.User;
 
 pub const description = "Kills the player";
-pub const usage = "/kill";
+pub const usage =
+	\\/kill
+	\\/kill @<playerIndex>
+;
 
 const Args = union(enum) {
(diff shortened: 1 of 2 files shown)


src/server/command/kill.zig
```diff
@@ -1,15 +1,29 @@
 const std = @import("std");
 
 const main = @import("main");
+const command = main.server.command;
 const User = main.server.User;
 
 pub const description = "Kills the player";
 pub const usage = "/kill";
```
Argmaster: There seems to be hidden usage of `/kill @playerIndex`?
So actually there are two allowed forms, where `/kill` alone is for suicide and `/kill @playerIndex` for killing others?
This needs documentation.
I also have to point out that most likely there will be difference between usage generated for optionals and for branched approach like in `/time`

Wunka: ups thanks for noticing. I added the @playerindex functionality. WHy did I forget to edit the usage :/

Argmaster: > Why did I forget to edit the usage :/

Yeah that's why redundant information is bad.

IntegratedQuantum (approved):

Issue #3043, opened 2026-05-06
`/count` command (worldedit)


IntegratedQuantum:
This would count the blocks inside a chunk for each block type.

Extracted from https://github.com/PixelGuys/Cubyz/pull/2585#discussion_r3190340056

PR #3111, opened 2026-05-22
Add Selection struct

### LINKED ISSUE #2816: Add a helper function to get worldedit positions

Argmaster:
This pull request adds a `Selection` struct which is now passed to `Blueprint.capture` instead of separate vectors. Additionally some utility functions were added for obtaining `Selection` objects from user data and for undo/redo histories.

A possible future improvement would be to extract failure dispatching from `Blueprint.capture` since it also seems to be boilerplate in case of most WE commands which have to do it.

Resolves: #2816


Argmaster:
It is preferable to apply this change before migrating WE commands to argparse.
It is preferable to apply this change before continuing with #3043

src/blueprint.zig
```diff
@@ -74,18 +79,25 @@ pub const Blueprint = struct {
 		failure: struct { pos: Vec3i, message: []const u8 },
 	};
 
-	pub fn capture(allocator: NeverFailingAllocator, pos1: Vec3i, pos2: Vec3i) CaptureResult {
-		const startX = @min(pos1[0], pos2[0]);
-		const startY = @min(pos1[1], pos2[1]);
-		const startZ = @min(pos1[2], pos2[2]);
+	pub const Selection = struct {
+		minPos: Vec3i,
+		maxPos: Vec3i,
```
IntegratedQuantum, line 86: Please specify that both sides are inclusive, since this is counter to the usual model (left inclusive, right exclusive).

src/blueprint.zig
```diff
@@ -74,18 +79,25 @@ pub const Blueprint = struct {
...
 
-		const endX = @max(pos1[0], pos2[0]);
-		const endY = @max(pos1[1], pos2[1]);
-		const endZ = @max(pos1[2], pos2[2]);
+		pub fn init(pos1: Vec3i, pos2: Vec3i) Selection {
+			return .{.minPos = @min(pos1, pos2), .maxPos = @max(pos1, pos2)};
+		}
+		pub fn size(self: Selection) @Vector(3, u32) {
+			return @intCast(self.maxPos - self.minPos + @as(Vec3i, @splat(1)));
```
IntegratedQuantum: Please make sure this works at the edge of the world
```suggestion
			return @intCast(self.maxPos -% self.minPos + @as(Vec3i, @splat(1)));
```

src/server/server.zig
```diff
@@ -56,6 +56,9 @@ pub const WorldEditData = struct {
 				main.globalAllocator.free(self.message);
 				self.blueprint.deinit(main.globalAllocator);
 			}
+			pub fn selection(self: Value) Blueprint.Selection {
+				return .init(self.position, self.position + self.blueprint.extent() - @as(Vec3i, @splat(1)));
```
```suggestion
				return .init(self.position, self.position +% self.blueprint.extent() -% @as(Vec3i, @splat(1)));
```

src/server/command/worldedit/paste.zig
```diff
@@ -32,11 +32,8 @@ pub fn execute(args: []const u8, source: *User) void {
 		const pos: Vec3i = @floor(source.player().pos);
 		source.sendMessage("Pasting: {}", .{pos});
 
-		const undo = Blueprint.capture(main.globalAllocator, pos, .{
-			pos[0] + @as(i32, @intCast(clipboard.blocks.width)) - 1,
-			pos[1] + @as(i32, @intCast(clipboard.blocks.depth)) - 1,
-			pos[2] + @as(i32, @intCast(clipboard.blocks.height)) - 1,
-		});
+		const selection: Blueprint.Selection = .init(pos, pos + clipboard.extent() - @as(Vec3i, @splat(1)));
```
IntegratedQuantum:
```suggestion
		const selection: Blueprint.Selection = .init(pos, pos +% clipboard.extent() -% @as(Vec3i, @splat(1)));
```

IntegratedQuantum (changes requested):

src/blueprint.zig
Argmaster: Will it make more sense to change it? Lemme see.
Otherwise I would add a comment, is that fine?

IntegratedQuantum: It should be in the name, otherwise it's easy to miss this nuance when reviewing code, or when using autocomplete.

Argmaster: ~~Hey but they do follow the normal convention!
Min is inclusive and max is exclusive - look at the `size()` - it adds 1 and doesn't fail out of bounds assertion in `Array3D.set`, no?
Also in rendering we add 1 too: `@floatFromInt(topRight - bottomLeft + Vec3i{1, 1, 1})`~~

Argmaster: Never mind, im just dumb xD

Argmaster: Ok, after brief confusion I have updated it to be inclusive-exclusive which imo is healthier long term?
Although logically its kind of rough to explain, since max is now "maximal position but outside of the selection"

IntegratedQuantum: It makes a lot of sense if you think about it in floating point coordinates. e.g. The block at position `0` spans the coordinate range from `0.0` to `1.0`, a selection from 0.0 to 1.0 contains only this one block.

Argmaster: Thankfully all of that is completely abstracted now, so i have to only bother with iterating from 0 to size() and exact min and max are whatever.
Something to note and possibly improve on in the future: selectionPosition1 and selectionPosition2 which are separate concepts are still both inclusive and may need a note about that state. There is no good way to make them asymmetric tho, since there is no actual ordering between them, but I would consider a comment.

src/server/command/worldedit/paste.zig
```diff
@@ -32,11 +32,8 @@ pub fn execute(args: []const u8, source: *User) void {
 		const pos: Vec3i = @floor(source.player().pos);
 		source.sendMessage("Pasting: {}", .{pos});
 
-		const undo = Blueprint.capture(main.globalAllocator, pos, .{
-			pos[0] + @as(i32, @intCast(clipboard.blocks.width)) - 1,
-			pos[1] + @as(i32, @intCast(clipboard.blocks.depth)) - 1,
-			pos[2] + @as(i32, @intCast(clipboard.blocks.height)) - 1,
-		});
+		const selection: Blueprint.Selection = .init(pos, pos +% clipboard.extent());
```
IntegratedQuantum: I think this is wrong, given that init takes inclusive bounds, but extent returns size and inclusive lower bound + size == exclusive upper bound.
Maybe it would be best to have two variants of the function, e.g. here we could use a init from extent.

IntegratedQuantum (changes requested):

src/blueprint.zig
IntegratedQuantum: You could also make selectionPosition `Selection`s themselves which only include one block. Then when computing the full thing you'd just merge them together.

src/server/command/worldedit/paste.zig
Argmaster: Yeah it magically increases the size by 1
I have separated `initFromInclusive` and `initFromExtent`

src/blueprint.zig
Argmaster: That's kind of wasteful

src/blueprint.zig
```diff
@@ -74,18 +79,38 @@ pub const Blueprint = struct {
...
-		const startY = @min(pos1[1], pos2[1]);
-		const startZ = @min(pos1[2], pos2[2]);
+	pub const Selection = struct {
+		/// Minimal position of a block, inclusive.
+		minPos: Vec3i,
+		/// Maximal position of a block, exclusive.
+		maxPos: Vec3i,
+
+		pub fn init(pos1Inclusive: Vec3i, pos2Exclusive: Vec3i) Selection {
+			return .{.minPos = @min(pos1Inclusive, pos2Exclusive), .maxPos = @max(pos1Inclusive, pos2Exclusive)};
```
IntegratedQuantum: This is wrong, you cannot swap them like this when one of them is exclusive and the other inclusive.
I'd suggest to remove this one entirely in favor of explicitly assigning the struct fields.

IntegratedQuantum (changes requested):

src/blueprint.zig
Argmaster: Ok, somehow I missed that I do not need to min max them in extent constructor. Well.

IntegratedQuantum (approved):

PR #3112, opened 2026-05-22
Perm command update

Wunka:

### DIFF
--- a/src/server/command/permission/perm.zig
+++ b/src/server/command/permission/perm.zig
@@ -26,7 +26,7 @@ const Args = union(enum) {
 	@"/perm <playerIndex> <permissionPath>": struct { playerIndex: ?command.PlayerIndex, permissionPath: Path },
 };
 
-const ArgParser = main.argparse.Parser(Args, .{.commandName = "/time"});
+const ArgParser = main.argparse.Parser(Args, .{.commandName = "/perm"});
 
 pub fn execute(args: []const u8, source: *User) void {
 	var errorMessage: main.ListUnmanaged(u8) = .{};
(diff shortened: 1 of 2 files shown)


src/server/command/permission/perm.zig
```diff
@@ -1,102 +1,82 @@
...
-		};
+const Path = struct {
+	path: []const u8,
 
-		if (permissionPath[0] != '/') {
-			source.sendMessage("#ff0000Permission paths always begin with a \"/\", got: {s}", .{arg});
-			return error.InvalidArgs;
+	pub fn parse(allocator: NeverFailingAllocator, name: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!Path {
+		if (arg[0] != '/') {
+			errorMessage.print(allocator, "Permission path got for <{s}> doens't begin with a \"/\", got: {s}", .{name, arg});
```
```suggestion
			errorMessage.print(allocator, "Permission path for <{s}> doesn't begin with a \"/\", got: {s}", .{name, arg});
```

IntegratedQuantum (changes requested):
finally one that is noticeably smaller

IntegratedQuantum (approved):

src/server/command/permission/perm.zig
```diff
@@ -1,102 +1,82 @@
...
-		};
+const Path = struct {
+	path: []const u8,
 
-		if (permissionPath[0] != '/') {
-			source.sendMessage("#ff0000Permission paths always begin with a \"/\", got: {s}", .{arg});
-			return error.InvalidArgs;
+	pub fn parse(allocator: NeverFailingAllocator, name: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!Path {
+		if (arg[0] != '/') {
+			errorMessage.print(allocator, "Permission path for <{s}> doesn't begin with a \"/\", got: {s}", .{name, arg});
```
Argmaster, line 77: I did not think about it before, but this is wrong allocator: it should use `main.stackAllocator` which is weird, because it outlives this function call.
Coincidentally allocator is `main.stackAllocator` but that is not the same; allocator is defined as allocator used to allocate `Args` type whereas `errorMessage` always uses `main.stackAllocator`.

PR #3113, opened 2026-05-22
update invite command

Wunka:

### DIFF
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -168,14 +168,6 @@ pub const Target = struct {
 	}
 };
 
-pub const String = struct {
-	string: []const u8,
-
-	pub fn parse(_: NeverFailingAllocator, _: []const u8, arg: []const u8, _: *ListUnmanaged(u8)) error{ParseError}!String {
-		return .{.string = arg};
-	}
-};
-
 pub const PlayerIndex = struct {
 	index: usize,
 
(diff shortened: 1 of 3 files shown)


src/server/command/invite.zig
```diff
@@ -2,24 +2,31 @@ const std = @import("std");
 
 const main = @import("main");
 const User = main.server.User;
+const command = main.server.command;
 
 pub const description = "Invite a player";
 pub const usage = "/invite <IP>";
 
+const Args = union(enum) {
+	@"/invite <IP>": struct { IP: command.String },
```
IntegratedQuantum: name convention
```suggestion
	@"/invite <IP>": struct { ip: command.String },
```

src/server/command.zig
```diff
@@ -117,3 +119,11 @@ pub const Target = struct {
 		if (self.increasedRefCount) self.user.decreaseRefCount();
 	}
 };
+
+pub const String = struct {
```
IntegratedQuantum: In my opinion we should handle the `[]const u8` case in the parser instead of as a struct.

IntegratedQuantum (changes requested):

src/server/command.zig
Wunka: done.

IntegratedQuantum (approved):

PR #3117, opened 2026-05-22
help command update

Wunka:

### DIFF
--- a/src/server/command/help.zig
+++ b/src/server/command/help.zig
@@ -10,6 +10,7 @@ pub const description = "Shows info about all the commands.";
 pub const usage = "/help\n/help <command>";
 
 const Args = union(enum) {
+	@"/help <bobik>": struct { bobik: enum { Bobik, bobik } },
 	@"/help <command>": struct { command: Cmd },
 	@"/help": struct {},
 };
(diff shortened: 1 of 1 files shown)


src/server/command/help.zig
```diff
@@ -1,44 +1,77 @@
...
+			msg.append('\n');
+		},
 	}
 	if (msg.items[msg.items.len - 1] == '\n') _ = msg.pop();
 	source.sendMessage("{s}", .{msg.items});
 }
+
+const Cmd = union(enum) {
+	cmd: command.Command,
+	bobik: void,
```
IntegratedQuantum: I would suggest to implement bobik as a separate entry in the Args union.
Then it would also have a help entry once we autogenerate them, which would make it possible to discover this easter egg in-game, but it's still reasonably hidden: Who is going to use `/help help` anyways?

IntegratedQuantum (changes requested):

src/server/command/help.zig
Wunka: Thats actually how I had it before. But I wanted to "hide" Bobik a liitle bit more. I reverted the "hide Bobik" commit

IntegratedQuantum (approved):

PR #3120, opened 2026-05-22
Jungle hills biome

careeoki:
Another dew grass biome, a more sparse jungly area. It features large plateaus and an extra sparse subbiome.
image
image


Argmaster:
I don't get what's the point of Jingle being sparse really. Isn't jungle about being dense and hard to traverse? If it's not dense is just another forest just with different trees. I don't like how this biome is really similar to other biomes. For me its very annoying to traverse a world where every biome looks so similar. I would want to see new things instead of another permutation of what I have already seen.
The hills look neat, but the could be just another jungle sub biome and imo that would work better.


---

I am still missing some structure in how biomes transition. It's clearly visible that biomes have same baseline height and that there is no smoothly changing global base height, so when biomes change, changes are mostly abrupt.
Even if we make biomes that have higher baseline height or some other properties that make them blend with tall mountains better, due to scale it usually still looks like perfectly flat surface and then nearly vertical lonely mountain.

 I think it would be helpful if our baseline height was a sin function or alike, like tectonic plates. So instead of 0 + biome settings it would be 0 + tectonic(x, y) + biome settings, with mountain biomes having soft preference to be higher up. This would give biomes natural slope towards mountains. Tectonic could be just sin (that might tile too visibly at large scale tho, or some repeating precomputed low resolution noise.

If we are doing that already, it is not visible, based on my recent exploration of world.

Argmaster:
There is another reason to have global tectonic plates function - we could  base river direction on it and they automatically direction  to oceans, since oceans would prefer lower base height.

IntegratedQuantum (changes requested):
I like the idea of a jungle with plateaus, but I have to agree with Argmaster: Jungles should be more dense.

careeoki:
I'll revisit this in the future, but will probably start from scratch.

PR #3124, opened 2026-05-23
Bigger Oceans

TyrrelMeter:
This is an attempt at making the oceans bigger to more match real life. It just makes the ocean rarity less rare so that there is more ocean that land. I don't really expect this to get merged but I thought I'd put it out there.
Cubyz 0 3 0-dev 5_23_2026 6_13_57 PM
Cubyz 0 3 0-dev 5_23_2026 6_13_49 PM
Cubyz 0 3 0-dev 5_23_2026 6_13_46 PM
Cubyz 0 3 0-dev 5_23_2026 6_13_27 PM
Cubyz 0 3 0-dev 5_23_2026 6_13_25 PM
Cubyz 0 3 0-dev 5_23_2026 6_38_30 PM

### DIFF
--- a/assets/cubyz/biomes/ocean/cold/archipelago.zig.zon
+++ b/assets/cubyz/biomes/ocean/cold/archipelago.zig.zon
@@ -1,5 +1,5 @@
 .{
-	.chance = 0.1,
+	.chance = 1,
 	.properties = .{
 		.cold,
 		.ocean,
--- a/assets/cubyz/biomes/ocean/cold/base.zig.zon
+++ b/assets/cubyz/biomes/ocean/cold/base.zig.zon
@@ -1,4 +1,5 @@
 .{
+	.chance = 30,
 	.properties = .{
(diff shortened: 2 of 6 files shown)


Argmaster:
That could work if gliding was the main way to move...

With rings of wind you have to fly through that give you speed boost

TyrrelMeter:
Currently this isn't a very good idea to impliment. There isn't that much to do in the oceans and there isn't a good way to traverse them.

Crepestrom:
This would make a good alternate world preset

TyrrelMeter:
it would, but not the way I did this PR. I might have to experiment with that

Wbjitscool:
this could work as a world preset? hmm or we can have like sections where there's massive oceans and other's are like smaller oceans with more land mixed and the larger oceans have more islands

TyrrelMeter:
Might be good to have an issue instead of a PR

PR #3129, opened 2026-05-24
Rename List to ListManaged

IntegratedQuantum:
I still left an alias for the old List, so we can have a soft migration period. In a few weeks I would like to remove it.

I also tried to avoid some contended regions, to reduce conflict potential for now.

I also did migrate some Lists to unmanaged, but only in the simple cases.

progress towards #1181


H41ogen:
on the topic of this, why do we call it `List` instead of `ArrayList`?

IntegratedQuantum:
I called it `List` to give it a different name from the `std.ArrayList`, it made the migration a bit easier back then, and also today I think it makes it easier to spot incorrect usage in review, but that part could maybe also be done with the linter now.

PR #3133, opened 2026-05-29
Shadow maps are actually doable (no raymarching)

codemob-dev:
Very WIP.

image

TODO:
- [ ] Fix culling issue
- [x] Fix jaggies
- [x] Fix peter panning
- [ ] Improve performance
- [x] ~Add entity support~ (separate PR)
- [x] Get shadows on block models in the inventory working correctly
- [x] Make light sources remove shadows
- [ ] Try different sampling methods
- [x] Tweak shadow color
- [x] ~Cascaded shadow maps~ (separate PR)


IntegratedQuantum:
> Fix jaggies

I think the main problem is that the shadowmap is not aligned with the pixels, instead it's rotated in 3d space. I'd suggest to align it to a plane in 3d space to fix this (either x-y or x-z plane depending on the angle). This can be achieved with non-orthogonal projection. You do lose some area coverage, up to 30% by doing this, but it should give a significantly more consistent appearance.
map

IntegratedQuantum:
I think part of your performance issues come from an implementation detail of the occlusion culling:
It first draws the chunks that were visible in the previous frames, then it draws them and determines which chunks are visible (and need to be drawn) in the current frame. This is then stored in the buffer of chunks, but you did not assign an extra data region of this for the shadow map data.
So instead of drawing the chunks previously visible in the shadow map, you draw the chunks which were visible in the view of the player. This makes occlusion culling ineffective for chunks in the suns view but outside the players view for shadow map rendering and vice versa for normal rendering.

Argmaster:
Pardon me if it's already implemented or doesn't make sense, but:
Can we cache the shadow map?
Would it be possible to generate shadow map relative to coordinates of current chunk in which player is located, rather than current player location?
As far as I understand that would mean we have to only consider when we have to invalidate that cache rather than regenerate the shadow map every frame?
Invalidation events would be sun progressing 1 degree (arbitrary choice), player moving into new chunk, one of included chunks (simply chunk in range of x chunks) having it's mesh updated.
We could, with CSM, defer and desync regeneration of further shadow cascades from frame generation(?) the highest res one would have to remain synced to avoid glitches when player edits nearby blocks (outdated shadows)
That idea kind of breaks apart due to sun moving, especially at shallow angles, when shadows lengthen quickly and blockiness of animation would be clearly visible.
Another issue would be animated foliage or really anything animated that changes models/transparency.
It would require a lot of effort to fine tune which parts of shadow map have to be regenerated and may defy the point.

codemob-dev:
> Can we cache the shadow map?

Technically yes, but we want entities to have shadows as well. Plus, the performance really isn't all that bad right now at least.

codemob-dev:
> I think part of your performance issues come from an implementation detail of the occlusion culling: It first draws the chunks that were visible in the previous frames, then it draws them and determines which chunks are visible (and need to be drawn) in the current frame. This is then stored in the buffer of chunks, but you did not assign an extra data region of this for the shadow map data. So instead of drawing the chunks previously visible in the shadow map, you draw the chunks which were visible in the view of the player. This makes occlusion culling ineffective for chunks in the suns view but outside the players view for shadow map rendering and vice versa for normal rendering.

That should be fixed now!
Here's another pretty picture:
image

ikabod-kee:
image
Are you going to have transparent shadows too?

codemob-dev:
> Are you going to have transparent shadows too?

Probably not, but it might be possible (very big might) using the other color channels.

ikabod-kee:
I'm begging you

Argmaster:
I really love it!
But this is concerning:

image

OneAvargeCoder193:
> Are you going to have transparent shadows too?

im pretty sure for transparent shadows to work the shadows would need to be stored in a per pixel linked list or we would need to not have shadows on transparent objects.

Argmaster:
> im pretty sure for transparent shadows to work the shadows would need to be stored in a per pixel linked list or we would need to not have shadows on transparent objects.

It's called ray tracing

OneAvargeCoder193:
an optimization you can do with csm is that you can enable GL_DEPTH_CLAMP, which then means you can render the objects behind the near plane which you would have with the frustum, and you save the floating point precision for the objects that are actually in the player's frustum

Wbjitscool:
Also will you make the shadows move with day/night cycle oh wait will that require a sun/moon so you can tell which way the direction of the sun/moon is going?

Xandertron:
> It's called ray tracing

this gave me a good laugh cause we could just do hardware raytracing when the vulkan renderer finally exists, but these look good as it is, really adds some nice polish to the game

IntegratedQuantum:
What's the current state on this?
Maybe we could already get some of the refactoring changes merged to prevent this from rotting too much?

codemob-dev:
I just improved the performance a bit and fixed the culling issues (:

codemob-dev:
I don't have a good way to profile the shader on my computer, so if anybody could do so I would appreciate it.

IntegratedQuantum:
Does the F5 screen not work for you?

codemob-dev:
> Does the F5 screen not work for you?

I mean line-by-line in the shaders.

IntegratedQuantum:
As far as I know only nsight can do that for OpenGL (all AMD tooling is Vulkan only), so you have to find an nvidia user.

Wbjitscool:
I Use Nvidia

codemob-dev:
Okay, something went horribly wrong during the merge and now the shadows are completely gone. I have no clue why.

codemob-dev:
> Okay, something went horribly wrong during the merge and now the shadows are completely gone. I have no clue why.

It was a byte alignment mismatch...

codemob-dev:
The culling problems *should* be completely fixed now!

Argmaster:
- [ ] daylight cycle
- [ ] add toggle to graphics settings

Wbjitscool:
it's already working better then before and fps seems better this time too

Wbjitscool:
looks like 1 check failed

IntegratedQuantum:
Interestingly the gaps in the shadow seem to grow with increased shadow map resolution, there is definitely still an issue there Screenshot at 2026-07-30 16-15-56

IntegratedQuantum:
Also at low res there are stripes on my grass:
Screenshot at 2026-07-30 16-19-40
Apart from these two issues, it's much better, and also feels a bit faster than before.

PR #3136, opened 2026-05-30
fix component init and deinit ordering and clean up bag deinit code

RanPix:
_clearing up quantums mess_

### DIFF
--- a/src/entity.zig
+++ b/src/entity.zig
@@ -100,31 +100,31 @@ pub fn unloadComponent(comptime side: main.sync.Side, componentId: EntityCompone
 
 pub const client = struct {
 	pub fn init() void {
-		inline for (@typeInfo(components).@"struct".decls) |decl| {
-			@field(components, decl.name).client.init();
-		}
 		inline for (@typeInfo(systems).@"struct".decls) |decl| {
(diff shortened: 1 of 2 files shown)


src/entity.zig
```diff
@@ -100,31 +100,31 @@ pub fn unloadComponent(comptime side: main.sync.Side, componentId: EntityCompone
...
 	}
 	pub fn deinit() void {
+		main.client.entity_manager.deinit();
 		inline for (@typeInfo(components).@"struct".decls) |decl| {
 			@field(components, decl.name).client.deinit();
 		}
 		inline for (@typeInfo(systems).@"struct".decls) |decl| {
 			@field(systems, decl.name).client.deinit();
 		}
-		main.client.entity_manager.deinit();
```
IntegratedQuantum, line 118: @tillpp was there a reason you did this at the end and not at the start?
With the way it's done here we save some cleanup code, since the components can assume that all the entities had their components removed

tillpp: `@tillpp was there a reason you did this at the end and not at the start?`

hmm, has been a while, but i don't think there was a reason

IntegratedQuantum (approved):

PR #3138, opened 2026-05-30
Update `/blueprint` command to use `argparse.zig`

Argmaster:
Related-to: #3073

### DIFF
--- a/src/server/command/worldedit/blueprint.zig
+++ b/src/server/command/worldedit/blueprint.zig
@@ -7,7 +7,6 @@ const vec = main.vec;
 const Vec3i = vec.Vec3i;
 
 const Dir = main.files.Dir;
-const ListUnmanaged = main.ListUnmanaged;
 const Block = main.blocks.Block;
 const Blueprint = main.blueprint.Blueprint;
 const NeverFailingAllocator = main.heap.NeverFailingAllocator;
@@ -61,7 +60,7 @@ const Args = union(enum) {
 const ArgParser = main.argparse.Parser(Args, .{.commandName = "/blueprint"});
 
 pub fn execute(args: []const u8, source: *User) void {
-	var errorMessage: main.ListUnmanaged(u8) = .{};
+	var errorMessage: ListUnmanaged(u8) = .{};
 	defer errorMessage.deinit(main.stackAllocator);
 
 	const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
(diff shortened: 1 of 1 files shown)


src/server/command/worldedit/blueprint.zig
```diff
@@ -19,73 +20,77 @@ pub const usage =
...
+		}
+	},
+	@"/blueprint load <file-name>": struct {
+		_: enum { load },
+		path: FilePath,
+
+		fn deinit(self: @This(), allocator: NeverFailingAllocator) void {
+			self.path.deinit(allocator);
+		}
+	},
```
Wunka: Okay 2 points on that:
1. why are there so many structs? Why not make: `@"/blueprint <subCommand> <file-name>`? 
2. I think there should be no naming difference between what the command says and what is in the struct (`file-name` vs `path`) (I even thought this was already the case). For me there is the direct question of what is now given to `parse` for the name? At the end I think `path` fits better because you allow paths and not only file names

(only after starting to make this comment I realized this is a draft)

Wunka (changes requested):

src/server/command/worldedit/blueprint.zig
Argmaster: > why are there so many structs? Why not make: @"/blueprint <subCommand> <file-name>?

`/blueprint <subCommand> <file-name>` is insufficient as `list` does not accept a `file-name` parameter and thus it makes no sense to include it. Moreover each of those commands has different expected evolution path since they perform very distinct operations.

But yeah I was considering it in form of `/blueprint <save/load/delete>`. I am worried about eg. compatibility params or skipping entities which would make sense for load and save but not delete. 

> I think there should be no naming difference between what the command says and what is in the struct (file-name vs path) (I even thought this was already the case). For me there is the direct question of what is now given to parse for the name? At the end I think path fits better because you allow paths and not only file names

fair.

> (only after starting to make this comment I realized this is a draft)

no issue.

Argmaster: Overall I would prefer for `deinit` to requite less boilerplate rather than to make it more awkward to use multiple structs for each of the commands. I was wondering if it would make sense to implement similar recursive `deinit` as the `parse` is recursive, so `ArgParser.deinit(result)` would fit into that clean interface.

src/server/command/worldedit/blueprint.zig
```diff
@@ -1,91 +1,95 @@
...
-		return std.meta.stringToEnum(BlueprintSubCommand, string) orelse {
-			if (string.len == 0) return .empty;
-			return .unknown;
-		};
+const Args = union(enum) {
+	@"/blueprint save <filePath>": struct {
+		_: enum { save },
+		filePath: FilePath,
+
+		fn deinit(self: @This(), allocator: NeverFailingAllocator) void {
```
IntegratedQuantum, line 27: (nothing to change) This should be removed again after we switch to an arena allocator.

src/server/command/worldedit/blueprint.zig
```diff
@@ -179,8 +155,32 @@ fn blueprintLoad(args: []const []const u8, source: *User) void {
...
+const FilePath = struct {
+	path: []const u8,
+
+	pub fn deinit(self: FilePath, allocator: NeverFailingAllocator) void {
+		allocator.free(self.path);
+	}
+
+	pub fn parse(allocator: NeverFailingAllocator, name: []const u8, arg: []const u8, errorMessage: *ListUnmanaged(u8)) error{ParseError}!FilePath {
+		if (std.mem.containsAtLeast(u8, arg, 1, "..")) {
+			errorMessage.print(main.stackAllocator, "File path for <{s}> is not allowed to contain parent directory references (..) for security reasons: '{s}'", .{name, arg});
```
IntegratedQuantum: This is insufficient. There are also global paths (leading / on linux, driverletter: on windows).
And Windows also has some fun other variants, like `\\.\`.

Instead of doing this here, and adding a false sense of security, please make an issue where we can discuss how to best approach this problem.

IntegratedQuantum (changes requested):

src/server/command/worldedit/blueprint.zig
Argmaster: Fair, I found the original issue: #1263

src/server/command/worldedit/blueprint.zig
Argmaster: Made #3156, idk if its a duplicate

IntegratedQuantum (approved):
Seems to work, apart from the list command, but that is a different issue since you didn't touch it #3157.

PR #3139, opened 2026-05-31
add Quat struct

RanPix:
needed for #2824
i couldn't find any uses for quaternions in the code so its just a struct with a helper function from #2824 added

### DIFF
--- a/src/entityModel.zig
+++ b/src/entityModel.zig
@@ -14,6 +14,7 @@ const Mat4f = vec.Mat4f;
 const Vec3d = vec.Vec3d;
 const Vec3f = vec.Vec3f;
 const Vec4f = vec.Vec4f;
+const Quat = vec.Quat;
 const NeverFailingAllocator = main.heap.NeverFailingAllocator;
 
 const c = @import("c");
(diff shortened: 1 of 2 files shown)


src/entityModel.zig
```diff
@@ -228,12 +229,12 @@ pub const EntityModel = struct {
 		};
 	}
 
-	fn convertCoordinateSystemQuat(q: Vec4f, sys: CoordinateSystem) Vec4f {
+	fn convertCoordinateSystemQuat(q: Vec4f, sys: CoordinateSystem) Quat {
```
IntegratedQuantum: This parameter is also a quaternion right?
```suggestion
	fn convertCoordinateSystemQuat(q: Quat, sys: CoordinateSystem) Quat {
```

IntegratedQuantum (changes requested):

src/entityModel.zig
RanPix: this really just adds an unnecessary conversion step for gltf quaternions as this function is only used in this file and only for changing the gltf node quaternions, i think maybe renaming it to `gltfQuat` would be better

IntegratedQuantum: If it's intended for gltf, then please use the C type and not Vec4f.

RanPix: now good?

IntegratedQuantum (approved):

PR #3144, opened 2026-05-31
fix /avatar typo

tillpp:
grafik

### DIFF
--- a/src/server/command/entity/avatar.zig
+++ b/src/server/command/entity/avatar.zig
@@ -30,7 +30,7 @@ pub fn execute(args: []const u8, source: *User) void {
 			model.server.put(source.id, .{
 				.entityModel = params.entityModel.index,
 			});
-			source.sendMessage("#00ff00You're EntityModel was changed to {s}.", .{params.entityModel.index.get().entityModelId});
+			source.sendMessage("#00ff00Your EntityModel was changed to {s}.", .{params.entityModel.index.get().entityModelId});
 		},
 		.@"/avatar" => {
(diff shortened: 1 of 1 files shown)


src/server/command/entity/avatar.zig
```diff
@@ -30,7 +30,7 @@ pub fn execute(args: []const u8, source: *User) void {
 			model.server.put(source.id, .{
 				.entityModel = params.entityModel.index,
 			});
-			source.sendMessage("#00ff00You're EntityModel was changed to {s}.", .{params.entityModel.index.get().entityModelId});
+			source.sendMessage("#00ff00Your EntityModel was changed to {s}.", .{params.entityModel.index.get().entityModelId});
```
RanPix:
```suggestion
			source.sendMessage("#00ff00Y'ro'u EntityModel was changed to {s}.", .{params.entityModel.index.get().entityModelId});
```

RanPix (commented):
can be even correcter

src/server/command/entity/avatar.zig
IntegratedQuantum: While you are changing this line:
```suggestion
			source.sendMessage("#00ff00Your entity model was changed to {s}.", .{params.entityModel.index.get().entityModelId});
```

IntegratedQuantum (changes requested):

src/server/command/entity/avatar.zig
IntegratedQuantum: this should be a language

IntegratedQuantum (approved):

PR #3145, opened 2026-05-31
Update `/copy` command to use `argparser.zig`

Argmaster:
Related-to: #3073

### DIFF
--- a/src/server/command/worldedit/copy.zig
+++ b/src/server/command/worldedit/copy.zig
@@ -10,11 +10,21 @@ const Blueprint = main.blueprint.Blueprint;
 pub const description = "Copy selection to clipboard.";
 pub const usage = "/copy";
 
+const Args = union(enum) {
+	@"/copy": struct {},
+};
+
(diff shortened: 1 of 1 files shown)


src/server/command/worldedit/copy.zig
```diff
@@ -10,11 +10,21 @@ const Blueprint = main.blueprint.Blueprint;
 pub const description = "Copy selection to clipboard.";
 pub const usage = "/copy";
 
+const Args = union(enum) {
+	@"/copy": struct {},
+};
+
+const ArgParser = main.argparse.Parser(Args, .{.commandName = "/kill"});
```
Wunka:
```suggestion
const ArgParser = main.argparse.Parser(Args, .{.commandName = "/copy"});
```
It seems I am not th only one who just copied it over and then forgot to change it

Wunka (changes requested):

IntegratedQuantum (approved):

Issue #3121, opened 2026-05-22
Subbiomes are sometimes cut off at the map border


IntegratedQuantum:
<img width="256" height="512" alt="Image" src="https://github.com/user-attachments/assets/ec0eeca8-548b-4d4a-97ba-2ed8d59859af" />

This bug was introduced in 62f6c5a106804942f8797d8aa349e8ad9bce9f48, odd that we didn't see it earlier

The problem is that the check whether a biome fits doesn't account for points outside the map fragment in generation, the fix would require checking against the map which has the margin

PR #3154, opened 2026-06-03
fix(terrain): account for biome map margins in SdfCaveGenerator

entiflags:
Fixed a bug where **CaveBiomeMapView** was initialized with 0 margin, causing calculation artifacts or missing biome points outside the immediate map fragment boundaries during SDF cave generation. [()](https://github.com/PixelGuys/Cubyz/issues/3121)


codemob-dev (changes requested):
Please use the [Cubyz formatter](github.com/PixelGuys/Cubyz-formatter) and not the one included with zig. You might need to reconfigure your editor to not reformat the code on save.

entiflags:
tell please how to do that its very annoying

codemob-dev:
> tell please how to do that its very annoying

If you are using vscode you can go into settings and set `zig.formattingProvider` to `off`

PR #3161, opened 2026-06-04
Update `/mask` command

Argmaster:
Related-to: 3073

### DIFF
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -2,6 +2,7 @@ const std = @import("std");
 
 const main = @import("main");
 const Blueprint = main.blueprint.Blueprint;
+const Mask = main.blueprint.Mask;
 const NeverFailingAllocator = main.heap.NeverFailingAllocator;
 const List = main.List;
 const User = main.server.User;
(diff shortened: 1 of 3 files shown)


src/server/command.zig
```diff
@@ -220,3 +221,20 @@ pub const EntityModel = struct {
 		}
 	}
 };
+
+pub const MaskExpression = struct {
+	// Always uses global allocator to guarantee arbitrary lifetime.
```
IntegratedQuantum: In my opinion this should be done in the command through copying. Parser structs should not care about possible use cases. e.g. for the replace commands this means a bunch of extra memory management code which really doesn't belong there.

IntegratedQuantum (changes requested):

src/server/command.zig
Argmaster: As you command, with all the implications.

src/server/command/worldedit/mask.zig
```diff
@@ -1,27 +1,56 @@
...
+		source.sendMessage("#ff0000{s}", .{errorMessage.items});
 		return;
 	};
-	source.worldEditData.mask = mask;
-	source.sendMessage("#00ff00Mask set.", .{});
+	defer result.deinit(main.stackAllocator);
+
+	switch (result) {
+		.@"/mask <mask>" => |cmd| {
+			source.worldEditData.mask = cmd.mask.mask.clone(main.globalAllocator);
```
IntegratedQuantum, line 49: Doesn't this also need to free the old entry?

IntegratedQuantum (changes requested):

src/server/command/worldedit/mask.zig
Argmaster: Yeah it probably should

IntegratedQuantum (approved):

PR #3163, opened 2026-06-04
Fix item migration issues by migrating items on load

IntegratedQuantum:
We saw everyone's iron turn into an uncraftable and unstackable version after the last migration in #2095, this was not severe since iron is mainly used in tool crafting, so no resources are lost.

For #2117 this would be more annoying since you possibly (depending on how your palette is ordered) couldn't craft your logs into anything.

Since there is no nice solution for this and I don't want to iterate through everything, I decided to go with the easiest/least effort solution: Just migrate duplicate ids on load using a direct map to the real index. This should only have a small performance impact since item loading is done rarely.

I'd be happy to hear if anyone has a better idea. (Also did we not have an issue for this problem?)

### DIFF
--- a/src/items.zig
+++ b/src/items.zig
@@ -981,7 +981,8 @@ pub const Item = union(ItemType) { // MARK: Item
 		const typ = try reader.readEnum(ItemType);
 		switch (typ) {
 			.baseItem => {
-				return .{.baseItem = try reader.readEnum(BaseItemIndex)};
+				const index = try reader.readEnum(BaseItemIndex);
+				return .{.baseItem = realItemIndices[@intFromEnum(index)]};
 			},
 			.proceduralItem => {
 				return .{.proceduralItem = try ProceduralItem.fromBytes(reader)};
@@ -1187,6 +1188,9 @@ var modifierRestrictions: std.StringHashMapUnmanaged(*const ModifierRestriction.
 pub var itemList: [65536]BaseItem = undefined;
 pub var itemListSize: u16 = 0;
 
+// Due to migrations multiple indices can map to the same item. This must be resolved during inventory loading using this map.
+var realItemIndices: [65536]BaseItemIndex = undefined;
+
 var recipeList: main.ListManaged(Recipe) = .init(main.worldArena);
 
 pub fn hasRegistered(id: []const u8) bool {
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
This by the way should also retroactively fix the iron issue in existing worlds

src/items.zig
```diff
@@ -1187,6 +1188,9 @@ var modifierRestrictions: std.StringHashMapUnmanaged(*const ModifierRestriction.
 pub var itemList: [65536]BaseItem = undefined;
 pub var itemListSize: u16 = 0;
 
+// Due to migrations multiple indices can map to the same item. This must be resolved during inventory loading using this map.
+var realItemIndices: [65536]BaseItemIndex = undefined;
```
Argmaster:
```suggestion
var deduplicatingItemMap: [65536]BaseItemIndex = undefined;
```

Or anything less general than `real`? IMO current name explains nothing really.

src/items.zig
```diff
@@ -1241,7 +1245,11 @@ pub fn register(_: []const u8, texturePath: []const u8, replacementTexturePath:
 	defer itemListSize += 1;
 
 	newItem.init(main.worldArena, texturePath, replacementTexturePath, id, zon);
-	reverseIndices.put(main.worldArena.allocator, newItem.id, @enumFromInt(itemListSize)) catch unreachable;
+	const result = reverseIndices.getOrPut(main.worldArena.allocator, newItem.id) catch unreachable;
```
Argmaster, line 1248: Following edge case comes randomly to my mind and I don't exactly have time to verify it rn:

Consider migration `cubyz:bar` -> `cubyz:foo` with two opposite palette states:

```
items:
...
33: cubyz:foo
...
37: cubyz:bar
...
```

`cubyz:foo` 33 is initialized first, then `cubyz:bar`  (which is now `cubyz:foo`)  37 is deduplicated into `cubyz:foo` 33

Opposite situation:

```
items:
...
33: cubyz:bar
...
37: cubyz:foo
...
```

`cubyz:bar` 33 is initialized first, then `cubyz:foo` 37 is deduplicated into `cubyz:bar` 33 (which is now `cubyz:foo` too... 
I'm not sure I understand implications if there are any.

Argmaster: Ad. 2: Yeah I think it doesn't matter right as long as we override `cubyz:bar` with `cubyz:foo` its fine.

IntegratedQuantum: > I'm not sure I understand implications if there are any.

There should be none. Both instances load the same asset, both will point at the same index after loading

Argmaster (approved):

PR #3165, opened 2026-06-05
Replace `.{}` with `.empty` for `List`

### LINKED ISSUE #3130: Use `.empty` instead of `.{}` for `List`
We are currently are using the old standard library way by using `.{}`
We should switch to using `.emtpy` where `empty` has the default values

Wunka:
Fixes: #3130


IntegratedQuantum:
Please make this a two-step migration: Allow both for a week before making the hard cut. This should reduce breakage on master.

Wunka:
> Please make this a two-step migration: Allow both for a week before making the hard cut. This should reduce breakage on master.

Only allow or should I also revert the migration?

IntegratedQuantum:
Allow and do the migration, so that people can see it in the codebase already.

Wunka:
I brought back the defaults

IntegratedQuantum (approved):

Issue #1365, opened 2025-04-27
Thin ice


careeoki:
- It breaks when you fall on it with enough velocity.
- The surface of water freezes into thin ice in freezing biomes. 
- It can be melted into water with heat sources.
- You can use it to craft regular ice, which cannot be melted (for building purposes)

IntegratedQuantum:
Might need some adjustments to the `onTouch` functions to allow capturing the velocity.

ikabod-kee:
I have an idea on how this could be implemented: Fall damage you take can directly translate to block damage. If you fall really far onto some dirt, the dirt will break. Thin ice is like that, except since it's so brittle, it'll take less damage for it to break.

This is a stupid idea because of the amount of lag it would bring, but I thought I'd put it out there.

IntegratedQuantum:
No this is a good idea, and I don't think it would lag too much. We'd need #2938 for that though

PR #3168, opened 2026-06-05
Add Cracked Pyrolite

ikabod-kee:
Cracked Pyrolite: A block found within the chasms of the Mantle. Be careful when placing torches around in the darkness, for you might fall into this deadly trap.

https://github.com/user-attachments/assets/13a6bc0d-0274-42dd-a78d-e1d110f40681


TyrrelMeter:
That's an amazing idea. Just one question - will it be possible for someone to obtain the block in survival for traps?

Crepestrom:
No because player placed cracked pyrolite will not decay like that

TyrrelMeter:
I wish it would, it would be fun for traps

du82:
I like it! Maybe it could be like quicksand rather than evaporating? The more blocks the better

TyrrelMeter:
A quicksand like block could be a different block I think. Cracked pyrolite is supposed to shatter, not for poor snakes to fall through.

careeoki (changes requested):
1) This has to be obtainable in survival somehow. And with how placing decaying blocks work, there should be a way to place them so that they actually decay, for making your own traps.
2) I think this would be a lot more fun if it would break if you fell on it (like #1365), so I think it would be best it we waited until then. 

Also, why shouldn't other stones have cracked variants as well? It could make caving more interesting in general, you would have to be more cautious where you mine, or maybe we could hide caves or structures behind cracked walls. (obviously there wouldn't be as many traps as there would be in the mantle)
And I'm sure players wouldn't mind more stone textures to build with.
If we did then we would probably want a less bloaty solution, maybe a "chipper" item that turns stones into their cracked variants or whatever.

TyrrelMeter:
I don't think we need velocity breaking to implement this. It already has fine mechanics and velocity breaking can always be added to the block later. There is no reason why it has to wait till then, is there? It already seems like a threat/hazard to me, which is what it is meant to be. No reason to wait for a possibly long away PR.

careeoki:
> No reason to wait for a possibly long away PR.

I will always rather wait to do something right the first time than use half-measures. There are many many things I want to add that _technically_ we could do right now, but I don't because we just aren't there yet.
And what usually happens is that people get complacent with the half-measure and then there is less motivation to actually finalize the feature.

TyrrelMeter:
This isn't a half measure. It is fully functional using systems already present in the game, and we can add the velocity break mechanic in the same PR as the ice breaking whenever that is added. This doesn't make it take any longer to add ice breaking does it?

ikabod-kee:
> I think this would be a lot more fun if it would break if you fell on it (like https://github.com/PixelGuys/Cubyz/issues/1365)

Why should cracked stone act like thin ice? Wouldn't that defeat the purpose of having it since they'd act identically?

TyrrelMeter:
Maybe if it acted like thin ice but you needed a higher velocity to break through? Though personally I think if you were going to have it break I'd want it when placed/stepped on. But it doesn't need to act like thin ice, and even if it does that isn't a reason to postpone it.

ikabod-kee:
Yes, the point of the trap is to catch you off guard when placing torches in dark areas.

ikabod-kee:
> And I'm sure players wouldn't mind more stone textures to build with.

Does that mean we can get cobbled variants of the different stones? 🥺

ikabod-kee:
> If we did then we would probably want a less bloaty solution, maybe a "chipper" item that turns stones into their cracked variants or whatever.

I can actually see that working. You could apply it to stone blocks like an ore :P

careeoki:
> Does that mean we can get cobbled variants of the different stones? 🥺

Yeah, yeah, im already planning on it

careeoki:
> Why should cracked stone act like thin ice?

Not saying it should, I was thinking something similar to terrarias cracked dungeon bricks. But they would both use the same system to detect the player's movement

Issue #3125, opened 2026-05-24
A proposal for command modding and ambiguous commands


IntegratedQuantum:
Commands are a bit of a special-case:
Mods will inevitably want to add commands with the same name. For all other assets this is solved by adding `mod:` in front of it, but for commands this would be rather annoying. Autocomplete would make this bearable, but it would still make commands needlessly long.

Beyond that there is a second problem: Ambiguous commands. As commands get more complex we (or a mod author) may find ourselves in a situation where multiple results may be parsed. One example of this would `/tp <biome>` and (hypothetical) `/tp <structure>`, if any biome id matches a structure id, it is unclear where to tp the player.

To solve this I propose to send an error to the player when this happens and give them an option to resolve the conflicts, possible message:
```
Found ambiguous arguments for command ..., could be
/tp[1] <biome>
/tp[4] <structure>
```
The numbers would be the index based on the order of the command arguments union (and the order of mods), to select one variant you could then just run `/tp[1] cubyz:xyz`

Wunka:
While this is a solution. I still like the `mod:` solution more. This would give the player a better idea about what the index means. I suggest maybe something in this direction:
```
Found ambiguous arguments for command ..., could be
/tp <biome>
/tp <structure>
please type either /tp biome:<id> or /tp structure:<id>
```
this could then also later be used for non index based optional arguments. 
We probably need some kind of reserved key to indicate that the start is talking about the argument name.
For overloaded commands you can do:
```
Found ambiguous arguments for command ..., could be
/cubyz:tp <biome>
/mod:tp <biome>
```
While this does "bloat" the commands. I think it makes it more explicit and would pave the way for a maybe later more complex system

IntegratedQuantum:
Adding biome: to the string may in itself also cause problems, e.g. if there is an addon with the same name, then it could cause more issues.

I do however agree that `mod:` is nicer. So let me extend on that:
```
/cubyz:tp:biome
/mod:tp:structure
```
Where the last part would be the name of the command in the code (currently it's `@"/tp ..."`, but once we auto-generate these, we could replace them with a better name).

Wunka:
> Adding biome: to the string may in itself also cause problems, e.g. if there is an addon with the same name, then it could cause more issues.

That's why I said this:
> We probably need some kind of reserved key to indicate that the start is talking about the argument name.

Something like `/tp $biome:cubyz:rare/rainbow_forest`
Or mabye if we don't use `=` somewhere:
[...]

PR #3170, opened 2026-06-06
Add set argument support for argparser

Wunka:
This adds support specified in [#3125](https://github.com/PixelGuys/Cubyz/issues/3125#issuecomment-4533560687)
Example:
`/gamemode .playerIndex=@0 .mode=survival`


IntegratedQuantum (changes requested):
It should be allowed to specify flags in an arbitrary order.

PR #3182, opened 2026-06-06
fix host losing its bag

tillpp:
maybe we should add even more asserts?

### DIFF
--- a/src/server/Entity.zig
+++ b/src/server/Entity.zig
@@ -61,10 +61,9 @@ pub fn save(self: *const @This(), allocator: NeverFailingAllocator, audience: ma
 		zon.put("name", name);
 	}
 
+	// TMP
+	std.debug.assert(main.entity.components.@"cubyz:model".server.get(self.id) != null);
 
-	// TMP 
-	std.debug.assert(main.entity.components.@"cubyz:model".server.get(self.id)!=null);
-	
 	return zon;
 }
 pub fn deinit(self: *@This(), comptime side: main.sync.Side) void {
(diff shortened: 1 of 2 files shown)


src/server/Entity.zig
```diff
@@ -60,6 +60,10 @@ pub fn save(self: *const @This(), allocator: NeverFailingAllocator, audience: ma
 	if (self.name) |name| {
 		zon.put("name", name);
 	}
+
+	// TMP
+	std.debug.assert(main.entity.components.@"cubyz:model".server.get(self.id) != null);
```
IntegratedQuantum, line 65: This assert gives no meaningful information, we already know that it doesn't have it when saving, and we already know the stacktrace when it saves.

src/server/server.zig
```diff
@@ -167,6 +167,7 @@ pub const User = struct { // MARK: User
 
 	pub fn deinit(self: *User) void {
 		std.debug.assert(self.refCount.load(.monotonic) == 0);
+		std.debug.assert(self.id != std.math.maxInt(u32));
```
IntegratedQuantum: This assertion does fail when I disconnect the player before it sent over its name (add the following code to the handshake in protocols.zig):
```diff
				.signatureRequest => {
+					conn.disconnect();
					const signature1Len = try reader.readVarInt(usize);
```

src/server/Entity.zig
tillpp: the game stores the player every few seconds.
With this you will notice immediatly when you lost your component, which might help find a  cause

IntegratedQuantum: Well, but with the other assert triggered we already have a plausible cause

tillpp:
already fixed with https://github.com/PixelGuys/Cubyz/pull/3193 and #3190

PR #3184, opened 2026-06-07
Sort Button

Crepestrom:
Sorts items in your inventory

Currently only compresses items in inventory but that will be changed

### DIFF
--- a/assets/cubyz/blocks/basalt/_defaults.zig.zon
+++ b/assets/cubyz/blocks/basalt/_defaults.zig.zon
@@ -1,5 +1,5 @@
 .{
-	.tags = .{.stone, .mineable, .basalt},
+	.tags = .{.basalt, .stone, .mineable},
 	.blockHealth = 30,
 	.blockResistance = 1,
 	.drops = .{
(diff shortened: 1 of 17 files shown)


Wunka (changes requested):
Ok after looking at your code again. I really think that this should be clientside where your `swaps `are then just DepositOrSwap (your `move `too). 

Also I think this code should have comments explaining what you are exactly doing.
If I understand correcltly you first do `compress` (making for example 1 Slate, 1 Slate into 2 Slate. Then you sort first the ProcduralTools by: moving them to the start and then sorting them. Then you are doing the same for baseItems. 

I also think that you don't need to sort this yourself.
For example if you look at https://ziglang.org/documentation/0.16.0/std/#std.sort.insertion you can define your own context struct with a lessThan and a swap method and zig will do the rest.
so something like:
```zig
const Context = struct {
        inventory: Inventory,

        pub fn lessThan(ctx: @This(), a: usize, b: usize) bool {
            // bla bla bla
        }

        pub fn swap(ctx: @This(), a: usize, b: usize) void {
            // use depositOrSwap to swap
        }
    };
```
and then you can just run `std.sort.(whateEverAlgorithm which doens't use memory)Context(...)` 
It doesn't do compress for you so you would still need to do this but else this would make your code much better

Crepestrom:
closed due to a new pr with better changes #3194

PR #3194, opened 2026-06-08
Sort button attempt 2

Crepestrom:
Sort button now functions as normal

Before sorting it will compress all items into their max stacks and push them all to the lowest index getting rid of any holes

Then the sort button will first seperate items into procederal items and nonprocedural items
then they will be sorted by tag
items that share the same first tag will be internally sorted by their second tag and so on and so forth
procedural items are sorted by durability

image

Also allows buttons to be hidden
if a button is hidden the icon within it is extended to the size of the whole button (its scaled up).


Wunka (changes requested):
I still think that `std.sort` is probably way nicer and easier. Here is an example (not tested) based on your current code (but of course not completly mapped and I with an added `getTags` method for the `Item` struct):
```zig
pub fn sortItems(source: ClientInventory, ignoredSlotCount: usize) void {
	compressItems(source);
	const ctx: SortContext = .{.inv = source};
	std.sort.insertionContext(ignoredSlotCount, source.super._items.len, ctx);
}

const SortContext = struct {
	inv: ClientInventory,

	pub fn lessThan(ctx: @This(), a: usize, b: usize) bool {
		const itemA = ctx.inv.getItem(a);
		const itemB = ctx.inv.getItem(b);

		if(itemA == .null) return false;
		if(itemB == .null) return true;

		const itemATags = itemA.getTags().?;
		const itemBTags = itemB.getTags().?;

		for(0..@min(itemATags.len, itemBTags.len)) |i| {
			if(itemATags[i] == itemBTags[i]) continue;
			return std.mem.lessThan(u8, itemATags[i].getName(), itemBTags[i].getName());
		}
		if(itemATags.len != itemBTags.len) return itemATags.len < itemBTags.len;

		return std.mem.lessThan(u8, itemA.id().?, itemB.id().?);
	}

	pub fn swap(ctx: @This(), a: usize, b: usize) void {
		main.sync.client.executeCommand(.{.depositOrSwap = .{
			.dest = .{.inv = ctx.inv.super, .slot = @intCast(a)}, 
			.source = .{.inv = ctx.inv.super, .slot = @intCast(b)},
		}});
	}
};
```

Crepestrom:
also fixes #3195

src/Inventory.zig
```diff
@@ -531,6 +532,150 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+		const tagCount = findHighestTagCount(source);
+		sortProceduralItemsByTag(source, 0, ignoredSlotCount);
+		for (0..tagCount) |index| {
+			sortBaseItemsByTag(source, index, ignoredSlotCount);
+		}
+	}
+
+	pub fn compressItems(source: ClientInventory) void {
+		for (source.super._items, 0..) |invStack, slot| {
+			for (source.super._items, 0..) |checkedInvStack, checkedSlot| {
```
IntegratedQuantum, line 578: If the item is filled with mostly the same items, this will spawn O(n²) inventory commands. That is excessive.

src/Inventory.zig
```diff
@@ -531,6 +532,150 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+				if (std.meta.eql(invStack.item, checkedInvStack.item)) {
+					main.sync.client.executeCommand(.{.depositOrSwap = .{.dest = .{.inv = source.super, .slot = @intCast(checkedSlot)}, .source = .{.inv = source.super, .slot = @intCast(slot)}}});
```
IntegratedQuantum: deposit should be enough

src/Inventory.zig
```diff
@@ -531,6 +532,150 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 		main.sync.client.executeCommand(.{.craftProceduralItem = .init(destinations, workbenchInv)});
 	}
 
+	pub fn sortItems(source: ClientInventory, ignoredSlotCount: usize) void {
+		compressItems(source);
+		orderItems(source, ignoredSlotCount);
+		const tagCount = findHighestTagCount(source);
+		sortProceduralItemsByTag(source, 0, ignoredSlotCount);
+		for (0..tagCount) |index| {
+			sortBaseItemsByTag(source, index, ignoredSlotCount);
```
IntegratedQuantum: Instead of spamming sort operations, please make a single `lessThan` function that contains this logic and then sort once using it.

src/Inventory.zig
```diff
@@ -531,6 +532,150 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			for (source.super._items, 0..) |checkedInvStack, checkedSlot| {
+				if (std.meta.eql(invStack.item, checkedInvStack.item)) {
+					main.sync.client.executeCommand(.{.depositOrSwap = .{.dest = .{.inv = source.super, .slot = @intCast(checkedSlot)}, .source = .{.inv = source.super, .slot = @intCast(slot)}}});
+				}
+			}
+		}
+	}
+
+	pub fn orderItems(source: ClientInventory, ignoredSlotCount: usize) void {
+		// put proceduralItems first then BaseItems
```
IntegratedQuantum: This should also be included in the lessThan function

src/Inventory.zig
IntegratedQuantum: Also in my opinion the `lessThan` function should be provided by the caller, since there is no single obvious order.

IntegratedQuantum (changes requested):
About the UI: An icon like this is hard to understand, there is no such thing as a universal sorting icon, and a funnel is often used for filtering.
I think Terraria has a simple and intuitive UI for this:

Image

Crepestrom:
> I still think that `std.sort` is probably way nicer and easier. Here is an example (not tested) based on your current code (but of course not completly mapped and I with an added `getTags` method for the `Item` struct):
> 
> ```zig
> pub fn sortItems(source: ClientInventory, ignoredSlotCount: usize) void {
> 	compressItems(source);
> 	const ctx: SortContext = .{.inv = source};
> 	std.sort.insertionContext(ignoredSlotCount, source.super._items.len, ctx);
> }
> 
> const SortContext = struct {
> 	inv: ClientInventory,
> 
> 	pub fn lessThan(ctx: @This(), a: usize, b: usize) bool {
> 		const itemA = ctx.inv.getItem(a);
> 		const itemB = ctx.inv.getItem(b);
> 
> 		if(itemA == .null) return false;
> 		if(itemB == .null) return true;
> 
> 		const itemATags = itemA.getTags().?;
> 		const itemBTags = itemB.getTags().?;
> 
> 		for(0..@min(itemATags.len, itemBTags.len)) |i| {
> 			if(itemATags[i] == itemBTags[i]) continue;
> 			return std.mem.lessThan(u8, itemATags[i].getName(), itemBTags[i].getName());
> 		}
> 		if(itemATags.len != itemBTags.len) return itemATags.len < itemBTags.len;
> 
> 		return std.mem.lessThan(u8, itemA.id().?, itemB.id().?);
> 	}
> 
> 	pub fn swap(ctx: @This(), a: usize, b: usize) void {
> 		main.sync.client.executeCommand(.{.depositOrSwap = .{
> 			.dest = .{.inv = ctx.inv.super, .slot = @intCast(a)}, 
> 			.source = .{.inv = ctx.inv.super, .slot = @intCast(b)},
> 		}});
> 	}
> };
> ```

I realized how to turn tags into numbers so I’m gonna try this

src/Inventory.zig
Crepestrom: ok i made it a bit more effecient

Crepestrom:
is ready for rereview

src/Inventory.zig
```diff
@@ -531,6 +532,68 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			return &[_]Tag{};
+		} else if (givenItem == .proceduralItem) {
+			return givenItem.proceduralItem.type.tags();
+		} else if (givenItem == .baseItem) {
+			return givenItem.baseItem.tags();
+		} else {
+			std.log.err("getSortingTag: Could not find item class {}", .{givenItem});
+			return &[_]Tag{};
+		}
+	}
```
Wunka: This should be in the `Item` struct and use a `switch(self)` so you don't need the else at the end

Wunka (changes requested):
Nice. Now the code looks much better

src/Inventory.zig
```diff
@@ -531,6 +532,68 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			if ((itemA == .proceduralItem) and (itemB != .proceduralItem)) return true;
+
+			const itemATags = getTagsFromItem(itemA);
+			const itemBTags = getTagsFromItem(itemB);
+
+			for (0..@min(itemATags.len, itemBTags.len)) |i| {
+				if (itemATags[i] == itemBTags[i]) continue;
+				return std.mem.lessThan(u8, itemATags[i].getName(), itemBTags[i].getName());
+			}
+			if (itemATags.len != itemBTags.len) return itemATags.len < itemBTags.len;
```
Wunka: this is probably not the best thing (I know it is from me but it was at that time made up on the spot). We probably need to enforce that the tags are sorted so that for example when ItemA has a,b tags and ItemB has c,a,b It gets compared with a,b and a,b,c. Either by sorting them when we load the tags or here. Or some other way.

src/Inventory.zig
```diff
@@ -531,6 +532,68 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+
+			return std.mem.lessThan(u8, itemA.id().?, itemB.id().?);
```
Wunka: I think it would be nice to add here that proceduralItems are sorted by durability, when even the id is the same.

Wunka (changes requested):

src/Inventory.zig
Argmaster: Definitely on load.

Crepestrom: If I do this I think it would cause read order issues if I make it check if the first tag of ItemA exists in ItemB because they could have different orders
I could fix this with forcing them to read based on the enum int value but I like what I have now because it gives control to developers on how items are catogorized and sorted

you could create tags that are used in nothing but sorting so that all flowers are sorted together under the category of cuttable

Wunka: > If I do this I think it would cause read order issues if I make it check if the first tag of ItemA exists in ItemB because they could have different orders I could fix this with forcing them to read based on the enum int value but I like what I have now because it gives control to developers on how items are catogorized and sorted
> 
> you could create tags that are used in nothing but sorting so that all flowers are sorted together under the category of cuttable

We didn't mean that. We meant that the tags should be already sorted so what you at the start said doesn't need to be done. This sorting algorithm could then also be changed by a user if they want to. 
You just need to add in `src/tag.zig` 
```diff
pub fn loadTagsFromZon(_allocator: main.heap.NeverFailingAllocator, zon: main.ZonElement) []Tag {
	const result = _allocator.alloc(Tag, zon.toSlice().len);
	for (zon.toSlice(), 0..) |tagZon, i| {
		result[i] = Tag.find(tagZon.as([]const u8) orelse blk: {
			std.log.err("Tag array field {s} has incorrect type, expected string", .{@tagName(tagZon)});
			break :blk "incorrect";
		});
	}
+	// some kind of sorting before returning the result
	return result;
}

Comment (Crepestrom) [reply src/Inventory.zig]: I don’t understand what you mean by sorting the tags

Comment (Crepestrom) [reply src/Inventory.zig]: Oh I see
You mean so that
 C A B
B A C
C B A

all get sorted together A B C
Right?

Comment (Wunka) [reply src/Inventory.zig]: > Oh I see You mean so that C A B B A C C B A
> 
> all get sorted together A B C Right?

yes exactly.

Comment (Crepestrom) [reply src/Inventory.zig]: No
Id rather give the developer control on sorting rather than enforce is based on the order tags were read into memory
Even if that means mistakes can make sorting look weird

Comment (Wunka) [reply src/Inventory.zig]: But the sorting of the tags is also clientside. So it can also be changed.

File: src/Inventory.zig (hunk 7 above)
Comment (Crepestrom) [reply src/Inventory.zig]: added

**Comment by Crepestrom** [comment]
this is ready to review again

**Comment by Crepestrom** [comment]
added this to chest as well because im bored

File: src/Inventory.zig [hunk 8]
```diff
@@ -538,6 +539,67 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			const itemATags = getTagsFromItem(itemA);
+			const itemBTags = getTagsFromItem(itemB);
+
+			for (0..@min(itemATags.len, itemBTags.len)) |i| {
+				if (itemATags[i] == itemBTags[i]) continue;
+				return std.mem.lessThan(u8, itemATags[i].getName(), itemBTags[i].getName());
+			}
+			if (itemATags.len != itemBTags.len) return itemATags.len < itemBTags.len;
+			if ((itemA == .proceduralItem) and (itemB == .proceduralItem)) {
+				std.log.debug("checking durability", .{});
```
Wunka: you forgot to remove your debug message here

src/Inventory.zig
```diff
@@ -538,6 +539,67 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+				return (itemA.proceduralItem.durability > itemB.proceduralItem.durability);
+			}
+
+			return std.mem.lessThan(u8, itemA.id().?, itemB.id().?);
```
Wunka: I would say first compare the id than the durability. That way procedural item types are sorted together.

src/Inventory.zig
```diff
@@ -538,6 +539,67 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+		}
+	};
+
+	pub fn getTagsFromItem(givenItem: Item) []const Tag {
+		switch (givenItem) {
+			.null => return &[_]Tag{},
+			.proceduralItem => return givenItem.proceduralItem.type.tags(),
+			.baseItem => return givenItem.baseItem.tags(),
+		}
+	}
```
Wunka: I still think this should be in the item struct

Wunka (changes requested):

Wunka:
But it really looks nice! this will be a great addition to the game

src/Inventory.zig
Crepestrom: dont worry it compares by tags then by durability
procudural items have tags

Wunka: mhh. okay, then I will let it go.

src/Inventory.zig
Crepestrom: ill move that now

Crepestrom:
thanks for reviewing

src/Inventory.zig
Crepestrom: ok i did that now

src/Inventory.zig
Crepestrom: removed

src/gui/components/Button.zig
Wunka: I think you forgot to remove that here? This would then render the child two times. Once here and once in the render function.
(I just realized that, because you added code to render but didn't remove it anywhere)

Wunka (changes requested):
Everything else looks good.

src/gui/components/Button.zig
Crepestrom: i fixed it rendering twice now

src/gui/components/Button.zig
```diff
@@ -170,7 +185,4 @@ pub fn render(self: *Button, mousePosition: Vec2f) void {
 
 	const oldColor = draw.setColor(if (self.disabled) 0xff808080 else 0xffffffff);
 	defer draw.restoreColor(oldColor);
```
Wunka: this can also be removed. this does nothing now

Wunka (changes requested):

src/gui/components/Button.zig
Crepestrom: gonna test to make sure deleting it doesnt change visuals
yep same bejavopr

Crepestrom:
thanks for catching that so quantum doesnt have to

Wunka:
> thanks for catching that so quantum doesnt have to

did you forget to push the remove?

Crepestrom:
ye

Wunka (approved):

src/gui/components/Button.zig
```diff
@@ -148,6 +152,17 @@ pub fn mainButtonReleased(self: *Button, mousePosition: Vec2f) void {
...
+
+	const oldColor = draw.setColor(if (self.disabled) 0xff808080 else 0xffffffff);
+	defer draw.restoreColor(oldColor);
+	const textPos = self.pos + self.size/@as(Vec2f, @splat(2.0)) - self.child.size()/@as(Vec2f, @splat(2.0));
+	self.child.mutPos().* = textPos;
+	if (self.hidden) self.child.mutSize().* = self.size;
+	self.child.render(mousePosition - self.pos);
+}
+
+pub fn renderButton(self: *Button, mousePosition: Vec2f) void {
```
IntegratedQuantum: I think this name would be better.
```suggestion
fn renderBackground(self: *Button, mousePosition: Vec2f) void {
```

src/gui/components/Button.zig
```diff
@@ -58,6 +58,7 @@ size: Vec2f,
 disabled: bool = false,
 pressed: bool = false,
 hovered: bool = false,
+hidden: bool = false,
```
IntegratedQuantum: hidden could mean something different.
```suggestion
hideBackground: bool = false,
```

src/gui/windows/chest.zig
```diff
@@ -29,6 +29,12 @@ pub var window = GuiWindow{
 const padding: f32 = 8;
 var itemSlots: main.List(*ItemSlot) = .empty;
 
+var sortIcon: Texture = undefined;
+
+pub fn init() void {
+	sortIcon = Texture.initFromFile("assets/cubyz/ui/inventory/sort_icon.png");
```
IntegratedQuantum, line 35: Please also deinit the texture.

src/gui/windows/chest.zig
```diff
@@ -39,7 +45,16 @@ pub fn setInventory(selectedInventory: main.items.Inventory.ClientInventory) voi
 	openInventory = selectedInventory;
 }
 
+pub fn sortItems(target: main.items.Inventory.ClientInventory) void {
```
IntegratedQuantum:
```suggestion
fn sortItems(target: main.items.Inventory.ClientInventory) void {
```

src/gui/windows/chest.zig
```diff
@@ -39,7 +45,16 @@ pub fn setInventory(selectedInventory: main.items.Inventory.ClientInventory) voi
...
+	const ignoredSlotCount = 0;
+	target.sortItems(ignoredSlotCount);
+}
+
 pub fn onOpen() void {
+	const sortCallback: main.callbacks.SimpleCallback = .{.inner = @ptrCast(&sortItems), .data = &openInventory};
```
IntegratedQuantum: Please use the init function. ptrCast is unsafe (and you are already using it incorrectly, the functions takes a value, and you cast it to a function that takes a pointer)

src/gui/windows/inventory.zig
```diff
@@ -30,9 +30,11 @@ pub var window = GuiWindow{
 const padding: f32 = 8;
 
 var craftingIcon: Texture = undefined;
+var sortIcon: Texture = undefined;
 
 pub fn init() void {
 	craftingIcon = Texture.initFromFile("assets/cubyz/ui/inventory/crafting_icon.png");
+	sortIcon = Texture.initFromFile("assets/cubyz/ui/inventory/sort_icon.png");
```
IntegratedQuantum, line 37: deinit this

src/gui/windows/inventory.zig
```diff
@@ -41,14 +43,23 @@ pub fn deinit() void {
...
+	window.titleBar.?.add(Button.initIcon(.{0, 0}, .{9, 9}, sortIcon, .{.onAction = sortCallback, .hidden = true}));
+
 	const list = VerticalList.init(.{padding, padding + 16}, 300, 0);
 	// Some miscellanious slots and buttons:
 	// TODO: armor slots, backpack slot + stack-based backpack inventory, other items maybe?
 	{
 		const row = HorizontalList.init();
 		blk: {
-			row.add(GuiComponent.BagSlot.init(.{0, 0}, main.entity.components.@"cubyz:bag".client.getBag(main.game.Player.id) orelse break :blk));
+			row.add(GuiComponent.BagSlot.init(.{32, 0}, main.entity.components.@"cubyz:bag".client.getBag(main.game.Player.id) orelse break :blk));
```
IntegratedQuantum: unintended change I assume?

src/assets.zig
```diff
@@ -425,6 +425,7 @@ fn assignBlockItem(stringId: []const u8) !void {
 	const index = items.BaseItemIndex.fromId(stringId).?;
 	const item = &items.itemList[@intFromEnum(index)];
 	item.block = block;
+	item.tags = blocks.parseBlock(stringId).tags();
```
IntegratedQuantum: This is questionable, I'm not sure if we want to merge block and item tags, it may have unintended side-effects.
Also it's wrong, this would fully overwrite any item tags.

(also you forgot to mention #3195 in the PR description)

src/Inventory.zig
```diff
@@ -538,6 +539,58 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			}
+			if (itemATags.len != itemBTags.len) return itemATags.len < itemBTags.len;
+			if ((itemA == .proceduralItem) and (itemB == .proceduralItem)) {
+				return (itemA.proceduralItem.durability > itemB.proceduralItem.durability);
+			}
+
+			return std.mem.lessThan(u8, itemA.id().?, itemB.id().?);
+		}
+
+		pub fn swap(ctx: @This(), a: usize, b: usize) void {
```
IntegratedQuantum: Resources are not free, and sending O(n²) swap operations to the server is a waste of network and server resources.

Instead please sort an array of indicies into the inventory. Then you can just swap them with their target location, this only sends O(n) operations worst case, O(0) best case.

IntegratedQuantum (changes requested):

src/Inventory.zig
IntegratedQuantum: I'm not sure if sorting by tags even is the right approach. It seems fragile, and just one weird tag in one item could make it sort it completely differently.

The only thing I could see working is defining priorities for specific tags and then sorting by the tag with highest priority.

But for now I'd suggest to just sort by name.

src/assets.zig
Crepestrom: ok it has been changed to combine the tags instead of overwriting them

src/Inventory.zig
Crepestrom: would it be fine if i added a seperate tag system just for sorting then?
like a extra line to each block telling how it sorts

Crepestrom: that way its more stable than tags or folders

src/Inventory.zig
Crepestrom: ok i made a better algorithem that takes O(n) operations instead

src/gui/components/Button.zig
```diff
@@ -91,6 +92,7 @@ fn defaultOnAction(_: usize) void {}
 const Options = struct {
 	onAction: main.callbacks.SimpleCallback = .{},
 	disabled: bool = false,
+	hidden: bool = false,
```
IntegratedQuantum:
```suggestion
	hideBackground: bool = false,
```

src/gui/windows/chest.zig
```diff
@@ -39,7 +46,16 @@ pub fn setInventory(selectedInventory: main.items.Inventory.ClientInventory) voi
 	openInventory = selectedInventory;
 }
 
+fn sortItems() void {
+	const target = Player.inventory;
+	const ignoredSlotCount = 0;
+	target.sortItems(ignoredSlotCount);
```
IntegratedQuantum: I assume you made ignoredSlotCount a variable to make it more readable at the callsite?
I appreciate this, but I would suggest something even better: An options struct `options: struct {ignoredSlotCount: usize}`, then at the callsite you can pass `target.sortItems(.{.ignoredSlotCount = 0});`

src/gui/windows/chest.zig
```diff
@@ -39,7 +46,16 @@ pub fn setInventory(selectedInventory: main.items.Inventory.ClientInventory) voi
...
+}
+
 pub fn onOpen() void {
+	window.titleBar = HorizontalList.init();
+	window.titleBar.?.add(Button.initIcon(.{0, 0}, .{9, 9}, sortIcon, .{.onAction = .init(sortItems), .hidden = true}));
```
IntegratedQuantum: this is leaking. Please run the game in debug mode and close it normally to check for leaks in the future.

src/server/command/clear copy.zig
IntegratedQuantum, line 1: ?

src/utils/list.zig
```diff
@@ -165,6 +165,12 @@ pub fn ListManaged(comptime T: type) type {
 			return old;
 		}
 
+		pub fn swap(self: *@This(), a: usize, b: usize) void {
```
IntegratedQuantum: We have std.mem.swap for this.

src/Inventory.zig
```diff
@@ -538,6 +539,82 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 		main.sync.client.executeCommand(.{.craftProceduralItem = .init(destinations, workbenchInv)});
 	}
 
+	pub fn sortItems(source: ClientInventory, ignoredSlotCount: usize) void {
+		compressItems(source);
+		const InventorySize: usize = source.super.size() - ignoredSlotCount;
```
IntegratedQuantum: This is not a type
```suggestion
		const inventorySize: usize = source.super.size() - ignoredSlotCount;
```

src/Inventory.zig
```diff
@@ -538,6 +539,82 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+		var SortList = main.ListManaged(usize).init(main.globalAllocator);
```
IntegratedQuantum: This is not a type, local allocations should use the stackAllocator
```suggestion
		var sortList = main.ListManaged(usize).init(main.stackAllocator);
```

src/Inventory.zig
```diff
@@ -538,6 +539,82 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+		var iteratorList = main.ListManaged(usize).init(main.globalAllocator);
```
IntegratedQuantum: This list is used once and the values are easy to compute, which makes it kind of pointless.

src/Inventory.zig
```diff
@@ -538,6 +539,82 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+		defer SortList.deinit();
+		defer iteratorList.deinit();
+		for (0..InventorySize) |i| {
+			SortList.append(i + ignoredSlotCount);
+			iteratorList.append(i + ignoredSlotCount);
+		}
+		var ctx: SortContext = .{.inv = source, .sortlist = SortList};
+		std.sort.insertionContext(0, InventorySize, &ctx);
```
IntegratedQuantum: Just use `insertion` with lessThan function, there is no need for the context version.

src/Inventory.zig
```diff
@@ -538,6 +539,82 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+		// we sort the default (12..inventorysize) array into a sorted array
```
IntegratedQuantum: this comment appears to be outdated

src/Inventory.zig
```diff
@@ -538,6 +539,82 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 		main.sync.client.executeCommand(.{.craftProceduralItem = .init(destinations, workbenchInv)});
 	}
 
+	pub fn sortItems(source: ClientInventory, ignoredSlotCount: usize) void {
+		compressItems(source);
```
IntegratedQuantum: Shouldn't compression also account for ignored slots?

IntegratedQuantum (changes requested):

src/Inventory.zig
IntegratedQuantum: I think it would be best to propose that in a separate PR.

Crepestrom: names are a worse way to sort these
currently with tags all stone blocks sort together like basalt and slate
without this basalt and slate would not be grouped together
if you dont want a tag to be used for sorting you add it to the end of the tags list

src/Inventory.zig
Crepestrom: do you mean iterator list or sort list?

src/gui/windows/chest.zig
Crepestrom: i do not know what is leaking

src/Inventory.zig
Crepestrom: both are decently important
the sortlist says which slot things have to be at
the iterator list helps figure out how to swap items to match the sortlist

IntegratedQuantum: iteratorList, it's never changed and trivially constructed.

Crepestrom: no it is changed

Crepestrom: also i changed the name to intermidiaryList

Crepestrom: in the most recent version on line 568

src/gui/windows/chest.zig
Crepestrom: well it says its not leaking to me anymore

src/Inventory.zig
IntegratedQuantum: That only works because, by chance, the authors layed it out that way. And tag order is not something I want to review. And it's going to be a mess for addons since the authors wouldn't know that the order of tags is relevant for anything.

Crepestrom: ok i implemented it i will make a seperate pr then for more control

src/gui/windows/chest.zig
```diff
@@ -39,7 +46,15 @@ pub fn setInventory(selectedInventory: main.items.Inventory.ClientInventory) voi
 	openInventory = selectedInventory;
 }
 
+fn sortItems() void {
+	const target = Player.inventory;
+	target.sortItems(.{.ignoredSlotCount = 0});
```
IntegratedQuantum, line 51: Isn't this supposed to sort the chest inventory?

IntegratedQuantum (changes requested):
Before my next review please figure out how to get leak checking working. It still leaks.
To use leak checking you need to build in debug mode, test your feature, and then you need to exit the game normally (do not kill it through Ctrl+C or whatever), then you will see the leaks in the terminal output.
Please ask someone for help if you can't figure it out on your own. But we cannot have you as a contributor if every time I review I need to run the leak check for you.

Crepestrom:
ok what allocator is leaking?
ill see if i can fix that check

du82:
I don't think the sort button should be that big. The UI / UX is pretty bad on it. Maybe it should live in the toolbar as a small filter icon

Wunka:
> I don't think the sort button should be that big. The UI / UX is pretty bad on it. Maybe it should live in the toolbar as a small filter icon

Are you looking at the old pictures?

src/gui/components/Button.zig
```diff
@@ -145,6 +149,17 @@ pub fn mainButtonReleased(self: *Button, mousePosition: Vec2f) void {
 }
 
 pub fn render(self: *Button, mousePosition: Vec2f) void {
+	if (!self.hideBackground) renderBackground(self, mousePosition);
+
+	const oldColor = draw.setColor(if (self.disabled) 0xff808080 else 0xffffffff);
+	defer draw.restoreColor(oldColor);
+	const textPos = self.pos + self.size/@as(Vec2f, @splat(2.0)) - self.child.size()/@as(Vec2f, @splat(2.0));
+	self.child.mutPos().* = textPos;
+	if (self.hideBackground) self.child.mutSize().* = self.size;
```
IntegratedQuantum: Please do this change during the initialization. There is no need to change it ever single frame, and it may not react correctly in the first frame.

IntegratedQuantum (changes requested):
Not sure how you missed it, but it's still leaking.

Also stack merging doesn't work properly:
Image
when clicking the button becomes
Image

src/gui/components/Button.zig
IntegratedQuantum: Please do this

src/assets.zig
```diff
@@ -425,6 +426,8 @@ fn assignBlockItem(stringId: []const u8) !void {
 	const index = items.BaseItemIndex.fromId(stringId).?;
 	const item = &items.itemList[@intFromEnum(index)];
 	item.block = block;
+	const combinedTags = std.mem.concat(main.worldArena.allocator, Tag, &.{blocks.parseBlock(stringId).tags(), item.tags}) catch unreachable;
```
IntegratedQuantum: I'm still not sure if we want this.
Also it's not needed now that sorting no longer uses tags

src/items.zig
```diff
@@ -1238,6 +1238,14 @@ pub const Item = union(ItemType) { // MARK: Item
 			}
 		}
 	}
+
+	pub fn getTags(self: Item) []const Tag {
```
IntegratedQuantum: not used

src/Inventory.zig
```diff
@@ -538,6 +539,75 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+
+			if (itemA == .null) return false;
+			if (itemB == .null) return true;
+			if ((itemA != .proceduralItem) and (itemB == .proceduralItem)) return false;
+			if ((itemA == .proceduralItem) and (itemB != .proceduralItem)) return true;
+
+			return std.mem.lessThan(u8, itemA.id().?, itemB.id().?);
+		}
+
+		pub fn swap(ctx: *@This(), a: usize, b: usize) void {
```
IntegratedQuantum: unused

src/Inventory.zig
```diff
@@ -538,6 +539,75 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+		ignoredSlotCount: usize,
+	};
+
+	pub fn sortItems(source: ClientInventory, options: SortOptions) void {
+		compressItems(source, options);
+		const InventorySize: usize = source.super.size() - options.ignoredSlotCount;
+		var SortList = main.ListManaged(usize).init(main.stackAllocator);
+		var intermediaryList = main.ListManaged(usize).init(main.stackAllocator);
+		defer SortList.deinit();
+		defer intermediaryList.deinit();
```
IntegratedQuantum, line 551: defer should always be directly after the line that does the init

src/Inventory.zig
```diff
@@ -538,6 +539,75 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
 	}
 
+	pub const SortOptions = struct {
+		ignoredSlotCount: usize,
+	};
+
+	pub fn sortItems(source: ClientInventory, options: SortOptions) void {
+		compressItems(source, options);
+		const InventorySize: usize = source.super.size() - options.ignoredSlotCount;
+		var SortList = main.ListManaged(usize).init(main.stackAllocator);
```
```suggestion
		var sortList = main.ListManaged(usize).init(main.stackAllocator);
```

IntegratedQuantum (changes requested):

IntegratedQuantum:
I don't see any changes since my last review. Did you forget to push?

src/Inventory.zig
Crepestrom: renamed

src/Inventory.zig
Crepestrom: moved

src/items.zig
Crepestrom: removed

src/assets.zig
Crepestrom: removed

src/Inventory.zig
Crepestrom: removed

Crepestrom:
testing finished

src/gui/components/Button.zig
```diff
@@ -99,7 +101,9 @@ pub fn initText(pos: Vec2f, width: f32, text: []const u8, options: Options) *But
 		.onAction = options.onAction,
 		.child = label.toComponent(),
 		.disabled = options.disabled,
+		.hideBackground = options.hideBackground,
 	};
+	if (self.hideBackground) self.child.mutSize().* = self.size;
```
IntegratedQuantum: Instead of changing it afterwards, please set it to the right size to begin with.

src/assets.zig
```diff
@@ -14,6 +14,7 @@ const NeverFailingAllocator = main.heap.NeverFailingAllocator;
 const NeverFailingArenaAllocator = main.heap.NeverFailingArenaAllocator;
 const List = main.List;
 const files = main.files;
+const Tag = main.Tag;
```
IntegratedQuantum: unintended change

src/Inventory.zig
```diff
@@ -7,6 +7,7 @@ const Item = main.items.Item;
 const ItemStack = main.items.ItemStack;
 const ProceduralItem = main.items.ProceduralItem;
 const utils = main.utils;
+const Tag = main.Tag;
```
IntegratedQuantum: unintended change

src/Inventory.zig
```diff
@@ -538,6 +539,71 @@ pub const ClientInventory = struct { // MARK: ClientInventory
 		main.sync.client.executeCommand(.{.craftProceduralItem = .init(destinations, workbenchInv)});
 	}
 
+	pub const SortOptions = struct {
```
IntegratedQuantum: Can be private

IntegratedQuantum (changes requested):
Full stacks should always be placed left of partial stacks of the same type, currently it behaves like this instead:
Image

src/Inventory.zig
Crepestrom: gone

src/assets.zig
Crepestrom: gone

src/gui/components/Button.zig
Crepestrom: made correct size

src/Inventory.zig
Crepestrom: privated

Crepestrom:
funny ennough that happens because items were not being sorted by their ammounts

> Full stacks should always be placed left of partial stacks of the same type, currently it behaves like this instead: Image

src/Inventory.zig
```diff
@@ -538,6 +538,82 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+				if (itemADurabilityPercent > itemBDurabilityPercent) return true;
+				if (itemADurabilityPercent < itemBDurabilityPercent) return false;
+				const itemADps: f32 = itemA.proceduralItem.getProperty(.damage)*itemA.proceduralItem.getProperty(.swingSpeed);
+				const itemBDps: f32 = itemB.proceduralItem.getProperty(.damage)*itemB.proceduralItem.getProperty(.swingSpeed);
+				if (itemADps < itemBDps) return true;
+				if (itemADps > itemBDps) return false;
+			}
+			if ((ctx.inv.getAmount(a) > ctx.inv.getAmount(b)) and std.mem.eql(u8, itemA.id().?, itemB.id().?)) return true;
+
+			return std.mem.lessThan(u8, itemA.id().?, itemB.id().?);
```
IntegratedQuantum: They should still sort by id first (you did accound for that with items, but not tools).

src/gui/components/Button.zig
```diff
@@ -112,7 +116,9 @@ pub fn initIcon(pos: Vec2f, iconSize: Vec2f, iconTexture: Texture, options: Opti
 		.onAction = options.onAction,
 		.child = icon.toComponent(),
 		.disabled = options.disabled,
+		.hideBackground = options.hideBackground,
 	};
+	if (self.hideBackground) self.child.mutSize().* = self.size;
```
IntegratedQuantum: Same here https://github.com/PixelGuys/Cubyz/pull/3194#discussion_r3740401652

src/Inventory.zig
```diff
@@ -538,6 +538,82 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			const itemA: Item = ctx.inv.getItem(a);
+			const itemB: Item = ctx.inv.getItem(b);
+
+			if (itemA == .null) return false;
+			if (itemB == .null) return true;
+			if ((itemA != .proceduralItem) and (itemB == .proceduralItem)) return false;
+			if ((itemA == .proceduralItem) and (itemB != .proceduralItem)) return true;
+			if ((itemA == .proceduralItem) and (itemB == .proceduralItem)) {
+				const itemADurabilityPercent: f32 = @as(f32, @floatFromInt(itemA.proceduralItem.durability))/itemA.proceduralItem.getProperty(.maxDurability);
+				const itemBDurabilityPercent: f32 = @as(f32, @floatFromInt(itemB.proceduralItem.durability))/itemB.proceduralItem.getProperty(.maxDurability);
```
IntegratedQuantum, line 607: I'm not sure if this is reliable.
Does the floating point spec guarantee that x/x is always exactly 1?

IntegratedQuantum (changes requested):

src/Inventory.zig
Crepestrom: does it matter?
at the point that happens the durabilty will be close enough anyway

IntegratedQuantum: If it doesn't matter, then why do you have a separate sort criterion afterwards?

Also I think it does matter at tools with full durability (which is why I'm asking whether they are guaranteed to be exactly 1).

src/gui/components/Button.zig
Crepestrom: oops forgot to remove that line

src/Inventory.zig
Crepestrom: i think so
they are IEEE compliant

Crepestrom: unless we are using @setFloatMode(.Optimized)

src/Inventory.zig
Crepestrom: sorts by id first now

src/Inventory.zig
```diff
@@ -538,6 +538,87 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			const itemB: Item = ctx.inv.getItem(b);
+			// Sorts between ProceduralItems and Baseitems
+			// Then by Id
+			// then by the items properties (eg; durability, ammount, dps)
+			if (itemA == .null) return false;
+			if (itemB == .null) return true;
+			if ((itemA != .proceduralItem) and (itemB == .proceduralItem)) return false;
+			if ((itemA == .proceduralItem) and (itemB != .proceduralItem)) return true;
+			if (std.mem.lessThan(u8, itemA.id().?, itemB.id().?)) return true;
+			if (!std.mem.lessThan(u8, itemA.id().?, itemB.id().?) and !std.mem.eql(u8, itemA.id().?, itemB.id().?)) return false;
```
Wunka: on the second line the first argument is already proven to be false (because of the first line). So either do:

```suggestion
			if (std.mem.lessThan(u8, itemA.id().?, itemB.id().?)) return true;
			if (!std.mem.eql(u8, itemA.id().?, itemB.id().?)) return false;
```
or

```suggestion
			if (!std.mem.eql(u8, itemA.id().?, itemB.id().?)) return std.mem.lessThan(u8, itemA.id().?, itemB.id().?):
```

src/Inventory.zig
```diff
@@ -538,6 +538,87 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			if ((itemA == .proceduralItem) and (itemB == .proceduralItem)) {
+				if (std.mem.lessThan(u8, itemA.id().?, itemB.id().?)) return true;
+				if (!std.mem.lessThan(u8, itemA.id().?, itemB.id().?) and !std.mem.eql(u8, itemA.id().?, itemB.id().?)) return false;
```
Wunka: this is already checked a few lines above

Wunka (changes requested):

src/Inventory.zig
Crepestrom: oops fixed

src/Inventory.zig
Crepestrom: applied suggesion

Crepestrom:
thanks for your help wunka

src/gui/components/Button.zig
```diff
@@ -88,30 +89,35 @@ pub fn globalDeinit() void {
...
 		.disabled = options.disabled,
+		.hideBackground = options.hideBackground,
 	};
 	return self;
 }
 
-pub fn initIcon(pos: Vec2f, iconSize: Vec2f, iconTexture: Texture, options: Options) *Button {
-	const icon = Icon.init(undefined, iconSize, iconTexture);
+pub fn initIcon(pos: Vec2f, givenIconSize: Vec2f, iconTexture: Texture, options: Options) *Button {
+	const calculatedIconSize: Vec2f = if (options.hideBackground) givenIconSize + @as(Vec2f, @splat(3*border)) else givenIconSize;
```
IntegratedQuantum, line 111: This looks wrong, if there is no background, then it should not add any border around the supplied size.

src/Inventory.zig
```diff
@@ -538,6 +538,85 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			const itemB: Item = ctx.inv.getItem(b);
+			// Sorts between ProceduralItems and Baseitems
+			// Then by Id
+			// then by the items properties (eg; durability, ammount, dps)
+			if (itemA == .null) return false;
+			if (itemB == .null) return true;
+			if ((itemA != .proceduralItem) and (itemB == .proceduralItem)) return false;
+			if ((itemA == .proceduralItem) and (itemB != .proceduralItem)) return true;
+			if (std.mem.lessThan(u8, itemA.id().?, itemB.id().?)) return true;
+			if (!std.mem.eql(u8, itemA.id().?, itemB.id().?)) return false;
```
IntegratedQuantum: I'd prefer it like this:
```suggestion
			if (!std.mem.eql(u8, itemA.id().?, itemB.id().?)) return std.mem.lessThan(u8, itemA.id().?, itemB.id().?);
```

src/Inventory.zig
```diff
@@ -538,6 +538,85 @@ pub const ClientInventory = struct { // MARK: ClientInventory
...
+			if ((itemA == .proceduralItem) and (itemB == .proceduralItem)) {
+				const itemADurabilityPercent: f32 = @as(f32, @floatFromInt(itemA.proceduralItem.durability))/itemA.proceduralItem.getProperty(.maxDurability);
+				const itemBDurabilityPercent: f32 = @as(f32, @floatFromInt(itemB.proceduralItem.durability))/itemB.proceduralItem.getProperty(.maxDurability);
```
IntegratedQuantum, line 606: I think durability should be the last property to consider (→after dps), otherwise items with the identical recipe will not be sorted next to each other.

IntegratedQuantum (changes requested):

src/gui/components/Button.zig
Crepestrom: nope getting rid of the border just makes the button smaller or the icon smaller than it needs to be

src/Inventory.zig
Crepestrom: ok done

PR #3203, opened 2026-06-10
Disabled buttons

### LINKED ISSUE #2556: Social options while in-game could use clarification

Rather than omit the name change option entirely, it should instead have a note on the gui telling the user they have to be on the main menu to change their username.

IntegratedQuantum:
With a reference use-case (showing the social settings that cannot be changed when in the world, previously they were hidden, now they are shown as disabled):
Screenshot at 2026-06-10 08-51-22

To achieve this I made a small change to the button interface to allow for optional arguments

replaces #2008

fixes #2556


tillpp:
Please rework the design of the grayed out button. (and honestly it's terrible in its current state)

OneAvargeCoder193:
I think it would make more sense if disabled buttons always drew the pressed texture

OneAvargeCoder193:
image

Wunka:
> image
> 

I don't like this. This looks more like a bug then a disabled button feature

Argmaster:
Full gray scale is the answer you are looking for.

Crepestrom:
if you look at my recolored slots pr you can use that to make darker slots

IntegratedQuantum:
@careeoki can you help me with the texture once you are back from your trip?

careeoki:
> @careeoki can you help me with the texture once you are back from your trip?

I was home monday so I can help right now: https://github.com/IntegratedQuantum/Cubyz/pull/1

IntegratedQuantum:
> I think a good change to make it feel even more disabled would be to make the text a bit darker:

Sadly that is not possible currently. I could make it transparent, but it looks kind of odd:
Screenshot at 2026-06-11 21-41-43

IntegratedQuantum:
Alright here we go:
Screenshot at 2026-06-13 10-38-20

IntegratedQuantum:
And it works on icons too
Screenshot at 2026-06-13 10-41-20

PR #3207, opened 2026-06-11
Voidstone overhaul

careeoki:
The current voidstone set is sad looking, and it is hard to use in builds since it is so dark and low contrast.
With these new textures I gave them brighter highlights and a saturated deep blue color. It's also much smoother, almost crystalline. 
I think this all helps voidstone feel more like **void**stone, not just a very dark regular stone. With how hard this block is to obtain, it should feel special and give you cool blocks to build with.
image
image
image
also the old color palette was awful and i hated working with it when adding new stone building blocks


Argmaster:
Honestly, I see no problem with having both. As of now there is very little blocks which could be used for shading builds, like you would do in most fancy builds in certain other voxel game.

image

careeoki:
> Honestly, I see no problem with having both

The problem is that they would be too similar in color and in design. And I really do not like the old one

> As of now there is very little blocks which could be used for shading builds

Yes, this will be improved over time. My main focus right now is covering the large gaps in color we are missing, rather than adding blocks of similar color to what we already have.
I would rather add a new stone set that is in between basalt and voidstone in terms of darkness instead of adding another near-black set that is similar to what we already have.

IntegratedQuantum (approved):
I like the new textures. The old ones were hard to build with in my opinion, since it didn't seem to fit with other blocks.

One thing that might be problematic is how similar it looks to obsidian though. Maybe it would make sense to also give it a small refresh.

Issue #3202, opened 2026-06-10
Proposal: remove the "replacement" syntax from zon parsing


H41ogen:
It's super easy to write and honestly more readable for methods like `.get` to return an optional type, and for the replacement to be done as `orelse <replacement>` directly after. additionally, there are tons of instances in the codebase of `.get(?T, something, null)` and siblings. This would also match the behavior of std-lib things like hashmaps, whose getter functions return `null` when nothing is found.

Wunka:
I never thought about that, but I agree. 
If we look at the `get` method this would be an easy change and I currently see no reason why we are even doing it that way
```diff
-pub fn get(self: *const ZonElement, comptime _type: type, key: []const u8, replacement: _type) _type {
-	if (self.* != .object) {
-		return replacement;
-	} else {
-		if (self.object.get(key)) |elem| {
-			return elem.as(_type, replacement);
-		} else {
-			return replacement;
-		}
-	}
-}
+pub fn get(self: *const ZonElement, comptime _type: type, key: []const u8) ?_type {
+	if (self.* != .object) {
+		return null;
+	}
+	if (self.object.get(key)) |elem| {
+		return elem.as(_type, replacement);
+	}
+	return null;
+}
```

H41ogen:
side note but `_type` is awful, use `T`. and also `innerType` in `.as` should be `InnerType`

Wunka:
That was a fast write up. Thats why I didn't rename them. Its just copied over from the current code

IntegratedQuantum:
We've talked about this many times, I guess just no one made an issue for it.
The pattern still comes from java which does not have optionals.

PR #3208, opened 2026-06-11
Remove replacement parameter from zon.as

IntegratedQuantum:
I also added some more error handling in some places where it was missing and easy to add (this also uncovered a minor configuration error: The lamps didn't specify a model)
first step towards #3202


H41ogen:
the replacement syntax is also used in zon.get, as well as another place i can't think of off the top of my head

IntegratedQuantum:
Yes I know, but that's why I'm not closing the issue with this PR.

PR #3210, opened 2026-06-11
Only deinit the player entity if there is one

### LINKED ISSUE #3179: Entity component data disappears for the host sometimes
4 weeks ago in the play session I observed that my bag was gone after a crash, I first suspected saving, but
In today's play session I noticed that I cannot seem to interact with my bag at all, the game saves no component data, other players cannot see my model.

It seems the server has completely forgotten all the components that my player entity has.

@tillpp could you try looking into this again?

IntegratedQuantum:
Should fix #3179 (after #3193 it still would have been an assertion failure)

### DIFF
--- a/src/server/server.zig
+++ b/src/server/server.zig
@@ -186,7 +186,9 @@ pub const User = struct { // MARK: User
 
 		self.worldEditData.deinit();
 
-		self.player().deinit(.server);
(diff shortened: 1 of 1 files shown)


tillpp (approved):
i approve of this.

PR #3212, opened 2026-06-12
replace _type with T

H41ogen:
This is the more common name for type parameters and also follows the naming conventions


H41ogen:
after this i'll remove the replacement syntax from the rest of the codebase

IntegratedQuantum (approved):

PR #3213, opened 2026-06-12
Remove getAtIndex and getChildAtIndex from zon.zig

H41ogen:
`getAtIndex` was used in 2 places in item.zig, so those locations are a bit more clunky now. `getChildAtIndex` was used nowhere though.


Wunka:
This sadly has now merge conflicts

PR #3217, opened 2026-06-12
Remove replacement syntax from zon.get

### LINKED ISSUE #3202: Proposal: remove the "replacement" syntax from zon parsing

H41ogen:
this took so long

closes #3202


src/server/terrain/sdf_models/rectangular_cuboid.zig
```diff
@@ -21,8 +21,8 @@ const Instance = struct {
 
 pub fn initAndGetExtend(zon: ZonElement) sdf.SdfModel.InitResult {
 	const self = main.worldArena.create(@This());
-	self.minRadii = zon.get(Vec3f, "minSideLengths", @splat(32))/@as(Vec3f, @splat(2));
-	self.maxRadii = zon.get(Vec3f, "maxSideLengths", self.minRadii*@as(Vec3f, @splat(2)))/@as(Vec3f, @splat(2));
+	self.minRadii = zon.get(Vec3f, "minSideLengths") orelse @as(Vec3f, @splat(32))/@as(Vec3f, @splat(2));
+	self.maxRadii = zon.get(Vec3f, "maxSideLengths") orelse self.minRadii*@as(Vec3f, @splat(2))/@as(Vec3f, @splat(2));
```
IntegratedQuantum: You need parenthesis around the left side, orelse has lower precedence

src/server/terrain/simple_structures/SimpleTreeModel.zig
```diff
@@ -48,30 +48,30 @@ branched: bool,
...
-		.height0 = parameters.get(i32, "height", 6),
-		.deltaHeight = parameters.get(u31, "height_variation", 3),
-		.leafRadius = parameters.get(f32, "leafRadius", (1 + parameters.get(f32, "height", 6))/2),
-		.deltaLeafRadius = parameters.get(f32, "leafRadius_variation", parameters.get(f32, "height_variation", 3)/2),
-		.leafElongation = parameters.get(f32, "leafElongation", 1),
-		.deltaLeafElongation = parameters.get(f32, "deltaLeafElongation", 0),
-		.branched = parameters.get(bool, "branched", true),
+		.height0 = parameters.get(i32, "height") orelse 6,
+		.deltaHeight = parameters.get(u31, "height_variation") orelse 3,
+		.leafRadius = parameters.get(f32, "leafRadius") orelse (1 + (parameters.get(f32, "height") orelse 6))/2,
```
IntegratedQuantum: missing parenthesis

src/server/terrain/simple_structures/SimpleTreeModel.zig
```diff
@@ -48,30 +48,30 @@ branched: bool,
...
+		.deltaLeafRadius = parameters.get(f32, "leafRadius_variation") orelse (parameters.get(f32, "height_variation") orelse 3)/2,
```
IntegratedQuantum: missing parenthesis

src/blocks.zig
```diff
@@ -183,43 +183,43 @@ pub fn register(_: []const u8, id: []const u8, zon: ZonElement) u16 {
...
-	_bounciness[size] = zon.get(f32, "bounciness", 0.0);
-	_density[size] = zon.get(f32, "density", main.physics.airDensity);
-	_terminalVelocity[size] = zon.get(f32, "terminalVelocity", 90);
-	_mobility[size] = zon.get(f32, "mobility", 1.0);
-	_allowOres[size] = zon.get(bool, "allowOres", false);
+	_replaceable[size] = zon.get(bool, "replaceable") orelse false;
+	_transparent[size] = zon.get(bool, "transparent") orelse false;
+	_collide[size] = zon.get(bool, "collide") orelse true;
+	_alwaysViewThrough[size] = zon.get(bool, "alwaysViewThrough") orelse false;
+	_viewThrough[size] = zon.get(bool, "viewThrough") orelse false or _transparent[size] or _alwaysViewThrough[size];
```
IntegratedQuantum: missing parenthesis

src/zon.zig
```diff
@@ -51,18 +51,15 @@ pub const ZonElement = union(enum) { // MARK: Zon
...
-				if (@typeInfo(T) == .optional) {
-					return elem.as(@typeInfo(T).optional.child) orelse replacement;
-				}
-				return elem.as(T) orelse replacement;
-			} else {
-				return replacement;
+	pub fn get(self: *const ZonElement, comptime T: type, key: []const u8) ?T {
+		if (self.* != .object) return null;
+		if (self.object.get(key)) |elem| {
+			if (@typeInfo(T) == .optional) {
```
IntegratedQuantum: Should no longer be needed, please remove this case

IntegratedQuantum (changes requested):
found some mistakes

src/server/terrain/simple_structures/SimpleTreeModel.zig
H41ogen: i thought `/` had higher precedence than `orelse`. I'll add them nonetheless for clarity

src/zon.zig
H41ogen: Something is still needed for parsing optional fields. I added functionality to `zon.as` to deal with optionals in a way that allows an optional field manually set to null to be unwrapped once, but not twice, so the `orelse` syntax with `zon.get` still works as expected

H41ogen:
explanation for my use of double optionals and `orelse`:
```zig
const foo: ?u32 = null;
var bar: ??u32 = foo orelse null;
std.debug.assert(bar == null);
bar = foo;
std.debug.assert(bar != null);
std.debug.assert(bar.? == null);
```

H41ogen:
Would it also make sense to add a test for the optional type `.get`/`.as`?

src/zon.zig
IntegratedQuantum: We should not allow optional fields here, since the result is by design already optional. I do not want to see a double optional in the code.

src/server/terrain/simple_structures/SimpleTreeModel.zig
IntegratedQuantum: Nevermind, I thought the the /2 was meant to come after the `.get`, but looking at it again, that doesn't seem to be the case.

src/zon.zig
```diff
@@ -170,6 +164,10 @@ pub const ZonElement = union(enum) { // MARK: Zon
 					else => return null,
 				}
 			},
+			.optional => |info| {
```
IntegratedQuantum: Should be a compiler error.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):
Since you were inactive I decided to take over this PR before it gets totally demolished by conflicts.

PR #3218, opened 2026-06-12
Sword crafting

### LINKED ISSUE #3147: Weapon crafting
### Description

could someone implement and right the right slot weights to the procedural weapon crafting?

careeoki:
closes #3147

Adds procedural swords, but all you can do with them is attack some leaves. You can make a variety of shapes: (stats in this image are old)

image

I tried my best to balance it but it's kinda impossible without combat to test it with. Also I have a hard time getting the numbers how I'd like with the new sum and averaging stats. But as you would expect, light swords are fast with lower damage, heavy swords are slow with higher damage.

"shortsword" shape
image

"broadsword" shape
image

**Other things:**
Currently we have a damage parameter for tools. It's scaled for blocks, and the numbers would be way too high for a weapon.
Should we split "damage" into "block damage" and "attack damage?" Because I think you should be able to use tools to attack, for the sensible ones like axes and sickles, but also the silly ones like your pickaxe or shover.
And also vice versa, your sword should let you cut through foliage.

It would be nice if we could have #1068 so you would craft swords at an anvil instead of the workbench. But I don't know if it's a blocker.


DellieDelta:
If possible I'd keep damage the same for blocks and attacks, to me that makes sense. But that would mean adjusting the scaling for tools

IntegratedQuantum:
I just player around a bit, and overall it's pretty good. How it maps to the texture is a bit odd, but I think it is fine given the restrictions (we should consider a larger grid for swords in the future).

One thing that stuck out to me though is that you can get much lower swing time if you move stuff to the tip, e.g. compare the following two swords:
Screenshot at 2026-06-13 18-12-15
Screenshot at 2026-06-13 18-11-25

Another small suggestion: Currently it is not viable to place connections on the outer sides, since they are only one pixel wide. Maybe you could use the overlay to increase their size to make them viable connections?

Argmaster:
Can we make the slots responsible for the quillon optional so we can make swords that look like katanas?

careeoki:
> Can we make the slots responsible for the quillon optional so we can make swords that look like katanas?

Sure yeah
image

careeoki:
> One thing that stuck out to me though is that you can get much lower swing time if you move stuff to the tip, e.g. compare the following two swords:

I really tried to fix this but I can't figure it out. I don't know how to get the numbers to do what I want them to do. I would appreciate someone else's help

> Another small suggestion: Currently it is not viable to place connections on the outer sides, since they are only one pixel wide. Maybe you could use the overlay to increase their size to make them viable connections?

The reason I didn't do this is because it made these two shapes look identical when they are mapped to the texture:
And that just felt weird, so I made the outer slot pixels 1-wide. But if you want, I can change it.

image
image

IntegratedQuantum:
Hmm, I think that's not a big deal, as long as they look different when they contain different materials.

careeoki:
> Hmm, I think that's not a big deal, as long as they look different when they contain different materials.

Alright, I got that to work

image

IntegratedQuantum:
Can you add this one missing pixel?
Screenshot at 2026-06-13 22-01-33

careeoki:
> Can you add this one missing pixel?

Done, also fixed a few other edge cases

ZigguratV:
> If possible I'd keep damage the same for blocks and attacks, to me that makes sense. But that would mean adjusting the scaling for tools.

I think creatures/entities could start with some defence, making the tool and weapons use the same damage parameter but generally higher damage on weapons. Perhaps axes and hands and logs might need a rebalance somehow even in such a situation...

Crepestrom:
i think we could take a bit from groundeds combat
its sytem of blocks parrys and fighting just is quite fun

assets/cubyz/tools/sword.zig.zon
```diff
@@ -0,0 +1,110 @@
+.{
+	.tags = .{.cuttable},
```
IntegratedQuantum: This makes swords usable as a sickle replacement. I do not think we should do this.

assets/cubyz/tools/sword.zig.zon
```diff
@@ -0,0 +1,110 @@
...
+				0x0, 0.1, 0.1, 0x0, 0.1,
+			},
+			.factor = 0.8,
+			.method = .average,
+		},
+		.{
+			.source = .swingSpeed,
+			.destination = .swingSpeed,
+			.matrix = .{
+				0.1, 0.2, 0.1, 0x0, 0x0,
```
IntegratedQuantum: > I really tried to fix this but I can't figure it out. I don't know how to get the numbers to do what I want them to do. I would appreciate someone else's help

0.1 and 0.2 here mean that the front pieces have almost no impact on swingspeed.
whereas they have factors like 3.0 for damage. I'd suggest to use similar matrices for swingSpeed and damage.

IntegratedQuantum (changes requested):

IntegratedQuantum:
Also @tillpp please confirm or deny whether this is sufficient to close #3147

tillpp:
> Also @tillpp please confirm or deny whether this is sufficient to close https://github.com/PixelGuys/Cubyz/issues/3147

yeah i think so

assets/cubyz/tools/sword.zig.zon
careeoki: Oh, I guess I had the numbers backwards in my head. I was thinking lower number = less speed but obviously that's not it.

The example from before now works as you would expect:
amber at top: 1.4 sps
amber at hilt: 3.9 sps

assets/cubyz/tools/sword.zig.zon
careeoki: I do think it's fun to chop leaves and plants away with a sword, but it is kinda broken since its just straight up better than a sickle. Removed it

IntegratedQuantum (approved):

assets/cubyz/tools/sword.zig.zon
IntegratedQuantum: I do agree, but if we do it, I think it at least shouldn't drop anything.

careeoki: That would make sense I think.

PR #3219, opened 2026-06-12
Reload Part 1 | /server <stop/restart> command

### LINKED ISSUE #2293: Closing the game gracefully in headless mode results in a panic

```
thread 581016 panic: attempt to use null value
/home/boysanic/git/Cubyz/.zig-cache/o/4acb3828c29a2d123b580add4b5925f0/cimport.zig:6684:32: 0x150fbda in deinit (Cubyz)
    return glad_glDeleteBuffers.?(arg_444, arg_445);
                               ^
/home/boysanic/git/Cubyz/src/models.zig:810:17: 0x14d47ca in deinit (Cubyz)
 quadSSBO.deinit();
                ^
/home/boysanic/git/Cubyz/src/main.zig:550:21: 0x14ccdf1 in main (Cubyz)
 defer models.deinit();
                    ^
/home/boysanic/git/Cubyz/compiler/zig/lib/std/start.zig:672:22: 0x14cc7ad in main (Cubyz)
            root.main();
                     ^
???:?:?: 0x7fb43b227634 in ??? (libc.so.6)
Unwind information for `libc.so.6:0x7fb43b227634` was not available, trace may be incomplete

???:?:?: 0x7fb43b2276e8 in ??? (libc.so.6)
???:?:?: 0x1423b54 in ??? (???)
./debug_linux.sh: line 27: 581016 Aborted                    (core dumped) ./zig-out/bin/Cubyz
```

I suspect something isn't inited when we launch with headless but we deinit it even if we're in headless.

I did not catch this before because there wasn't a graceful exit from headless before. You'd just Ctrl+C to close it.

tillpp:
stops the server
restarts the server

TODO list:

- [x]  /server stop with head
- [x]  /server stop with head and then rejoining a world
- [x]  /server restart  work on headfull server
- [x]  /server stop     on headless (with a client connected)
- [x]  /server restart on headless (with a client connected)

btw. restart and reload are not the same. restart does kick all players except the host, reload doesn't kick any player 

fixes #2293


tillpp:
i have to say, that i am very unfamiliar with this part of the codebase.(anything with items)
So expect this PR to be bad. I am not sure either what i am doing.
```zig
main.threadPool.clear(); // <-- this prevents
//  leaked user, but should i really be doing this here?

main.items.clearRecipeCachedInventories(); // <-- this removed
// a memory Leak, but should i really be doing this here?
```

if you got better idea, improvements, feel free to review & comment
Good night

src/items.zig
```diff
@@ -1382,8 +1382,10 @@ pub fn registerRecipes(zon: ZonElement) void {
 }
 
 pub fn clearRecipeCachedInventories() void {
-	for (recipeList.items) |recipe| {
+	for (recipeList.items) |*recipe| {
 		main.globalAllocator.free(recipe.sourceItems);
```
tillpp: this is really confusing
recipe.sourceItems sometimes uses main.worldArena (like in "items.zig:parseRecipe") to be allocated but sometimes main.globalAllocator (like in items.zig:registerRecipes calling "recipes.zig:parseRecipe calling "recipes.zig:addRecipe")

also just noticed that item.zig:parseRecipe is never used

Wunka (changes requested):
Command code looks good. I don't know the rest of code so I cannot good review it.
But as restart doesn't work for me my requested change is to make it work

tillpp:
>  But as restart doesn't work for me my requested change is to make it work

when i did request a review from you, restart wasn't there yet. It also isn't finished yet.
I am more interested in the 

main.threadPool.clear();
main.items.clearRecipeCachedInventories();

1. I know that you too struggled with "Leaked Player", right?
Might the missing main.threadPool.clear() be the same cause for your problem?

2.main.threadPool.clear(); does also clear the tasks from the client. Is that a problem?
or is there a way to clear only the task of the server?


those questions are not only towards you, but also everyone else reviewing this PR

Wunka:
> those questions are not only towards you, but also everyone else reviewing this PR

Those questions can be answered but other people because I have no idea. I have never touched those parts of the codebase (yes I did something with recipes but not that part. It seems #1824 did that).And the pr that fixed the leaked player issue was this: #3019

IntegratedQuantum:
I find both of these odd. These are a sign that the server is not cleaned up properly in deinit (see #2293), I'd suggest to address #2293 first, then you should be able to just do deinit here as normally.

> clearRecipeCachedInventories

This function looks completely broken, it frees entries with the wrong allocator, I'll investigate this, but I think we can just remove it entirely.

tillpp:
ready for review now.
I am not exactly sure what I am doing myself, reordering the deinit/inits
but it makes the crashes go away.

src/server/command/server.zig
```diff
@@ -0,0 +1,36 @@
...
+const main = @import("main");
+const User = main.server.User;
+
+pub const description = "Stop the server.";
+pub const usage =
+	\\/server <stop/restart>
+;
+
+const Args = union(enum) {
+	@"/server <restart>": struct { restart: enum { stop, restart } },
```
Wunka:
```suggestion
	@"/server <action>": struct { action: enum { stop, restart } },
```
or some other naming

src/block_entity.zig
```diff
@@ -217,7 +217,8 @@ pub const BlockEntityTypes = struct { // MARK: BlockEntityTypes
 		pub fn init() void {
 			StorageServer.init();
 		}
-		pub fn deinit() void {
+		pub fn deinit(headless: bool) void {
+			_ = headless;
```
Wunka:
```suggestion
		pub fn deinit(_: bool) void {
```

src/items.zig
```diff
@@ -1130,8 +1130,8 @@ pub const ItemStack = struct { // MARK: ItemStack
 };
 
 pub const Recipe = struct { // MARK: Recipe
-	sourceItems: []BaseItemIndex,
-	sourceAmounts: []u16,
+	sourceItems: []BaseItemIndex = &.{},
+	sourceAmounts: []u16 = &.{},
```
Wunka: is this actually now doing something or is it just an old fix?

Wunka (changes requested):

Wunka:
Does this now fix: #2293?

src/items.zig
tillpp: prevents a crash, because somehow somewhere on a magic space these get deinited without being given any value

src/server/command/server.zig
tillpp: ah forgot to rename, good catch

src/block_entity.zig
tillpp: i prefer it to have name, so it is easier to see what this parameter is supposed to be for and a reminder

tillpp:
> Does this now fix: https://github.com/PixelGuys/Cubyz/issues/2293?

no idea how they closed the headless server with ctrl+c,  but assuming they do it the proper way, by setting the running variable to false, then there should be a graceful exit.

tillpp:
ok "/server restart" also works on headfull servers, now!
And you can even switch the world you play on
like ``/server restart world2`` switches to the world2

tillpp:
after talkng with quantum we decided that world switching is too dangerous + too little use case.
got removed

src/server/command/server.zig
```diff
@@ -0,0 +1,32 @@
+const std = @import("std");
+
+const main = @import("main");
+const User = main.server.User;
+
+pub const description = "Stop the server.";
+pub const usage =
+	\\/server <stop/restart> ?world
```
IntegratedQuantum:
```suggestion
	\\/server <stop/restart>
```

src/server/command/server.zig
```diff
@@ -0,0 +1,32 @@
...
+		source.sendMessage("#ff0000{s}", .{errorMessage.items});
+		return;
+	};
+	switch (result.@"/server <action>".action) {
+		.stop => {},
+		.restart => {
+			main.server.restart.store(true, .release);
+		},
+	}
+	main.server.running.store(false, .release);
```
IntegratedQuantum: Should call server.stop, which does this internally.

src/server/command/server.zig
```diff
@@ -0,0 +1,32 @@
...
+	defer errorMessage.deinit(main.stackAllocator);
+
+	const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
+		source.sendMessage("#ff0000{s}", .{errorMessage.items});
+		return;
+	};
+	switch (result.@"/server <action>".action) {
+		.stop => {},
+		.restart => {
+			main.server.restart.store(true, .release);
```
IntegratedQuantum: Doesn't need to be atomic. Also in my opinion this should be done inside `server.stop`

src/server/server.zig
```diff
@@ -581,6 +584,9 @@ fn init(name: []const u8, singlePlayerPort: ?u16) void { // MARK: init()
 }
 
 fn deinit() void {
+	main.threadPool.clear();
+	main.items.clearRecipeCachedInventories();
```
IntegratedQuantum: was just removed, please rebase

src/server/reload.zig
IntegratedQuantum, line 1: It seems weird to have a new file for what is basically just 5 lines of code. I would prefer to have this code local in server.zig. All of this could even be local to the function that does the restart loop.

src/server/server.zig
```diff
@@ -697,25 +705,34 @@ fn update() void { // MARK: update()
 pub fn startFromNewThread(name: []const u8, port: ?u16) void {
 	main.initThreadLocals();
 	defer main.deinitThreadLocals();
-	startFromExistingThread(name, port);
+	const _name = main.globalAllocator.dupe(u8, name);
+	defer main.globalAllocator.free(_name);
```
IntegratedQuantum: Why are you duping it, then duping it again inside `startFromExistingThread`?

src/server/server.zig
```diff
@@ -697,25 +705,34 @@ fn update() void { // MARK: update()
...
+	startFromExistingThread(_name, port);
+	main.heap.GarbageCollection.syncPoint();
```
IntegratedQuantum: Why is this syncpoint needed?

src/main.zig
```diff
@@ -476,22 +474,27 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
 	if (!headless) graphics.init();
 	defer if (!headless) graphics.deinit();
 
-	if (!headless) audio.init() catch std.log.err("Failed to initialize audio. Continuing the game without sounds.", .{});
-	defer if (!headless) audio.deinit();
-
 	utils.initDynamicIntArrayStorage();
 	defer utils.deinitDynamicIntArrayStorage();
 
+	defer heap.GarbageCollection.forceAllFreeItemsFromList();
```
IntegratedQuantum: This should already be done in `defer deinitThreadLocals`, why is this needed here?

src/main.zig
```diff
@@ -476,22 +474,27 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
...
+
+	threadPool = utils.ThreadPool.init(globalAllocator, settings.cpuThreads orelse @max(1, (std.Thread.getCpuCount() catch 4) -| 1));
+	defer threadPool.deinit();
+
+	if (!headless) audio.init() catch std.log.err("Failed to initialize audio. Continuing the game without sounds.", .{});
+	defer if (!headless) audio.deinit();
```
IntegratedQuantum, line 480: Could you explain why you moved audio and threadpool init?

src/models.zig
```diff
@@ -887,8 +887,8 @@ pub fn reset() void {
 	nameToIndex.put("none", Model.init(&.{})) catch unreachable;
 }
 
-pub fn deinit() void {
-	quadSSBO.deinit();
+pub fn deinit(headless: bool) void {
+	if (!headless) quadSSBO.deinit();
```
IntegratedQuantum: I would suggest to make `quadSSBO` optional instead.

src/block_entity.zig
```diff
@@ -349,12 +350,14 @@ pub const BlockEntityTypes = struct { // MARK: BlockEntityTypes
...
 			}
 		}
-		pub fn deinit() void {
+		pub fn deinit(headless: bool) void {
 			while (textureDeinitList.popOrNull()) |texture| {
 				texture.deinit();
 			}
 			textureDeinitList.deinit(main.globalAllocator);
-			pipeline.deinit();
+			if (!headless) {
```
IntegratedQuantum: Please use the same style as in init which uses `if (!main.settings.launchConfig.headlessServer) {`

src/server/server.zig
```diff
@@ -581,6 +584,9 @@ fn init(name: []const u8, singlePlayerPort: ?u16) void { // MARK: init()
 }
 
 fn deinit() void {
+	main.threadPool.clear();
```
IntegratedQuantum, line 589: This has a race condition. A client may start new tasks before `connectionManager.deinit`.

Also I think it would be best to use a slightly less invasive approach. I think you only want to remove tasks from players that are disconnected, so I propose to only clear the playerJobQueue and wait until all threads finish their current task.

src/server/server.zig
```diff
@@ -605,6 +611,8 @@ fn deinit() void {
 	main.items.Inventory.server.deinit();
 	main.entity.server.deinit();
 
+	main.entityModel.reset();
```
IntegratedQuantum: This belongs in `assets.reset`, you can thank till++ for putting it in the wrong location :​P

src/server/server.zig
```diff
@@ -697,25 +705,34 @@ fn update() void { // MARK: update()
...
+				main.io.sleep(newTime.durationTo(lastTime.addDuration(updateTime)), .awake) catch {};
+				lastTime = lastTime.addDuration(updateTime);
+			} else {
+				std.log.warn("The server is lagging behind by {d:.1} ms", .{@as(f32, @floatFromInt(newTime.nanoseconds -% lastTime.nanoseconds -% updateTime.nanoseconds))/1000000.0});
+				lastTime = newTime;
+			}
+			update();
 		}
-		update();
+		if (!main.settings.launchConfig.headlessServer) return;
```
IntegratedQuantum: I think this should be checked in the command and print a nice error saying that restart is only allowed on headless servers.

IntegratedQuantum (changes requested):

src/server/server.zig
tillpp: because otherwise crash

src/main.zig
tillpp: to prevent crash, utils.deinitDynamicIntArrayStorage was needed by some function on some thread

src/main.zig
tillpp: to prevent crash, if i remember correctly audio deinit accesses the threadPool so it still needs to be inited

src/server/reload.zig
tillpp: this file will keep growing.
My plan is to store all the stuff temporary into reload during server.deinit
then ask reload.zig if there are any stored values during server.init 

its nice to have the entire reload code at one place

src/server/server.zig
tillpp: no. restart is also allowed in headfull servers.

it just works in a different way:

in headless server this while just keeps looping
in headfull server this while loop exists, but gets restarten from clientMain^^

src/server/server.zig
tillpp: if i remeber correctlly there might have been a memory leak here otherwise

src/server/server.zig
tillpp: you mean assets.unloadAssets :P

src/server/server.zig
tillpp: removed

src/server/server.zig
tillpp: I know to little about the ThreadPool to know how i would do this

tillpp: i think i figure it out.

src/server/server.zig
```diff
@@ -538,6 +538,8 @@ var userConnectList: main.utils.ConcurrentQueue(*User) = undefined;
 pub var connectionManager: *ConnectionManager = undefined;
 
 pub var running: std.atomic.Value(bool) = .init(false);
+pub var restart: std.atomic.Value(bool) = .init(true);
```
IntegratedQuantum: This does not need to be atomic. This read before the server starts, and only set after the server is running.
After it is set the server only stops after running is set to false, and restart is only read afterwards.

src/server/server.zig
```diff
@@ -550,10 +552,20 @@ fn init(name: []const u8, singlePlayerPort: ?u16) void { // MARK: init()
 	userDeinitList = .init(main.globalAllocator, 16);
 	userConnectList = .init(main.globalAllocator, 16);
 	lastTime = main.timestamp();
-	connectionManager = ConnectionManager.init(main.settings.defaultPort, false) catch |err| {
-		std.log.err("Couldn't create socket: {s}", .{@errorName(err)});
-		@panic("Could not open Server.");
-	}; // TODO Configure the second argument in the server settings.
+
+	if (main.reload.connectionManager) |_connManager| {
```
IntegratedQuantum: If it's not reset inside the loop, then it should be initialized and freed outside the loop to avoid branching in here.
e.g. at the start of `startFromExistingThread` you have a call to a new `init`/`deinit` function pair that initializes all persistent state, which includes world name and connection manager.

src/server/server.zig
```diff
@@ -698,28 +723,38 @@ pub fn startFromNewThread(name: []const u8, port: ?u16) void {
...
+			}
+			update();
 		}
-		update();
+		if (!main.settings.launchConfig.headlessServer) return;
 	}
 }
 
-pub fn stop() void {
+pub fn stop(_restart: bool) void {
```
IntegratedQuantum: An `enum {stop, restart}` would make this more readable at the callsite

src/server/reload.zig
IntegratedQuantum, line 1: If you want to keep this file I'd suggest the following:
- init/deinit should init/deinit the relevant state that persists, there should be no switches in server.init/deinit
- init/deinit should be called in `server.startFromExistingThread` and not in `main`
- the file should be moved to the `server` folder

src/utils.zig
```diff
@@ -866,9 +866,13 @@ pub const ThreadPool = struct { // MARK: ThreadPool
 
 		self.allocator.free(self.currentTasks);
 		self.allocator.free(self.threads);
-		self.allocator.destroy(self);
 	}
+	pub fn reset(self: *ThreadPool) void {
+		const threadCount = self.threads.len;
 
+		self.deinit();
```
IntegratedQuantum: This is very invasive. I think we shouldn't restart all background threads, it's already bad enough that you are clearing it.

IntegratedQuantum (changes requested):
Could you explain your strategy regarding keeping the connection manager?
Why not just keep running but not accept any new connections?

src/server/server.zig
IntegratedQuantum: But syncPoint itself doesn't give any guarantees that resources are cleaned. You probably want `waitForFreeCompletion` which actually waits until all the garbage is freed.

src/main.zig
IntegratedQuantum: I don't think you need this, if you call `waitForFreeCompletion` at the right points

src/main.zig
IntegratedQuantum: Why did you move the ThreadPool though?

The ThreadPool should be one of the first things to init, so other components can use it to parallelize in the future to reduce loading times. And the ThreadPool should not depend on anything above it. I suspect the problem was that you still had tasks in the ThreadPool, that it wasn't properly cleared.

IntegratedQuantum:
`/server restart` shouldn't give you the main menu:
Screenshot at 2026-06-14 19-38-16

src/main.zig
tillpp: i reverted the threadpool, there are no crashes anymore

> I suspect the problem was that you still had tasks in the ThreadPool, that it wasn't properly cleared.

this was probably the reason

src/server/server.zig
tillpp: waitForFreeCompletion
make the game freeze
grafik
I have been waiting for a minute without a reaction yet

IntegratedQuantum: Can you figure out where the other threads are blocked?

src/main.zig
```diff
@@ -476,9 +479,6 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
 	if (!headless) graphics.init();
 	defer if (!headless) graphics.deinit();
 
-	if (!headless) audio.init() catch std.log.err("Failed to initialize audio. Continuing the game without sounds.", .{});
-	defer if (!headless) audio.deinit();
-
```
IntegratedQuantum, line 481: Please just keep it where it was

src/utils.zig
```diff
@@ -868,7 +868,6 @@ pub const ThreadPool = struct { // MARK: ThreadPool
 		self.allocator.free(self.threads);
 		self.allocator.destroy(self);
 	}
-
```
IntegratedQuantum, line 871: unintended change

src/network.zig
```diff
@@ -528,27 +528,48 @@ pub const ConnectionManager = struct { // MARK: ConnectionManager
...
 
-		self.running.store(false, .monotonic);
-		self.thread.join();
 		self.socket.deinit();
 		self.connections.deinit(main.globalAllocator);
 		for (self.requests.items) |request| {
 			request.requestNotifier.signal();
 		}
 		self.requests.deinit(main.globalAllocator);
+		self.requests = .empty;
```
IntegratedQuantum: I suspect you added this in the mids of implementing continue/pause. I don't think this is needed (and if it is, it would be hiding a bug)

IntegratedQuantum (changes requested):

src/server/server.zig
tillpp: > e.g. at the start of startFromExistingThread you have a call to a new init/deinit function pair that initializes all persistent state, which includes world name and connection manager.

but that would destroy the persistent states once startFromExistingThread returning

IntegratedQuantum: Then we should find a way to handle the client side without returning here (and we talked about this yesterday, we need this for reloading either way).
So for the sake of keeping this PR simple, I'd suggest to drop the headfull restart functionality for now.

tillpp: done and tested

src/server/server.zig
```diff
@@ -538,6 +538,8 @@ var userConnectList: main.utils.ConcurrentQueue(*User) = undefined;
 pub var connectionManager: *ConnectionManager = undefined;
 
 pub var running: std.atomic.Value(bool) = .init(false);
+pub var restart: bool = true;
```
IntegratedQuantum: Can be private now I think

src/server/server.zig
```diff
@@ -702,25 +708,49 @@ pub fn startFromNewThread(name: []const u8, port: ?u16) void {
...
+			update();
 		}
-		update();
+		if (!main.settings.launchConfig.headlessServer) return;
 	}
 }
 
-pub fn stop() void {
-	running.store(false, .monotonic);
+pub const stopType = enum { stop, restart };
```
```suggestion
pub const StopType = enum { stop, restart };
```

src/network.zig
```diff
@@ -506,6 +506,7 @@ pub const ConnectionManager = struct { // MARK: ConnectionManager
 		errdefer main.globalAllocator.destroy(result);
 		result.* = .{};
 		result.packetSendRequests = .initContext({});
+		result.running.store(true, .monotonic);
```
IntegratedQuantum: The value was already initialized to true.

src/network.zig
```diff
@@ -528,16 +529,24 @@ pub const ConnectionManager = struct { // MARK: ConnectionManager
 		}
 		return result;
 	}
-
+	pub fn @"continue"(result: *ConnectionManager) !void {
```
IntegratedQuantum, line 530: Since did this with deinit, how about also making `init` call `continue`?

IntegratedQuantum (changes requested):

src/network.zig
tillpp: Done.

IntegratedQuantum (approved):

IntegratedQuantum (changes requested):
Actually it crashes sporadically when I do `/server stop` in headfull
```
thread 60237 panic: index out of bounds: index 0, len 0
/home/mint/Downloads/tillpp/src/blocks.zig:700:44: 0x16c7822 in fogDensity (main.zig)
  return textureFogData.items[animationData[textureIndices[block.typ][0]].startFrame].fogDensity;
                                           ^
/home/mint/Downloads/tillpp/src/renderer.zig:462:32: 0x16c6d1e in render (main.zig)
  extractImageDataAndDownsample(playerBlock, playerPos, viewMatrix);
                               ^
/home/mint/Downloads/tillpp/src/renderer.zig:296:15: 0x16b61d3 in renderWorld (main.zig)
  Bloom.render(lastWidth, lastHeight, playerBlock, playerPos, game.camera.viewMatrix);
              ^
/home/mint/Downloads/tillpp/src/renderer.zig:161:13: 0x16b40b6 in render (main.zig)
 renderWorld(game.world.?, ambient, game.fog.skyColor, playerPosition);
            ^
/home/mint/Downloads/tillpp/src/main.zig:620:20: 0x1771b6c in clientMain (main.zig)
    renderer.render(game.Player.getEyePosBlocking(), deltaTime);
                   ^
/home/mint/Downloads/tillpp/src/main.zig:531:13: 0x15b11a6 in main (main.zig)
  clientMain();
            ^
/home/mint/Cubyz/compiler/zig/lib/std/start.zig:699:88: 0x15b08ae in callMain (std.zig)
                                                                                       ^
../sysdeps/nptl/libc_start_call_main.h:58:16: 0x78df6fe29d8f in __libc_start_call_main (../sysdeps/x86/libc-start.c)
../csu/libc-start.c:392:3: 0x78df6fe29e3f in __libc_start_main_impl (../sysdeps/x86/libc-start.c)
???:?:?: 0x1e2c004 in ??? (???)
```

tillpp:
wait found a bug in headless, don't merge

tillpp:
done and tested !

IntegratedQuantum:
still crashes (43690 is `undefined` by the way)
```
thread 22734 panic: index out of bounds: index 43690, len 4080
/home/mint/Downloads/tillpp/src/models.zig:89:22: 0x16d847b in quadInfo (main.zig)
  return &quads.items[@intFromEnum(self)];
                     ^
/home/mint/Downloads/tillpp/src/models.zig:625:34: 0x1cd7c22 in getRawFaces (main.zig)
    var quad = quadIndex.quadInfo().*;
                                 ^
/home/mint/Downloads/tillpp/src/rotation.zig:68:25: 0x1cd9950 in rayModelIntersection (main.zig)
   modelData.getRawFaces(&quadList);
                        ^
/home/mint/Downloads/tillpp/src/rotation.zig:60:31: 0x1cd96bf in rayIntersection (main.zig)
   return rayModelIntersection(blocks.meshes.model(block), relativePlayerPos, playerDir);
                              ^
/home/mint/Downloads/tillpp/mods/cubyz/rotations/stairs.zig:318:54: 0x1cf478f in rayIntersection (rotations.zig)
 return RotationMode.DefaultFunctions.rayIntersection(block, item, relativePlayerPos, playerDir);
                                                     ^
/home/mint/Downloads/tillpp/src/renderer.zig:949:37: 0x16fbfeb in select (main.zig)
    if (block.mode().rayIntersection(block, item, relativePlayerPos, _dir)) |intersection| {
                                    ^
/home/mint/Downloads/tillpp/src/renderer.zig:229:22: 0x16b5588 in renderWorld (main.zig)
 MeshSelection.select(playerPos, direction, game.Player.inventory.getItem(game.Player.selectedSlot));
                     ^
/home/mint/Downloads/tillpp/src/renderer.zig:161:13: 0x16b40b6 in render (main.zig)
 renderWorld(game.world.?, ambient, game.fog.skyColor, playerPosition);
            ^
/home/mint/Downloads/tillpp/src/main.zig:626:20: 0x1771b6c in clientMain (main.zig)
    renderer.render(game.Player.getEyePosBlocking(), deltaTime);
                   ^
/home/mint/Downloads/tillpp/src/main.zig:531:13: 0x15b11a6 in main (main.zig)
  clientMain();
            ^
/home/mint/Cubyz/compiler/zig/lib/std/start.zig:699:88: 0x15b08ae in callMain (std.zig)
                                                                                       ^
```

How about only unloading the assets in headless mode, and giving that responsibility to the client in headfull?

tillpp:
~~does it crash in stop or restart or both.
please be more descriptive in the future when posting a crash~~
(right, i remember, we removed restart in headfull )

tillpp:
can you try again? It should work now
if it doesn't we can try this next:

> How about only unloading the assets in headless mode, and giving that responsibility to the client in headfull?

IntegratedQuantum:
Yes, it does work now, but could you at least consider my suggestion, stopping rendering is in my opinion not a particularly good solution.

tillpp:
> stopping rendering is in my opinion not a particularly good solution.

funny you say that, since the idea to freeze the rendering comes from your clientside() function code :D

> but could you at least consider my suggestion, 

I did consider it, but it is really paintful and complex since everything is so interconnected and **i can't test the crash on my machine. So i can only GUESS what the cause of the crash is** _<< important point i made bold_
using my debugger i could figure out, that on my machine :
1.the client disconnects first, 
2. leading to a world .deinit() on the client first, 
3. then doing the .deinit() on the server.

both client.deinit and server.deinit do call destroyWorldArena, but only at the last call it does actually unload the assets (i.e. destroy the worldArena). So it is always on the server, since it calls the destroyWorldArena last. 

But it looks like for some reason on your machine the disconnect packet takes more time (since game.world!=null) , so i guess on your PC it is:
0. disconnect packet is send but not received yet
1. the server calls deinits first
<crash?>
2. disconnect packet gets received by the client? 
3. then doing the .deinit() on the client?

but on the other hand worldArena is only called one time at the time of your crash, so it shouldn't deinit the world assets yet ?????? very confusing

can you test my hypothesis by adding std.debug.log("server deinit"); in serverworld.deinit and std.debug.log("client deinit"); in world.deinit? and using the commit it crashed at: https://github.com/PixelGuys/Cubyz/pull/3219/commits/2ecf5271d70a50ffefb4199107aca99f03a4883e

IntegratedQuantum:
Yes, server deinit comes first.

You could try to delay the network packets by adding a small sleep where it receives the disconnect packet in Connection.tryReceive

tillpp:
> ding a small sleep where it receives the disconnect packet in Connection.tryReceive

tbh I don't think this is a better solution then what i came up with.

IntegratedQuantum:
I meant for reproducing this issue

tillpp:
i see, i hope this fixes it too.
i replaced the clientState variable with a reference Counter in Assets

src/assets.zig
```diff
@@ -771,6 +774,9 @@ pub fn loadWorldAssets(assetFolder: []const u8, blockPalette: *Palette, itemPale
 }
 
 pub fn unloadAssets() void { // MARK: unloadAssets()
+	const prevVal = refCount.fetchSub(1, .monotonic);
+	std.debug.assert(prevVal != 0);
+	if (prevVal != 1) return;
 	if (!loadedAssets) return;
 	loadedAssets = false;
```
IntegratedQuantum: Can this bool be removed then? It was already doing reference counting but without the count.

IntegratedQuantum (changes requested):

IntegratedQuantum:
Also it seems to work now.

src/assets.zig
tillpp: yes (done)

tillpp: (it was reference ing)

IntegratedQuantum (approved):
I also marked this as fixing #2293 which you forgot to add to the PR description.

PR #3220, opened 2026-06-13
Stacked rotation mode

Crepestrom:
its like hanging where you have to stack things on top of one another to get new textures

allows for making stalagmites and other stacks of ect.


Wunka:
Sorry for not putting this earlier on the board!

mods/cubyz/rotations/stacked.zig
```diff
@@ -0,0 +1,59 @@
...
+
+fn transform(_: *main.models.QuadInfo) void {}
+
+pub fn init() void {}
+pub fn deinit() void {}
+pub fn reset() void {}
+
+pub fn createBlockModel(_: Block, _: *u16, zon: ZonElement) ModelIndex {
+	const topModelIndex = main.models.getModelIndex(zon.get([]const u8, "top", "cubyz:cube"));
+	const bottomModelIndex = main.models.getModelIndex(zon.get([]const u8, "bottom", "cubyz:cube"));
```
Wunka: you are on a too old version. Please update to master the default parameter was removed in #3217

Wunka (changes requested):

IntegratedQuantum:
Can we somehow combine this with the hanging rotation mode? I'd rather not have 3 incompatible rotation modes (e.g. you couldn't turn a stacked or hanging rotation mode into one that does both) when we only need one that's configurable.

PR #3223, opened 2026-06-13
Change source for commands from User to Source

Wunka:
Changes the source for commands to Source instead of directly User, this will allow to add other types like Console. The commands based on if the require something specifc of the user.


IntegratedQuantum (changes requested):
Please wait with this until the other command rewrite is done.

Wunka:
@IntegratedQuantum can I start with this or should I wait until the rest of #3073 is completly done? (give Args argument insteaf of []const u8 for example)
Else this is ready for review

IntegratedQuantum:
No, I still think the other rewrite should be finished first.

Wunka:
@IntegratedQuantum just to get your opinion, not to already start review. 
Should I like I currently do force every command to use the new Source or should I allow either Source or *User? through comptime magic this could be done. (kinda like how the main function can expect either nothing or init / init.Minimal)

IntegratedQuantum:
I would suggest to force using Source right away, but make a separate PR that just introduces `Source` with only `*User` as a parameter. Then you can get most of the migration done quickly without causing any conflicts.

src/server/command.zig
```diff
@@ -53,11 +63,11 @@ pub fn deinit() void {
 	commands.deinit();
 }
 
-pub fn execute(msg: []const u8, source: *User) void {
+pub fn execute(msg: []const u8, source: Source) void {
 	const end = std.mem.indexOfScalar(u8, msg, ' ') orelse msg.len;
 	const command = msg[0..end];
 	if (commands.get(command)) |cmd| {
-		if (!main.entity.components.@"cubyz:permissions".server.hasPermission(source.id, cmd.permissionPath)) {
+		if (source == .user and !main.entity.components.@"cubyz:permissions".server.hasPermission(source.user.id, cmd.permissionPath)) {
```
IntegratedQuantum: Please make a helper function for this on the union, so we don't forget it when new Sources are added

src/server/command.zig
```diff
@@ -89,22 +99,30 @@ pub const Coordinate = union(enum) {
...
+	if (source != .user and (x == .relative or y == .relative or z == .relative)) {
+		source.sendMessage("Command was run without a user; unable to interpret relative coordinates.", .{});
+		return error.InvalidArg;
+	}
 	return .{
 		// TODO: Remove clamp after #310 is implemented
-		std.math.clamp(if (x == .relative) player.player().pos[0] + x.relative else x.absolute, -1e9, 1e9),
-		std.math.clamp(if (y == .relative) player.player().pos[1] + y.relative else y.absolute, -1e9, 1e9),
-		std.math.clamp(if (z == .relative) player.player().pos[2] + z.relative else z.absolute, -1e9, 1e9),
+		std.math.clamp(if (x == .relative and source == .user) source.user.player().pos[0] + x.relative else x.absolute, -1e9, 1e9),
```
IntegratedQuantum: You already checked that above
```suggestion
		std.math.clamp(if (x == .relative) source.user.player().pos[0] + x.relative else x.absolute, -1e9, 1e9),
```

src/server/command/tp.zig
```diff
@@ -21,7 +21,14 @@ pub const Args = union(enum) {
 	@"/tp <playerIndex>": struct { playerIndex: command.PlayerIndex },
 };
 
-pub fn execute(args: Args, source: *User) void {
+const ArgParser = main.argparse.Parser(Args, .{.commandName = "/tp"});
```
IntegratedQuantum: Mistake in rebasing?

src/server/command/entity/avatar.zig
```diff
@@ -15,7 +15,12 @@ pub const Args = union(enum) {
 	@"/avatar <entityModel>": struct { entityModel: command.EntityModel },
 };
 
-pub fn execute(args: Args, source: *User) void {
+pub fn execute(args: Args, _source: Source) void {
+	if (_source != .user) {
```
IntegratedQuantum: (not a requirement)
Maybe instead of checking for the source here, each command could get a new constant that mentions what sources are supported? That way it would be harder to forget checking things.

IntegratedQuantum (changes requested):

src/server/command.zig

src/server/command/entity/avatar.zig
Wunka: I don't know. there is also the problem that some commands may only be restricted on specifc paths. Maybe add an error catch wich prints a nice message 
So people can do:
`if(_source != .user) return error.InvalidSource` ?

If I did the constant I would like to do the inverse and only mention which are not supported.

IntegratedQuantum: Yeah, having an error set would be a good idea. For this PR, let's keep it as it is I guess.

src/server/command/entity/avatar.zig
```diff
@@ -15,7 +15,12 @@ pub const Args = union(enum) {
...
+		_source.sendMessage("Command cannot be run without a user", .{});
+		return;
+	}
+	const source = _source.user;
```
IntegratedQuantum: Please name it `user` to avoid the name conflict, I know it's a bit more work, but I think it would be better like that.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #3224, opened 2026-06-13
Color -> ANSI for chat in terminal

### LINKED ISSUE #2610: Strip formatting characters from log messages (sort of)
### Description

I feel like this could be useful if implemented correctly, and I'm willing to do it, but I want to get ideas and feedback before I start doing stuff.

For example, `Chat: #ff7700H#ffb800a#fff900l#c4ff00o#83ff00g#42ff00e#00ff00n_#ffff00 left` is a bit hard to read. Perhaps `Chat: #ff7700H#ffb800a#fff900l#c4ff00o#83ff00g#42ff00e#00ff00n_#ffff00 (Halogen) left` would be better.

Wunka:
Finnally printing ANSI inside the terminal

fixes #2610

### DIFF
--- a/src/log.zig
+++ b/src/log.zig
@@ -18,8 +18,14 @@ pub const Level = enum {
 	info,
 	/// Debug: messages only useful for debugging.
 	debug,
-	/// Chat message
+	/// server messages
+	server,
+	/// chat messages
 	chat,
+
+	fn isColorCoded(self: Level) bool {
+		return self == .chat or self == .server;
+	}
 };
 
 var logFile: ?std.Io.File = undefined;
@@ -58,7 +64,8 @@ noinline fn runtimeLogFn(level: Level, format: []const u8, args: []const fmt.For
 		.info => "",
 		.warn => "\x1b[33m",
 		.debug => "\x1b[37;44m",
-		.chat => "",
+		.server => "\x1b[34mserver\x1b[0m: ",
+		.chat => "\x1b[36mchat\x1b[0m: ",
 	};
 	const colorReset = "\x1b[0m\n";
 	const filePrefix = switch (level) {
@@ -66,6 +73,7 @@ noinline fn runtimeLogFn(level: Level, format: []const u8, args: []const fmt.For
 		.warn => "warning",
 		.info => "info",
 		.debug => "debug",
+		.server => "server",
 		.chat => "chat",
 	};
 	const fileSuffix = "\n";
@@ -136,8 +144,8 @@ fn logToStdErr(level: Level, comptime format: []const u8, args: anytype) void {
 	const allocator = fba.allocator();
 
 	const _string = std.fmt.allocPrint(allocator, format, args) catch format;
-	const string = if (level == .chat and supportsANSIColors) convertColorToANSI(_string) else _string;
-	defer if (level == .chat and supportsANSIColors) main.stackAllocator.free(string);
+	const string = if (level.isColorCoded() and supportsANSIColors) convertColorToANSI(_string) else _string;
+	defer if (level.isColorCoded() and supportsANSIColors) main.stackAllocator.free(string);
 
 	const writer = std.debug.lockStderr(&.{});
 	defer std.debug.unlockStderr();
@@ -207,6 +215,14 @@ pub fn convertColorToANSI(text: []const u8) []const u8 {
 		list.appendSlice(main.stackAllocator, testBuff[0..len]);
 	}
 	return list.toOwnedSlice(main.stackAllocator);
+}
+
+pub fn server(comptime format: []const u8, args: anytype) void {
+	var runtimeArgs: [args.len]fmt.FormatArg = undefined;
+	inline for (0..args.len) |i| {
+		runtimeArgs[i] = .fromAnytype(@TypeOf(args[i]), &args[i]);
+	}
+	runtimeLogFn(.server, format, &runtimeArgs);
 }
 
 pub fn chat(comptime format: []const u8, args: anytype) void {
--- a/src/log.zig
+++ b/src/log.zig
@@ -187,10 +187,8 @@ fn convertColorToANSI(allocator: main.heap.NeverFailingAllocator, text: []const 
 
 		if (fontEffect.color != currentFontEffect.color) {
 			list.appendSlice(allocator, "38;2;");
-			var shift: u5 = 16;
-			while (true) : (shift -= 8) {
+			for ([3]u5{16, 8, 0}) |shift| {
 				list.print(allocator, "{d};", .{@as(u8, @truncate(fontEffect.color >> shift))});
-				if (shift == 0) break;
 			}
 		}
 		if (fontEffect.bold != currentFontEffect.bold) {
--- a/src/log.zig
+++ b/src/log.zig
@@ -148,7 +148,7 @@ fn logToStdErr(level: Level, comptime format: []const u8, args: anytype) void {
 	const allocator = fba.allocator();
 
 	const _string = std.fmt.allocPrint(allocator, format, args) catch format;
-	const string = if (level.isColorCoded() and supportsANSIColors) convertColorToANSI(_string) else _string;
+	const string = if (level.isColorCoded() and supportsANSIColors) convertColorToANSI(main.stackAllocator, _string) else _string;
 	defer if (level.isColorCoded() and supportsANSIColors) main.stackAllocator.free(string);
 
 	const writer = std.debug.lockStderr(&.{});
@@ -156,15 +156,15 @@ fn logToStdErr(level: Level, comptime format: []const u8, args: anytype) void {
 	nosuspend writer.file_writer.interface.writeAll(string) catch {};
 }
 
-fn convertColorToANSI(text: []const u8) []const u8 {
+fn convertColorToANSI(allocator: main.heap.NeverFailingAllocator, text: []const u8) []const u8 {
 	var list: List(u8) = .empty;
 
 	var parser = graphics.TextBuffer.Parser{
 		.unicodeIterator = std.unicode.Utf8Iterator{.bytes = text, .i = 0},
 		.currentFontEffect = .{},
-		.parsedText = .init(main.stackAllocator),
-		.fontEffects = .init(main.stackAllocator),
-		.characterIndex = .init(main.stackAllocator),
+		.parsedText = .init(allocator),
+		.fontEffects = .init(allocator),
+		.characterIndex = .init(allocator),
 		.showControlCharacters = false,
 	};
 	defer parser.fontEffects.deinit();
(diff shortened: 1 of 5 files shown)


Wunka:
I think this is ready. I don't know where to put he function. Before I tried to do it directly on log but the `stackallocator` is not always there when logging, thats why its currently in main.zig

src/main.zig
```diff
@@ -109,6 +109,99 @@ pub const std_options: std.Options = .{ // MARK: std_options
...
+	var bold: bool = false;
+	var underlined: bool = false;
+	var strikeThroughed: bool = false;
+	var lastColor: [3]u8 = @splat(0);
+	var unicodeIterator = std.unicode.Utf8Iterator{.bytes = text, .i = 0};
+
+	loop: while (true) {
+		var curSlice = unicodeIterator.nextCodepointSlice() orelse break :loop;
+		inner: switch (std.unicode.utf8Decode(curSlice) catch unreachable) {
+			'#' => {
```
IntegratedQuantum: Please reuse the parsing code we already have in graphics

src/server/server.zig
```diff
@@ -824,7 +824,9 @@ pub fn messageFrom(msg: []const u8, source: *User) void { // MARK: message
 fn sendRawMessage(msg: []const u8) void {
 	chatMutex.lock();
 	defer chatMutex.unlock();
-	std.log.info("Chat: {s}", .{msg}); // TODO use color \033[0;32m
+	const result = main.convertColorToANSI(msg);
+	defer main.stackAllocator.free(result);
+	std.log.info("Chat: {s}", .{result});
```
IntegratedQuantum: I think it's time we introduce a log level for this e.g. `main.log.chat()`.
The current solution wouldn't look good in the log file. You can only convert to ANSI when in the terminal.

IntegratedQuantum (changes requested):

src/main.zig
Wunka: I think I did what you meant

src/server/server.zig
Wunka: I don't quite know how to tackle this as we completly override std.log so even if I did a wrapper it would still write to file.

IntegratedQuantum: Add a function `main.log.chat` that does
```
			var runtimeArgs: [args.len]fmt.FormatArg = undefined;
			inline for (0..args.len) |i| {
				runtimeArgs[i] = .fromAnytype(@TypeOf(args[i]), &args[i]);
			}
			runtimeLogFn(.chat, format, &runtimeArgs); // Passing .chat
```
Change `main.log.runtimeLogFn` to take an enum with one more value than `std.log.Level`.
Make `logToStdErr` also take the enum and switch in there to convert the color codes if in `.chat` (and if the console allows it).

Wunka: done. I don't know how but this also fixes an issue where debug messages would set their background color instead of text color (maybe some reset??):

before:
image

after:
image

Wunka: I made it a bit better (at least in my mind):
image
This also fixes a wierd behavior in the logs as we then had Chat double
Before:
```
[chat]: Chat: [#db5757Wunka§#ffffff] 1
[chat]: User "#db5757Wunka@0" executed command "time 0"
[chat]: Chat: #db5757Wunka§#ffff00 left
```
Now:
```
[chat]: #db5757Wunka§#ffff00 joined
[server]: User "#db5757Wunka@0§#ffffff" executed command "help"
[chat]: [#db5757Wunka§#ffffff] Hello
[chat]: #db5757Wunka§#ffff00 left
```

src/log.zig
```diff
@@ -0,0 +1,234 @@
...
+	}
+};
+
+var logFile: ?std.Io.File = undefined;
+var logFileTs: ?std.Io.File = undefined;
+var supportsANSIColors: bool = undefined;
+var openingErrorWindow: bool = false;
+
+// overwrite the log function:
+pub const std_options: std.Options = .{ // MARK: std_options
```
IntegratedQuantum: The std_options need to be in the root file, otherwise they have no effect. Please move them back (but probably define the function here.

src/log.zig
```diff
@@ -0,0 +1,234 @@
...
+	const _string = std.fmt.allocPrint(allocator, format, args) catch format;
+	const string = if (level.isColorCoded() and supportsANSIColors) convertColorToANSI(_string) else _string;
+	defer if (level.isColorCoded() and supportsANSIColors) main.stackAllocator.free(string);
+
+	const writer = std.debug.lockStderr(&.{});
+	defer std.debug.unlockStderr();
+	nosuspend writer.file_writer.interface.writeAll(string) catch {};
+}
+
+pub fn convertColorToANSI(text: []const u8) []const u8 {
```
IntegratedQuantum: Should be private
```suggestion
fn convertColorToANSI(text: []const u8) []const u8 {
```

src/server/server.zig
IntegratedQuantum: > I don't know how but this also fixes an issue where debug messages would set their background color instead of text color (maybe some reset??):

It was intentional, and your "fix" threw out our custom log function entirely and reverted to Zig's implementation (which doesn't write to the log file).

IntegratedQuantum (changes requested):

src/log.zig
Wunka: ups. In a before test I had that. Instead of moving it completly back I did: `std_options = log.std_options`

IntegratedQuantum: I think it belongs to main, since it contains other options (currently not used by us) that have nothing to do with logging.

Wunka: :+1: moved back

src/main.zig
IntegratedQuantum: Please move this function to log.zig, so you don't have to expose the `runtimeLogFn` publicly.

src/main.zig
```diff
@@ -104,104 +102,11 @@ pub const std_options: std.Options = .{ // MARK: std_options
 			inline for (0..args.len) |i| {
 				runtimeArgs[i] = .fromAnytype(@TypeOf(args[i]), &args[i]);
 			}
-			runtimeLogFn(level, format, &runtimeArgs);
+			log.runtimeLogFn(@enumFromInt(@intFromEnum(level)), format, &runtimeArgs);
```
IntegratedQuantum: I don't think this is a good idea, please use a switch statement or a string.

src/log.zig
```diff
@@ -0,0 +1,215 @@
...
+	const _string = std.fmt.allocPrint(allocator, format, args) catch format;
+	const string = if (level.isColorCoded() and supportsANSIColors) convertColorToANSI(_string) else _string;
+	defer if (level.isColorCoded() and supportsANSIColors) main.stackAllocator.free(string);
+
+	const writer = std.debug.lockStderr(&.{});
+	defer std.debug.unlockStderr();
+	nosuspend writer.file_writer.interface.writeAll(string) catch {};
+}
+
+fn convertColorToANSI(text: []const u8) []const u8 {
```
IntegratedQuantum: Functions that return an allocated result to be freed by the caller should always take an allocator as a parameter.

src/log.zig
```diff
@@ -0,0 +1,215 @@
...
+		.fontEffects = .init(main.stackAllocator),
+		.characterIndex = .init(main.stackAllocator),
+		.showControlCharacters = false,
+	};
+	defer parser.fontEffects.deinit();
+	defer parser.parsedText.deinit();
+	defer parser.characterIndex.deinit();
+	parser.parse();
+
+	parser.currentFontEffect = .{};
```
IntegratedQuantum: Please make this a local variable instead of recycling internal state of the parser.

src/log.zig
```diff
@@ -0,0 +1,215 @@
...
+			list.appendSlice(main.stackAllocator, "\x1b[38;2");
+			var shift: u5 = 16;
+			while (true) : (shift -= 8) {
+				list.print(main.stackAllocator, ";{d}", .{@as(u8, @truncate(parser.fontEffects.items[i].color >> shift))});
+				if (shift == 0) break;
+			}
+			list.append(main.stackAllocator, 'm');
+		}
+		if (parser.fontEffects.items[i].bold != parser.currentFontEffect.bold) {
+			list.appendSlice(main.stackAllocator, "\x1b[");
```
IntegratedQuantum: I think ANSI supports multiple entries in one row.
To reduce the amount of output we have I'd suggest to combine the sequences. This should make the code simpler too, since you can just do a single `if (parser.fontEffects.items[i] == parser.currentFontEffect) continue;` at the start of the loop.

src/log.zig
```diff
@@ -0,0 +1,215 @@
...
+		if (parser.fontEffects.items[i].underline != parser.currentFontEffect.underline) {
+			list.appendSlice(main.stackAllocator, "\x1b[");
+			if (parser.currentFontEffect.underline) {
+				list.append(main.stackAllocator, '2');
+			}
+			list.appendSlice(main.stackAllocator, "4m");
+		}
+		parser.currentFontEffect = parser.fontEffects.items[i];
+		var testBuff: [3]u8 = undefined;
+		const len = std.unicode.utf8Encode(@truncate(parser.parsedText.items[i]), &testBuff) catch continue;
```
IntegratedQuantum: We can assert this
```suggestion
		const len = std.unicode.utf8Encode(@intCast(parser.parsedText.items[i]), &testBuff) catch continue;
```

IntegratedQuantum (changes requested):

src/main.zig
Wunka: all done

src/log.zig
```diff
@@ -0,0 +1,250 @@
...
+		.characterIndex = .init(allocator),
+		.showControlCharacters = false,
+	};
+	defer parser.fontEffects.deinit();
+	defer parser.parsedText.deinit();
+	defer parser.characterIndex.deinit();
+	parser.parse();
+
+	var currentFontEffect: graphics.TextBuffer.FontEffect = .{};
+	for (0..parser.parsedText.items.len) |i| {
```
IntegratedQuantum:
```suggestion
	for (parser.parsedText.items, parser.fontEffects.items) |unicodeChar, fontEffect| {
```

src/log.zig
```diff
@@ -0,0 +1,250 @@
...
+		// add actual text at the end
+		defer blk: {
+			var testBuff: [3]u8 = undefined;
+			const len = std.unicode.utf8Encode(@intCast(parser.parsedText.items[i]), &testBuff) catch break :blk;
```
IntegratedQuantum: I think the parser already guarantees that these are valid unicode.
```suggestion
			const len = std.unicode.utf8Encode(@intCast(parser.parsedText.items[i]), &testBuff) catch unreachable;
```

src/log.zig
```diff
@@ -0,0 +1,250 @@
...
+	defer parser.fontEffects.deinit();
+	defer parser.parsedText.deinit();
+	defer parser.characterIndex.deinit();
+	parser.parse();
+
+	var currentFontEffect: graphics.TextBuffer.FontEffect = .{};
+	for (0..parser.parsedText.items.len) |i| {
+		// add actual text at the end
+		defer blk: {
+			var testBuff: [3]u8 = undefined;
```
IntegratedQuantum: I think they can be up to 4 chars long
```suggestion
			var testBuff: [4]u8 = undefined;
```

src/log.zig
```diff
@@ -0,0 +1,250 @@
...
+			const len = std.unicode.utf8Encode(@intCast(parser.parsedText.items[i]), &testBuff) catch break :blk;
+			list.appendSlice(allocator, testBuff[0..len]);
+		}
+		if (parser.fontEffects.items[i] == currentFontEffect) continue;
+
+		list.appendSlice(allocator, "\x1b[");
+		defer list.append(allocator, 'm');
+
+		var addedEffect: bool = false;
```
IntegratedQuantum: To simplify this, how about adding the semicolon at the end and then removing the last one (asserting that it has one)?

src/log.zig
```diff
@@ -0,0 +1,250 @@
...
+				list.appendSlice(allocator, "22");
+			}
+			addedEffect = true;
+		}
+		if (parser.fontEffects.items[i].italic != currentFontEffect.italic) {
+			if (addedEffect) list.append(allocator, ';');
+			if (parser.currentFontEffect.italic) {
+				list.append(allocator, '2');
+			}
+			list.append(allocator, '3');
```
IntegratedQuantum: I don't like this style of relying on the previous char.
I'd suggest just two branches on with `"23;"` and one with `"3;"`

IntegratedQuantum (changes requested):

src/log.zig
Wunka: that looks better. done

src/log.zig
Wunka: done. The other suggestions are also applied

src/log.zig
```diff
@@ -0,0 +1,243 @@
...
+		// add actual text at the end
+		defer {
+			var testBuff: [4]u8 = undefined;
+			const len = std.unicode.utf8Encode(@intCast(unicodeChar), &testBuff) catch unreachable;
+			list.appendSlice(allocator, testBuff[0..len]);
+		}
+		if (fontEffect == currentFontEffect) continue;
+
+		list.appendSlice(allocator, "\x1b[");
+		defer list.items[list.items.len - 1] = 'm';
```
IntegratedQuantum: Please also assert that there is a semicolon at the end, this would also catch the addition of new font effects.

src/log.zig
```diff
@@ -0,0 +1,243 @@
...
+
+		if (fontEffect.color != currentFontEffect.color) {
+			list.appendSlice(allocator, "38;2;");
+			var shift: u5 = 16;
+			while (true) : (shift -= 8) {
+				list.print(allocator, "{d};", .{@as(u8, @truncate(fontEffect.color >> shift))});
+				if (shift == 0) break;
+			}
```
IntegratedQuantum: How about
```suggestion
			for ([3]u5{16, 8, 0}) |shift| {
				list.print(allocator, "{d};", .{@as(u8, @truncate(fontEffect.color >> shift))});
			}
```

IntegratedQuantum (changes requested):

src/log.zig
Wunka: That looks nice. I copied the other version from `graphics.zig`

src/log.zig
Wunka: done

IntegratedQuantum (approved):
Thanks, this looks so much nicer now!

I also linked the related issue which I guess you forgot to search for.

Issue #3192, opened 2026-06-08
The "seed phrase" system is confusing and annoying


du82:
### Steps to Reproduce and Observed Behavior

The seed phrase system is pretty annoying. I want to download and run the game without setting up or knowing what a private key / seed phrase is.

It doesn't even fix the problem of needing to exchange IP addresses, so it just seems to annoy users, some who don't even care about multiplayer.

du82:
For me it's not confusing, but for others I've played with, they don't understand why its useful or that the seed phrase is a password of sorts.

ikabod-kee:
I agree completely.

IntegratedQuantum:
Yes, we are aware of these issues, but how do you propose we fix them?
We do need to have some way of identifying a player in way such that no one else can impersonate them.

Crepestrom:
rename it to account id so players can intuit its function more

IntegratedQuantum:
It is already called Account Code, I don't see how Account id is any better.

ikabod-kee:
I think maybe it shouldn't be the first thing that shows up when the player boots up for the first time. I think there should be an "Account"/"Create Account" button in the menus, and have a pop-up when joining a multiplayer server that you must create an account first.

The reason for this is because not every player will want to play multiplayer.

IntegratedQuantum:
Yes that is true, but I fear that this will open a bunch of new edges with regards to your singleplayer worlds. We could make a "local" account that gets used when you play locally, but what happens if you share the world? When I was playing Terraria with a friend a couple years ago we regularly switch who hosts it, if there was such a thing as a local player, then the first host would have lost access to the local account with no obvious explanation as to why.
Sure we could automatically bind the local account to the player account when they first login, but when you move that world to a PC where you aren't logged in then you would again lose access to it.

ikabod-kee:
We could use the Minecraft method: the host always gets the items.

Also, if you don't have an account, you can't host.

IntegratedQuantum:
Then there is also another problem:
Lets say you played Cubyz for a while in singleplayer and now want to play with your some friends. Your friends are waiting for you: Would you take the time to properly store your Account code somewhere, or would you just store it quickly somewhere in a random text file and forget where and what it is by the end of the day?

ikabod-kee:
Honestly, I haven't stored my account code anywhere. I just let the program do it for me because it was an option. When I get logged out, I have no idea how to get my account back, so I just make another one.

This is a problem Quantum.

[...]

PR #3229, opened 2026-06-13
Improve the Account setup process

IntegratedQuantum:
The start screen is much simpler:
Screenshot at 2026-06-13 21-47-37
Account creation now starts with a small explanation, that is unskippable (the button is disabled for 8 seconds)
Screenshot at 2026-06-13 21-47-46
Then you have a set of options how you store it:
Screenshot at 2026-06-13 21-51-28
Password manager and write it down lead to a similar screen, again the button to continue is disabled, for 25 seconds this time, just to be sure that the player does something:
Screenshot at 2026-06-13 21-47-58
The save as file does not have a timer, but you have to enter a filename to continue.
It is also missing a file select dialog (you have to enter the path manually, which might be annoying, but I will make an issue for this):
Screenshot at 2026-06-13 21-49-07
Then you return to the login screen, after login there are two more windows for storing the Account Code:
Screenshot at 2026-06-13 21-49-51
Screenshot at 2026-06-13 21-49-59

progress towards #3192


IntegratedQuantum:
also the clipboard is no longer cleared after the screen that lets you copy your account code.

ikabod-kee:
I have a """few""" criticisms:
- [x] 1. The timer should be visible on timer buttons (Example: "`Return to Login (5)`")
- [x] 2. The timer should be a bit shorter on the "This is your Account Code:" screens or have an "Are you sure?" prompt.
- [x] 3. "Are you sure?" prompt if you press "cancel" during the "Encrypt it with a password (recommended)" screen
- [ ] 4. "Encrypt it with a password (recommended)" should be opt-in rather than opt-out (for the sake of the player's convenience, because who actually uses this feature? The most niche scenario for using this is if you share a computer with someone), and said screen should have a short timer.
- [x] 5. "Encrypt it with a password (recommended)" screen's "confirm" option should be disabled if no password is inserted.
- [x] 6. There should be a "back" button to view the last screen you were on, just in case you choose the wrong option on accident.
- [x] 7. The Login button should be disabled if there is no text in the "Please enter your Account Code:" screen instead of having an error prompt pop up.
- [x] 8. The player has no way to change the music volume while on this screen, which can be especially annoying for first-time players. You could have a prompt to let the player set this if it's their first time playing. #3192
- [x] 9. The "Save as file" option does not tell you where the file is stored or let you choose where to store the file. #3232
- [x] 10. There is an error if you do not enter a file name for the "Save as file" option.
- [ ] 11. The "Save as file" option saves your Account Code as a file without a file type.
- [x] 12. You should be able to import your Account Code if you saved it as a file. #3232
- [ ] 13. The "Password Manager" and "Write it down yourself" options are functionally the same.
- [x] 14. Seeing this screen on first boot-up is uninviting. It would be nice if it said "Welcome to Cubyz!" somewhere. #3192
- [x] 15. Emphasize in the Login screen that an account is needed for singleplayer worlds. #3192
- [ ] 16. "Please enter your local password!" screen's "unlock" button should be renamed to "`Login`"

ikabod-kee:
17. "Set your name" prompt should open every time when making a new account instead of just once for first-time players.

IntegratedQuantum:
> 3. "Are you sure?" prompt if you press "cancel" during the "Encrypt it with a password (recommended)" screen

I have changed this to a "Back" button, that seems better than adding a new dialog that explains to you that cancelling won't store it.

> 4. The most niche scenario for using this is if you share a computer with someone

What about the rather common "you have an untrustworthy application installed" scenario? On most operating systems any application can read almost any file.

> 6. There should be a "back" button to view the last screen you were on, just in case you choose the wrong option on accident.

I assume you meant this only for the Account Code screen, since that is the only case where you made a decision that mattered.

> .8. The player has no way to change the music volume while on this screen
> .14. Seeing this screen on first boot-up is uninviting. It would be nice if it said "Welcome to Cubyz!" somewhere.
> .15. Emphasize in the Login screen that an account is needed for singleplayer worlds.

will be fixed by not having this at the start of the game, as mentioned in https://github.com/PixelGuys/Cubyz/issues/3192#issuecomment-4652768962. I will do this in a separate PR.

> 9. The "Save as file" option does not tell you where the file is stored or let you choose where to store the file.

This will be fixed by adding a file selection dialog, I will make an issue

> 10. There is an error if you do not enter a file name for the "Save as file" option.

fixed this by disabling the button

> 11. The "Save as file" option saves your Account Code as a file without a file type.

Does this matter? I'm pretty sure all operating systems handle that reasonably well. And if you want a file type, you can add one.

> 12. You should be able to import your Account Code if you saved it as a file.

Nice idea, but since you can save the Account Code multiple times in different files, this would also require the mentioned file selection dialog. I will make an issue

> 13. The "Password Manager" and "Write it down yourself" options are functionally the same.

Yes, they are intended to prompt the user to think about how they are going to store it ahead of time. There is only a minor difference in the text.

> 16. "Please enter your local password!" screen's "unlock" button should be renamed to "Login"

I wanted to emphasize that this is not the login, you are just decrypting the account details. If you don't like "unlock" then maybe propose another alternative.

> 17. "Set your name" prompt should open every time when making a new account instead of just once for first-time players.

The name is completely independent from your Account though. I can see the point of allowing you to change your name before you join a multiplayer world though, but I think adding a text field to the Multiplayer button would be the way to go. I will make an issue.

ikabod-kee:
> Does this matter? I'm pretty sure all operating systems handle that reasonably well. And if you want a file type, you can add one.

At least give them the option to save it as a .txt or something then XD

IntegratedQuantum:
Just add .txt to the end?

IntegratedQuantum:
Correct me if that's different for you, but on Linux Mint there is no visible difference between a txt and a file without extension:
Screenshot at 2026-06-14 13-34-52

codemob-dev:
> Correct me if that's different for you, but on Linux Mint there is no visible difference between a txt and a file without extension:

On windows there's a quite annoying pop-up that asks you what program you want to open it with. By default, file extensions are hidden so many people won't even think to save it as a text file.

ikabod-kee:
What Codemob said

IntegratedQuantum:
done

ikabod-kee:
Cool! I'll check this out again in a bit.

ikabod-kee:
> What about the rather common "you have an untrustworthy application installed" scenario? On most operating systems any application can read almost any file.

Then just don't run "RealCubyz.exe" lol

Anyways it should explain that encrypting with a password will make you have to input it *every time you open the game.*
Which 9 times out of 10, people don't want to deal with.

ikabod-kee:
Also I do think the **Account Code** is too long to be reasonably written down manually. It's silly to ask someone to write down a code with over 100 characters.

ikabod-kee:
But that might be a seperate issue.

ikabod-kee:
Anyways anyways, I checked out the new version and it's much better. 
image
On this screen though, 15 seconds would be better than 20. That's really all my criticisms addressed that you can fit within this PR, other than the opt-out and encryption explanation.

IntegratedQuantum:
Done

> It's silly to ask someone to write down a code with over 100 characters.

It takes me less than a minute to write it on a piece of paper. But write it down can also mean that you copy paste it to the document that has all your passwords.

ikabod-kee (approved):
This gets 5 big smiles from me.

PR #3230, opened 2026-06-13
Configurable titlebar for windows

Wunka:
Added example which will be removed when this is mergable

### DIFF
--- a/src/gui/GuiWindow.zig
+++ b/src/gui/GuiWindow.zig
@@ -512,6 +512,9 @@ pub fn drawIcons(self: *const GuiWindow) void {
 	zoomOutTexture.render(.{x, 0}, .{iconWidth, titleBarHeight});
 	x -= iconWidth;
 	zoomInTexture.render(.{x, 0}, .{iconWidth, titleBarHeight});
+
+	const oldClip = graphics.draw.setClip(.{x, titleBarHeight});
+	defer graphics.draw.restoreClip(oldClip);
 	if (self.titleBar) |titleBar| titleBar.render(.{0, 0});
 }
 
--- a/src/gui/GuiWindow.zig
+++ b/src/gui/GuiWindow.zig
@@ -513,7 +513,7 @@ pub fn drawIcons(self: *const GuiWindow) void {
 	x -= iconWidth;
 	zoomInTexture.render(.{x, 0}, .{iconWidth, titleBarHeight});
 
-	const oldClip = graphics.draw.setClip(.{x, titleBarHeight - 2});
+	const oldClip = graphics.draw.setClip(.{x, titleBarHeight});
 	defer graphics.draw.restoreClip(oldClip);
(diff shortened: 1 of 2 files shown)


src/gui/GuiWindow.zig
```diff
@@ -502,6 +512,7 @@ pub fn drawIcons(self: *const GuiWindow) void {
 	zoomOutTexture.render(.{x, 0}, .{iconWidth, titleBarHeight});
 	x -= iconWidth;
 	zoomInTexture.render(.{x, 0}, .{iconWidth, titleBarHeight});
+	if (self.titleBar) |titleBar| titleBar.render(.{0, 0});
```
IntegratedQuantum, line 518: Please add a clip rect so it can't stretch into the window contents

IntegratedQuantum (changes requested):

src/gui/GuiWindow.zig
Wunka: done

src/gui/GuiWindow.zig
```diff
@@ -502,6 +512,10 @@ pub fn drawIcons(self: *const GuiWindow) void {
 	zoomOutTexture.render(.{x, 0}, .{iconWidth, titleBarHeight});
 	x -= iconWidth;
 	zoomInTexture.render(.{x, 0}, .{iconWidth, titleBarHeight});
+
+	const oldClip = graphics.draw.setClip(.{x, titleBarHeight});
```
Wunka, line 516:
```suggestion
	const oldClip = graphics.draw.setClip(.{x, titleBarHeight-2});
```
so it doesn't even strech into the border

ikabod-kee:
Does this mean...
image

Wunka:
> Does this mean..

Added to the example

src/gui/GuiWindow.zig
Wunka: @IntegratedQuantum this was kinda meant as a question. How do we want it? allow going into the border or should we not even allow this?

IntegratedQuantum: Nothing should stretch into the border.

Wunka: > Nothing should stretch into the border.

:+1: okay, I applied my suggestion

IntegratedQuantum: Oh wait, I thought you meant the window border. The border at the bottom of the titlebar is part of the texture and we shouldn't cut anything off before it. The other icons are also explicitly allowed to overlap with it.

Wunka: Soo I should revert it? I reverted it.

IntegratedQuantum (changes requested):
Please remove the example then

Wunka:
removed

IntegratedQuantum (approved):

PR #3238, opened 2026-06-15
Distinguish between joining a world as singleplayer and hosting a world for multiplayer

IntegratedQuantum:
This now distinguishes between joining a world in singleplayer and hosting a world for multiplayer.
There is now a new Window when you click multiplayer that allows you to host the game:
Screenshot at 2026-06-15 16-26-50
If you open a world through the singleplayer options instead, it will not allow you to invite any players:
Screenshot at 2026-06-15 16-26-32
This distinction is required for local accounts for #3192, and I think this will also make hosting more logical, as you no longer go through singleplayer to do so.

### DIFF
--- a/src/gui/windows/_list.zig
+++ b/src/gui/windows/_list.zig
@@ -24,6 +24,7 @@ pub const invite = @import("invite.zig");
 pub const main = @import("main.zig");
 pub const players = @import("players.zig");
 pub const multiplayer = @import("multiplayer.zig");
+pub const multiplayer_join = @import("multiplayer_join.zig");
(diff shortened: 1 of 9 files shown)


ikabod-kee:
Ah! Just like Terraria :)

ikabod-kee:
There is no distinction between the "Select World" screens for singleplayer and multiplayer. On the top of the "Host World" world selection screen, it should instead say "Select World to Host" so the player can tell them apart.

IntegratedQuantum:
done

ikabod-kee (approved):
approved

Issue #2312, opened 2025-11-17
generalize tooltip rendering into a single function


SpellDigger:
turned into a issue from #2203 

**General Idea**
to make rendering a tooltip like element on the screen easier instead of it being hardcoded
it should be generalized into a function that renders it

the function should be written in such a way, that it can render multiple tooltips at once.
the idea of a background texture is optional on this, requires a generic draw function for 9 slice texture to avoid bloat

the function to render a tooltip takes 3 arguments
**component:** ***GuiComponent** which is the component thats drawn over the tooltip background
this is where the function should know the tooltip size from
**mousePos: Vec2f** the position of the mouse, it should be the thing that determines where the tooltip is draw
and a third argument that is the alignment of the tooltip, by default it should be drawn on the right side of the mouse
the options are
.right
.left
.center

to give more freedom to the person calling the function, the function should not account for any offsets or limiting the tooltip from drawing itself out of the window bounds

thats the idea
**priority: low** as of writing this, expect related prs to be on low priority as its not important for now and other things should be focused on (like that 9 slice function if it still isnt made)
this is more on the contributor friendly side tho quantum will not go easy on u :clearly:

PR #3242, opened 2026-06-15
GuiWindow based tooltips

### LINKED ISSUE #2527: Allow any GUI component to create a tooltip on hover instead of tying it to item slots.

Wunka:
This is not resolving: #2312
But Fixes: #2527

With more and more disabled buttons (for example: #3238). I think Buttons need a tooltip to render why they are disabled. As I didn't want to do everything which is in #2312. I took the for now easier route, `getTooltip` on `GuiWindows` instead of on items only. 

This PR can be kinda seen (if needed even split into seperate PRs) as two things.
1. Add `getTooltip` on windows and move the rendering of tooltips around.
2. Add `disabledReason` variable to buttons which will be shown in the tooltip of buttons if they are disabled. (Also added two example texts)


IntegratedQuantum:
This has been broken by #3276.
I'd suggest to make your getTooltip return a component to make this more generically usable.

PR #3244, opened 2026-06-15
Make Surface Caves not Lead to dead ends anymore

Crepestrom:
This increases the chance of caves on the surface so that most large caves on the surface actually lead underground instead of stopping halfway


TyrrelMeter:
We already have a lot of weird hills+mountains because of surface caves, won't this make *even* more of those weird formations?

IntegratedQuantum:
I think the main issue is not actually the chance (which by the way you made higher than in the underground), but the fact that we have these odd hemispheres everywhere that mislead the player.

IntegratedQuantum:
I will take it from here.

Crepestrom:
ok

PR #3252, opened 2026-06-18
Bypass authentication in singleplayer and only ask the player to login when they click the multiplayer button

### LINKED ISSUE #3192: The "seed phrase" system is confusing and annoying

IntegratedQuantum:
Also comes with some minor changes to some windows to reflect the new position:
This one has an extra sentence at the start
Screenshot at 2026-06-18 21-59-16
This one now mentions decryption, and I decided to rename the unlock button to decrypt, since there were some complaints about the name (is that better now?)
Screenshot at 2026-06-18 22-03-31

Implements the remaining steps mentioned in https://github.com/PixelGuys/Cubyz/issues/3192#issuecomment-4652768962
fixes #3192


ikabod-kee:
This is good! Only complaint is that this window is not closable- though I think it shouldn't be closeable via the "X" button on the toolbar, since that could lead to it seeming like a pop-up that you want to just ignore.

image

IntegratedQuantum:
> I think it shouldn't be closeable via the "X" button on the toolbar, since that could lead to it seeming like a pop-up that you want to just ignore.

Most of our menus have the X button. How would it be different here?

ikabod-kee:
On second thought, the X button is good enough

ikabod-kee:
If you can exit out of the login screen, then you should also be able to go back to it when you want. The following screens should have a "Back" button:

image
image

IntegratedQuantum:
What if we also make them close-able?

ikabod-kee:
I mean, you could (and should), but having a back button would be handy too.

IntegratedQuantum:
Would be nice, but I can't be bothered to implement that right now, it requires carrying a bunch of extra state to ensure that you get the same account code if you go forward again, and I'd rather not worry about that right now.

IntegratedQuantum:
If you really want it you can also make an issue

ikabod-kee:
Fine then. Merge this and I'll make an issue for it.

IntegratedQuantum:
by the way, did you check this on one of your world, to see if you still got your inventory, just to be sure?

ikabod-kee:
Let me check.

ikabod-kee:
Just checked by making a singleplayer world while logging in and filling my inventory with grass blocks. Then I exited the world, logged out, and re-entered without logging back in. The grass blocks are still there, so it seems like it worked.

IntegratedQuantum:
I would like to know if an old world still works, one you used before this PR.

ikabod-kee:
I did unfortunately delete all my worlds after uploading them to the Discord server, but I do have some old worlds from even older versions where worlds were still saved in the same folder as the game. I'll check.

ikabod-kee:
Alright, I made a world in the latest master, filled my inventory with leaves, then left. Now I'm in this PR and the leaves are still here, so it seemed to work!

IntegratedQuantum:
I thought some more about the closing thing, and I think this isn't a good idea. everything that can be reached via back button from "this is your account code" should be uncloseable so you cannot bypass the timer by going back and closing (or just going back).

ikabod-kee:
Exactly

ikabod-kee (approved):

PR #3255, opened 2026-06-19
Use clusters in sky/base caves to reduce single caves and increase connectivity

IntegratedQuantum:
Clusters also now recognize child amounts.
Also adjusted the fractal cave spawn rate in surface layer.

This should make it easier to find a decent cave from the surface.

Before:
Screenshot at 2026-06-19 20-57-33
After:
Screenshot at 2026-06-19 21-18-28

related to #3244

This does not fully fix the issue, but I hope that it makes it rarer without compromising on the underground structure too much, and without spamming caves all around the surface.

### DIFF
--- a/assets/cubyz/biomes/cave/sky.zig.zon
+++ b/assets/cubyz/biomes/cave/sky.zig.zon
@@ -7,11 +7,19 @@
 	.ground_structure = .{},
 	.caveModels = .{
 		.{
-			.id = "cubyz:partial_sphere",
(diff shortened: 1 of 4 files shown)


ikabod-kee (approved):
This seems way better. I'll let @careeoki give her opinion before it's moved because I do want to know it.

careeoki (approved):
Yeah, this is better, there's less tiny air pockets everywhere which is nice.

PR #3262, opened 2026-06-20
fix asset loading for models

RanPix:
required for #2824


RanPix:
i found that now it breaks when trying to host 2 times during a single game session

RanPix:
> i found that now it breaks when trying to host 2 times during a single game session

fixed

src/game.zig
```diff
@@ -297,12 +297,9 @@ pub const World = struct { // MARK: World
...
 
-		main.Window.setMouseGrabbed(true);
-
-		main.blocks.meshes.generateTextureArray();
-		main.particles.ParticleManager.generateTextureArray();
-		main.models.uploadModels();
-		main.entityModel.loadModelsAndTexture();
+		try self.finishHandshake(main.network.protocols.handShake.handshakeZon);
+		main.network.protocols.handShake.assetsLoadedCondition.signal();
+		main.network.protocols.handShake.hasFinishedLoadingAssets = true;
```
IntegratedQuantum: You need to set this before signalling, otherwise the signalled thread may still read the old value and go back to sleep. Also I'd prefer to have this in a helper function so you can make all these helper variables private.

src/network/protocols.zig
```diff
@@ -63,6 +63,9 @@ pub fn onReceive(conn: *Connection, protocolIndex: u8, data: []const u8) !void {
 
 pub const handShake = struct { // MARK: handShake
 	pub const id: u8 = 1;
+	pub var assetsLoadedCondition: main.utils.Condition = .{};
+	pub var handshakeZon: ZonElement = undefined;
```
IntegratedQuantum: Instead of making this public, please return it from `clientSide()` instead.

src/game.zig
IntegratedQuantum: Can be private now?

IntegratedQuantum (changes requested):

RanPix:
done

IntegratedQuantum (approved):

PR #3266, opened 2026-06-20
Mod change from iterator to walker to allow trees

Wunka:
In preparation for moving other `_list.zig` ~~this implements my understanding of https://github.com/PixelGuys/Cubyz/pull/1630#discussion_r2173316900 for the rotations.~~ (It doesn't anymore, as we saw a problem with it)


mods/cubyz/rotations.zig
```diff
@@ -0,0 +1,14 @@
...
+pub const hanging = @import("rotations/hanging.zig");
+pub const torch = @import("rotations/torch.zig");
+pub const decayable = @import("rotations/decayable.zig");
+pub const direction = @import("rotations/direction.zig");
+pub const planar = @import("rotations/planar.zig");
+pub const log = @import("rotations/log.zig");
+pub const carpet = @import("rotations/carpet.zig");
+pub const branch = @import("rotations/branch.zig");
+pub const fence = @import("rotations/fence.zig");
+pub const sign = @import("rotations/sign.zig");
```
IntegratedQuantum: I think this should be generated as well, forgetting an entry is otherwise going to be quite a common bug report.

Also I would prefer to have everything in one file, to make easier for an addon creator to inspect what is available to them and to avoid add more pollution from generated files inside the mods.
I was imagining something like this:
```zig
pub const cubyz = struct {
    pub const stairs = @import("...");
    ...
    pub const a_sub_folder = struct {
        pub const ... = ...;
    };
};
pub const anti_cubyz = struct {
    ...
};
...
```

Wunka: How would that exactly work? When can I know that I need to go deeper? 
I don't mean when looking at the files but for example My created `_list.zig` is:
```zig
pub const cubyz = struct {
    pub const stairs = @import("...");
    ...
    pub const a_sub_folder = struct {
        pub const exmple = @import(...);
    };
};
...
```
How do I know that example is the deepest? What if the file of example looks like this:
```zig
// some rotation functions
pub fn init()
pub fn deinit()

pub const Helper = struct {
    ...
};
```
How do I know that helper should not be registered?
(so not how to create the file but how to parse it with my `getFeatures()` function)

One way would be to just ignore those who cannot be parsed, but this makes it less obvious for mod creators why their mod is not working

Wunka: what do you think of special casing the "go deeper" folders? for example through adding a `_` at the start or end of the name?

IntegratedQuantum: Yeah, that is a problem.
I think we have multiple options here:
- Accept the status quo and use @""
- Make two data structures, one for iteration, one for access by the code, I don't really like this one as it gives us two ways to address it.
- Tag the structs that were generated (→could add a named field), but that would make iteration and generation code even more complex.

Honestly I'm beginning to wonder if we should just stick with the `@"mod:path/name"` syntax, to be honest it doesn't look that bad, especially when it's being imported by `name` without mod/path (which we could enforce through the linter).

Wunka: I mean the syntax is not really the problem, but if there is no fix to this we can not even generate all `_list.zig` files we want, that would mean we have to go back to `_list.zig` files in general. 
I do like the `@"mod:path/name"` syntax. with my other suggestions of adding `_` to the start of a "go deeper" folder this can even be kept (the `_` can be automaticly removed for the `@"mod:path/name"` syntax.
And I think if we do it right the "get it by id" should even become the norm and import special rotations should be a "don't do that" thing. Like
```zig
const branch = main.rotation.rotations.@"cubyz:branch"
``` 
feels wrong. I think it should be something like: 
```zig
const branch = main.rotation.getRotationById("branch")
```
that way we can allow mods to overwrite the default.

IntegratedQuantum: I don't understand what you mean by a "go deeper" folder.

> that way we can allow mods to overwrite the default.

If we want to support this then that should be done at build-time, so that we wouldn't need to include overwritten things in the binary.

Wunka: > I don't understand what you mean by a "go deeper" folder.

for example the `worldedit` folder of the commands would be called: `_worldedit` then the generated `commands.zig` would look like this:
```zig
pub const cubyz = struct {
    pub const spawn = @import("...");
    ...
    pub const _worldedit = struct {
        pub const paste = @import(...);
    };
};
...
```
I can then know that I should not look if any structs are in `paste`
Mod creators can also in that way sort out helper folders where only helper structs live by just not adding the `_` at the start. But this is also just a random idea to still have the auto generation. I myself don't know how much I like / dislike the idea

IntegratedQuantum: > I do like the @"mod:path/name" syntax. with my other suggestions of adding _ to the start of a "go deeper"

But how does this work together with keeping the `@""` syntax. Then you wouldn't even have that problem as everything would be listed by its full path.

Wunka: > > I do like the @"mod:path/name" syntax. with my other suggestions of adding _ to the start of a "go deeper"
> 
> But how does this work together with keeping the `@""` syntax. Then you wouldn't even have that problem as everything would be listed by its full path.

I think I may misunderstood? When I talk about the `@""` syntax. I more talk about how the id looks. I am currently from the folder structure creating the id which then looks like this `[modName|cubyz]:feature`.

IntegratedQuantum: When I talked about the `@""` syntax I meant to say maybe we should keep the current structure of having only one file with a bunch of @"mod:name" in them.

Wunka: > When I talked about the `@""` syntax I meant to say maybe we should keep the current structure of having only one file with a bunch of @"mod:name" in them.

Yeah... I don't know. I do like the "auto create the file" but I also see the current problems... maybe @Argmaster has a good opinion there?

IntegratedQuantum: It would still be auto-created?

Wunka: > It would still be auto-created?

OHHHHH you mean I should "lower" the tree when autogenerating? That makes sense. YES I like that

Wunka: Ok. Now this pr just changes the iterator to a walker
I don't know if we want to keep them like this? (auto generated with test files to see how it looks)
```zig
pub const @"cubyz:fence" = @import("cubyz/rotations/fence.zig");
pub const @"cubyz:sign" = @import("cubyz/rotations/sign.zig");
pub const @"cubyz:special/test_rotation" = @import("cubyz/rotations/special/test_rotation.zig");
pub const @"cubyz:special/evenmorespecial/test_rotation2" = @import("cubyz/rotations/special/evenmorespecial/test_rotation2.zig");
```

IntegratedQuantum: The lines can get quite long, but I think it's fine.

IntegratedQuantum (changes requested):
Do you think you could also sort these alphabetically? The file system order is usually not sorted.

Wunka:
> Do you think you could also sort these alphabetically? The file system order is usually not sorted.

Added.

Wunka:
I didn't like how different "mods" were cramped together so I did this:
```zig

// MARK: anti-cubyz
pub const @"anti-cubyz:branch" = @import("anti-cubyz/rotations/branch.zig");
pub const @"anti-cubyz:carpet" = @import("anti-cubyz/rotations/carpet.zig");
pub const @"anti-cubyz:decayable" = @import("anti-cubyz/rotations/decayable.zig");
...

// MARK: cubyz
pub const @"cubyz:branch" = @import("cubyz/rotations/branch.zig");
pub const @"cubyz:carpet" = @import("cubyz/rotations/carpet.zig");
pub const @"cubyz:decayable" = @import("cubyz/rotations/decayable.zig");
...
```

build.zig
```diff
@@ -102,28 +102,41 @@ pub fn makeModFeature(io: std.Io, step: *std.Build.Step, name: []const u8) !void
...
 					name,
-					featureEntry.name,
+					featureEntry.path,
 				},
 			));
 		}
 	}
+	std.mem.sort([]const u8, featureList.items, {}, struct {
+		fn lessThanFn(_: void, _lhs: []const u8, _rhs: []const u8) bool {
+			const lhs = if (std.mem.startsWith(u8, _lhs, "\n// MARK: ")) _lhs[10..] else _lhs[12..];
```
IntegratedQuantum: I think it would be enough to sort within the mods, then you wouldn't need this special casing.

IntegratedQuantum (changes requested):

build.zig
Wunka: done

build.zig
```diff
@@ -102,24 +102,41 @@ pub fn makeModFeature(io: std.Io, step: *std.Build.Step, name: []const u8) !void
...
 			));
 		}
+		std.mem.sort([]const u8, modFeatureList.items, {}, struct {
+			fn lessThanFn(_: void, lhs: []const u8, rhs: []const u8) bool {
+				return std.mem.lessThan(u8, lhs, rhs);
+			}
+		}.lessThanFn);
+
+		try featureList.appendSlice(step.owner.allocator, step.owner.fmt(
+			\\
```
IntegratedQuantum: Could you also remove the empty line at the start of the file?

build.zig
```diff
@@ -102,24 +102,41 @@ pub fn makeModFeature(io: std.Io, step: *std.Build.Step, name: []const u8) !void
...
-					featureEntry.name,
+					featureEntry.path,
 				},
 			));
 		}
+		std.mem.sort([]const u8, modFeatureList.items, {}, struct {
+			fn lessThanFn(_: void, lhs: []const u8, rhs: []const u8) bool {
+				return std.mem.lessThan(u8, lhs, rhs);
+			}
+		}.lessThanFn);
```
IntegratedQuantum, line 135: I think this should work
```suggestion
		std.mem.sort([]const u8, modFeatureList.items, u8, std.mem.lessThan);
```

build.zig
```diff
@@ -102,24 +102,41 @@ pub fn makeModFeature(io: std.Io, step: *std.Build.Step, name: []const u8) !void
...
+
+		try featureList.appendSlice(step.owner.allocator, step.owner.fmt(
+			\\
+			\\// MARK: {s}
+			\\
+		, .{modEntry.name}));
+		try featureList.appendSlice(step.owner.allocator, try std.mem.join(step.owner.allocator, "\n", modFeatureList.items));
```
IntegratedQuantum: Please just use a for loops instead of making a needless intermediate allocation.

IntegratedQuantum (changes requested):

build.zig
Wunka: sadly not. the context is given by runtime while std.mem.lesstThan expects the slice type by comptime

build.zig
Wunka: done

build.zig
Wunka: I don't quite get what you exactly mean? Do you want me to use a `[][]const u8` completly and only join at the end or what? I cannot operate directly on the slice going to the file as I am sorting on the `[][]const u8` and not `[]const u8`

IntegratedQuantum: I mean to call `appendSlice` in a for loop instead of using `join` to prevent the needless intermediate allocation

build.zig
IntegratedQuantum: Instead of blindly discarding the first character at the end, I'd prefer a conditional here:
```
if (featureList.items.len != 0) featureList.append('\n');
```

Wunka: done

build.zig
Wunka: ohhh. I see now. Yeah it does a needless intermediate allocation. I made it now into a for loop

build.zig
```diff
@@ -102,24 +102,45 @@ pub fn makeModFeature(io: std.Io, step: *std.Build.Step, name: []const u8) !void
...
 
-			try featureList.appendSlice(step.owner.allocator, step.owner.fmt(
+			try modFeatureList.append(step.owner.allocator, step.owner.fmt(
 				\\pub const @"{s}:{s}" = @import("{s}/{s}/{s}");
-				\\
 			,
 				.{
 					modEntry.name,
-					featureEntry.name[0 .. featureEntry.name.len - 4],
+					featureEntry.path[0 .. featureEntry.path.len - 4],
```
IntegratedQuantum: I think on Windows the path contains `\` instead of `/` and should be normalized. Maybe you could also get someone to test this on Windows.

IntegratedQuantum (changes requested):

build.zig
Wunka: Good that I have windows laptop around. fixed and tested on windows

IntegratedQuantum (approved):

PR #3268, opened 2026-06-20
Make days longer than nights

### LINKED ISSUE #1211: Change day-cycle times

AfmanS:
Closes https://github.com/PixelGuys/Cubyz/issues/1211.

Also adds dusk and dawn as time commands.

Tested, animates well and doesn't crash on transitions between times of day.

This changes how the day cycle is calculated substantially, but in my view it's much more readable and flexible. In future, the dawn and dusk could have different colors, for example.


src/game.zig
```diff
@@ -397,63 +404,120 @@ pub const World = struct { // MARK: World
...
-			return 1 - @as(f32, @floatFromInt(dayTime - (dayCycleLength/4 - dayCycleLength/16)))/@as(f32, @floatFromInt(dayCycleLength/8));
+			periodTime -= dayDuration;
+
+			if (periodTime < duskDuration) {
+				return 1 - @as(f32, @floatFromInt(duskDuration - periodTime))/@as(f32, @floatFromInt(duskDuration));
+			}
+
+			periodTime -= duskDuration;
+
+			if (periodTime < nightDuration) {
```
IntegratedQuantum: Since you repeatedly do this in multiple functions, I'd suggest to put this into a helper function that returns a tagged union which gives you the progress (0 to 1) for each of the day phases:
```zig
union(enum) {
	day: f32,
	dawn: f32,
	night: f32,
	dusk: f32,
}
```
Then here you'd just do
```zig
return switch(getDayPhase()) {
	.day => 0,
	.night => 1,
	.dusk => |dusk| 1 - dusk,
	.dawn => |dawn| dawn,
};
```

IntegratedQuantum (changes requested):
In order for stars (and the future sun) to be correct they'd need to rotate around a different axis. Could you also calculate that axis here and apply it to the star rotation?

src/game.zig
AfmanS: Since we're going to adjust the stars, should the day be calculated from characteristics such as orbital inclination, latitude and planetary rotational speed?
Or should it just calculate the celestial axis from hardcoded day phase values?
I ask because you said you wanted the moon to be a possible location eventually and this logic could be reused.

IntegratedQuantum: No I don't think we need to worry about orbital mechanics.

The visitable moon would be a static sphere of blocks in the existing world without gravity even, so it wouldn't have different values.

AfmanS:
Here's the tilt:
output
Also added phase to debug screen
image

src/game.zig
Argmaster: I can't tell if ur both serious or not

src/gui/windows/debug.zig
```diff
@@ -63,7 +63,7 @@ pub fn render() void {
 		y += 8;
 		draw.print("EyePos: {d:.1} EyeVelocity: {d:.1} EyeCoyote: {d:.3}", .{player.getEyePosBlocking(), player.getEyeVelBlocking(), @max(0, player.getEyeCoyoteBlocking())}, 0, y, 8, .left);
 		y += 8;
-		draw.print("Game Time: {} Day Time: {}", .{main.game.world.?.gameTime.load(.monotonic), main.game.world.?.dayTime.dayTime}, 0, y, 8, .left);
+		draw.print("Game Time: {} Day Time: {} Day Phase: {}", .{main.game.world.?.gameTime.load(.monotonic), main.game.world.?.dayTime.dayTime, main.game.world.?.dayTime.dayPhase}, 0, y, 8, .left);
```
IntegratedQuantum: In my opinion with the Day Phase, the Day Time doesn't really have much value to be displayed anymore.

src/game.zig
```diff
@@ -382,85 +382,152 @@ pub const World = struct { // MARK: World
...
+		pub const duskStart = dayStart + dayDuration;
+		pub const nightStart = duskStart + duskDuration;
+		pub const dawnStart = nightStart + nightDuration;
+
+		const dayCycleLength = dawnStart + dawnDuration; // 24 minutes or 14400 ticks
+
+		const lightFraction = @as(f32, @floatFromInt(dawnDuration/2 + dayDuration + duskDuration/2))/@as(f32, @floatFromInt(dayCycleLength));
+		const halfDayArc: f32 = std.math.pi*lightFraction;
+		const axialTilt = 0.409106; // same as earth
+		pub const celestialTilt = std.math.atan(-@cos(halfDayArc)/@tan(axialTilt));
```
IntegratedQuantum, line 434: Please link the source of the formula (doesn't need to be in the code, but it would make it easier to review, I do think the presence of the axial tilt is a bit odd)

src/game.zig
```diff
@@ -382,85 +382,152 @@ pub const World = struct { // MARK: World
...
 		}
 
-		pub fn getStarOpacity(self: *DayTime) f32 {
-			const dayTime = @abs(self.dayTime - dayCycleLength/2);
-			if (dayTime < dayCycleLength/4 - dayCycleLength/16) {
-				return 1;
-			}
-			if (dayTime > dayCycleLength/4 + dayCycleLength/16) {
-				return 0;
+		fn updatePhaseOfDay(self: *DayTime) void {
```
IntegratedQuantum: I think this should be combined with `updateTimeOfDay`

src/game.zig
```diff
@@ -382,85 +382,152 @@ pub const World = struct { // MARK: World
...
-			if (dayTime < dayCycleLength/4 - dayCycleLength/16) {
-				self.ambientLight = 0.1;
+			if (periodTime < duskDuration) {
+				self.dayPhase = .{.dusk = 1 - @as(f32, @floatFromInt(duskDuration - periodTime))/@as(f32, @floatFromInt(duskDuration))};
 				return;
 			}
-			if (dayTime > dayCycleLength/4 + dayCycleLength/16) {
-				self.ambientLight = 1;
+
+			periodTime -= duskDuration;
```
IntegratedQuantum: Instead of subtracting these everywhere, why not just check against the `duskStart` and related constants you declared above?

IntegratedQuantum (changes requested):

src/gui/windows/debug.zig
AfmanS: done

src/game.zig
AfmanS: done

src/game.zig
AfmanS: you're right, i changed it to be much faster

src/game.zig
AfmanS: done

src/game.zig
```diff
@@ -416,81 +416,142 @@ pub const World = struct { // MARK: World
...
+		pub const nightStart = duskStart + duskDuration;
+		pub const dawnStart = nightStart + nightDuration;
+
+		const dayCycleLength = dawnStart + dawnDuration; // 24 minutes or 14400 ticks
+
+		// The celestial sphere spins about an axis raised celestialPoleAltitude above the horizon,
+		// with the sun solarDeclination radians off the sphere's equator.
+		// We can use the sunrise equation when the sun sits on the horizon:
+		//     cos(sunsetHourAngle) = -tan(celestialPoleAltitude)*tan(solarDeclination)
+		const daylightFraction = @as(f32, @floatFromInt(2*dayDuration + dawnDuration + duskDuration))/@as(f32, @floatFromInt(2*dayCycleLength));
```
IntegratedQuantum: I think this would be more readable:
```suggestion
		const daylightFraction = @as(f32, @floatFromInt(dayDuration + (dawnDuration + duskDuration)/2))/@as(f32, @floatFromInt(dayCycleLength));
```

src/game.zig
```diff
@@ -416,81 +416,142 @@ pub const World = struct { // MARK: World
...
+		const sunsetHourAngle: f32 = std.math.pi*daylightFraction;
+		const solarDeclination = 0.409106; // equals Earth's axial tilt since it's always the solstice, can later be used to position sun in the sky
```
IntegratedQuantum: I think this would be more readable with std.math.degreesToRadians and the degree value.

src/game.zig
```diff
@@ -416,81 +416,142 @@ pub const World = struct { // MARK: World
...
+					if (dusk < 0.5) {
+						skyColorFactor[0] = 1;
+					} else {
+						skyColorFactor[0] = (1 - dusk)/0.5;
+					}
+
+					return skyColorFactor;
+				},
+				.night => @splat(0),
+				.dawn => |dawn| {
```
IntegratedQuantum: Why not combine dusk and dawn?
```
.dusk, .dawn => |progress| {
    const brightness = if (self.dayPhase == .dusk) 1 - progress else progress;
    // Then use brightness to do the calculation
}
```

IntegratedQuantum (changes requested):

IntegratedQuantum:
Also something is wrong with the debug display (during night it shows negative dawn):
Screenshot at 2026-09-06 10-54-30

src/game.zig
AfmanS: done

src/game.zig
AfmanS: done

IntegratedQuantum (approved):
looks good, thank you

PR #3254, opened 2026-06-19
Cobble variants

careeoki:
closes #3196 

Whoever added all these stone types is sure making my life a lot harder when it comes to adding these new variants.

I did it in the way I originally proposed to show what I would like and to make it faster while I was texturing. Either way I will have to do a bunch of migrating later.
Now that we have a PR, we need a final decision for how we are doing these. Are we going with the smooth and rough stone idea I proposed, or what? And the thing about recipes in the issue is still an open question.

image
image

- [x] we need a final decision on how to do these cobbles
- [x] textures final pass
- [x] the great migration

PR #3273, opened 2026-06-21
Collapse transitive migrations

IntegratedQuantum:
should fix the problem described in https://github.com/PixelGuys/Cubyz/pull/3254#issuecomment-4762715964

I also decided to add an error for circular migrations:

Screenshot at 2026-06-21 21-56-31


careeoki:
Yay, nice.

PR #3276, opened 2026-06-22
Generalize Tooltip Rendering into one function, Also add 9 slice drawing function

### LINKED ISSUE #2312: generalize tooltip rendering into a single function

SpellDigger:
closes: https://github.com/PixelGuys/Cubyz/issues/2312

Tooltip Example (i added backgroud texture)
{52F36E4C-8F71-4810-BA7F-F0D6AAF1CC8D}

buttons also now use bound9SliceImage to draw its outline instead of it being hardcoded
(they are drawn with slice scale 2 so it matches the original)
{A6A21B87-C8C7-4280-8448-1756F5686CB9}

waiting to get slimed by quantum


SpellDigger:
Made the tooltip texture less curved

{0218E725-1F0E-46D0-9D79-626F9D02FB78}

Wunka:
As I atleast think we need more tooltips (see #3242) I will move this to high priority so it gets soon looked at)

src/graphics.zig
```diff
@@ -470,6 +470,34 @@ pub const draw = struct { // MARK: draw
 		c.glDrawArrays(c.GL_TRIANGLE_STRIP, 0, 4);
 	}
 
+	pub fn drawSlice(_destMin: Vec2f, _destMax: Vec2f, _uvMin: Vec2f, _uvMax: Vec2f) void {
+		boundSubImage(_destMin, _destMax - _destMin, _uvMin, _uvMax - _uvMin);
+	}
+
+	pub fn bound9SliceImage(_pos: Vec2f, _dim: Vec2f, _textureSize: Vec2f, _sliceCenter: Vec4f, _sliceScale: f32) void {
```
IntegratedQuantum: Please don't use underscores everywhere.
Also I'd suggest to rename `_sliceCenter` to `edgeSize`, also is there any reason this is a `Vec4f` instead of a simple float? In my opinion we shouldn't support asymmetrical texture sizes.

Also the existence of `sliceScale` is just odd. What does this mean? Why is it needed? It looks like it's hiding a bug.

src/gui/components/Button.zig
```diff
@@ -162,23 +162,10 @@ pub fn render(self: *Button, mousePosition: Vec2f) void {
...
-		graphics.draw.boundSubImage(self.pos + self.size - cornerSize, cornerSize, upperTexture, cornerSizeUV);
-		// Edges:
-		graphics.draw.boundSubImage(self.pos + Vec2f{cornerSize[0], 0}, Vec2f{self.size[0] - 2*cornerSize[0], cornerSize[1]}, .{lowerTexture[0], 0}, .{upperTexture[0] - lowerTexture[0], cornerSizeUV[1]});
-		graphics.draw.boundSubImage(self.pos + Vec2f{cornerSize[0], self.size[1] - cornerSize[1]}, Vec2f{self.size[0] - 2*cornerSize[0], cornerSize[1]}, .{lowerTexture[0], upperTexture[1]}, .{upperTexture[0] - lowerTexture[0], cornerSizeUV[1]});
-		graphics.draw.boundSubImage(self.pos + Vec2f{0, cornerSize[1]}, Vec2f{cornerSize[0], self.size[1] - 2*cornerSize[1]}, .{0, lowerTexture[1]}, .{cornerSizeUV[0], upperTexture[1] - lowerTexture[1]});
-		graphics.draw.boundSubImage(self.pos + Vec2f{self.size[0] - cornerSize[0], cornerSize[1]}, Vec2f{cornerSize[0], self.size[1] - 2*cornerSize[1]}, .{upperTexture[0], lowerTexture[1]}, .{cornerSizeUV[0], upperTexture[1] - lowerTexture[1]});
-	}
+
+	textures.outlineTexture.bindTo(0);
+	graphics.draw.bound9SliceImage(self.pos, self.size, textures.outlineTextureSize, .{2, 2, 2, 2}, 2);
```
IntegratedQuantum: Previously this number was determined from the texture data. Allowing for arbitrary resolution textures. Here you have hardcoded it. Texture information should never be hardcoded.

src/gui/tooltip.zig
```diff
@@ -0,0 +1,45 @@
...
+pub fn globalDeinit() void {
+	tooltipTexture.deinit();
+}
+
+pub fn render(guicomponent: *GuiComponent, pos: Vec2f, alignment: graphics.TextBuffer.Alignment) void {
+	const size = guicomponent.size();
+	tooltipTexture.bindTo(0);
+
+	var renderpos = pos;
+	switch (alignment) {
```
IntegratedQuantum:
```suggestion
	const renderpos = switch (alignment) {
```

src/gui/tooltip.zig
```diff
@@ -0,0 +1,45 @@
...
+		.right => {
+			renderpos = pos + Vec2f{tooltipSliceCenter[0], 0};
+		},
```
IntegratedQuantum:
```suggestion
		.right => pos + Vec2f{tooltipSliceCenter[0], 0},
```

src/gui/components/Label.zig
```diff
@@ -12,7 +12,7 @@ const GuiComponent = gui.GuiComponent;
 
 const Label = @This();
 
-const fontSize: f32 = 16;
+pub const fontSize: f32 = 16;
```
IntegratedQuantum: No, this is not the place for a global font size.
```suggestion
const fontSize: f32 = 16;
```

assets/cubyz/ui/tooltip_background.png
IntegratedQuantum, line 1: Why is this slightly transparent? I cannot see through it at all.

src/gui/gui.zig
```diff
@@ -784,38 +787,29 @@ pub const inventory = struct { // MARK: inventory
 		// Draw tooltip:
 		const hovered = hoveredItemSlot orelse return;
 		if (carried.getAmount(0) == 0) {
-			if (hovered.inventory.getItem(hovered.itemSlot).getTooltip()) |tooltip| {
-				var textBuffer = graphics.TextBuffer.init(main.stackAllocator, tooltip, .{}, false, .left);
-				defer textBuffer.deinit();
-				const fontSize = 16;
-				var size = textBuffer.calculateLineBreaks(fontSize, 300);
+			if (hovered.inventory.getItem(hovered.itemSlot).getTooltip()) |tooltipContent| {
+				var label = GuiComponent.Label.init(Vec2f{0, 0}, 300, tooltipContent, .left);
```
IntegratedQuantum: I'd suggest to make a convenience function in tooltip.zig to render it as text.

IntegratedQuantum (changes requested):

SpellDigger:
woah alot of stuff to do time to get to work

assets/cubyz/ui/tooltip_background.png
SpellDigger: honestly i dont know myself i just made it like that cuz it looks nicer, i can make it non transparent or more transparent if you like, i can see through it just fine but maybe its cuz of my monitor settings

src/graphics.zig
SpellDigger: the reason why _sliceCenter is a Vec4f is to allow for drawing more complex 9 slice images, having a float edgeSize locks away more complex textures from being drawn for which some sides and corners may have different proportions, if you really insist i can make it a single float

sliceScale is not a bug, idk what r u talking about but the whole purpose of sliceScale is quite literally what it says to scale up the size of the slices i think you are forgetting this is a general function for simplicity i can tell u goodluck drawing the buttons outline without it because u wont get the same result as the original i may remind u that the original code for drawing the buttons outline was also scaled x2

also the underscores for the u and v arguments are only there cuz zig compiler was screaming at me and refused to compile i could probably rename them to something else

src/gui/components/Button.zig
SpellDigger: im assuming you are talking about the slice edges here, well i sure can make the slice edges be determined from the texture

assets/cubyz/ui/tooltip_background.png
IntegratedQuantum: I would suggest to just make it opaque.

src/gui/components/Button.zig
IntegratedQuantum: yes

src/graphics.zig
IntegratedQuantum: But why is the button scaling the texture, but the tooltip isn't? Shouldn't everything have the same scale?

src/gui/components/Button.zig
SpellDigger: im doing the same to the tooltip background just incase

src/graphics.zig
IntegratedQuantum: I'd prefer a single float, or 2 floats if you really want to have differences between x and y sizes. Anything else seems just weird and shouldn't be allowed in my opinion.

SpellDigger: oki doki (i will make it a Vec2f)
the button is being scaled purely so it looks the same as the original buttons did, if you actually prefer a smaller outline i can set the scale to 1

IntegratedQuantum: But why doesn't the tooltip texture need to be scaled by the same amount?

SpellDigger: tooltip has scale of 1? its just the buttons who have scale 2 since the original button code scaled them like that

IntegratedQuantum: In my opinion we should have the same pixel size for the tooltip outline as for the button outline

SpellDigger: oki doki ur the boss i dont really care enough

SpellDigger:
Stuff changed:

- funni quantum requests
- tooltip background is now opaque
- added renderFromText
- tooltip now handles switching between alignments dynamically when the provided one leaves the screen
- the dynamic corner thingy for buttons and tooltip textures
- tooltip offset is now not based off texture but a constant (before the offset was gotten from cornerSize which created issues when tooltip texture was big)

src/graphics.zig
```diff
@@ -470,6 +470,30 @@ pub const draw = struct { // MARK: draw
 		c.glDrawArrays(c.GL_TRIANGLE_STRIP, 0, 4);
 	}
 
+	pub fn drawSlice(_destMin: Vec2f, _destMax: Vec2f, _uvMin: Vec2f, _uvMax: Vec2f) void {
+		boundSubImage(_destMin, _destMax - _destMin, _uvMin, _uvMax - _uvMin);
+	}
+
+	pub fn bound9SliceImage(_pos: Vec2f, _dim: Vec2f, _textureSize: Vec2f, _slices: Vec2f, _sliceScale: f32) void {
```
IntegratedQuantum: Again please remove all unnecessary underscores inside this and all other functions.

src/gui/gui.zig
```diff
@@ -19,6 +19,7 @@ const ScrollBar = @import("components/ScrollBar.zig");
 const ContinuousSlider = @import("components/ContinuousSlider.zig");
 const DiscreteSlider = @import("components/DiscreteSlider.zig");
 const TextInput = @import("components/TextInput.zig");
+const tooltip = @import("tooltip.zig");
```
IntegratedQuantum: Each file should be publicly imported in exactly one place so it's accessible from `main`.

src/gui/tooltip.zig
```diff
@@ -0,0 +1,71 @@
...
+
+fn posFromAlignment(pos: Vec2f, size: Vec2f, alignment: graphics.TextBuffer.Alignment) Vec2f {
+	return switch (alignment) {
+		.right => pos + Vec2f{offsetFromMouse, 0},
+		.left => pos - Vec2f{size[0] + cornerSize[0]*2, 0},
+		.center => pos - Vec2f{size[0]/2, 0},
+	};
+}
+
+pub fn render(guicomponent: *GuiComponent, pos: Vec2f, alignment: graphics.TextBuffer.Alignment) void {
```
IntegratedQuantum: I don't think we should allow for alignment here. Tooltips should always be rendered in the same way in my opinion.

This would mean you can simplify the logic to a single if, like for the y coordinate.

src/gui/tooltip.zig
```diff
@@ -0,0 +1,71 @@
...
+	var label = GuiComponent.Label.init(Vec2f{0, 0}, 300, text, .left);
+	var size = label.text.calculateLineBreaks(fontSize, 300);
+	size[0] = 0;
+	for (label.text.lineBreaks.items) |lineBreak| {
+		size[0] = @max(size[0], lineBreak.width);
+	}
+	label.size = size;
+
+	var component = GuiComponent{.label = label};
+	defer component.deinit();
```
IntegratedQuantum: `defer deinit` should always be next to the init function, not 10 lines below it.

src/gui/tooltip.zig
```diff
@@ -0,0 +1,71 @@
...
+pub fn renderFromText(text: []const u8, pos: Vec2f, alignment: graphics.TextBuffer.Alignment) void {
+	var label = GuiComponent.Label.init(Vec2f{0, 0}, 300, text, .left);
+	var size = label.text.calculateLineBreaks(fontSize, 300);
+	size[0] = 0;
+	for (label.text.lineBreaks.items) |lineBreak| {
+		size[0] = @max(size[0], lineBreak.width);
+	}
+	label.size = size;
+
+	var component = GuiComponent{.label = label};
```
IntegratedQuantum: use `label.toComponent()`

src/gui/tooltip.zig
```diff
@@ -0,0 +1,71 @@
...
+	} else if (renderpos[0] < 0) {
+		renderpos = posFromAlignment(pos, size, .right);
+	}
+	if (renderpos[1] + size[1] > windowSize[1]) {
+		renderpos[1] += windowSize[1] - (renderpos[1] + size[1]);
+	}
+
+	draw.bound9SliceImage(renderpos, size, @floatFromInt(tooltipTexture.size()), cornerSize, 1);
+
+	guicomponent.mutPos().* = renderpos + Vec2f{cornerSize[0], cornerSize[1]};
```
IntegratedQuantum: Instead of modifying the component, please use `draw.setTranslation` and `draw.setClip` to define the drawing area and pass an adjusted mouse position.

IntegratedQuantum (changes requested):

src/graphics.zig
SpellDigger: maybe list the unnecessary underscores so i actually know

IntegratedQuantum: I'm pretty sure most of them are unnecessary.
You only need underscores when you have name conflicts, and I'm pretty sure no similar names exist in scope.

src/gui/gui.zig
SpellDigger: im assuming you mean make it public

src/gui/tooltip.zig
SpellDigger: i would argue otherwise incase someone would want to render 1 tooltip on the left and 1 tooltip on the right for whatever reason there could be, but since u say so

src/gui/tooltip.zig
SpellDigger: im not following what ur trying to make me do here

IntegratedQuantum: Check out VerticalList.render

SpellDigger: im assuming you mean doing it like this?
```zig
const adjustedpos = renderpos + Vec2f{cornerSize[0], cornerSize[1]};
const oldTranslation = draw.setTranslation(adjustedpos);
defer draw.restoreTranslation(oldTranslation);
const oldClip = draw.setClip(guicomponent.size());
defer draw.restoreClip(oldClip);
guicomponent.render(adjustedpos);
```

SpellDigger: nvm i forgot to consider guicomponent would still use its internal position this might be more correct if we are going for mouse being inside the component
```zig
const adjustment = Vec2f{cornerSize[0], cornerSize[1]};
const oldTranslation = draw.setTranslation(renderpos + adjustment);
defer draw.restoreTranslation(oldTranslation);
const oldClip = draw.setClip(guicomponent.size());
defer draw.restoreClip(oldClip);
guicomponent.render(guicomponent.pos() + adjustment);
```

src/graphics.zig
```diff
@@ -470,6 +470,30 @@ pub const draw = struct { // MARK: draw
 		c.glDrawArrays(c.GL_TRIANGLE_STRIP, 0, 4);
 	}
 
+	pub fn drawSlice(destMin: Vec2f, destMax: Vec2f, uvMin: Vec2f, uvMax: Vec2f) void {
```
IntegratedQuantum: Helper functions should be private
```suggestion
	fn drawSlice(destMin: Vec2f, destMax: Vec2f, uvMin: Vec2f, uvMax: Vec2f) void {
```

src/gui/tooltip.zig
```diff
@@ -0,0 +1,63 @@
...
+pub fn globalInit() void {
+	tooltipTexture = Texture.initFromFile("assets/cubyz/ui/tooltip_background.png");
+	cornerSize = (@as(Vec2f, @floatFromInt(tooltipTexture.size())) - Vec2f{1, 1})/Vec2f{2, 2};
+}
+
+pub fn globalDeinit() void {
+	tooltipTexture.deinit();
+}
+
+pub fn render(guicomponent: *GuiComponent, pos: Vec2f) void {
```
IntegratedQuantum:
```suggestion
pub fn render(guiComponent: *GuiComponent, pos: Vec2f) void {
```

src/gui/tooltip.zig
```diff
@@ -0,0 +1,63 @@
...
+	const size = guicomponent.size() + Vec2f{cornerSize[0]*2, cornerSize[1]*2};
+	tooltipTexture.bindTo(0);
+
+	const windowSize = main.Window.getWindowSize()/@as(Vec2f, @splat(gui.scale));
+	var renderpos = pos + Vec2f{offsetFromMouse, 0};
```
IntegratedQuantum:
```suggestion
	var renderPos = pos + Vec2f{offsetFromMouse, 0};
```

src/gui/tooltip.zig
```diff
@@ -0,0 +1,63 @@
...
+	if (renderpos[0] + size[0] > windowSize[0]) {
+		renderpos = pos - Vec2f{size[0] + cornerSize[0]*2, 0};
+	}
```
IntegratedQuantum, line 36: I'd suggest to also do a check for the left side of the screen, just in case the tooltip is larger than half the screen
```suggestion
	}
	renderPos = @max(0, renderPos);
```

IntegratedQuantum (changes requested):

SpellDigger:
task completed.

IntegratedQuantum (approved):

PR #3278, opened 2026-06-22
Reload Part 2 | no autokick when /server restart (headless) 

tillpp:
https://github.com/user-attachments/assets/c3aad82a-5264-413a-af14-b0100489bec5

TODOs:

- [x] refCount of User doesnt hit zero when leaving the server?!
- [x] handle old packets not being received after restart


tillpp:
ready for review

Wunka:
Is that not almost #2400? :laughing:

tillpp:
> Is that not almost https://github.com/PixelGuys/Cubyz/issues/2400? 😆

Quantum forced me to split it into multiple PRs because one was to big :(
https://github.com/PixelGuys/Cubyz/pull/3219 was basicly the first step towards reload
this is the second^^

and third would be to also be able to do /server restart on headfull (but not sure if i feel motivated on working on the third part)

src/network/protocols.zig
```diff
@@ -45,7 +47,23 @@ pub fn init() void { // MARK: init()
...
+		} else {
+			try Reload.clientReceiveWithChannelId(conn, &reader, channelId);
+		}
+		return;
+	}
+	std.debug.assert(@intFromEnum(channelId) < 3);
+	std.debug.assert(channelId == ChannelId.lossy or channelId == ChannelId.secure or channelId == ChannelId.slow);
+
+	// Throw away everything from before the restart
+	if (conn.restartChannelCounter[@intFromEnum(channelId)] != conn.restartCounter) return;
```
IntegratedQuantum: I'd prefer if you'd do all this in network.zig, protocols.zig should not need to know that the data is sent through different channels.

src/network/protocols.zig
```diff
@@ -1056,3 +1085,53 @@ pub const EntityComponentUpdate = struct { // MARK: EntityComponentUpdate
 		conn.send(.secure, id, writer.data.items);
 	}
 };
+
+pub const Reload = struct { // MARK: Reload
```
IntegratedQuantum: To emphasize that this is part of the low-level network protocol (since it does change quite a few things on the `Connection`), I think it would make more sense to have this as a separate `ChannelId` so the code can be executed in `network.zig`

src/server/server.zig
```diff
@@ -582,6 +606,16 @@ fn init(name: []const u8, singlePlayerPort: ?u16, mode: ServerWorld.Mode) void {
...
 	};
+
+	connectionManager.@"continue"() catch |err| {
+		std.log.err("Couldn't create thread: {s}", .{@errorName(err)});
+		@panic("Could not open Server.");
+	};
+	for (connectionManager.connections.items) |conn| {
+		main.network.protocols.Reload.informClientOfRestart(conn);
+		conn.handShakeState.store(.signatureResponse, .monotonic);
+	}
```
IntegratedQuantum: I think it would make sense to put this directly inside the continue function.

src/server/server.zig
```diff
@@ -154,20 +156,53 @@ pub const User = struct { // MARK: User
...
 		self.increaseRefCount();
-		self.worldEditData = .init();
-		self.permissions = .init(main.globalAllocator);
 		network.protocols.handShake.serverSide(self.conn);
 		return self;
 	}
+	pub fn wakeup(self: *User) void {
+		if (self.wokeup) return;
+
+		// persistent data
```
IntegratedQuantum: I'd suggest to put the persistant data into a substruct.

src/server/server.zig
```diff
@@ -154,20 +156,53 @@ pub const User = struct { // MARK: User
...
 		std.debug.assert(self.refCount.load(.monotonic) == 0);
 
+		self.wakedown();
+
+		self.conn.deinit();
+		main.globalAllocator.free(self.name);
+		if (self.newKeyString) |str| main.globalAllocator.free(str);
+		main.globalAllocator.destroy(self);
+	}
+	pub fn wakedown(self: *User) void {
```
IntegratedQuantum: I'M not happy with the naming here.
I personally would suggest to use the same function names as for the ConnectionManager.

src/network.zig
```diff
@@ -551,7 +571,7 @@ pub const ConnectionManager = struct { // MARK: ConnectionManager
 		self.packetSendRequests.deinit(main.globalAllocator.allocator);
 
 		for (self.connections.items) |conn| {
-			conn.disconnect();
+			conn.hibernate();
```
IntegratedQuantum: Same here, why not use the same naming as for the ConnectionManager?

src/game.zig
```diff
@@ -281,21 +281,24 @@ pub const World = struct { // MARK: World
 	itemDrops: ClientItemDropManager = undefined,
 	playerBiome: Atomic(*const main.server.terrain.biomes.Biome) = undefined,
 
-	pub fn init(self: *World, ip: []const u8, manager: *ConnectionManager) !void {
+	pub fn init(self: *World, ip: []const u8, manager: *ConnectionManager, restart: bool) !void {
```
IntegratedQuantum: Instead of a flag I'd suggest to also make a separate pair of functions for restarting.

src/network.zig
```diff
@@ -1441,6 +1465,9 @@ pub const Connection = struct { // MARK: Connection
 	secureChannel: SecureChannel,
 	slowChannel: Channel,
 
+	restartChannelCounter: [3]u32 = .{0, 0, 0},
+	restartCounter: u32 = 0,
```
IntegratedQuantum, line 1477: Please init this with a random number, to avoid potential issues when one packet arrives later after a client reconnect and two server restarts.

src/main.zig
```diff
@@ -496,11 +498,33 @@ pub fn clientMain() void { // MARK: clientMain()
...
+					std.log.err("Encountered error while opening world: {s}", .{@errorName(err)});
+					gui.windowlist.notification.raiseNotification("Encountered error while opening world: {s}", .{@errorName(err)});
+					game.world = null;
+					continue;
+				};
+				settings.save();
+
+				for (gui.openWindows.items) |openWindow| {
+					gui.closeWindowFromRef(openWindow);
+				}
```
IntegratedQuantum: Isn't this already done by world.deinit?

src/main.zig
```diff
@@ -496,11 +498,33 @@ pub fn clientMain() void { // MARK: clientMain()
...
+
+				network.protocols.Reload.informServerOfRestart(world.conn);
+				world.conn.handShakeState.store(.start, .monotonic);
+				game.testWorld.init(&.{}, game.testWorld.manager, true) catch |err| {
+					std.log.err("Encountered error while opening world: {s}", .{@errorName(err)});
+					gui.windowlist.notification.raiseNotification("Encountered error while opening world: {s}", .{@errorName(err)});
+					game.world = null;
+					continue;
+				};
+				settings.save();
```
IntegratedQuantum: Why is this happening? What setting changed here?

src/main.zig
```diff
@@ -496,11 +498,33 @@ pub fn clientMain() void { // MARK: clientMain()
...
 			gui.windowlist.gpu_performance_measuring.stopQuery();
 		}
+		if (shouldRestart.load(.acquire)) {
+			shouldRestart.store(false, .monotonic);
+			if (game.world) |world| {
+				world.deinit(true);
+				std.debug.assert(game.world == &game.testWorld);
+
+				network.protocols.Reload.informServerOfRestart(world.conn);
+				world.conn.handShakeState.store(.start, .monotonic);
```
IntegratedQuantum: Shouldn't this start with the reload step?

src/main.zig
```diff
@@ -496,11 +498,33 @@ pub fn clientMain() void { // MARK: clientMain()
...
+				game.testWorld.init(&.{}, game.testWorld.manager, true) catch |err| {
+					std.log.err("Encountered error while opening world: {s}", .{@errorName(err)});
+					gui.windowlist.notification.raiseNotification("Encountered error while opening world: {s}", .{@errorName(err)});
+					game.world = null;
+					continue;
```
IntegratedQuantum: A random continue in a big loop like this is rather weird. I'd suggest to move this whole thing into a function. Maybe in game.zig?
Also to work correctly this needs `gui.openWindow("main");`

IntegratedQuantum (changes requested):
> and third would be to also be able to do /server restart on headfull

Could you please turn #2400 into a checklist then, so it's easier to track what's missing?
e.g. the other thing that's missing is asset reloading (since currently they are cached on start).

src/network/protocols.zig
tillpp: what exactly should have a different ChannelId?
This thing works because it uses the same channelIds as the other packets

IntegratedQuantum: Yeah, you are right, I forgot about this.

Still I would prefer to see all of the action be catched earlier, in network.zig, and just have a dummy here that makes sure no one else uses this protocol id for anything.

src/server/server.zig
tillpp: nothing really gets paused or continued here, but i can rename it anyway

src/network.zig
tillpp: that isn't an issue, since the client will then be at a higher restartCounter, ignoring older packets.

src/server/server.zig
IntegratedQuantum: Yeah, but it kind of belongs to the network pausing, so I would like to see consistent naming between those.

Maybe suspend/resume could fit more?

src/network.zig
IntegratedQuantum: Not when the client reconnects manually which produces a new `Connection`.

What do you think about putting the restartCounter into the ConnectionManager?

src/main.zig
tillpp: what do you mean? this is the start of the reload step?

IntegratedQuantum: No, `.start` is restarting the whole handshake, but you implemented `.reload` as a separate step, I assumed you meant to use it here?

src/server/server.zig
tillpp: are you sure you want me to move user.conn to user.persistentData.conn?

IntegratedQuantum: Hmm, yeah I guess it's used in a lot of places.
Let's leave it for now then.

src/main.zig
tillpp: ah yes,

src/network.zig
tillpp: I think you might be under the misconception that there is only one restartCounter per Server?
Each Connection has it's own unique restartCounter^^

IntegratedQuantum: Does it need to be unique though?

tillpp: Having a restartCounter for each user gives a lot of advantages, like simpler code.
And a global restartCounter does provide no benefit and only more complex code to handle.
for example i would have transmit  the starting restartCounter to the client when first joining.

The important part is, the Issue you are describing is not happening, since the new Connection doesn't count the global restartCounter but it's own restartCounter. 

(i also like it for some secret side project.)

IntegratedQuantum: The real reason why my issue can't happen is because a new connection has new channels with new byte offsets, so if it receives a packet from the previous connection, it will be discarded elsewhere anyways.
Anyways you are right, we shouldn't add more complexity if there are no issues.

Then I see no reason to make it so big. `u8` is probably enough.

tillpp: > The real reason 

**BUT This is exactly what i am already Refering to! AARRGRGH!** 

since a new connection dismissises old packages, i don't need to keep a global restartCounter, if i would have a globalRestartcounter then i would have the issue that you described. Because without sending the state of the globalrestart counter when joining:

global rc 0
user1 joins with rc 0 
server restarts
global rc 1
user1 rc 1
user manually restarts, rc back at 0.
server restarts
global rc 2
user1 rc 1
user1 receives old packets! error!

  What is more maddening is that i was about to write "the real reason...", but stopped since it is obvious to both of us, so concluded you must meant the case above.

> Then I see no reason to make it so big. u8 is probably enough.

u8 would only allow a connection to restart 255, not much, especially if you testing something frquently (imagine an addon tester for example).
There is no good reason to shorten this variable, just to save a few bytes where it doesn't matter.

src/network/protocols.zig
tillpp: i removed the receive functions, but kept the send functions, hope that is dummy enough

src/network.zig
IntegratedQuantum: But you can use overflow arithmetic.

> There is no good reason to shorten this variable, just to save a few bytes where it doesn't matter.

It takes more space in the struct. You are making it 12 bytes bigger for no reason.
Of course it probably won't have any measurable impact in any benchmark, but it could be the 12 bytes that push another field into a new cache line. I don't any reason to make it worse, especially since both implementations would be equally complex.

IntegratedQuantum: In fact switching to u8 would even make the implementation simpler, since you wouldn't need a reader/writer anymore.

src/game.zig
tillpp: do you mean two init functions? or continue & pause like in connectionManager? because i have problems with both ideas here

src/network.zig
tillpp: > But you can use overflow arithmetic.

how do i use > < comparison on overflow arithmetic?
What if i restart the server 255 times, and my friend having a bad ping, catching up and is at the same id but one entire 255 in the past?

IntegratedQuantum: Right, good point, I didn't consider that a ton of data could get stuck in the network queue.

src/game.zig
IntegratedQuantum: I think I would prefer a separate struct that holds the Connection+Manager with its own init/deinit pair.

tillpp: hmm world.conn is used at a lot of places, are you sure to put it in a separate struct?

IntegratedQuantum: For now you can put it in both places and make an issue to remove the duplicate.

tillpp: how about the way i just commited?

src/server/server.zig
tillpp: ConnectionManager is also used by client. So this doesn't feel right

src/game.zig
IntegratedQuantum: I don't like this because it has two init functions that require different deinits.

What I imagined was something like this:

```
pub const server_connection = struct {
    var manager...
    pub var conn...
    
    pub fn init(manager: ...) {...}
    pub fn deinit(...) {...}
};
pub const World = struct {
    ...
    conn: *Connection,
    
    pub fn init(...) {
         ...
        self.conn = server_connection.conn; // TODO: Rename use-cases
   }
   ...
}
```

src/network/protocols.zig
```diff
@@ -44,7 +44,6 @@ pub fn init() void { // MARK: init()
 		}
 	}
 }
-
```
IntegratedQuantum, line 47: unintended change

src/network/protocols.zig
```diff
@@ -209,16 +213,23 @@ pub const handShake = struct { // MARK: handShake
...
+			}
+			prefix = [1]u8{@intFromEnum(Connection.HandShakeState.userData)};
+			try conn.secureChannel.startTlsHandshake();
+			conn.secureChannel.finishedCollectingClientVerificationData = true;
+		}
+		if (conn.handShakeState.load(.monotonic) == .reload) {
+			prefix = [1]u8{@intFromEnum(Connection.HandShakeState.reload)};
 		}
-		const prefix = [1]u8{@intFromEnum(Connection.HandShakeState.userData)};
 		const data = zonObject.toStringEfficient(main.stackAllocator, &prefix);
```
IntegratedQuantum: Why are you sending an empty zon object with the reload packet?
I'd suggest to move the sending code into the branches above as well.

src/network/protocols.zig
```diff
@@ -209,16 +213,23 @@ pub const handShake = struct { // MARK: handShake
...
+			zonObject.putOwnedString("version", settings.version.version);
+			zonObject.putOwnedString("name", name);
+			if (main.network.authentication.KeyCollection.initialized) {
+				zonObject.put("keys", main.network.authentication.KeyCollection.getPublicKeys(main.stackAllocator));
+			}
+			prefix = [1]u8{@intFromEnum(Connection.HandShakeState.userData)};
+			try conn.secureChannel.startTlsHandshake();
+			conn.secureChannel.finishedCollectingClientVerificationData = true;
+		}
+		if (conn.handShakeState.load(.monotonic) == .reload) {
```
IntegratedQuantum: I'd prefer a switch case, so we can assert with `else => unreachable` that the other cases don't happen.

src/network/protocols.zig
```diff
@@ -1056,3 +1067,28 @@ pub const EntityComponentUpdate = struct { // MARK: EntityComponentUpdate
 		conn.send(.secure, id, writer.data.items);
 	}
 };
+
+pub const Reload = struct { // MARK: Reload
```
IntegratedQuantum: This is a namespace
```suggestion
pub const reload = struct { // MARK: reload
```

src/network/protocols.zig
```diff
@@ -156,11 +155,16 @@ pub const handShake = struct { // MARK: handShake
...
-						try conn.user.?.verifySignatures(reader);
+				.signatureResponse, .reload => {
+					if (newState != .reload) {
+						if (main.server.world.?.mode != .singleplayer) {
+							try conn.user.?.verifySignatures(reader);
+						}
+						conn.user.?.keysVerified = true;
+					} else {
+						// check if player is already logged in.
+						if (!conn.user.?.keysVerified) return error.keysNotVerified;
```
IntegratedQuantum, line 176: I think this is a bit odd. We are reloading it, just to kick the player again. I'd suggest to send extra data with the reload packet to tell the client whether it should do a reload or restart the handshake entirely.

src/server/server.zig
```diff
@@ -131,12 +131,14 @@ pub const User = struct { // MARK: User
...
 	key: network.authentication.PublicKey = undefined,
 	legacyKey: ?network.authentication.PublicKey = null,
+	keysVerified: bool = false,
 
 	inventoryClientToServerIdMap: std.AutoHashMap(InventoryId, InventoryId) = undefined,
 	inventory: ?InventoryId = null,
 	handInventory: ?InventoryId = null,
 
 	connected: Atomic(bool) = .init(true),
+	wokeup: bool = false,
```
IntegratedQuantum: I'd prefer an enum to handle both states here
```suggestion
	state: enum {awaitingKeyVerification, connected, awaitingReload} = .awaitingKeyVerification,
```
On restart .connected state goes to .awaitingReload, .awaitingKeyVerification stays where it is.

src/game.zig
```diff
@@ -757,3 +764,21 @@ pub fn update(deltaTime: f64) void { // MARK: update()
...
 	particles.ParticleSystem.update(@floatCast(deltaTime));
 }
+pub fn restart() void {
+	if (world) |_world| {
+		_world.pause();
+		std.debug.assert(_world == &testWorld);
+
+		network.protocols.Reload.informServerOfRestart(_world.conn);
+		_world.conn.handShakeState.store(.reload, .monotonic);
+		testWorld.@"continue"() catch |err| {
```
IntegratedQuantum: Why are you using testWorld here?
```suggestion
		_world.@"continue"() catch |err| {
```

src/game.zig
```diff
@@ -757,3 +764,21 @@ pub fn update(deltaTime: f64) void { // MARK: update()
 	world.?.update(deltaTime);
 	particles.ParticleSystem.update(@floatCast(deltaTime));
 }
+pub fn restart() void {
+	if (world) |_world| {
+		_world.pause();
+		std.debug.assert(_world == &testWorld);
```
IntegratedQuantum: Shouldn't be needed if you don't use `testWorld` anywhere

src/game.zig
IntegratedQuantum: Oh I see you changed it from the commit, I guess pause and continue is fine.

src/network.zig
```diff
@@ -525,13 +525,28 @@ pub const ConnectionManager = struct { // MARK: ConnectionManager
...
 		result.packetSendRequests = .initContext({});
 		result.running.store(true, .monotonic);
 		result.thread = try std.Thread.spawn(.{}, run, .{result});
 		result.thread.setName(main.io, "Network Thread") catch |err| std.log.err("Couldn't rename thread: {s}", .{@errorName(err)});
 	}
 	pub fn deinit(self: *ConnectionManager) void {
 		if (self.running.load(.monotonic)) self.pause();
+
+		for (self.connections.items) |conn| {
+			conn.@"continue"();
```
IntegratedQuantum: Why does it continue before a disconnect? This seems a little odd.

src/network.zig
```diff
@@ -1524,6 +1549,50 @@ pub const Connection = struct { // MARK: Connection
...
+			self.connectionState.store(.connected, .monotonic);
+			if (self.user) |user| {
+				user.connected.store(true, .monotonic);
+			}
+		}
+	}
+	pub fn checkRestartCounter(conn: *Connection, protocolIndex: u8, data: []const u8, channelId: ChannelId) !bool { // MARK: onReceive()
+		// Reload protocol bypasses everything else
+		std.debug.assert(@intFromEnum(channelId) < 3);
+		std.debug.assert(channelId == ChannelId.lossy or channelId == ChannelId.secure or channelId == ChannelId.slow);
```
IntegratedQuantum: These two asserts are redundant. I'd suggest to only keep the bottom one.
Also

```suggestion
		std.debug.assert(channelId == .lossy or channelId == .secure or channelId == .slow);
```

src/network.zig
```diff
@@ -1524,6 +1549,50 @@ pub const Connection = struct { // MARK: Connection
...
+	}
+	pub fn @"continue"(self: *Connection) void {
+		if (self.connectionState.load(.monotonic) == .paused) {
+			self.connectionState.store(.connected, .monotonic);
+			if (self.user) |user| {
+				user.connected.store(true, .monotonic);
+			}
+		}
+	}
+	pub fn checkRestartCounter(conn: *Connection, protocolIndex: u8, data: []const u8, channelId: ChannelId) !bool { // MARK: onReceive()
```
```suggestion
	pub fn checkRestartCounter(conn: *Connection, protocolIndex: u8, data: []const u8, channelId: ChannelId) !bool { // MARK: checkRestartCounter()
```

IntegratedQuantum (changes requested):

src/network/protocols.zig
tillpp: > We are reloading it, just to kick the player again.

An honest player should never be kicked here, only someone who is trying to cheat

IntegratedQuantum: What about someone who was in the process of joining when you restarted it?

src/network.zig
tillpp: > These two asserts are redundant.

I kind of thought about the future, just in case someone adds another channel between lossy and secure or something

IntegratedQuantum: How is a redundant check helpful in the future?

src/network.zig
tillpp: because server.disconnect expects the user to be connected to work:
grafik

tillpp: let me try to replace it with state as mentioned above

src/server/server.zig
tillpp: i assume you want it to be atomic too?
(since connected is an atomic)

or wait, maybe you didn't mean to replace connected with both ?

IntegratedQuantum: Is it accessed from multiple threads?

tillpp: yeah from ChunkLoadTask or  LightMapLoadTask .connected is being used from multiple threads

tillpp: the more i think  about this, the more i  dislike this. there would be no way for me to know if continue() was already called or not, leading to mutliple inits or deinits or variables

IntegratedQuantum: How could it even be called twice?

IntegratedQuantum: Why should we allow calling it twice? In my opinion this just looks like a mistake.

src/network/protocols.zig
tillpp: should now handle this case too

src/server/server.zig
tillpp: paused() for example is called in server.deinit()
and then again at the end at user.deinit()

Do you want me to removed paused() from deinit()?

IntegratedQuantum: Yeah, I think that would make sense. Allowing calling these functions twice is in my opinion just going to invite more issues in the future.

src/network.zig
tillpp: didn't work, all of the sub functions of .disconnect expect the connection to be connected.
since we only pretended to be disconnected we have stop pretending to be disconnected, to actually disconnect

src/server/server.zig
tillpp: ok removed

src/network.zig
tillpp: I would like to keep the continue, before the disconnect at this position, if u don't mind

IntegratedQuantum: I actually do mind. The pretending to be disconnected is weird.
I think we should implement a state machine (→an enum) here, with well-defined state transitions (→switch case). Otherwise this will turn into a mess quickly and no one will know what `connected` actually means.

tillpp: i solved it, but a bit different than what you had in mind:

https://github.com/PixelGuys/Cubyz/pull/3278/commits/21ad67a8305062e7ce30b06450f4000a736731c9

src/network/protocols.zig
```diff
@@ -207,19 +213,30 @@ pub const handShake = struct { // MARK: handShake
...
+
+				zonObject.putOwnedString("version", settings.version.version);
+				zonObject.putOwnedString("name", name);
+				if (main.network.authentication.KeyCollection.initialized) {
+					zonObject.put("keys", main.network.authentication.KeyCollection.getPublicKeys(main.stackAllocator));
+				}
+				try conn.secureChannel.startTlsHandshake();
+				conn.secureChannel.finishedCollectingClientVerificationData = true;
+
+				const prefix: [1]u8 = [1]u8{@intFromEnum(Connection.HandShakeState.userData)};
```
IntegratedQuantum:
```suggestion
				const prefix: [1]u8 = .{@intFromEnum(Connection.HandShakeState.userData)};
```

src/network/protocols.zig
```diff
@@ -207,19 +213,30 @@ pub const handShake = struct { // MARK: handShake
...
+				const data = zonObject.toStringEfficient(main.stackAllocator, &prefix);
+				defer main.stackAllocator.free(data);
+
+				conn.send(.secure, id, data);
+			},
+			.reload => {
+				conn.send(.secure, id, &[1]u8{@intFromEnum(Connection.HandShakeState.reload)});
```
IntegratedQuantum:
```suggestion
				conn.send(.secure, id, &.{@intFromEnum(Connection.HandShakeState.reload)});
```

src/network/protocols.zig
```diff
@@ -1056,3 +1073,30 @@ pub const EntityComponentUpdate = struct { // MARK: EntityComponentUpdate
...
+
+		writer.writeInt(u32, conn.restartCounter);
+		writer.writeEnum(main.server.User.State, conn.user.?.state);
+
+		conn.send(.secure, id, writer.data.items);
+		conn.send(.lossy, id, writer.data.items);
+		conn.send(.slow, id, writer.data.items);
+	}
+	pub fn informServerOfRestart(conn: *Connection) void {
+		// unlock!
```
IntegratedQuantum: I don't think this comments helps.

src/server/server.zig
```diff
@@ -154,20 +158,53 @@ pub const User = struct { // MARK: User
...
+		// persistent data
+		const conn = self.conn;
+		const name = self.name;
+		const newKeyString = self.newKeyString;
+		const playerIndex = self.playerIndex;
+		const keysVerified = self.keysVerified;
+		const state = self.state;
+
+		// reset
+		self.* = .{};
```
IntegratedQuantum: The following is legal and well-defined by the way
```zig
		self.* = .{
				.conn = self.conn,
				...
		};
```

src/server/server.zig
```diff
@@ -582,6 +609,15 @@ fn init(name: []const u8, singlePlayerPort: ?u16, mode: ServerWorld.Mode) void {
 		std.log.err("Failed to generate world: {s}", .{@errorName(err)});
 		@panic("Can't generate world.");
 	};
+
+	connectionManager.@"continue"() catch |err| {
+		std.log.err("Couldn't create thread: {s}", .{@errorName(err)});
+		@panic("Could not open Server.");
+	};
+	for (connectionManager.connections.items) |conn| {
```
IntegratedQuantum: You need to lock the mutex when accessing the connections.
~~I'd suggest to move this code into connectionManager.continue.~~

Actually why is this not just part of `User.@"continue`?

src/server/server.zig
```diff
@@ -600,19 +636,26 @@ fn deinit() void {
 	main.threadPool.clear();
 
 	users.clearAndFree();
+
+	for (connectionManager.connections.items) |conn| {
```
IntegratedQuantum: Same here

src/game.zig
```diff
@@ -282,14 +282,18 @@ pub const World = struct { // MARK: World
 	playerBiome: Atomic(*const main.server.terrain.biomes.Biome) = undefined,
 
 	pub fn init(self: *World, ip: []const u8, manager: *ConnectionManager) !void {
+		const conn = try Connection.init(manager, ip, null);
+		self.conn = conn;
```
```suggestion
		self.conn = try Connection.init(manager, ip, null);
```

src/game.zig
```diff
@@ -282,14 +282,18 @@ pub const World = struct { // MARK: World
...
+		self.manager = manager;
+		try self.@"continue"();
+	}
+	pub fn @"continue"(self: *World) !void {
 		main.heap.allocators.createWorldArena();
 		errdefer main.heap.allocators.destroyWorldArena();
-		self.* = .{
```
IntegratedQuantum, line 287: Please keep `self.* = .{...}` somewhere to catch missing initializations

src/main.zig
```diff
@@ -298,6 +298,9 @@ pub var lastFrameTime = std.atomic.Value(f64).init(0);
 pub var lastDeltaTime = std.atomic.Value(f64).init(0);
 
 var shouldExitToMenu = std.atomic.Value(bool).init(false);
+pub var shouldRestart = std.atomic.Value(bool).init(false);
+pub var shouldReload: bool = false;
```
IntegratedQuantum: I think we should avoid global variables in main, they pollute the namespace as well. I'd suggest to move them inside the world. I'd also suggest to move the logic that checks it into game.update().

src/network.zig
```diff
@@ -871,12 +886,14 @@ pub const Connection = struct { // MARK: Connection
 		buffer: main.utils.FixedSizeCircularBuffer(u8, receiveBufferSize),
 		header: ?Header = null,
 		protocolBuffer: main.List(u8) = .empty,
+		channelId: ChannelId = undefined,
```
```suggestion
		channelId: ChannelId,
```

src/network.zig
```diff
@@ -1524,6 +1548,49 @@ pub const Connection = struct { // MARK: Connection
 		main.globalAllocator.destroy(self);
 	}
 
+	// pretending the connection is closed
+	pub fn pause(self: *Connection) void {
+		if (self.connectionState.load(.monotonic) == .connected) {
+			self.connectionState.store(.paused, .monotonic);
```
IntegratedQuantum, line 1566: This is not used anywhere as far as I can tell, what do you need this for?

IntegratedQuantum (changes requested):

src/network.zig
tillpp: oh why remove the undefined?

IntegratedQuantum: because you explicitly initialize it, so the default isn't needed and could cause bugs.

src/network.zig
tillpp: if you mean the function, it is used in ConnectionManager.pause

or do you mean .paused?

IntegratedQuantum: I mean .paused

tillpp: for example:
ChunkLoadTask isStillNeeded

grafik

src/server/server.zig
tillpp: this cant be part of connectionmanager.pause because main.threadPool.clear has to run first

tillpp: no actually, i did it. it works now.

src/network.zig
```diff
@@ -1524,6 +1559,51 @@ pub const Connection = struct { // MARK: Connection
 		main.globalAllocator.destroy(self);
 	}
 
+	// pretending the connection is closed
+	pub fn pause(self: *Connection) void {
+		if (self.connectionState.load(.monotonic) == .connected) {
```
IntegratedQuantum, line 1565: Please assert that it's not already paused

src/network.zig
```diff
@@ -1524,6 +1559,51 @@ pub const Connection = struct { // MARK: Connection
...
+			self.connectionState.store(.paused, .monotonic);
+		}
+	}
+	pub fn @"continue"(self: *Connection) void {
+		if (self.connectionState.load(.monotonic) == .paused) {
```
IntegratedQuantum, line 1571: Please assert that it's not connected

src/game.zig
```diff
@@ -547,6 +562,10 @@ pub fn getBlockWithSide(comptime side: main.sync.Side, x: i32, y: i32, z: i32) ?
 }
 
 pub fn update(deltaTime: f64) void { // MARK: update()
+	if (world.?.shouldRestart.load(.monotonic)) {
```
IntegratedQuantum: Should be `.acquire`, otherwise the previous store to `shouldReload` may not be observed

src/server/server.zig
```diff
@@ -596,23 +622,21 @@ fn init(name: []const u8, singlePlayerPort: ?u16, mode: ServerWorld.Mode) void {
 }
 
 fn deinit() void {
-	connectionManager.pause();
 	main.threadPool.clear();
-
+	connectionManager.pause();
```
IntegratedQuantum, line 623: Doing this after clear adds a race condition, since new threadpool tasks may be added by connected players.
Why does player pausing need to happen after threadpool clear?

IntegratedQuantum (changes requested):

src/server/server.zig
tillpp: i dont remember all interconnections, but  user.pause  does deinit stuff, that is still needed by threads on the main.threadPool

tillpp: we could do main.threadPool.clear twice, one before and after

tillpp: or  i move it where i moved it now

IntegratedQuantum: I'm not happy with the outcome here, and I think it would be best to use a more robust mechanism, as proposed in #3321.

src/network/protocols.zig
```diff
@@ -1071,3 +1088,28 @@ pub const EntityComponentUpdate = struct { // MARK: EntityComponentUpdate
 		conn.send(.secure, id, writer.data.items);
 	}
 };
+
+pub const reload = struct { // MARK: Reload
+	pub const id: u8 = 16;
```
IntegratedQuantum, line 1093: Could you give this a more special id? I think id 0 is free, which would put this before the handshake in this file.

src/network/protocols.zig
```diff
@@ -1071,3 +1088,28 @@ pub const EntityComponentUpdate = struct { // MARK: EntityComponentUpdate
 		conn.send(.secure, id, writer.data.items);
 	}
 };
+
+pub const reload = struct { // MARK: Reload
```
IntegratedQuantum, line 1092:
```suggestion
pub const reload = struct { // MARK: reload
```

src/server/server.zig
```diff
@@ -146,6 +148,8 @@ pub const User = struct { // MARK: User
 
 	permissions: permission.Permissions = undefined,
 
+	pub const State = enum { awaitingKeyVerification, connected, awaitingReload };
```
IntegratedQuantum, line 151: Don't both of these have their keys verified? Do we really needed a separate bool for `keysVerified`?
```suggestion
	pub const State = enum { awaitingKeyVerification, connectedAndKeysVerified, awaitingReloadAndKeysVerified };
```

src/server/server.zig
```diff
@@ -154,20 +158,45 @@ pub const User = struct { // MARK: User
...
+			.newKeyString = self.newKeyString,
+			.playerIndex = self.playerIndex,
+			.keysVerified = self.keysVerified,
+			.state = self.state,
+		};
 
+		self.inventoryClientToServerIdMap = .init(main.globalAllocator.allocator);
+		self.worldEditData = .init();
+		self.permissions = .init(main.globalAllocator);
+		self.jobQueue = .init(main.globalAllocator);
```
IntegratedQuantum, line 181:
```suggestion

			.inventoryClientToServerIdMap = .init(main.globalAllocator.allocator),
			.worldEditData = .init(),
			.permissions = .init(main.globalAllocator),
			.jobQueue = .init(main.globalAllocator),
		};
```

src/server/world.zig
```diff
@@ -218,7 +218,7 @@ pub const ChunkManager = struct { // MARK: ChunkManager
 
 		pub fn isStillNeeded(self: *ChunkLoadTask) bool {
 			switch (self.source) { // Remove the task if the player disconnected
-				.user => |user| if (!user.connected.load(.monotonic)) return false,
+				.user => |user| if (!user.connected.load(.monotonic) or user.conn.connectionState.load(.monotonic) == .paused) return false,
```
IntegratedQuantum, line 221: this (whole line) shouldn't even be needed anymore, since all player tasks are removed now.

src/game.zig
```diff
@@ -314,6 +326,10 @@ pub const World = struct { // MARK: World
 		defer main.threadPool.@"continue"();
```
IntegratedQuantum, line 326: This should be done inside of the pause function, otherwise other invocations of it won't have this paused.

src/network.zig
```diff
@@ -541,17 +558,26 @@ pub const ConnectionManager = struct { // MARK: ConnectionManager
...
 		}
 		self.packetSendRequests.deinit(main.globalAllocator.allocator);
 
 		for (self.connections.items) |conn| {
-			conn.disconnect();
+			conn.pause();
+			conn.restartCounter += 1;
+			if (conn.user) |user| {
+				user.pause();
+			}
```
IntegratedQuantum, line 580: I think all of this should be done inside `conn.pause`

src/network.zig
```diff
@@ -1524,6 +1559,53 @@ pub const Connection = struct { // MARK: Connection
...
+			if (conn.restartChannelCounter[@intFromEnum(channelId)] < restartCounter) {
+				conn.restartChannelCounter[@intFromEnum(channelId)] = restartCounter;
+			}
+			return false;
+		}
+
+		// Throw away everything from before the restart
+		if (conn.restartChannelCounter[@intFromEnum(channelId)] != conn.restartCounter) return false;
+		return true;
```
IntegratedQuantum, line 1606: I would prefer to have this check in the inverse if at the start of the function.

src/network.zig
```diff
@@ -1524,6 +1559,53 @@ pub const Connection = struct { // MARK: Connection
...
+	}
+	pub fn @"continue"(self: *Connection) void {
+		std.debug.assert(self.connectionState.load(.monotonic) != .connected);
+		if (self.connectionState.load(.monotonic) == .paused) {
+			self.connectionState.store(.connected, .monotonic);
+		}
+		main.network.protocols.reload.informClientOfRestart(self);
+		self.handShakeState.store(.signatureResponse, .monotonic);
+	}
+	pub fn checkRestartCounter(conn: *Connection, protocolIndex: u8, data: []const u8, channelId: ChannelId) !bool { // MARK: checkRestartCounter()
```
IntegratedQuantum, line 1577: Should return an enum. Without inspecting the function it's not clear what the return value means.

IntegratedQuantum (changes requested):
I would also like to give you an extra test case to check:
What happens when the client times out after the reload, and it gets cleared in the `.awaitingReload` state?

PR #3281, opened 2026-06-24
Sound effects system Part 1: Assets, Stereo sound

RanPix:
requires #3367

this is mostly intended to be a base for other people to maybe continue on this

has a bit of renaming and test code, i still have to figure out what we will need

the questions i have for now: 
- do we want to be able to change the position of sounds? that would require creating some sort of audio soruce with a handle to a current playing sound which could allow for some cool things like a moving breathing of some scary monster
- how would/should we integrate the sound system with music later?
- in future, do we want to have dynamically changable special effects like pitch, reverb or distortion? so that for example when you break the same block all the time it has some pitch variation to make it less repetative

and i also would want to discuss how the ideal api for that would look like, and maybe what tools do we give to the artists

todo:
- [x] ~~doppler effect~~ - separate pr
- [x] ~~simd sound mixing?~~ - separate pr probably
- [x] ~~playing sounds limit~~ - separate pr
- [x] ~~volume control based on how many of the same sounds are playing at the same time?~~ - separate pr
- [x] find other optimizations
- [x] sound zon files should be able to specify the sound file they are using so we can reuse sounds in some cases? (could also be a separate pr if we need that)
- [x] dont process sounds that are too far / too silent
- [x] fix sounds sounding wrong
- [x] fix windows ikabod sound issue


RanPix:
also for now you can tp to  0 0 0 to hear a testing sound thats constantly playing for stereo

Wunka:
CI failed but I also don't understand if this is supposed to be reviewed?

RanPix:
i think this is ready for review, just a few things to be done before merging:
- remove testing sound/s
- fix windows sounds going insane

ikabod-kee:
https://github.com/user-attachments/assets/bedb5ec3-338c-4c35-959f-f7bb373368d8

Still broken on my side

ikabod-kee:
Could you record how it's supposed to sound on Linux?

ikabod-kee:
I bring you good news
https://github.com/user-attachments/assets/e149d0c6-7759-4813-b974-33923733dc0e

RanPix:
> I bring you good news https://github.com/user-attachments/assets/e149d0c6-7759-4813-b974-33923733dc0e

YEEEEEEESSSSSSSSSSS, thank god its just a file issue

RanPix:
ready for review

Crepestrom:
finally we can have sound effects
thanks for that

Crepestrom:
are music and sound effects seperated?
so that you can turn down sound effects while having music be loud

RanPix:
> are music and sound effects seperated? so that you can turn down sound effects while having music be loud

yes, look at the settings for audio

Crepestrom:
ah good

Crepestrom:
can you add a field along with volume that slightly randomizes the pitch/playback speed of the sound effect
or will that be a seperate pr
this helps with making sounds less repetitive

RanPix:
> can you add a field along with volume that slightly randomizes the pitch/playback speed of the sound effect
or will that be a seperate pr
this helps with making sounds less repetitive

i tried implementing this several times and all of them were horrible, this will be a separate pr

RanPix:
refer to this on sound effect system features #3364

src/gui/windows/audio.zig
```diff
@@ -39,11 +44,18 @@ fn musicFormatter(allocator: NeverFailingAllocator, value: f32) []const u8 {
 	return allocator.print("Music volume:", .{});
 }
 
+fn soundFormatter(allocator: NeverFailingAllocator, value: f32) []const u8 {
+	const percentage = 100*deziBelToLinear(value);
```
IntegratedQuantum, line 48: Please also print it

src/assets.zig
```diff
@@ -42,6 +43,8 @@ pub const Assets = struct {
 	worldPresets: ZonHashMap,
 	entityModelDescriptions: ZonHashMap,
 	entityModelMigrations: ZonHashMap,
+	audioDatas: main.List([]const u8),
```
IntegratedQuantum, line 46: I think we should load sounds dynamically (with a cache) instead of loading them upfront, sound files can get quite big

assets/cubyz/sounds/audio/block_break.ogg
IntegratedQuantum, line 1: Sound effects should be in <https://github.com/PixelGuys/Cubyz-Assets/> together with the music

src/audio.zig
```diff
@@ -286,6 +306,76 @@ pub fn setMusic(music: []const u8) void {
...
+			std.log.err("Sound Data audio could not be found: {s} ({s}) audio ID: {s}", .{id, assetsFolder, audioId});
+			return;
+		},
+		.volume = zon.get(f32, "volume") orelse 1.0,
+	});
+
+	std.log.debug("Registered sound data: {s}", .{id});
+}
+
+fn addSound(id: []const u8, pos: Vec3f, maxDistance: f32, isSpatial: bool) void {
```
IntegratedQuantum, line 359: Don't use bools as function parameters
```suggestion
fn addSound(id: []const u8, pos: Vec3f, maxDistance: f32, isSpatial: enum {spatial, ???}) void {
```

src/audio.zig
```diff
@@ -333,6 +423,76 @@ fn mixMusic(buffer: []f32) void {
...
+
+		if (sound.isSpatial) {
+			const toSound = sound.pos - playerPos;
+			const distance: f32 = vec.length(toSound);
+
+			if (distance > sound.maxDistance) {
+				sound.bufPos += @intCast(if (audioData.channelType == .mono) @divFloor(buffer.len, 2) else buffer.len);
+				if (sound.bufPos >= soundBuffer.len) {
+					soundCount -= 1;
+					activeSounds.items[i] = activeSounds.items[soundCount];
```
IntegratedQuantum, line 460: This is called swapRemove on the list

src/audio.zig
```diff
@@ -333,6 +423,76 @@ fn mixMusic(buffer: []f32) void {
...
+	const playerRight = vec.normalize(vec.cross(playerForward, Vec3f{0, 0, 1}));
+
+	var i: u32 = 0;
+	var soundCount = activeSounds.items.len;
+	main: while (i < soundCount) {
+		var sound = activeSounds.items[i];
+		const audioData = audios.items[sound.audioIndex];
+		const soundBuffer = audioData.data;
+
+		const notMonoInt: u32 = @intFromBool(audioData.channelType != .mono);
```
IntegratedQuantum, line 446: Can we forbid stereo tracks entirely? Ideally enforced through the conversion script in Cubyz-Assets.

src/audio.zig
```diff
@@ -286,6 +306,76 @@ pub fn setMusic(music: []const u8) void {
...
+		.volume = zon.get(f32, "volume") orelse 1.0,
+	});
+
+	std.log.debug("Registered sound data: {s}", .{id});
+}
+
+fn addSound(id: []const u8, pos: Vec3f, maxDistance: f32, isSpatial: bool) void {
+	const idx = soundDataIdMap.get(id) orelse return;
+	const soundData = soundDatas.items[idx];
+	activeSounds.append(main.globalAllocator, PlayingSound{
```
IntegratedQuantum, line 365: Needs to lock the mutex, otherwise there is a race condition with the other thread.

IntegratedQuantum (changes requested):

src/audio.zig
RanPix: @ikabod-kee  do we want stereo sounds? (not the directional spatial stuff in the game but the audio itself being stereo), having no stereo sounds would make things simplier

Crepestrom:
is this still being worked on?

RanPix:
> is this still being worked on?

at the moment no, but you can overtake this pr if you are willing to finish it

Crepestrom:
I am bored and willing to contribute

Crepestrom:
good job making this pr

RanPix:
moved to #3626

PR #3285, opened 2026-06-26
Stealing wunka's PR >:D | AAA Part 3

tillpp:
https://github.com/PixelGuys/Cubyz/pull/2898

TODO:
- [ ] MeshSelection.keyPress just passes it's data to sync.client.ExecuteCommand...keypress without branching
- [ ] onInteract has a left,right button parameter
- [ ] keyPress command: 
```
if item has onLeft/onRight 
     .handle => done
if not shift and block has onInteract    
     .handle => done
break/place block (same as updateblock)
```
- [ ] unite UpdateBlock with keyPress, making them one command. (i.e. updateBlock will be removed)
- [ ] make sure block type, selected item, etc. are send with the keyPressed command and checked for deviations.
  don't forget to correct client when missmatch


tillpp:
i wrote my approach into the PR description @Wunka please tell me if you have any objections with this

Wunka:
> * [ ]  keyPress command:
> 
> ```
> if item has onLeft/onRight 
>      .handle => done
> if not shift and block has onInteract    
>      .handle => done
> break/place block (same as updateblock)
> ```

so you plan on doing a function that will look if a `onInteract` will run return `.handle`?

> * [ ]  make sure block type, selected item, etc. are send with the keyPressed command and checked for deviations.
>   don't forget to correct client when missmatch

that sounds like something not currently the goal? (I mean like it will be a future goal that doesn't need to be implemented currently) or what do you exactly want to compare?

Else this sounds for now good

tillpp:
> so you plan on doing a function that will look if a onInteract will run return .handle?

yeah

> that sounds like something not currently the goal? (I mean like it will be a future goal that doesn't need to be implemented currently) or what do you exactly want to compare?

the first half is already done in your PR, the second half (the checking of missmatching) can be done in a future PR i guess,  I am describing the entire keypress process here, so wrote it down here anyway. 
 
> Else this sounds for now good

you mean "otherwise this sounds good for now"? :D

tillpp:
@Wunka basicly i want to move the work currently done in main.renderer.MeshSelection,keyPress further back in the process, in sync.client.executeCommand(.keypress

tillpp:
i am mostly asking if this is ok with you, because i am wondering why you didn't do it this way?
 Was there a reason you disliked this solution? or did you just not think about it? or some other Issue?
 
 cause the way i see it, it would fix the problem you had with this PR this way

Issue #3092, opened 2026-05-18
allow addons to define arbitrary modifier restriction by defining a pattern


IntegratedQuantum:
this means we don't need to define all possible options in code for the base game.

extracted from #2693

requires #2818 to be usable

PR #3288, opened 2026-06-27
Arbitrary modifer restriction

Crepestrom:
Solves: #3092 

adds a restriction to a property that alows the user to give a abratrary sized grid and if its a perfect square and odd be able to scan a arbitrary pattern

Eg:
0, 1, 1
0, 0, 0
1, 0, 0

if a item in the array is 1 it is checked for that tag relative to the items position
the center tile of the grid is always where the item is (grid needs to be odd sized so that it has a center tile)


Crepestrom:
Can someone help me i am not sure how but the data in checkArray is corrupted when it goes from loadFromZon to satisfied

PR #3289, opened 2026-06-27
Disabled icon shadows inside a button

### LINKED ISSUE #3201: Remove shadow from button icons
since items don't have them anymore.
Edit: Actually there is some more reasoning. The icons are usually made with the intention of looking like 3d objects so it wouldn't make sense for them to have a flat shadow. I would also say that sometimes the shadow can make the shape of the icon look different than intended.

GasSpace630:
closes #3201
Before:
rm-shdw-pr-cubyz1

rm-shdw-pr-cubyz2

After:
rm-shdw-pr-cubyz4
rm-shdw-pr-cubyz3


ghost:
should the icon shadow logic be removed also? since its not used anywhere

careeoki (approved):

IntegratedQuantum (changes requested):
> should the icon shadow logic be removed also? since its not used anywhere

Yes please. If mods do want shadows it would probably be nicer to add it through the texture.

IntegratedQuantum (approved):

PR #3297, opened 2026-06-28
particle animation looping

RanPix:
useful for cyclical animations like leafs or butterfiles


src/particles.zig
```diff
@@ -552,15 +554,16 @@ pub const ParticleTypeLocal = struct {
...
 	rotVel: RandomRange(f32),
 	dragCoefficient: RandomRange(f32),
+	loopTime: ?RandomRange(f32),
 };
 
 pub const Particle = extern struct {
 	pos: [3]f32 align(16),
 	rot: f32 = 0,
 	lifeRatio: f32 = 1,
+	loopRatio: f32 = 1,
```
IntegratedQuantum: To avoid needless data upload I'd suggest to upload only the premultiplied lifeRatio*loopRatio.
(and why not also premultiply the `particleType.animationFrames`, and call it `animationFrame` instead of `ratio`)

src/particles.zig
```diff
@@ -552,15 +554,16 @@ pub const ParticleTypeLocal = struct {
 	density: RandomRange(f32),
 	rotVel: RandomRange(f32),
 	dragCoefficient: RandomRange(f32),
+	loopTime: ?RandomRange(f32),
```
IntegratedQuantum, line 558: This is never initialized to `null`

IntegratedQuantum (changes requested):

src/particles.zig
RanPix: it is set to null if the entry for it does not exist in the zon file
https://github.com/PixelGuys/Cubyz/pull/3297/changes#diff-c6cdbadc54ecfbdde364da6be10801f1a11b59370a9d2c6b61a61b15a8ef43fdR75

src/particles.zig
RanPix: i cant premultiply `particleType.animationFrames` because it is per type and i made it so the loop time is randomized per particle

IntegratedQuantum: You can multiply it after randomization.

RanPix: wait you confused me, `animationFrames` doesnt even exist, and as i understand you refer to `particle.lifeRatio` which is not in the  `particleType`, those are different things

IntegratedQuantum: It's called animationFrames in the shader, here it seems to be called frameCount.

RanPix:
done

src/particles.zig
```diff
@@ -560,7 +565,6 @@ pub const Particle = extern struct {
 	lifeRatio: f32 = 1,
 	light: u32 = 0,
 	typ: u32,
-	// 4 bytes left for use
```
IntegratedQuantum, line 563: they are free again

src/particles.zig
```diff
@@ -306,19 +307,22 @@ pub const ParticleSystem = struct {
...
 		const rot = if (properties.randomizeRotation) random.nextFloat(&main.seed)*std.math.pi*2 else 0;
 		const rotVel = particleType.rotVel.get(&main.seed);
 		const dragCoeff = particleType.dragCoefficient.get(&main.seed);
 
+		const loopTime = lifeTime/if (particleType.loopTime) |l| l.get(&main.seed) else lifeTime;
 		particles[particleCount] = Particle{
 			.pos = @as(Vec3f, @floatCast(pos - @as(Vec3d, @floatFromInt(previousPlayerPos)))),
 			.rot = rot,
 			.typ = typ,
+			.lifeRatio = loopTime,
```
IntegratedQuantum: The name lifeRatio does not fit anymore.
I'd suggest
`.currentFrame = loopTime*frameCount`
and lifeVelocity would be
`.frameRate = 1/lifeTime*loopTime*frameCount`

IntegratedQuantum (changes requested):

RanPix:
done

src/particles.zig
```diff
@@ -404,7 +404,8 @@ pub const DirectionMode = union(enum) {
 
 pub const Emitter = struct {
 	typ: u16 = 0,
-	particleType: ParticleTypeLocal,
+	particleTypeLocal: ParticleTypeLocal,
+	particleType: ParticleType,
```
IntegratedQuantum: Wait, why does it store a copy of these rather big structs? I think this should only store the type index, the rest can be read from the array dynamically.

IntegratedQuantum (changes requested):

RanPix:
done

src/particles.zig
```diff
@@ -503,7 +507,7 @@ pub const Emitter = struct {
 
 		return Emitter{
 			.typ = typ,
-			.particleType = ParticleManager.typesLocal.items[typ],
+			.particleTypeIndex = typ,
```
IntegratedQuantum: It's already stored literally one line above

IntegratedQuantum (changes requested):

src/particles.zig
RanPix: im very smart

IntegratedQuantum (approved):

PR #3301, opened 2026-06-29
Send block updates to client when a neighbor of the block changed its rotation mode

### LINKED ISSUE #3270: Fences fail to connect properly and grass fails to break when placing too fast
- set placing / breaking speed to 1ms
- take a fence
- hold place button

Image

Image

but that doesn't seem to be happening when breaking fences
another detail: when you see a broken fence its collision seemes to be actually correct but  when re entering the world fences correct themselves. on the other hand grass stays floating in the air

IntegratedQuantum:
fixes #3270

@RanPix can you test to confirm that this also fixes the issue for you?


RanPix:
image

this doesnt seem to happen with fences anymore but i managed to get one flower to fly, but it looks like its now more rare to do with flowers

IntegratedQuantum:
the flower issue is probably just #2491

PR #3306, opened 2026-07-01
add particle bounciness

RanPix:
i will later create a pr that takes block properties into account


IntegratedQuantum (changes requested):
I think we should add a callback for the collision behavior (e.g maybe you want to spawn multiple new particle, or modify other properties, or make them disappear, I think we already talked about this somewhere).

Collision doesn't happen too often, so adding a callback wouldn't even be that big of a performance issue.

PR #3307, opened 2026-07-01
add particle friction

RanPix:


src/particles.zig
```diff
@@ -251,31 +252,51 @@ pub const ParticleSystem = struct {
...
 						v3Pos[2] = box.min[2] - hitBox.max[2] - physics.epsilon;
 					}
 				}
+
+				const friction = @exp(-particleLocal.friction*deltaTime);
+
+				particleLocal.velAndRotationVel *= Vec4f{
+					friction*@intFromBool(addFrictionX) + 1*@intFromBool(!addFrictionX),
+					friction*@intFromBool(addFrictionY) + 1*@intFromBool(!addFrictionY),
+					friction*@intFromBool(addFrictionZ) + 1*@intFromBool(!addFrictionZ),
```
Wunka: does it really need the `1*`?

Wunka (changes requested):

src/particles.zig
RanPix: oops

IntegratedQuantum:
I think we should wait with this until after #1728

PR #3298, opened 2026-06-28
Migrate grass blocks

careeoki:
Migrates the old grasses to match `grass/dew` and `grass/sky`

image

PR #3309, opened 2026-07-01
Texture pile grass vegetation

careeoki:
requires #3298 
sequel to #2522 

Uses texture pile rotation to add a short variant of all the grasses. It makes grassy fields less uniform and more natural looking while still being player-controllable.

image

Only the grasslands was updated to use the new grass. Other biomes should come after this PR.

image

Also the grass seeds recipe now only produces 3 grass seeds, since you get about twice the amount of grass vegetation you did before. And also it was an odd number so I couldn't make it exactly half.


Argmaster:
With this change i would consider doubling the stack size for grass.

du82:
I like it!

IntegratedQuantum:
Sorry for leaving you waiting for long.

I do not think this is right solution. It adds more work for the biome creator. (and having to place it twice when building also seems annoying)

We definitely need a way to avoid excessive block drops and make it random by default in flower patches and on place.

I also do want to mention #63 which is possibly going to be the solution here.

> It makes grassy fields less uniform and more natural looking while still being player-controllable.

I know it can be somewhat jarring when you get the wrong grass texture (especially in minecraft where the block can be quite off-center too), but on the other hand not being controllable also means it's something you don't need to control. And sometimes I'd say less control is better for creativity, as you focus your energy into making something cool, and not into making a uniform grass field.

careeoki:
> And sometimes I'd say less control is better for creativity, as you focus your energy into making something cool, and not into making a uniform grass field.

Yeah, you're probably right. I tend to obsess over details.

> I also do want to mention https://github.com/PixelGuys/Cubyz/issues/63 which is possibly going to be the solution here.

Yes, that would work. ...So can I have random texture variants pretty please??

PR #3313, opened 2026-07-01
Add command to automatically grow and shrink selection

Argmaster:
This pull request adds `selection adjust` command:

- `/selection adjust <limit=32>`
- `/selection adjust <mask> <limit=32>`

This commands takes currently existing selection and performs grow + shrink sequence to exactly match the size of an object. Command works well on objects flying in the air, but including ground surface makes it ineffective.
Command optionally accepts `<mask>` parameter which can be used to (in some cases) aid mentioned problem or to specialize selection.
Tbd: maybe we should enable selection to be like the one from [Axiom](https://modrinth.com/mod/axiom) mod which seems to be a more modern solution that WE.

https://github.com/user-attachments/assets/d5c87e1f-b1e8-4d86-ac4e-b1fe933ae624


Argmaster:
As a follow-up I would want to move `/pos1` `/pos2` and `/deselect` to be subcommands of `/selection` command.

Wunka:
Your converting between u32 u31 and i32 has a problem: 
```bash
server: User "Wunka@0" executed command "selection cube 4294967295"
thread 186295 panic: integer overflow
```

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+const main = @import("main");
+const Block = main.blocks.Block;
+const command = main.server.command;
+const User = main.server.User;
+const Vec3i = main.vec.Vec3i;
+
+pub const description = "Operate on selection";
+pub const usage =
+	\\/selection normalize
+	\\  Ensure pos1 is set to minimal coordinates and pos2 is set to maximal coordinates from selection.
```
IntegratedQuantum: I don't think this should exist. The player should at no point need to worry about this. Every command should behave the same with and without normalization.

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+	\\/selection cube <radius=5>
+	\\  Create a cube selection with center at current player position.
+	\\/selection shrink <limit=32>
```
IntegratedQuantum: I don't think this needs a limit?

Actually in my opinion we shouldn't even keep these commands. We shouldn't clutter the help output with useless commands.

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+const Block = main.blocks.Block;
+const command = main.server.command;
+const User = main.server.User;
+const Vec3i = main.vec.Vec3i;
+
+pub const description = "Operate on selection";
+pub const usage =
+	\\/selection normalize
+	\\  Ensure pos1 is set to minimal coordinates and pos2 is set to maximal coordinates from selection.
+	\\/selection cube <radius=5>
```
IntegratedQuantum: > cube is a very convenient way to start selection wrapping process, instead of flying around and placing positions. Just fly up to the structure and make a large cube, then adjust - magic.

Why not just tag a single block with the select wand and then adjust from there?
I don't really get this use-case to be honest.

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+
+		fn getCurrentValue(self: Self, axis: Axis, candidate: Candidate) i32 {
+			const i: usize = @intFromEnum(axis);
+			return switch (candidate) {
+				.min => self.min[i],
+				.max => self.max[i],
+			};
+		}
+
+		fn shrink(self: *Self, stage: *Stage) void {
```
IntegratedQuantum: I think part of the complexity comes from your attempt to reuse code from shrinking and growing.
But shrinking is actually trivial to implement on it's own (and with much less block queries):
```
const box = ...;
for (x) for(y) for(z) {
  if (isSolid(x, y, z)) {
   box.min = @min(box.min, x,y,z);
   box.max = @max(box.min, x,y,z);
  }
}
```

As for expanding I probably would have gone for more of a breadth-first search approach, but that also may not be as easy given the caveat you mentioned (since on every increase you'd need to add new blocks to the search)

IntegratedQuantum (changes requested):

src/server/command/worldedit/selection.zig
Argmaster: The more abstract use case is to help you get a grasp of how big is eg. 31x31x31 block box (limit for structures)

Single selection is not enough to express the intent of the player. If there is not pos2 I guess it's doable to just pos2=pos1, but after you set pos2 unsettling it is awkward (you can only clear both). It's much more reliable to always work with a full box.

src/server/command/worldedit/selection.zig
Argmaster: Yes that's how I implemented it at first. However having two separate implementations is not my favorite approach, there is inherent symmetry of the processes. If I were to separate the implementations, realistically not much would change in growing and I would just take extra code for shrinking. 1.5x line count, no gain imo.

src/server/command/worldedit/selection.zig
Argmaster: > I don't think this needs a limit?

Symmetry. Also prevents long iteration on large selections - mind O(n^2) complexity of scanning perpendicular plane. Imo better than "uh so I will just freeze for 10 minutes now and you can't cancel me hf"

> Actually in my opinion we shouldn't even keep these commands. We shouldn't clutter the help output with useless commands.

There is at least one use case I can think of - you can use shrink when object touches ground from one direction to shrink selection in 5 other. directions, you can't use adjust this way

Rest is a matter of player creativity. There could be a case where s-g gives different result than g-s and one prefers one over the other.

src/server/command/worldedit/selection.zig
Argmaster: Fair.

src/server/command/worldedit/selection.zig
IntegratedQuantum: > but after you set pos2 unsettling it is awkward (you can only clear both)

You just both right and left click with the selection wand on the same block.

In a future ideally we'd have a unique wand for this so you just click and it automatically selects every connected block to the one you clicked.

> The more abstract use case is to help you get a grasp of how big is eg. 31x31x31 block box (limit for structures)

I think it's not good at that. It's ambiguous how big the box is going to be. If I type `/selection cube 5` is the result 9×9×9 or 11×11×11 or 10×10×10 and off-center?

src/server/command/worldedit/selection.zig
IntegratedQuantum: > There could be a case where s-g gives different result than g-s and one prefers one over the other.

I'd argue we should keep things simple and easy to understand for the average user, less options is better in my opinion since it makes the documentation smaller.

What you propose here is a really special use-case, and we should not worry about supporting it.

Now I guess I can see a point where you may only want to shrink or only want to expand, so maybe it's worth keeping.
Though then I kind of wonder what the value of `adjust` is, since it's basically just a shortcut to the other commands. If we add user-defined aliases, or the wand that does it for you, this shortcut would quickly become obsolete.

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+				var kLimit: i32 = 0;
+
+				var kIterator = kRange.iter();
+				while (kIterator.next()) |k| {
+					const x, const y, const z = Self.mapCoordinates(i, j, k);
+
+					if (main.server.world.?.getBlock(x, y, z)) |block| {
+						// Finding a non-air block in shrink mode means we have to stop contracting,
+						// but in growing mode it means we can continue expading to possibly find more.
+						if (block.typ != 0) return if (mode == .shrink) .failure else .success;
```
IntegratedQuantum: You don't need mode here if you instead return whether or not it has a matching block (in this case non-air, but it might make sense to add mask support in the future)

src/server/command/worldedit/adjust.zig
```diff
@@ -0,0 +1,359 @@
...
+
+					if (main.server.world.?.getBlock(x, y, z)) |block| {
+						// Finding a non-air block in shrink mode means we have to stop contracting,
+						// but in growing mode it means we can continue expading to possibly find more.
+						if (block.typ != 0) return if (mode == .shrink) .failure else .success;
+					}
+
+					kLimit += 1;
+					// We didn't even finish scanning one JK plane, so we can't return updated I
+					if (kLimit > limit) return .limitExceeded;
```
IntegratedQuantum, line 279: This limit is weird, I think it would make more sense to support arbitrary j/k sizes, and only limit the depth `i` of the "shell"

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+	main.network.protocols.genericUpdate.sendWorldEditPos(source.conn, .selectedPos1, pos1);
+
+	source.worldEditData.selectionPosition2 = pos2;
+	main.network.protocols.genericUpdate.sendWorldEditPos(source.conn, .selectedPos2, pos2);
+}
+
+const Range = struct {
+	start: i32,
+	stop: i32,
+	step: i32,
```
IntegratedQuantum: Step is always 1, should we really have it here?

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+		return .{.start = start, .stop = stop, .step = step};
+	}
+
+	const Iterator = struct {
+		current: i32,
+		range: Range,
+
+		fn next(self: *Iterator) ?i32 {
+			if (self.current != self.range.stop) {
+				defer self.current += self.range.step;
```
IntegratedQuantum: wrapping arithmetic

src/server/command/worldedit/adjust.zig
```diff
@@ -0,0 +1,359 @@
...
+	};
+
+	pub fn iter(self: Range) Iterator {
+		return .{.current = self.start, .range = self};
+	}
+};
+
+const ScannerMode = enum { shrink, grow };
+
+fn Scanner3D(comptime mode: ScannerMode) type {
```
IntegratedQuantum, line 99: To avoid code duplication (in the compiler), I'd suggest to only pass the parameter into the functions that need it.

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+			//      ┏━━━┓    ╋━━━╋━━━    ╋━━━━━━╋
+			//      ┃   ┃    ┃ * ┃       ┃ *    ┃
+			//      ┗━━━┛    ┗━━━┻━━━    ┗━━━━━━┻
+			//
+			// Now, this code does an extra effort of altering between directions until all are saturated.
+			// This part might not be necessary, but I don't think it changes much in the design, so I did it this way.
+
+			while (true) doScan: {
+				for (&scanningSequence) |*stage| {
+					if (stage.isComplete) continue;
```
IntegratedQuantum: Is this correct? if another stage progresses, this could complete a new stage right?

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+
+					switch (mode) {
+						.shrink => self.shrink(stage),
+						.grow => self.grow(stage),
+					}
+				}
+				for (scanningSequence) |stage| if (!stage.isComplete) break :doScan;
```
IntegratedQuantum: I think it would be better to check for whether the bounding box changed.

src/server/command/worldedit/adjust.zig
```diff
@@ -0,0 +1,359 @@
...
+	return struct {
+		const Self = @This();
+
+		min: [3]i32,
+		max: [3]i32,
```
IntegratedQuantum, line 104: Why not use Vec3i for all the vectors?

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,359 @@
...
+			const iterator: [3]Axis = .{.x, .y, .z};
+		};
+		const Candidate = enum(u1) {
+			min = 0,
+			max = 1,
+			const iterator: [2]Candidate = .{.min, .max};
+		};
+		const Stage = struct {
+			axis: Axis,
+			candidate: Candidate,
```
IntegratedQuantum: Have you considered using `main.chunk.Neighbor` instead of these two new enums? it already has some useful helper functions.

IntegratedQuantum (changes requested):

src/server/command/worldedit/selection.zig
Argmaster: This is a very generic abstraction, so, yeah we should not have **it** here, it should be elsewhere, eg. in utils.
I understood the topic of the question but intentionally changed the subject.
This is based on [Python's range()](https://docs.python.org/3.15/library/functions.html#func-range) which has proven to be very useful.

src/server/command/worldedit/adjust.zig
Argmaster: This is a micro optimization that requires me to rethink and rewrite part of the code that doesn't strictly violate any guidelines and seems to work.

src/server/command/worldedit/selection.zig
Argmaster: Axis and candidate are used separately in certain contexts (lines :238 and :260). This makes the code simpler than if I had to manually dispatch 6 values. Axis is also used independently in `getRange`

src/server/command/worldedit/selection.zig
Argmaster: I do not see any issues with this code, but this could be my limitation.
I cycle through all directions, check if direction is saturated, if it is, I skip it, if it is not, I go and try shrink/grow it.
Afterwards I check If all directions are marked as completed. Rinse and repeat.
What am I missing?

src/server/command/worldedit/selection.zig
Argmaster: I still need the boolean to avoid re-running the last iteration/check (that would always fail). It queries chunk data which is relatively expensive comparing to boolean check.

src/server/command/worldedit/adjust.zig
Argmaster: Because vector index has to be comptime known (thats what ZLS says) and it is not possible to (easily) guarantee that in all use cases.

image

src/server/command/worldedit/adjust.zig
Argmaster: I swear I explained this somewhere.
So, basically making j/k iteration infinite would (potentially) result in lag on large selections with no way to interrupt/cancel the command. This is inconvenient and highly undesired behavior as it forces you to restart the game. Therefore limit with sane default is used for every axis. Its not perfect solution, but IMO it maximizes gain with minimal number of knobs to configure.
One thing I could have included is an message indicating that the limit has been exceeded, so user knows why the expansion stopped.

src/server/command/worldedit/selection.zig
IntegratedQuantum: Neighbor is designed to handle cases where you need this information separately. This is handled through `vectorComponent` and `isPositive` helper functions

src/server/command/worldedit/adjust.zig
IntegratedQuantum: Maybe we should add a helper to vec.zig for getting a value by index then.
Once Zig provides glsl-style vectors this restriction should be dropped hopefully.

src/server/command/worldedit/adjust.zig
IntegratedQuantum: I still think it's unintuitive. I would prefer a check that either limits the depths of the expansion shell, or that limits the side length of the result (and input) bounding boxes.

src/server/command/worldedit/adjust.zig
IntegratedQuantum: I thought it was affecting more functions, but it only affects some small helpers, so it's probably fine.

src/server/command/worldedit/selection.zig
IntegratedQuantum: It's easy to reproduce:
Screenshot at 2026-07-27 17-06-05
Screenshot at 2026-07-27 17-06-11
You stop iterating towards the right, but your selection expands upwards revealing a new connection

src/server/command/worldedit/selection.zig
IntegratedQuantum: Together with the mention above you need to re-run anyways

src/server/command/worldedit/selection.zig
IntegratedQuantum: Then I think you should either put it there, or only give it the functions needed by this file.

Generally I think the Zig way is to just use a while loop in these cases, so it's probably not going to be used much outside. There also is no such thing in the standard library either.

src/server/command/worldedit/adjust.zig
Argmaster: Prior contradicts the point of doing the latter. Additionally even if I add `.get(x)` its nowhere close to convenience of `[x]`. Refactor issue seems a better fit IMO.

src/server/command/worldedit/selection.zig
Argmaster: Hm true, I missed that. Then bbox check it is.

src/server/command/worldedit/selection.zig
Argmaster: Ok, now that I revisit this after a month, I come to a conclusion I will just get rid of grow and shrink for now unless they are requested by community, they are simply not  that useful. They don't actually solve the issue of selecting things that are connected to each other, at least not well. This also means making it a subcommand no longer makes sense, so I will probably rename it to something like `/adjust`?

src/server/command/worldedit/adjust.zig
IntegratedQuantum: Then please make one

src/server/command/worldedit/selection.zig
```diff
@@ -0,0 +1,336 @@
...
+		return .{.start = start, .stop = stop, .step = step};
+	}
+
+	const Iterator = struct {
+		current: i32,
+		range: Range,
+
+		fn next(self: *Iterator) ?i32 {
+			if (self.current != self.range.stop) {
+				defer self.current +|= self.range.step;
```
IntegratedQuantum: You used saturating arithmetic, not wrapping arithmetic.
```suggestion
				defer self.current +%= self.range.step;
```

Argmaster: Wrapping will invert the selection, no? If I have p1 and p2 and wrap p1 to the other edge of the world, it will select almost whole world.

IntegratedQuantum: Ideally it should also wrap the selection box, to the smaller value. So that a box from integer limit - 1 to integer limit + 2 would be exactly 3 blocks long.

src/server/command/worldedit/selection.zig
Argmaster: Ok, I will remove the step, as this code is not sufficiently generic to be placed in utils (I would have to make the types parametrizeable).
Tho I don't think there is a clean way to remove use of Range completely, thus I won't try it.

src/server/command/worldedit/adjust.zig
Argmaster: https://github.com/PixelGuys/Cubyz/issues/3453

src/server/command/worldedit/selection.zig
Argmaster: Changed, no guarantees its bug free.

src/server/command/worldedit/selection.zig
Argmaster: Changed, seems to be working.

src/server/command/worldedit/adjust.zig
Argmaster: Ok, how about I add time based timeout?

src/server/command/worldedit/selection.zig
Argmaster: I still need to invert the mask between shrink and grow, since the condition is inverted, so its easier to just return inverted result. I will change the names tho, to `stopScan` and `continueScan` and I will add the mask support. Masks are quite good at selecting things that touch.

src/server/command/worldedit/adjust.zig
IntegratedQuantum: Generally I think timers are more tricky to implement (if you check it in every iteration, you end up with a lot of overhead from the timer itself, if you don't then it's inaccurate).
Furthermore timers are system and load dependent, the exact same command on the exact same volume called two times in a row, may fail once and succeed once.

I think selection size based check is better (and we'll need this anyways, once we add a server-side limit to the total selection size for all world edit commands)

src/server/command/worldedit/selection.zig
IntegratedQuantum: I'd prefer if this was a helper function, but at the end of the day, this is a command and we shouldn't waste too much time bikeshedding on the fine details, so I'd be fine with your way.

Argmaster:
Eh, I was experimenting with copilot on my fork and clearly it got ahead of itself committing changes without approval, I will clean this up soon

Edit:

AI slop removed.

src/server/command/worldedit/adjust.zig
Argmaster: I changed the limit to check ahead of time if current selection is bigger than limit, removed the iteration limit checks for perpendicular axes to shrink search. Depth check was preserved, so grow bounds are min-limit, max+limit

PR #3315, opened 2026-07-02
Add giant oak landmark

Argmaster:
This pull request adds a giant broken oak landmark.

Due to complexities coming from placement on slopes and intersection with other structures as of now the landmark is fine tuned to only appear in fairly flat biomes, only in forest clearings.
 
Screenshot 2026-07-02 141450

Screenshot 2026-07-02 141653


Argmaster:
I am not happy with how the roots and trunk look. i will have to re-do it.

PR #3317, opened 2026-07-04
Allow swords to break plants and leaves (but not collect them)

### LINKED ISSUE #3303: Allow swords to break plants and leaves (but not collect them)
As discussed here: https://github.com/PixelGuys/Cubyz/pull/3218#discussion_r3485817241

DellieDelta:
Fixes #3303


careeoki (approved):
Yeah looks good

IntegratedQuantum (approved):

PR #3330, opened 2026-07-05
Extend and rebalance torch brightness tiers

Argmaster:
This pull request tries to handle issue #1419 through following means:
- increasing brightness of sulfur torch 
- decreasing brightness of coal torch
- adding grass torch item that can be crafted without going underground hence completely removing luck factor from initial phase of caving

Grass torch has very low brightness compared to other torch types, with yellow tint and inefficient crafting recipe to discourage use after other resources are available.
Grass torch does not have a proper texture, right now it shares texture with coal torch. I am not sure if we should change coal torch or add new texture for grass torch, but imo both options are viable. If we end up going this route (actually adding grass torch) I hope for @careeoki help with textures <3
Rationale for coal and sulfur torch changes is that it makes sulfur useful as a higher tier light source.
Please check how the changes feel and share your feedback.

image

Picture 1: Grass torches are craftable with wood and one of vegetation items.

Screenshot 2026-07-05 183430

Picture 2: Torch progression, left to right: grass, coal, sulfur, lumi


careeoki:
Yeah, this seems like a reasonable way to solve the problem. Might make me actually want to mine sulfur.
I experienced the "no coal in exposed caves" thing last time I played so the grass torch would be nice.

> I am not sure if we should change coal torch or add new texture for grass torch, but imo both options are viable

I would change the grass torch, maybe give it a unique model so it's extra rustic looking. 
I think we could consider a different name, so it could possibly be crafted with other things as well. I thought of "tinder torch" but then I ruined it for myself when I remembered the dating app...

Argmaster:
Primitive torch?

Crepestrom:
could call it crude torch

careeoki:
Crude torch would work.

ikabod-kee:
I think this hurts micro-progressions a bit, since now the "Pre-Coal" age (in the Design Principles) can be ignored by just getting some nearby grass and some wood, which is in every spawn biome.

> Read this in the nerdiest voice, it'll be funny.

ikabod-kee:
Although the idea of having self-imposed biome challenges would be interesting for more experienced players who want a harder game.

careeoki:
> I think this hurts micro-progressions a bit, since now the "Pre-Coal" age (in the Design Principles) can be ignored by just getting some nearby grass and some wood, which is in every spawn biome.

Well, you need to craft a sickle to gather the grass, so it's not like you just instantly get torches. You have to interact with other game mechanics first.

And if we do want "pre-coal" to be a real thing, we should lean into it further, and push coal a bit deeper. The docs say "coal spawns low in the world", which is not true, it can appear up to +500 Z. 
Right now "pre-coal" lasts at most, like 3 minutes. Unless you're really unlucky and have to wander around for exposed coal (which is not necessarily fun)

ikabod-kee:
I guess that's true.

careeoki:
@IntegratedQuantum Your input here would be helpful for the game design aspects

TyrrelMeter:
I don't think coal should be a resource put to deep underground. Stone+Wood tools are pretty painfully slow and having to wait to use the copper is annoying.

ikabod-kee:
Crude torches could burn out.

codemob-dev:
> Crude torches could burn out.

No, if players want to build with them their builds shouldn't be temporary.

IntegratedQuantum:
I like the idea of a pre-coal phase with crude torches. I think it could be fun to have to enter your first cave with a very limited light source. Even just putting it 50 blocks underground could be enough to make it inaccessible without making crude torches first.

However I do worry that the recipe with grass would be difficult to discover naturally, since it requires multiple steps, and it's also not quite obvious why freshly cut grass would burn.

---

As for torch progression in general, I think the colored torches should be primarily decorative blocks and not part of the torch progression. They should be usable as light source, but not better. I personally wouldn't enjoy caving with colored light sources.

IntegratedQuantum:
Or instead of having the individual colored torches lumiflora/sulfur be brighter, maybe you can combine their materials with a green component to create the actual brightest torch.

careeoki:
> They should be usable as light source, but not better. I personally wouldn't enjoy caving with colored light sources.

I was thinking of that this morning, and yeah, I agree.

> However I do worry that the recipe with grass would be difficult to discover naturally, since it requires multiple steps, and it's also not quite obvious why freshly cut grass would burn.

Something involving branches could work, since you naturally end up with a bunch of them when you chop trees (and they are more crude than planks.) 
But it would probably need one more item for the recipe to make some sense. We could do something based on friction (rubbing sticks together) or have it use a firestriker (a hard rock like flint)

DellieDelta:
Resin would also make sense, but then we'd need an easy way to collect it

Another direction could be finding it. So things like glowing moss, fireflies, glowing mushrooms, which you can use like a torch

Argmaster:
Okay, what if we can make a small fire place with two sticks? Or planks and stick?
I hope players will figure out that this small fire place was indented to be used as torch...

Although that would be a crafting recipe that is just *always there*. We need some improvements in crafting inventory, I can't imagine scrolling past every combination of plank and branch :P

Argmaster:
> As for torch progression in general, I think the colored torches should be primarily decorative blocks and not part of the torch progression. They should be usable as light source, but not better. I personally wouldn't enjoy caving with colored light sources.

The default torch is colored too - its yellow-ish, not white. I think its just a mentality shift, nothing more. The side effect is that players are encouraged to view the world differently (ores look differently in different light colors) and to explore the features of the engine, instead of defaulting to the normal torch. We should make the ores recognizable regardless of the color, through shape, to maximize accessibility.

As for the crafting recipe, we could make it used plank + soil, since these materials are naturally obtained at the start of progression, but don't necessarily linger around your inventory 100% of the time, hence not bloating your crafting inventory (Realism comes after gameplay). There is also [Swedish torch](https://en.wikipedia.org/wiki/Swedish_torch), basically a campfire.

calebstenberg:
For end game caving I just use glow crystals, so I go straight from torch to whatever color glow crystal I can find. I don't think I would even bother with sulfur unless it was better than glow crystals, I have also never found lumi flowers

IntegratedQuantum:
> The default torch is colored too - its yellow-ish

It's yellow-ish, not yellow. If the sulfur torch was blue-ish and the lumi torch was pink-ish I wouldn't be complaining about this.

PR #3331, opened 2026-07-05
Move GuiWindows to the mods folder

### LINKED ISSUE #1537: [MODS] Load UI windows from mods

Wunka:
Fixes #1537


Wunka:
High Priority so there are no changes to any GuiWindow before this gets merged

IntegratedQuantum (changes requested):
Ideally mods shouldn't be allowed to use `c` stuff. Please make a separate PR that removes `c` from all windows.

Wunka:
> Ideally mods shouldn't be allowed to use `c` stuff. Please make a separate PR that removes `c` from all windows.

How else? Should we provide wrapper functions for those needed in `graphics.zig`? 
From https://github.com/PixelGuys/Cubyz/pull/3039#discussion_r3202244663 I understand you wouldn't like it if I just made `c` `pub` in `graphics.zig`

IntegratedQuantum:
> Should we provide wrapper functions for those needed in graphics.zig?

Ideally yes.
I think the only thing that it's needed for it to draw a line graph, and this should probably be turned into a function in graphics.draw.

Issue #2588, opened 2026-02-23
Send the whole permission tree to the user


Wunka:
### Description

This enables #2414 
Split of from #2530

PR #3332, opened 2026-07-05
Permission groups user changes

Wunka:
Continues from #2587


IntegratedQuantum (changes requested):
I think permissions and permission groups should be stored in a component of the ECS. This will also make it easier to share them with the client (#2588).

Wunka:
> I think permissions and permission groups should be stored in a component of the ECS. This will also make it easier to share them with the client (#2588).

Just wanted to note. The first pr for that is created #3356

Wunka:
Switched to ECS

src/entityComponent/permissions.zig
```diff
@@ -28,10 +28,18 @@ pub const client = struct {
...
+		permissionGroups: std.StringHashMapUnmanaged(*main.server.permission.Group),
 
 		pub fn save(self: Component, writer: *BinaryWriter, audience: main.entity.AudienceInfo) main.entity.ComponentSaveBehaviour {
 			if (audience != .disk) return .discard;
 			self.permissions.toBytes(writer);
+
+			writer.writeInt(u32, self.permissionGroups.count());
+			var it = self.permissionGroups.iterator();
+			while (it.next()) |entry| {
+				writer.writeSliceWithSize(entry.key_ptr.*);
```
IntegratedQuantum: Why is it storing the string? Isn't it supposed to access it by index?
I think we do maybe want to allow changing the name too.

src/entityComponent/permissions.zig
```diff
@@ -28,10 +28,18 @@ pub const client = struct {
 pub const server = struct {
 	pub const Component = struct {
 		permissions: main.server.permission.Permissions,
+		permissionGroups: std.StringHashMapUnmanaged(*main.server.permission.Group),
```
IntegratedQuantum: I think the hashmap from permission.zig should be used instead when needing to access by name.
Most accesses will want to iterate over groups to check if the user has permission, and iteration over hashmap is quite slow.

IntegratedQuantum (changes requested):

Wunka:
Okay changed to use ids. one nice thing is that I also don't save the permission Group pointer anymore, should hopefully make bugs in the future less likely

src/server/permission.zig
```diff
@@ -268,11 +268,21 @@ pub fn createGroup(name: []const u8) error{AlreadyExists}!void {
...
 	sync.threadContext.assertCorrectContext(.server);
 	return groups.items[groupNameToIdMap.get(name) orelse return error.GroupNotFound].?;
 }
 
+pub fn getGroupById(id: u32) error{GroupNotFound}!*Group {
+	sync.threadContext.assertCorrectContext(.server);
+	return groups.items[id] orelse error.GroupNotFound;
+}
+
+pub fn getGroupIdByName(name: []const u8) error{GroupNotFound}!u32 {
```
IntegratedQuantum, line 281: Or maybe we should make the Group be accessible directly through the id by making a wrapper as we've done e.g. for blocks, items, entities? Then we wouldn't expose the `Group*` at all externally (which also avoids the need to allocate them separately).

Having 3 different ways to use and handle groups is too much in my opinion.

src/entityComponent/permissions.zig
```diff
@@ -68,20 +86,46 @@ pub const server = struct {
 		return (getPermissions(entity) orelse return false).removePermission(listType, permissionPath);
 	}
 
+	pub fn addToGroupByName(entity: Entity, groupName: []const u8) error{GroupNotFound}!void {
```
IntegratedQuantum, line 89: Do we need both name and id functions here?
Wouldn't it be better to allow passing a Group* instead, so that the parent needs to do the validation?8

IntegratedQuantum (changes requested):

src/entityComponent/permissions.zig
```diff
@@ -28,10 +28,17 @@ pub const client = struct {
 pub const server = struct {
 	pub const Component = struct {
 		permissions: main.server.permission.Permissions,
+		permissionGroups: std.AutoHashMapUnmanaged(main.server.permission.Group, void),
 
 		pub fn save(self: Component, writer: *BinaryWriter, audience: main.entity.AudienceInfo) main.entity.ComponentSaveBehaviour {
 			if (audience != .disk) return .discard;
 			self.permissions.toBytes(writer);
+
+			writer.writeInt(u32, self.permissionGroups.count());
```
IntegratedQuantum: should use varint

src/entityComponent/permissions.zig
```diff
@@ -68,20 +85,39 @@ pub const server = struct {
...
 	pub fn loadFromData(entity: Entity, reader: *BinaryReader, version: u32) main.entity.EntityComponentLoadError!void {
 		if (version != entityComponentVersion) return error.InvalidComponentVersion;
-		const permissions = &components.add(main.globalAllocator, entity).permissions;
-		permissions.* = .init(main.globalAllocator);
-		permissions.fromBytes(reader) catch return error.UnreadableComponentData;
+		const component = components.add(main.globalAllocator, entity);
+		component.permissions = .init(main.globalAllocator);
+		component.permissions.fromBytes(reader) catch return error.UnreadableComponentData;
+		component.permissionGroups = .empty;
+		const len = reader.readInt(u32) catch return;
```
IntegratedQuantum: varint

src/server/permission.zig
```diff
@@ -196,35 +196,81 @@ pub const Group = struct { // MARK: Group
...
+pub const Group = enum(u32) { // MARK: Group
+	_,
 
-pub fn deinit() void {
-	sync.threadContext.assertCorrectContext(.server);
-	groupsArena.deinit();
-	groups = undefined;
-	groupNameToIdMap = .{};
-}
+	pub fn addGroupFromBin(self: Group, data: []const u8) void {
```
IntegratedQuantum: Why is this even public? I think it's just a helper for `loadGroups`, so maybe it should also not be in here then?

src/server/permission.zig
```diff
@@ -257,33 +303,18 @@ fn saveMetaData(allocator: NeverFailingAllocator) !void {
...
-	const result = groupNameToIdMap.getOrPut(groupsArena.allocator().allocator, name) catch unreachable;
-	if (result.found_existing) return error.AlreadyExists;
-
-	result.key_ptr.* = groupsArena.allocator().dupe(u8, name);
-	result.value_ptr.* = @truncate(groups.items.len);
-	groups.append(Group.init(groupsArena.allocator(), result.key_ptr.*));
-}
-
-pub fn getGroup(name: []const u8) error{GroupNotFound}!*Group {
+pub fn init(allocator: NeverFailingAllocator) void {
```
IntegratedQuantum: I think init/deinit should be above the other global functions, as it was before

src/server/permission.zig
```diff
@@ -196,35 +196,81 @@ pub const Group = struct { // MARK: Group
...
+		const group = try reader.readEnum(Group);
+		if (groups.items[@intFromEnum(group)] == null) return error.GroupNotFound;
+		return group;
+	}
+
+	pub fn toBytes(self: Group, writer: *main.utils.BinaryWriter) void {
+		writer.writeEnum(main.server.permission.Group, self);
+	}
+
+	pub fn getGroupByName(name: []const u8) error{GroupNotFound}!Group {
```
```suggestion
	pub fn getByName(name: []const u8) error{GroupNotFound}!Group {
```

src/server/permission.zig
```diff
@@ -196,35 +196,81 @@ pub const Group = struct { // MARK: Group
...
+		sync.threadContext.assertCorrectContext(.server);
+		return groupNameToIdMap.get(name) orelse error.GroupNotFound;
+	}
+
+	fn getGroupInstance(self: Group) *GroupInstance {
```
```suggestion
	fn getInstance(self: Group) *GroupInstance {
```

src/server/permission.zig
```diff
@@ -196,35 +196,81 @@ pub const Group = struct { // MARK: Group
...
+		return groups.items[@intFromEnum(self)].?;
+	}
+
+	pub fn deleteGroup(self: Group, allocator: NeverFailingAllocator) bool {
```
```suggestion
	pub fn delete(self: Group, allocator: NeverFailingAllocator) bool {
```

src/server/permission.zig
```diff
@@ -196,35 +196,81 @@ pub const Group = struct { // MARK: Group
...
+		sync.threadContext.assertCorrectContext(.server);
+		std.debug.assert(groupNameToIdMap.remove(self.getGroupInstance().name));
+		groups.items[@intFromEnum(self)] = null;
```
IntegratedQuantum, line 248: Actually, now that I think about this: What happens to the existing instances of this group?
Maybe we do need to switch to error checking for all functions.

IntegratedQuantum (changes requested):

src/server/permission.zig
Wunka: I don't quite understand this comment. The rest is done. But this remains

Wunka:
While updating the permission group command pr #3333 I noticed some bugs. Fixed.

src/server/permission.zig
IntegratedQuantum: When we delete a group, an online player may still have a reference to it which would then trigger the assert on .? in Group.getInstance().

Wunka: ah, So we should not do the .? Okay will do

Wunka: Done. It now uses an error. tested again with #3333

src/server/permission.zig
```diff
@@ -196,11 +193,94 @@ pub const Group = struct { // MARK: Group
...
+	pub fn toBytes(self: Group, writer: *main.utils.BinaryWriter) void {
+		writer.writeEnum(main.server.permission.Group, self);
+	}
+
+	pub fn getByName(name: []const u8) error{GroupNotFound}!Group {
+		sync.threadContext.assertCorrectContext(.server);
+		return groupNameToIdMap.get(name) orelse error.GroupNotFound;
+	}
+
+	pub fn getById(id: u32) error{GroupNotFound}!Group {
```
IntegratedQuantum: Where is this needed? Every in-code storage should use `Group` instead of `u32`, and every loading from binary should use `fromBytes`.

src/server/permission.zig
```diff
@@ -196,11 +193,94 @@ pub const Group = struct { // MARK: Group
...
+		if (id >= groups.items.len) return error.GroupNotFound;
+		if (groups.items[id] == null) return error.GroupNotFound;
+		return @enumFromInt(id);
+	}
+
+	fn getInstance(self: Group) error{GroupNotFound}!*GroupInstance {
+		return groups.items[@intFromEnum(self)] orelse return error.GroupNotFound;
+	}
+
+	pub fn delete(self: Group) bool {
```
IntegratedQuantum, line 243: The return value is always true, also it is quite unclear what it was supposed to mean (should have been an enum)

src/server/permission.zig
```diff
@@ -196,11 +193,94 @@ pub const Group = struct { // MARK: Group
...
+		return (try self.getInstance()).removePermission(allocator, self, listType, permissionPath);
+	}
+
+	pub fn hasPermission(self: Group, permissionPath: []const u8) error{GroupNotFound}!Permissions.PermissionResult {
+		return (try self.getInstance()).hasPermission(permissionPath);
+	}
+
+	pub fn format(self: Group, writer: *std.Io.Writer) std.Io.Writer.Error!void {
+		try writer.print("{s}", .{blk: {
+			break :blk (self.getInstance() catch break :blk "[Deleted]").name;
```
IntegratedQuantum, line 273: I think this should forward the error to the caller, which can then decide what to display.

src/server/permission.zig
```diff
@@ -444,24 +497,23 @@ test "permissionGroupToFromBytes" {
...
 
-	group.addPermission(main.heap.testingAllocator, .white, "/command/test");
-	group.addPermission(main.heap.testingAllocator, .white, "/command/spawn");
+	try group.addPermission(main.heap.testingAllocator, .white, "/command/test");
+	try group.addPermission(main.heap.testingAllocator, .white, "/command/spawn");
 
 	var writer: main.utils.BinaryWriter = .init(main.heap.testingAllocator);
 	defer writer.deinit();
-	group.toBytes(&writer);
+	(try group.getInstance()).toBytes(&writer);
```
IntegratedQuantum, line 510: Please also add a negative test, for failure of getInstance

IntegratedQuantum (changes requested):

src/server/permission.zig
Wunka: This is needed by the group command. As you should also be able to rename groups, you also need to a way to access them directly via id. But  I can also remove this and wait for the group command pr

src/server/permission.zig
Wunka: oh, yes. I did something wrong. I fixed that and made a doc comment. this works like the other remove functions in the std of zig. false means that nothing was removed and true that it was removed. Enum would also be possible but I would keep it as bool as it is like the std functions

src/server/permission.zig
Wunka: this is called by the std, I can't add another error. Or should I return `Writer.Error`?

src/server/permission.zig
Wunka: done

src/server/permission.zig
IntegratedQuantum: yes, please.
Also shouldn't players be able to access groups exclusively by name?
By using a hashmap we already require each group to have a different name anyways.

src/server/permission.zig
IntegratedQuantum: Oh right, I guess it's fine then.

src/server/permission.zig
Wunka: Removed.
And to answer your question: probably? I am not completly sure. I will probably just remove the id functionality from the group pr until someone needs it. it does make it more complicated, so its easier without.

IntegratedQuantum (approved):

PR #3334, opened 2026-07-06
Make swords effective against cloth

DellieDelta:
Related to #3303

### DIFF
--- a/assets/cubyz/blocks/cloth/_defaults.zig.zon
+++ b/assets/cubyz/blocks/cloth/_defaults.zig.zon
@@ -1,5 +1,5 @@
 .{
-	.tags = .{.cloth, .cuttable},
+	.tags = .{.cloth, .cuttable, .sliceable},
 	.blockHealth = 2,
 	.drops = .{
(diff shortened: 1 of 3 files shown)


IntegratedQuantum:
I'm not sure about this, I can see the arguments for plants, where the sword is like a machete, but I'm not sure if we can find a similar explanation for breaking cloth.

TyrrelMeter:
Maybe if you make the sword look like scissors it can cut cloth.

careeoki (changes requested):
> I'm not sure about this, I can see the arguments for plants, where the sword is like a machete, but I'm not sure if we can find a similar explanation for breaking cloth.

Yes, I would only want them to break plants.

ikabod-kee:
Just because a blade can cut cloth in real life doesn't mean it should in Cubyz.

gameplay > reality

DellieDelta:
Alright, I also wasn't sure if consistency with the sickle was better, or that swords should only be special in that they can cut foliage. I'll close this then

PR #3335, opened 2026-07-06
Uniform buffer

IntegratedQuantum:
The ultimate goal here is to have one global UBO that basically contains all the stuff you may need in most shaders, this PR aims to start with player position and matrices.

progress towards #102 (this is important for Vulkan, since we cannot store everything in push constants)

- [x] Should I use a #include for this?
- [x] ~~What Pipeline changes are needed for Vulkan to recognize the bindings?~~ Currently blocked since shaders don't compile for unrelated reasons


IntegratedQuantum:
@codemob-dev as a heads-up for #3133: This changes how the matrices and positions are passed to the shader, it now uses a uniform buffer that's constant for the entire frame.
To change it in the middle of the frame you can use a `graphics.frame_uniforms.StaticUbo`

PR #3340, opened 2026-07-08
Commands made mod-bound

MinikTombikZimik:
Commands are now bound by mod instead of hardcoded
Added a mod feature type called commands, moved the commands into mods/cubyz/commands/***.zig
Commands in source are now have their full names ex: "undo" is "cubyz:worldedit/undo" but in game and permissions it's still "undo"
Latter commands with the same names will overwrite the former since you call them based on their bare names, overwrition will log a warning. ex. "cubyz:tp" will overwrite "amod:tp"
also my first pull request yay ignore gitignore


MinikTombikZimik:
Sorry forgot about the naming scheme, gonna fix it

Wunka:
> Sorry forgot about the naming scheme, gonna fix it and try opening another one

you can also just push to this one more changes. They will be updated here. (I think the folder name fits. I would rename the variables in `commands.zig` to for example: `commandList` -> `commands` and `commands` -> `commandMap`

MinikTombikZimik:
yeah did commandList -> commands and commands -> registeredCommands

PR #3341, opened 2026-07-08
Made Commands Mod Bound

MinikTombikZimik:
Commands are now bound by mod instead of hardcoded
Added a mod feature type called commands, moved the commands into mods/cubyz/commands/***.zig
Commands in source are now have their full names ex: "undo" is "cubyz:worldedit/undo" but in game and permissions it's still "undo"
Latter commands with the same names will overwrite the former since you call them based on their bare names, overwrition will log a warning. ex. "cubyz:tp" will overwrite "amod:tp"
sorry for dupe


.gitignore
```diff
@@ -19,3 +19,5 @@ test.png
 !/assets/backgrounds/default_background.png
 /assets/cubyz/music
 /assets/cubyz/fonts
+.idea/editor.xml
+/.idea
```
SpellDigger: why exacly are we modifying the gitignore this is not getting past quantum

SpellDigger (changes requested):

.gitignore
Wunka, line 1: please don't edit the `.gitignore` file.

Wunka (changes requested):

.gitignore
Wunka: Funny. Reviewed at the same time 😆

SpellDigger: i was infact 2 minutes faster grrr

MinikTombikZimik:
sorry guys ts my first pull request 
did you guys make the changes or do i gotta edit it?

MinikTombikZimik:
oh got it

.gitignore
Argmaster: Because it's redundant. There are infinitely many fancy configuration files you can add to gitignore and we don't want this file to grow past what's specific to the project. Instead you can add global gitignore file in your home directory and handle your favourite code editor this way.

Argmaster:
> sorry guys ts my first pull request 
> did you guys make the changes or do i gotta edit it?

You have to edit your own pull request. 
The point of pull request is to contribute your work time towards open source project and to do that effectively you have to fix what's wrong with your changes. Otherwise you would only deliver extra work for maintainers.

.gitignore
SpellDigger: you have a horrible sense of humor

Argmaster: Well, it's GitHub, not a comedy club. Although sometimes I'm having doubts.

SpellDigger: well if what i said sounds too much of a question to you then i guess thats on me

Argmaster: Mh I might have misred the original comment. My bad.

Wunka (approved):
There is something with the "overwritting" where I am not sure if we want to do this that way (see #3125 for why) but else it looks good. 

For your understanding this PR will now be but on the project board where Quantum can see and review it, so its now a waiting game.

IntegratedQuantum:
You went straight for the hardest one.

> Commands in source are now have their full names ex: "undo" is "cubyz:worldedit/undo" but in game and permissions it's still "undo"

Screenshot at 2026-07-08 17-42-29

SpellDigger (approved):

Argmaster:
I see a functional issue with this - addon load order is arbitrary, so one has no control over which command they get. Even if the order was defined, eg. alphabetical, still we lack a tool to define which command we get. So overwriting commands should not be a thing.
IMO we should always allow invoking command through the full ID as a reliable fallback.
Again IMO the bare minimum would be to auto generate short aliases (last segment of ID) only for commands with no conflict and add `/alias foo=@spam:bar/dog` meta-command (one that is defined as special behavior in command line, rather than normal command) that can be used to manually assign aliases.

MinikTombikZimik:
oh yeah that's a good idea, i'll try to do that

MinikTombikZimik:
Btw still not ready cause
1- I gotta add persistence, user made aliases should be saved/read into/in server
2- Usages should refer to commands with their whole name
3- Command uses should refer to the command by how they are actually used in that case
Like /tp <playerindex> <x> <y> <z> should be told to you replacing tp by whatever alias/actual name you used to call cubyz:tp

PR #3343, opened 2026-07-08
Print help and refer to launchConfig.zon on any command line argument

### LINKED ISSUE #3284: On any command line argument the user should be refered to the `launchconfig.zon`

yel0h:
This adds a check in `main()` where if any argument beyond the executable path is present, it prints a short message explaining that it takes no CLI arguments, lists the fields supported by `launchConfig.zon`, and exits before any further initialization instead of proceeding to launch.

Closes #3284

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -335,22 +335,17 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
 		std.log.warn("Cubyz detected it's running on Windows. For optimal performance and reduced power usage please install Linux.", .{});
 	}
 
-	{
+	argCheck: {
 		var argIterator = args.args.iterateAllocator(stackAllocator.allocator) catch |err| {
 			std.log.err("Failed to read command line arguments: {s}", .{@errorName(err)});
-			@panic("Failed to read command line arguments");
+			break :argCheck;
 		};
 		defer argIterator.deinit();
 		_ = argIterator.skip();
 		if (argIterator.next() != null) {
 			std.debug.print(
 				\\Cubyz does not accept any command line arguments.
-				\\All launch-time configuration is done through the "launchConfig.zon" file in the game's working directory. Supported fields:
-				\\  cubyzDir - Directory used to store game data and world saves.
-				\\  autoEnterWorld - Name of a world to automatically enter on startup.
-				\\  headlessServer - Set to true to run as a headless server.
-				\\  preferredAuthenticationAlgorithm - Key algorithm used for server authentication (e.g. ed25519).
-				\\  vulkanTestingMode - Set to true to enable experimental Vulkan testing mode.
+				\\All launch-time configuration is done through the "launchConfig.zon" file in the game's working directory. See that file for the available options.
 				\\
 			, .{});
 			return;
--- a/src/main.zig
+++ b/src/main.zig
@@ -340,9 +340,8 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
 			std.log.info(
 				\\Cubyz does not accept any command line arguments.
 				\\All launch-time configuration is done through the "launchConfig.zon" file in the game's working directory. See that file for the available options.
-				\\
 			, .{});
-			return;
+			std.process.exit(0);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/main.zig
```diff
@@ -335,6 +335,28 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
...
+		_ = argIterator.skip();
+		if (argIterator.next() != null) {
+			std.debug.print(
+				\\Cubyz does not accept any command line arguments.
+				\\All launch-time configuration is done through the "launchConfig.zon" file in the game's working directory. Supported fields:
+				\\  cubyzDir - Directory used to store game data and world saves.
+				\\  autoEnterWorld - Name of a world to automatically enter on startup.
+				\\  headlessServer - Set to true to run as a headless server.
+				\\  preferredAuthenticationAlgorithm - Key algorithm used for server authentication (e.g. ed25519).
+				\\  vulkanTestingMode - Set to true to enable experimental Vulkan testing mode.
```
IntegratedQuantum: This list will quickly go out of date.
The information on available fields is already inside the file when you get it for the first time.

src/main.zig
```diff
@@ -335,6 +335,28 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
 		std.log.warn("Cubyz detected it's running on Windows. For optimal performance and reduced power usage please install Linux.", .{});
 	}
 
+	{
+		var argIterator = args.args.iterateAllocator(stackAllocator.allocator) catch |err| {
+			std.log.err("Failed to read command line arguments: {s}", .{@errorName(err)});
+			@panic("Failed to read command line arguments");
```
IntegratedQuantum: The game does not actually require this, and shouldn't crash when they cannot be read. Printing an error is enough, just break out of the block instead.

IntegratedQuantum (changes requested):

yel0h:
Okay, fixed both.

src/main.zig
IntegratedQuantum: Can you move the check before this? I think it looks odd to have this in the output too.

src/main.zig
```diff
@@ -335,6 +335,23 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
...
 
+	argCheck: {
+		var argIterator = args.args.iterateAllocator(stackAllocator.allocator) catch |err| {
+			std.log.err("Failed to read command line arguments: {s}", .{@errorName(err)});
+			break :argCheck;
+		};
+		defer argIterator.deinit();
+		_ = argIterator.skip();
+		if (argIterator.next() != null) {
+			std.debug.print(
```
IntegratedQuantum: I think we should also log this, just in case someone accidentally adds command line arguments to a shortcut or something and doesn't look at the console.

IntegratedQuantum (changes requested):

yel0h:
Moved and logged

src/main.zig
```diff
@@ -329,6 +329,23 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
...
+			std.log.err("Failed to read command line arguments: {s}", .{@errorName(err)});
+			break :argCheck;
+		};
+		defer argIterator.deinit();
+		_ = argIterator.skip();
+		if (argIterator.next() != null) {
+			std.log.info(
+				\\Cubyz does not accept any command line arguments.
+				\\All launch-time configuration is done through the "launchConfig.zon" file in the game's working directory. See that file for the available options.
+				\\
```
IntegratedQuantum: The trailing newline is automatically added by the logger.
```suggestion
```

src/main.zig
```diff
@@ -329,6 +329,23 @@ pub fn main(args: std.process.Init.Minimal) void { // MARK: main()
...
+			, .{});
+			return;
```
IntegratedQuantum: I also noticed a stray
```
Clearing global arena with 0 MiB
```
I'd suggest to exit the process instead of returning (which runs all the defers first).

IntegratedQuantum (changes requested):

yel0h:
Prints just the message now, no excess newline.

IntegratedQuantum (approved):

PR #3344, opened 2026-07-08
Add name field to Multiplayer menu

### LINKED ISSUE #3233: Add a text field to the Multiplayer menu that allows you to change your name before joining
Should be done after #3229 is fixed.

yel0h:
Adds a text field to the Multiplayer menu so players can change their display name right before hosting or joining a server, instead of having to back out to the Social panel first.

image

Reuses the same length validation and persistence logic as the existing Change Name window. The name is applied on Enter, when clicking "Host World" or "Join Server", and when the window is closed so no edit is lost.

Closes #3233

### DIFF
--- a/src/gui/windows/multiplayer_join.zig
+++ b/src/gui/windows/multiplayer_join.zig
@@ -9,6 +9,7 @@ const gui = @import("../gui.zig");
 const GuiComponent = gui.GuiComponent;
 const GuiWindow = gui.GuiWindow;
 const Button = @import("../components/Button.zig");
+const HorizontalList = @import("../components/HorizontalList.zig");
 const Label = @import("../components/Label.zig");
 const TextInput = @import("../components/TextInput.zig");
 const VerticalList = @import("../components/VerticalList.zig");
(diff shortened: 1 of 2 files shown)


IntegratedQuantum:
For graphical or user interface changes please add a screenshot to the PR description

yel0h:
Right, added.

IntegratedQuantum:
> The name is applied on Enter, when clicking "Host World" or "Join Server", and when the window is closed so no edit is lost.

I would prefer closing to reset it, otherwise there is no way to revert if you accidentally delete it or something.

yel0h:
Okay, now it reverts the name on close.

IntegratedQuantum:
I think having it here is a problem, e.g. you can click `Join Server`, then change it, then switch to the other menu and join the server.
So I think it would be better to have this field in the `multiplayer_join` window instead.

yel0h:
Moved it to `multiplayer_join`

image

src/gui/windows/multiplayer_join.zig
```diff
@@ -75,6 +88,9 @@ fn copyIp() void {
 
 pub fn onOpen() void {
 	const list = VerticalList.init(.{padding, 16 + padding}, 300, 16);
+	list.add(Label.init(.{0, 0}, width, "Name:", .center));
+	nameEntry = TextInput.init(.{0, 0}, width, 32, settings.playerName, .{.onNewline = .init(applyName)});
```
IntegratedQuantum: Could you please make it more compact, so that the label is on the same line as the text.
Also I would like to have the opinion of one of the artists on where to put the text field. I think having it at the very top is kind of weird.

IntegratedQuantum (changes requested):

yel0h:
Here it is at the top and above the button:

image

image

careeoki:
I think it being at the bottom looks best.

PR #3345, opened 2026-07-09
Move multiplayer connect off main thread

### LINKED ISSUE #2726: Multiplayer join freeze (move connection to another thread)
When joining someone via multiplayer, the main thread has to wait for the connection to succeed before it'll proceed, so the entire game window freezes.

We should move the connection attempt to a separate thread, so the main thread can continue updating the game window. Put a cancel button in the UI so the main thread can flag the child thread to exit if a cancellation is requested.

yel0h:
Moves the connect attempt off the main thread and onto a background thread, with a new "Connecting..." window shown while it runs.

image

Adds a Cancel button on that window that flags the background thread to bail out via a shared `cancelRequested` atomic, checked in both the TLS handshake retry loop and the handshake wait loop, where it could get blocked against an unresponsive server.

Cancelling or failing a join attempt stays on the Join screen instead of jumping to the main menu. `connecting.zig` restores the `ConnectionManager` via `multiplayer_join.restoreConnection()` on both the `.failed` and `.cancelled` paths. The background connect thread calls `GarbageCollection.syncPoint()` while spinning in the TLS retry loop.

Split the GPU-uploading `finishHandshake()` tail out into a new `World.finishGpuUpload()`, called on the main thread after a successful `init()`: from `connecting.zig`'s `.success` branch for multiplayer, and from `save_selection.zig`'s success path for singleplayer or hosting (which still is synchronous but would otherwise have lost its GPU upload).

Fixes #2726

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -285,7 +285,7 @@ pub const World = struct { // MARK: World
 	shouldRestart: std.atomic.Value(bool) = .init(false),
 	shouldReload: bool = false,
 
-	pub fn connect(self: *World) !ZonElement {
+	fn connect(self: *World) !ZonElement {
 		main.heap.allocators.createWorldArena();
 		errdefer main.heap.allocators.destroyWorldArena();
 
@@ -360,6 +360,8 @@ pub const World = struct { // MARK: World
 	}
 
 	pub fn finishHandshake(self: *World, zon: ZonElement) !void {
+		self.conn.manager.world = self;
+		main.game.world = self;
 		errdefer main.heap.allocators.destroyWorldArena();
 		errdefer self.conn.deinit();
 		errdefer self.itemDrops.deinit();
--- a/src/game.zig
+++ b/src/game.zig
@@ -285,12 +285,7 @@ pub const World = struct { // MARK: World
 	shouldRestart: std.atomic.Value(bool) = .init(false),
 	shouldReload: bool = false,
 
-	pub fn init(self: *World, ip: []const u8, manager: *ConnectionManager) !void {
-		self.conn = try Connection.init(manager, ip, null);
-		self.manager = manager;
-		try self.@"continue"();
-	}
-	pub fn @"continue"(self: *World) !void {
+	pub fn connect(self: *World) !ZonElement {
 		main.heap.allocators.createWorldArena();
 		errdefer main.heap.allocators.destroyWorldArena();
 
@@ -308,12 +303,17 @@ pub const World = struct { // MARK: World
 		self.itemDrops.init(main.globalAllocator);
 		errdefer self.itemDrops.deinit();
 
-		const handshakeZon = try network.protocols.handShake.clientSide(self.conn, settings.playerName);
+		return try network.protocols.handShake.clientSide(self.conn, settings.playerName);
+	}
 
-		try self.finishHandshake(handshakeZon);
-		main.network.protocols.handShake.signalLoadedAssets();
+	pub fn init(self: *World, ip: []const u8, manager: *ConnectionManager) !ZonElement {
+		self.conn = try Connection.init(manager, ip, null);
+		self.manager = manager;
+		return try self.connect();
+	}
 
-		self.paused = false;
+	pub fn @"continue"(self: *World) !void {
+		try self.finishHandshake(try self.connect());
 	}
 
 	pub fn deinit(self: *World) void {
@@ -359,7 +359,11 @@ pub const World = struct { // MARK: World
 		main.heap.allocators.destroyWorldArena();
 	}
 
-	fn finishHandshake(self: *World, zon: ZonElement) !void {
+	pub fn finishHandshake(self: *World, zon: ZonElement) !void {
+		errdefer main.heap.allocators.destroyWorldArena();
+		errdefer self.conn.deinit();
+		errdefer self.itemDrops.deinit();
+
 		// TODO: Consider using a per-world allocator.
 		self.blockPalette = try assets.Palette.init(main.globalAllocator, zon.getChild("blockPalette"), "cubyz:air");
 		errdefer self.blockPalette.deinit();
@@ -390,6 +394,9 @@ pub const World = struct { // MARK: World
 		main.entityModel.loadModelsAndTexture();
 
 		try Player.loadFrom(zon.getChild("player"));
+		main.network.protocols.handShake.signalLoadedAssets();
+
+		self.paused = false;
 	}
 
 	pub fn update(self: *World, deltaTime: f64) void {
--- a/src/gui/gui.zig
+++ b/src/gui/gui.zig
@@ -31,6 +31,7 @@ var windowList: ListManaged(*GuiWindow) = undefined;
 var hudWindows: ListManaged(*GuiWindow) = undefined;
 pub var openWindows: ListManaged(*GuiWindow) = undefined;
 var selectedWindow: ?*GuiWindow = null;
+var modalWindow: ?*GuiWindow = null;
 pub var selectedTextInput: ?*TextInput = null;
 var hoveredAWindow: bool = false;
 pub var reorderWindows: bool = false;
@@ -43,6 +44,7 @@ pub var hoveredItemSlot: ?*ItemSlot = null;
 const GuiCommandQueue = struct { // MARK: GuiCommandQueue
 	const Action = enum {
 		open,
+		openModal,
 		close,
 	};
 	const Command = struct {
@@ -70,6 +72,9 @@ const GuiCommandQueue = struct { // MARK: GuiCommandQueue
 				.open => {
 					executeOpenWindowCommand(command.window);
 				},
+				.openModal => {
+					executeOpenModalWindowCommand(command.window);
+				},
 				.close => {
 					executeCloseWindowCommand(command.window);
 				},
@@ -92,10 +97,20 @@ const GuiCommandQueue = struct { // MARK: GuiCommandQueue
 		selectedWindow = null;
 	}
 
+	fn executeOpenModalWindowCommand(window: *GuiWindow) void {
+		const alreadyOpen = std.mem.containsAtLeastScalar(*GuiWindow, openWindows.items, 1, window);
+		if (!alreadyOpen) setSelectedTextInput(null);
+		modalWindow = window;
+		executeOpenWindowCommand(window);
+	}
+
 	fn executeCloseWindowCommand(window: *GuiWindow) void {
 		defer updateWindowPositions();
 		if (selectedWindow == window) {
 			selectedWindow = null;
+		}
+		if (modalWindow == window) {
+			modalWindow = null;
 		}
 		for (openWindows.items, 0..) |_openWindow, i| {
 			if (_openWindow == window) {
@@ -327,6 +342,20 @@ pub fn openWindowFromRef(window: *GuiWindow) void {
 	GuiCommandQueue.scheduleCommand(.{.action = .open, .window = window});
 }
 
+pub fn openModalWindow(id: []const u8) void {
+	for (windowList.items) |window| {
+		if (std.mem.eql(u8, window.id, id)) {
+			openModalWindowFromRef(window);
+			return;
+		}
+	}
+	std.log.err("Could not find window with id {s}.", .{id});
+}
+
+pub fn openModalWindowFromRef(window: *GuiWindow) void {
+	GuiCommandQueue.scheduleCommand(.{.action = .openModal, .window = window});
+}
+
 pub fn toggleWindow(id: []const u8) void {
 	defer updateWindowPositions();
 	for (windowList.items) |window| {
@@ -471,6 +500,15 @@ pub fn mainButtonPressed(_: main.Window.Key.Modifiers) void {
 	setSelectedTextInput(null);
 	const mousePosition = main.Window.getMousePosition()/@as(Vec2f, @splat(scale));
 
+	if (modalWindow) |modal| {
+		if (@reduce(.And, mousePosition >= modal.pos) and @reduce(.And, mousePosition < modal.pos + modal.size)) {
+			if (modal.mainButtonPressed(mousePosition) == .handled) {
+				selectedWindow = modal;
+			}
+		}
+		return;
+	}
+
 	// reverse order of rendering, the last-rendered element is the first one that we should try to interact with
 	var i: usize = openWindows.items.len;
 	while (i > 0) {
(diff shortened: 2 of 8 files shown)


IntegratedQuantum (changes requested):
Instead of passing this cancelRequest stuff everywhere, I think this would be a perfect use-case for Zig's new async IO system, which does support cancellation of tasks.

yel0h:
Okay, removed `cancelRequested`, now uses a future and its `cancel()` method instead.

src/game.zig
```diff
@@ -362,13 +361,17 @@ pub const World = struct { // MARK: World
 		self.playerBiome = .init(main.server.terrain.biomes.getPlaceholderBiome());
 		main.audio.setMusic(self.playerBiome.raw.preferredMusic);
 
+		try Player.loadFrom(zon.getChild("player"));
+	}
+
+	pub fn finishGpuUpload(self: *World) void {
```
IntegratedQuantum: I think the entirety of asset loading should be kept together on the render thread.
(Also there is no point where it would cancel asset loading anyways)

src/network.zig
```diff
@@ -1299,7 +1299,8 @@ pub const Connection = struct { // MARK: Connection
 				const result = c.mbedtls_ssl_handshake(&self.sslContext);
 				self.mutex.unlock();
 				if (result == c.MBEDTLS_ERR_SSL_WANT_READ) {
-					main.io.sleep(.fromMilliseconds(10), .awake) catch {};
+					main.heap.GarbageCollection.syncPoint();
```
IntegratedQuantum, line 1325: That is correct and should be here, however I'd like to know if you added it because you saw it was missing or because you are trying to fix an error that came up during your implementation.

src/network/protocols.zig
```diff
@@ -231,6 +231,10 @@ pub const handShake = struct { // MARK: handShake
 
 		conn.mutex.lock();
 		while (true) {
+			main.io.checkCancel() catch |err| {
```
IntegratedQuantum: is Cancelled not part of the error set of timedWait?

src/gui/windows/save_selection.zig
```diff
@@ -71,9 +71,11 @@ pub fn openWorld(name: []const u8) void {
 	const ipPort = std.fmt.allocPrint(main.stackAllocator.allocator, "127.0.0.1:{}", .{main.server.connectionManager.localPort}) catch unreachable;
 	defer main.stackAllocator.free(ipPort);
 	main.game.world = &main.game.testWorld;
-	main.game.testWorld.init(ipPort, clientConnection) catch |err| {
+	if (main.game.testWorld.init(ipPort, clientConnection)) {
+		main.game.testWorld.finishGpuUpload();
+	} else |err| {
```
IntegratedQuantum: I think `catch` with early return (and this code should have an early return and not blindly open the game UI with no world active) is generally more readable.

src/gui/windows/connecting.zig
```diff
@@ -0,0 +1,120 @@
...
+	list.add(statusLabel);
+	list.add(Button.initText(.{0, 0}, 100, "Cancel", .{.onAction = .init(cancel)}));
+	list.finish(.center);
+	window.rootComponent = list.toComponent();
+	window.contentSize = window.rootComponent.?.pos() + window.rootComponent.?.size() + @as(Vec2f, @splat(padding));
+	gui.updateWindowPositions();
+}
+
+pub fn onClose() void {
+	if (connectFuture) |*future| {
```
IntegratedQuantum: I think ideally it should never be null at this point, the await should be done in other places.

IntegratedQuantum (changes requested):

src/network/protocols.zig
Wunka: Seems like it. We should maybe replace our implementation with the current zig one. (until 0.17.0 is released)

src/network.zig
yel0h: Yeah, when trying to connect to an unreachable IP, if you waited for >20 seconds before canceling, the GC would log an error about no sync point being executed, so that's why I added it.

src/network/protocols.zig
yel0h: It's not, there's only `Timeout`.

yel0h:
Addressed the issues. Split `World.init` into `connect` (which runs on the background thread) and `finishHandshake` (on the main thread, also merged `finishGpuUpload` back into this).

src/gui/windows/connecting.zig
```diff
@@ -0,0 +1,133 @@
...
+	switch (state.load(.acquire)) {
+		.connecting => {},
+		.connected => {
+			if (connectFuture) |*future| {
+				_ = future.await(main.io);
+				connectFuture = null;
+			}
+			main.game.testWorld.finishHandshake(handshakeZon) catch |err| {
+				errorMessage = main.globalAllocator.dupe(u8, @errorName(err));
+				state.store(.failed, .release);
```
IntegratedQuantum, line 98: Please use [switch continue](https://ziglang.org/documentation/master/#toc-Labeled-switch) here instead of waiting for the next frame

src/gui/windows/connecting.zig
```diff
@@ -0,0 +1,133 @@
...
+				return;
+			};
+			connectionManager.?.world = &main.game.testWorld;
+			state.store(.success, .release);
+		},
+		.success => {
```
IntegratedQuantum: I think this case can be removed. This also reduces the wait time by one frame and avoids edge cases when the button is clicked exactly in that extra frame.
```suggestion
```

IntegratedQuantum (changes requested):
I would also like you to think through some of the edge cases of having the rest of the UI interactive.
e.g. what happens when someone presses join again? Or what happens when someone enters a singleplayer world while connecting?

yel0h:
Okay, inlined the `.success` branch and used the labeled continue. To prevent the player from trying to load a save/join/host a server while already connecting to one I added a simple guard in `game.zig`, checked at world loading entry points, notifying the player with an error.

IntegratedQuantum:
> To prevent the player from trying to load a save/join/host a server while already connecting to one I added a simple guard in game.zig, checked at world loading entry points, notifying the player with an error.

Hmm, I don't really like your solution here, it's easy to reintroduce similar edge cases.
I think it would be best if the connecting screen is exclusive, such that the player cannot interact with anything else while this happens.

yel0h:
Ok, so I added modals to the gui, exclusively receiving all input as long as they're open and dimming the background to make this clear to the user, and made Connecting a modal.

image

src/gui/gui.zig
```diff
@@ -383,6 +384,15 @@ pub fn isWindowOpen(id: []const u8) bool {
 	return false;
 }
 
+fn getModalWindow() ?*GuiWindow {
+	var i = openWindows.items.len;
+	while (i > 0) {
+		i -= 1;
+		if (openWindows.items[i].modal) return openWindows.items[i];
```
IntegratedQuantum: Since you only allow one modal window at a time, making this a bool may lead to weird side effects.
Instead I'd prefer to have a variable in the gui system that stores the (optional) modal window together with a setter function to set it from the outside. This guarantees that there is only one and makes the workflow of accessing it simpler.

The setter function should also ensure that the modal window is actually on top.

Edit: Actually I think an inventory command for opening modal windows would be better, to avoid side effects when introducing this in the middle of the frame.

src/gui/gui.zig
```diff
@@ -542,15 +565,23 @@ pub fn updateAndRenderGui() void {
...
+			// reverse order of rendering, the last-rendered element is the first one that we should try to interact with
+			var i: usize = openWindows.items.len;
+			while (i != 0) {
+				i -= 1;
+				const window: *GuiWindow = openWindows.items[i];
+				if (GuiComponent.contains(window.pos, window.size, mousePos)) {
+					if (window.updateHovered(mousePos) == .handled) {
+						hoveredAWindow = true;
+						break;
+					}
```
IntegratedQuantum: Intead of duplicating the logic, I'd suggest to add an if with a continue, like you did in the button released case.

src/gui/gui.zig
```diff
@@ -79,6 +79,7 @@ const GuiCommandQueue = struct { // MARK: GuiCommandQueue
 
 	fn executeOpenWindowCommand(window: *GuiWindow) void {
 		defer updateWindowPositions();
+		if (window.modal) setSelectedTextInput(null);
```
IntegratedQuantum: This seems weird. What if the modal window contains a text field?

src/gui/windows/connecting.zig
```diff
@@ -0,0 +1,131 @@
...
+		}
+		return;
+	};
+	state.store(.connected, .release);
+}
+
+pub fn start(_ip: []const u8, manager: *ConnectionManager) void {
+	ip = main.globalAllocator.dupe(u8, _ip);
+	connectionManager = manager;
+	state.store(.connecting, .monotonic);
```
IntegratedQuantum: Does not need to be atomic.
```suggestion
	state = .init(.connecting);
```

src/gui/windows/connecting.zig
```diff
@@ -0,0 +1,131 @@
...
+
+fn connectFromNewThread() void {
+	main.initThreadLocals();
+	defer main.deinitThreadLocals();
+
+	handshakeZon = main.game.testWorld.connect(ip, connectionManager.?) catch |err| {
+		if (err == error.Canceled) {
+			state.store(.cancelled, .release);
+		} else {
+			errorMessage = main.globalAllocator.dupe(u8, @errorName(err));
```
IntegratedQuantum: `@errorName` returns a string that's in the read-only section of the executable. It's lifetime is ∞ and does not need to be allocated.

IntegratedQuantum (changes requested):

yel0h:
I replaced the bool with a `modalWindow` variable, set exclusively through the `openModal` command. It calls `executeOpenWindowCommand` so the window is guaranteed to end up on top, and the text input selection is no longer cleared when an already open window is promoted to a modal. Also the other fixes.

src/gui/gui.zig
```diff
@@ -92,11 +97,27 @@ const GuiCommandQueue = struct { // MARK: GuiCommandQueue
...
 	}
 
+	fn executeOpenModalWindowCommand(window: *GuiWindow) void {
+		var alreadyOpen = false;
+		for (openWindows.items) |_openWindow| {
+			if (_openWindow == window) {
+				alreadyOpen = true;
+				break;
+			}
+		}
```
IntegratedQuantum: I'd suggest to use std.mem.containsAtLeastScalar instead.

src/network/protocols.zig
```diff
@@ -231,6 +231,10 @@ pub const handShake = struct { // MARK: handShake
...
+				conn.mutex.unlock();
```
IntegratedQuantum: This is also missed in the `if (conn.connectionState.load(.monotonic) == .disconnected)` below.
I'd suggest to use a block so you can `defer mutex.unlock()` to handle all cases.
Then you could use `try` here.

IntegratedQuantum (changes requested):

IntegratedQuantum:
I think I just broke some things in #3359 for you. Sorry for this, but the other PR (originally authored by Till++ in https://github.com/PixelGuys/Cubyz/pull/3278) was more important. Please rebase your branch and fix the conflicts.

yel0h:
Ok, should be good, split `connect()` out of `@"continue"()` and replaced `connect()` with `init()` wherever that's used.

src/game.zig
```diff
@@ -285,12 +285,7 @@ pub const World = struct { // MARK: World
 	shouldRestart: std.atomic.Value(bool) = .init(false),
 	shouldReload: bool = false,
 
-	pub fn init(self: *World, ip: []const u8, manager: *ConnectionManager) !void {
-		self.conn = try Connection.init(manager, ip, null);
-		self.manager = manager;
-		try self.@"continue"();
-	}
-	pub fn @"continue"(self: *World) !void {
+	pub fn connect(self: *World) !ZonElement {
```
IntegratedQuantum: This can be private
```suggestion
	fn connect(self: *World) !ZonElement {
```

src/gui/gui.zig
```diff
@@ -570,6 +614,11 @@ pub fn updateAndRenderGui() void {
 		const oldScale = draw.setScale(scale);
 		defer draw.restoreScale(oldScale);
 		for (openWindows.items) |window| {
+			if (modalWindow == window) {
+				const modalOldColor = draw.setColor(0x80000000);
+				draw.rect(.{0, 0}, main.Window.getWindowSize());
+				draw.restoreColor(modalOldColor);
```
IntegratedQuantum: Please always use `defer` for restore operations and put them directly under the line that sets it.

src/gui/windows/connecting.zig
```diff
@@ -0,0 +1,129 @@
...
+		return;
+	};
+	state.store(.connected, .release);
+}
+
+pub fn start(_ip: []const u8, manager: *ConnectionManager) void {
+	ip = main.globalAllocator.dupe(u8, _ip);
+	connectionManager = manager;
+	state = .init(.connecting);
+	gui.openModalWindow("connecting");
```
IntegratedQuantum:
```suggestion
	gui.openModalWindowFromRef(&window);
```

src/gui/windows/connecting.zig
```diff
@@ -0,0 +1,129 @@
...
+
+const State = enum(u8) { connecting, connected, failed, cancelled };
+
+var connectionManager: ?*ConnectionManager = null;
+var ip: []const u8 = "";
+var connectFuture: ?std.Io.Future(void) = null;
+var handshakeZon: main.ZonElement = undefined;
+var state: std.atomic.Value(State) = .init(.connecting);
+var errorMessage: []const u8 = "";
+var statusLabel: *Label = undefined;
```
IntegratedQuantum: This label is not used for anything

IntegratedQuantum (changes requested):
Trying to join a world crashes, please thoroughly test your changes.

yel0h:
Right, I was testing on the optimized build, oops... The problem was I was setting the world pointers after `finishHandshake` but they were needed to initialize the inventory inside `finishHandshake`.

IntegratedQuantum:
looks good from a technical standpoint, now since this introduces a new UI concept I would also like to have an artistic review from @ikabod-kee

ikabod-kee:
Coming!!

ikabod-kee (approved):
Tested with bobik-man. It good 👍

Issue #2997, opened 2026-04-30
More inventory actions (right clicking, shift clicking) for the bag


IntegratedQuantum:
extracted from #2989

PR #3350, opened 2026-07-10
Added shift click in and out of bag

calebstenberg:
This lets you shift click items out of your bag and shift click into your bag if there is not a chest that it can go into
Fixes part of #2997


Cubyz-fmt-x86_64-windows.exe
Wunka, line 1: this is an unintended change. (the Cubyz.exe should also not be here

Wunka (changes requested):
Nice! this works good. One thing though

Cubyz-fmt-x86_64-windows.exe
calebstenberg: Whoops, fixed it

src/gui/gui.zig
```diff
@@ -708,6 +710,11 @@ pub const inventory = struct { // MARK: inventory
 						break;
```
IntegratedQuantum: This only breaks out of the while loop, so if there are any leftover items it would deposit them to the bag, which would be confusing.

src/gui/gui.zig
```diff
@@ -708,6 +710,11 @@ pub const inventory = struct { // MARK: inventory
...
 					}
 				}
+				if (@"cubyz:bag".client.getBag(main.game.Player.id)) |bag| {
+					if (bag.slots.items.len < bag.sizeLimit) {
```
IntegratedQuantum: This check is done inside the command, no need to have it here. (in fact this check here would make it impossible to deposit to a bag that has a partial stack at the top)

IntegratedQuantum (changes requested):

src/gui/gui.zig
calebstenberg: I removed this entirely

src/gui/gui.zig
calebstenberg: Checking with a bool instead

src/gui/gui.zig
```diff
@@ -702,12 +704,17 @@ pub const inventory = struct { // MARK: inventory
 		if (mainGuiButton.pressed and mainGuiButton.modsOnPress.shift) {
 			if (itemSlot.inventory.super.id == main.game.Player.inventory.super.id) {
 				var iterator = std.mem.reverseIterator(openWindows.items);
+				var toBag = true;
 				while (iterator.next()) |window| {
 					if (window.shiftClickableInventory) |inv| {
 						itemSlot.inventory.depositToAny(itemSlot.itemSlot, &.{inv}, itemSlot.inventory.getAmount(itemSlot.itemSlot));
+						toBag = false;
```
IntegratedQuantum: Please use a labelled break instead of a bool. ([see also](https://ziglang.org/documentation/master/#Blocks))

IntegratedQuantum (changes requested):

src/gui/gui.zig
```diff
@@ -702,11 +704,15 @@ pub const inventory = struct { // MARK: inventory
 		if (mainGuiButton.pressed and mainGuiButton.modsOnPress.shift) {
 			if (itemSlot.inventory.super.id == main.game.Player.inventory.super.id) {
```
Wunka: I think quantum meant that you do here:

```suggestion
			if (itemSlot.inventory.super.id == main.game.Player.inventory.super.id) blk: {
```
and then you can later do: `break :blk` so you don't need the `if(!deposited)`

Wunka (changes requested):

src/gui/gui.zig
calebstenberg: Thanks Wunka

Wunka: you should be able to just put `blk: ` before the  `{` of the if. then you don't need to open another scope

calebstenberg: Ah, ok!

src/gui/gui.zig
```diff
@@ -27,6 +27,8 @@ pub const tooltip = @import("tooltip.zig");
 pub const windowlist = @import("windows/_list.zig");
 const gamepad_cursor = @import("gamepad_cursor.zig");
 
+const @"cubyz:bag" = main.entity.components.@"cubyz:bag";
+
```
Wunka: unused import, please remove that. everything else looks good!

Wunka (changes requested):

Wunka (approved):

IntegratedQuantum (approved):

PR #3351, opened 2026-07-10
Add compass item/HUD

xdokux:
## What this adds

A horizontal scrolling compass.

- New 'cubyz:compass' item, craftable from 2 iron ingots + 1 ruby ore
- The compass HUD only appears while the compass item is the currently held/selected item — switching to anything else hides it
- 6 visual themes: Classic, Parchment, Forest, Stone, Iron, and Minimal. Went with somewhat earthy tones to "fit" in cubyz until i can get some textures made for them.
- A master on/off toggle and a theme-selection slider added to the Graphics settings menu
- Tick marks every 5°, degree numbers every 15°, and letter labels (N, NE, E, SE, S, SW, W, NW) at the 4 cardinal directions and the diagonals between them.

## Compass direction

Cubyz didn't previously have any concept of cardinal directions in the f3 menu, so this introduces it in the form of an item: yaw `0` (`camera.rotation[2] == 0`, facing `+Y`) is defined as North, with the heading increasing clockwise (N → E → S → W), matching how `camera.direction` is already computed elsewhere in the engine. This seemed like the most natural default given the existing rotation math, but very open to feedback if a different convention is preferred.

## Files that i edited/added

- `src/gui/windows/compass.zig` (new) — the HUD window and all 6 themes
- `assets/cubyz/items/compass.zig.zon` (new) — item definition
- `assets/cubyz/items/textures/compass.png` (new) — item texture
- `assets/cubyz/recipes/compass_recipes.zig.zon` (new) — crafting recipe
- `src/gui/windows/_list.zig` — registers the new window
- `src/settings.zig` — adds `compassEnabled` and `compassStyle` persisted settings
- `src/gui/windows/graphics.zig` — adds the Compass toggle checkbox and theme-selection slider to the Graphics menu

Happy to adjust the visual design, themes, or add texture-based backgrounds for the themes that would benefit from them (Parchment, Forest, Stone, Iron) if that's wanted — currently they're transparent colored rectangles rather than textured.


Wunka:
Please run Cubyz formatter (https://github.com/PixelGuys/Cubyz/blob/master/docs/CONTRIBUTING.md#disable-zigs-automatic-formatter) instead of the zig one

Wunka:
You need to run the formatter also on your zon files

Wbjitscool:
This pr will help out with exploring a lot

IntegratedQuantum:
I personally never used a compass like this in any game. Also the interface isn't particularly immersive, it would be nicer for the hand item to be animated.
Could outline the use-case? Maybe there are better solutions for it, like your compass could point towards a marked waypoint or something like that?

xdokux:
> I personally never used a compass like this in any game. Also the interface isn't particularly immersive, it would be nicer for the hand item to be animated.
> Could outline the use-case? Maybe there are better solutions for it, like your compass could point towards a marked waypoint or something like that?

I agree with you. Was just a rough proof of concept for the time being to get more aquatinted with zig.  I've got more ideas for it in the future once I have time. The waypoint idea would probably be cool. Maybe have different types of compasses  for hidden structures or something like that. Would love to hear your thoughts

codemob-dev:
With how big the render distance is, it's more immersive to just climb a rock and look around than to use a compass in your hud.

IntegratedQuantum:
> The waypoint idea would probably be cool. Maybe have different types of compasses for hidden structures or something like that. Would love to hear your thoughts

Yeah, it could be useful as a way to find secret structures.
Though I think right now the main blocker would be figuring out how best to support items like this.

PR #3352, opened 2026-07-10
Added Magma Blockset

Crepestrom:
image

image


Crepestrom:
ah magma/smooth probably need to be redone

H41ogen:
i love the textures, but shouldnt magma glow?

Crepestrom:
oops
here i fixed that

Crepestrom:
also note that all of these blocks hurt you

Wbjitscool:
cool textures

Wbjitscool:
maybe it can be used as some part of a certain ruins or something

assets/cubyz/blocks/magma/_defaults.zig.zon
```diff
@@ -0,0 +1,37 @@
...
+			.swingSpeed = 3.0,
+			.textureRoughness = 1.0,
+			.colors = .{
+				0xff6b231b, 0xff2c1b27, 0xff9b4522, 0xffcb6816, 0xff312834, 0xff4b3849,
+			},
+			.modifiers = .{
+				.{
+					.id = "good_at",
+					.strength = 0.5,
+					.tag = .hot,
```
SpellDigger:
```suggestion
					.tag = .magma,
```
making the tag .hot implies the effect would work for all blocks that can be considered to emit temperatures that are "hot" which wouldnt be the case
also would be a irregularity compared to other stone blocks where they are specific to that type of stone

in whatever case i would suggest changing it to magma instead for consistency

Crepestrom: i swapped it to stone
requring 3 precious
and only being a 20% buff
you never mine enough magma to need this specialized

calebstenberg: > i swapped it to stone requring 3 precious and only being a 20% buff you never mine enough magma to need this specialized

I feel like if it is almost never needed it should be easier to specialize rather than harder so it could be added as a small addition rather than a full tool commitment. Say I want to collect basalt but know I will also mine magma, then I would only add magma to my pick if it was a small change.
So I propose having it be 1 precious for a 30%-50% buff.

calebstenberg:
These textures are sick!

careeoki (changes requested):
I like the idea of lava-y building blocks, but I think we should approach it differently. 
I don't think magma makes sense as a stone type. It's not clear what the difference between "rough" and "smooth" magma would be (and your texture seems to just ignore that, but your comment suggests that you know already?)
And, while realism is not a priority, magma is molten rock. So I can't imagine it would be easy to work with.

**Instead:**
Give obsidian a stone set. 
Allow crafting these obsidian blocks with magma (or some alternative clutter-reducing method) to make "lava-y" versions of them, with lava running through the cracks like you have here. You could only do this with the bricks, tiles, walls, and pillars.

Benefits of doing it like this:

- There is now a non-lava version of the set, if you like the color/shapes, but not the lava
- You can mix the lava versions with the regular ones, for some nice effects when building (on it's own, a block set like this would be pretty niche)

You could still give magma a modifier if you want.
Whether or not you make new textures is up to you. Just know that I will probably end up redoing them. And that magma has been on my retexture hitlist for a while.

Crepestrom:
sure if you want to make the textures for the new obsidian blockset thats fine
i can easily setup the blocks once i have them
thanks for all the textures for the game you have already made

Crepestrom:
a cool blockset though i dont care to finish it

PR #3356, opened 2026-07-11
Permissions per ECS

Wunka:


Wunka:
You know what. the permission system has a lot of later goals. so this is high priority. I don't care

src/server/world.zig
```diff
@@ -990,7 +990,11 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 
 			user.gamemode = .init(self.settings.defaultGamemode);
 		} else {
-			user.permissions.fromZon(playerData);
+			// load old permissions from zon.
```
IntegratedQuantum: Any migrations of old data should be be done on world open for all entries in the world, so we can remove the migrating code in the future. (also it should be marked with a TODO and mentioned in the related issue)
In this case though I don't think it's necessary, there is no point in migrating permissions, as they are not particularly usable currently. Any servers that already use it will gladly switch to permission groups when they are finished.

src/server/server.zig
```diff
@@ -891,7 +888,11 @@ pub fn connectInternal(user: *User) void {
 	main.network.protocols.entity.send(user.conn, initialList);
 	main.stackAllocator.free(initialList);
 	sendMessage("{s}§#ffff00 joined", .{user.name});
-	user.permissions.addPermission(.white, "/command/avatar");
+
+	main.entity.components.@"cubyz:permissions".server.getPermissions(user.id).?.addPermission(.white, "/command/avatar");
```
IntegratedQuantum: Should only be done when the component doesn't exist already.

src/server/permission.zig
IntegratedQuantum: Is the old zon code still needed then?
Please remove it, that would also make the diff easier to review.

IntegratedQuantum (changes requested):

src/server/world.zig
Wunka: So I should just ignore the old permission completly?

IntegratedQuantum: Yes, I don't see a reason to care about it

src/server/permission.zig
Wunka: yes, because of groups. they are saved via zon

src/server/server.zig
Wunka: moved

src/server/permission.zig
IntegratedQuantum: Can you also change groups to binary then?

src/server/world.zig
Wunka: removed mirgration

Wunka:
I also realized that with this change, it will not be easy anymore to add permissions for a headless server. because it will all be in binary... So I would like to push #3223 after this. Or maybe even block this until that is done. Your descision @IntegratedQuantum

IntegratedQuantum:
As mentioned in voice chat, it's fine to go forward with this, I'm sure we can get #3223 before the next release, and also it is possible to do changes by opening the world without headless.

src/server/permission.zig
Wunka: done

src/server/server.zig
```diff
@@ -546,7 +549,7 @@ pub const User = struct { // MARK: User
 	}
 
 	pub fn hasPermission(user: *User, permissionPath: []const u8) bool {
```
IntegratedQuantum: This function should also go into the component (the goal is to remove this struct entirely in the end)

src/server/permission.zig
```diff
@@ -141,33 +140,35 @@ pub const Group = struct { // MARK: Group
 		allocator.destroy(self);
 	}
 
-	fn fromZon(allocator: NeverFailingAllocator, zon: ZonElement, id: u32, name: []const u8) *Group {
-		sync.threadContext.assertCorrectContext(.server);
+	pub fn fromBytes(allocator: NeverFailingAllocator, reader: *main.utils.BinaryReader, id: u32, name: []const u8) !*Group {
```
IntegratedQuantum: reading and writing should by symmetrical, so this should read (and allocate) `name` itself instead of taking it as a parameter.

src/server/permission.zig
```diff
@@ -12,23 +12,22 @@ const sync = main.sync;
...
 
-	pub fn fromZon(self: *PermissionMap, arena: NeverFailingAllocator, zon: ZonElement) void {
+	pub fn fromBytes(self: *PermissionMap, arena: NeverFailingAllocator, reader: *main.utils.BinaryReader) !void {
 		sync.threadContext.assertCorrectContext(.server);
-		for (zon.toSlice()) |item| {
-			const string = item.as([]const u8) orelse continue;
-			self.put(arena, string);
+		const len = try reader.readVarInt(usize);
+		for (0..len) |_| {
+			self.put(arena, try reader.readSliceWithSize());
```
IntegratedQuantum: Should reserver and use putAssumeCapacity

src/server/permission.zig
```diff
@@ -141,33 +140,35 @@ pub const Group = struct { // MARK: Group
...
-		self.permissions.toZon(allocator, zon);
+		writer.writeSliceWithSize(self.name);
+		self.permissions.toBytes(writer);
 	}
 
 	fn save(self: *Group, allocator: NeverFailingAllocator) void {
 		if (builtin.is_test) return;
 		sync.threadContext.assertCorrectContext(.server);
-		const path = std.fmt.allocPrint(allocator.allocator, "saves/{s}/groups/{d}.zon", .{main.server.world.?.path, self.id}) catch unreachable;
+		const path = std.fmt.allocPrint(allocator.allocator, "saves/{s}/groups/{d}.bin", .{main.server.world.?.path, self.id}) catch unreachable;
```
IntegratedQuantum: .bin is very generic. I think we should use our own extension .group or something, this makes it clearer what's inside and greppable where the parsing code is.
Also maybe consider changing the path to `permission/{d}.group` to make it clear that this has to do with permissions.

src/server/permission.zig
```diff
@@ -141,33 +140,35 @@ pub const Group = struct { // MARK: Group
...
-		var groupZon: ZonElement = .initObject(allocator);
-		defer groupZon.deinit(allocator);
-		groupZon.put("name", self.name);
-		self.toZon(allocator, &groupZon);
-		main.files.cubyzDir().writeZon(path, groupZon) catch |err| {
+
+		const writer: main.utils.BinaryWriter = .init(allocator);
+		defer writer.deinit();
+
+		self.toBytes(&writer);
```
IntegratedQuantum, line 176: Since this is a standalone file, I think we should add a version byte, just in case we ever need to migrate it.

IntegratedQuantum (changes requested):

src/server/server.zig
Wunka: done

src/server/permission.zig
Wunka: done

src/server/permission.zig
Wunka: done

src/server/permission.zig
Wunka: done

src/server/permission.zig
Wunka: done

IntegratedQuantum (approved):

PR #3357, opened 2026-07-11
Add player head movement (no raymarching)

RanPix:
adds the missing node position fields and gives snale the ability to node
only 2 issues
- in code its all kinda hardcoded to specific node names, im not sure if we wanna make someting to allow specifying how nodes should act
- if a node has rotated parents and rotated primitives then the rotation flips somehow and im not sure how to deal with it yet, like currently cubert turns his head in the opposite direction because of the way i fixed his coordinate system by rotating all of the nodes individually

[Screencast_20260711_153715.webm](https://github.com/user-attachments/assets/d3f67039-d94a-4d9a-b4d2-a15ef9be941f)


IntegratedQuantum (changes requested):
> (no raymarching)

Please implement this with raymarching instead.

src/entitySystem/modelRenderer.zig
```diff
@@ -121,13 +121,25 @@ pub const client = struct {
...
-			for (entModel.nodeParents, 0..) |parent, i| {
-				const parentMat = if (parent) |p| component.matrices[p].transpose() else Mat4f.identity();
+			const head = entModel.nodeIndexMap.get("Head");
+			if (entModel.nodeIndexMap.get("Eyestalks")) |eyestalksId| {
+				const stalkRot = ent.rot[0]*0.25;
+				const headRot = ent.rot[0]*0.75;
+				component.nodes[eyestalksId].rot = vec.Quat.quatFromAxisAngle(Vec3f{1, 0, 0}, stalkRot);
 
-				component.matrices[i] = parentMat.mul(entModel.nodePivots[i]).transpose();
+				const headId = head.?;
```
IntegratedQuantum: This is making an assertion about user data. We shouldn't crash the game here, instead please print an error message.

IntegratedQuantum (changes requested):

src/entitySystem/modelRenderer.zig
RanPix: are you sure about printing an error? maybe just continuing should be enough?

IntegratedQuantum: I'm not sure. Is that a use case that makes sense?

RanPix: i mean people would just have to know that this happens to those 2 nodes, unless we add some bindings or whatever to nodes that are going to get processed like `bind Head to PlayerRotation Y mul=0.75` but that would make things more complicated and idk if it is an actually good way of doing this and if we are going to use it for anything else, i could just not process this branch if the head doesnt exist. it would be great if you have other ideas for that

Argmaster: A headless golem is a legitimate model.

RanPix: should be better now

RanPix:
fixed cubert model

IntegratedQuantum:
> fixed cubert model

Did you try to add evil cubert again?

IntegratedQuantum:
Wait, you actually did add evil Cubert xD
Screenshot at 2026-07-11 21-52-39

IntegratedQuantum (changes requested):
Please fix the twig

RanPix:
DONE

IntegratedQuantum (approved):

PR #3358, opened 2026-07-11
Snela player model

careeoki:
Snela is another snale. They have a new model with a body type distinct from Snale.
They wear a blue headscarf, blue shorts, a brown knit sweater, and white leggings.

image
image


ikabod-kee (approved):
Yup.

IntegratedQuantum (approved):

PR #3360, opened 2026-07-12
Cave sprouts & surface caves

careeoki:
Cave sprouts are a hardy plant that thrive underground. They have `direction` rotation, so you can place them on all sides of a block.

image

Using the existing `surface_caves` layer, I added some new surface cave biomes which replace the regular `cave/slate/base`:

**Surface slate cave**: Features a lot of cave sprouts and some clay patches.
image

**Mossy slate cave**: Features cave sprouts and moss.
image

The other slate caves also have a chance for some cave sprouts to appear, but it's not as common as on the surface.
image


assets/cubyz/biomes/cave/slate/_defaults.zig.zon
```diff
@@ -20,5 +19,22 @@
...
+			.id = "cubyz:flower_patch",
+			.blocks = .{"cubyz:cave_sprouts"},
+			.chance = 0.03,
+			.width = 6,
+			.variation = 3,
+			.density = 0.06,
+			.priority = 0.1,
+		},
+		.{
+			.id = "cubyz:simple_vegetation",
```
IntegratedQuantum: Does the ceiling not support flower patches? (or did you explicitly not want to use them?)

IntegratedQuantum (changes requested):
What about marble caves? They only spawn until -512, this now effectively halves their occurences.

careeoki:
> What about marble caves? They only spawn until -512, this now effectively halves their occurences.

Did I miss something? It has `.{.shallow_cave_layer, .surface_caves_layer}` so shouldn't that let it spawn in the surface caves as well as the shallow caves?

IntegratedQuantum:
Ah, I guess I missed it then

assets/cubyz/biomes/cave/slate/_defaults.zig.zon
careeoki: I just assumed it didn't without trying it and used simple vegetation. But it appears to not work.

image
I'm happy with how it looks with simple vegetation though.

IntegratedQuantum: It's still a missing capability that should be addressed, this won't be the last time someone wants to put vegetation on the ceiling

careeoki: Yes, I just mean I don't really want it to be a blocker.

IntegratedQuantum: It's an easy fix #3373

careeoki: Alright, swapped them for flower patches

assets/cubyz/biomes/cave/slate/mossy.zig.zon
```diff
@@ -0,0 +1,64 @@
...
+			.blocks = .{"cubyz:cave_sprouts"},
+			.chance = 0.2,
+			.width = 5,
+			.variation = 3,
+			.density = 0.06,
+			.priority = 0.1,
+		},
+		.{
+			.id = "cubyz:simple_vegetation",
+			.block = "cubyz:cave_sprouts:1",
```
IntegratedQuantum, line 15: What about here?

assets/cubyz/biomes/cave/slate/surface.zig.zon
```diff
@@ -0,0 +1,64 @@
...
+			.blocks = .{"cubyz:cave_sprouts"},
+			.chance = 0.2,
+			.width = 6,
+			.variation = 3,
+			.density = 0.06,
+			.priority = 0.1,
+		},
+		.{
+			.id = "cubyz:simple_vegetation",
+			.block = "cubyz:cave_sprouts:1",
```
IntegratedQuantum, line 15: and here?

IntegratedQuantum (changes requested):

assets/cubyz/biomes/cave/slate/surface.zig.zon
careeoki: This is intentional for both. And you'll still get flower patches in these caves since it is a part of _defaults.

assets/cubyz/biomes/cave/slate/wide.zig.zon
```diff
@@ -1,4 +1,5 @@
 .{
+	.tags = .{.cave_layer},
 	.chance = 0.8,
 	.maxHeight = -100,
```
IntegratedQuantum: Can be removed, since it's already covered by the layer location.

assets/cubyz/biomes/cave/slate/low.zig.zon
```diff
@@ -1,4 +1,5 @@
 .{
+	.tags = .{.cave_layer},
 	.chance = 0.8,
 	.maxHeight = -100,
```
IntegratedQuantum: can be removed

IntegratedQuantum (changes requested):
Also I have a crazy idea: What if we treat cave sprouts as a grass type?

careeoki:
> Also I have a crazy idea: What if we treat cave sprouts as a grass type?

Honestly I did think of that. Just not sure what the soil would be though.

TyrrelMeter:
Maybe gravel or clay?

IntegratedQuantum:
Why not just slate?

TyrrelMeter:
I just can't imagine something growing on slate. Look at it, do you feel like there are any little crevices or easy places for roots to grow in. I think that slate should also be kept as a base stone without over complicating it. Lower down stones should have more features but slate should be kept basic in my opinion.

Crepestrom:
i mean moss grows on rocks

careeoki:
> Maybe gravel or clay?

Clay already has dew grass. Gravel would probably work. And it already exists too.

> Why not just slate?

With #556 fewer caves will be made of slate on the surface. I would also prefer if all of the grass blocks would break with a shover for consistency

IntegratedQuantum:
Hmm, maybe it could be a new category then, some kind of lichen or moss for the stone blocks?

Actually we do already have moss.

I guess for now we can keep it separate and migrate it later when needed.

careeoki:
> I guess for now we can keep it separate and migrate it later when needed.

Yeah, let's wait and think on it more.
Also did the thing

IntegratedQuantum:
You removed the wrong thing, I meant the height.
Removing the cavelayer seems like it would just remove the biome entirely (causing an error).

careeoki:
Oops. Github highlighted that line so I thought you were talking about that.

IntegratedQuantum (approved):

PR #3362, opened 2026-07-13
`replace_block_type` server callback

### LINKED ISSUE #2701: Add onTouch .replaceblocktype function which allows keeping the block data

if you make ontick .replace block convert a block like a log into a varient of a log it loses its rotation data when transforming.
Before:
Image
After:

Image

Test Addon:
[replacetest.zip](https://github.com/user-attachments/files/25833353/replacetest.zip)

Bamberghh:
Fixes https://github.com/PixelGuys/Cubyz/issues/2701.

### DIFF
--- a/src/callbacks/block/server/replace_block_type.zig
+++ b/src/callbacks/block/server/replace_block_type.zig
@@ -1,0 +1,42 @@
+const std = @import("std");
+
+const main = @import("main");
+const Block = main.blocks.Block;
+
+blockTyp: u16,
+
+pub fn init(zon: main.ZonElement, creator: main.callbacks.Creator) ?*@This() {
+	const replacedBlock = switch (creator) {
+		.block => |b| b,
+	};
+	const result = main.worldArena.create(@This());
+	const blockTyp = main.blocks.getTypeById(zon.get([]const u8, "block") orelse {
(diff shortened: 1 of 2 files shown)


Bamberghh:
Updated test addon: [replacetest.zip](https://github.com/user-attachments/files/29979326/replacetest.zip)

src/callbacks/block/server/replace_block_type.zig
```diff
@@ -0,0 +1,42 @@
...
+const Block = main.blocks.Block;
+
+blockType: u16,
+
+pub fn init(zon: main.ZonElement, creator: main.callbacks.Creator) ?*@This() {
+	const replacedBlock = switch (creator) {
+		.block => |b| b,
+	};
+	const result = main.worldArena.create(@This());
+	const blockType = main.blocks.getTypeById(zon.get([]const u8, "block") orelse {
```
IntegratedQuantum: Please use `getBlockById` which returns an error when the block is not found.
`getTypeById` is legacy code which shouldn't be used anymore.
In case of a missing block we do not want things to get replaced by air. We instead want the callback to not do anything.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #3363, opened 2026-07-14
Added hotbar/hover swapping

calebstenberg:
When you press a hotkey it swaps the hotbar slot with the hovered slot.

### DIFF
--- a/src/sync.zig
+++ b/src/sync.zig
@@ -226,6 +226,7 @@ pub const Command = struct { // MARK: Command
 		open = 0,
 		close = 1,
 		depositOrSwap = 2,
+		swap = 18,
 		deposit = 3,
 		takeHalf = 4,
 		drop = 5,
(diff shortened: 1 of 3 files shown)


IntegratedQuantum (changes requested):
The game crashes when hovering the same slot that should be swapped with

src/sync.zig
```diff
@@ -969,6 +971,32 @@ pub const Command = struct { // MARK: Command
 		}
 	};
 
+	const Swap = struct { // MARK: Swap
```
IntegratedQuantum, line 974: It's also unfortunate that you needed to add a new method to the interface

calebstenberg: Yes, but I think this is a very useful addition. It will make #3404 simple, and also add a lot of interesting custom client side mods like custom sorting or hotbar swapping.

IntegratedQuantum (dismissed):
Looks good to me, but I also want the artists opinion before I merge this.

careeoki (changes requested):
Trying to move items from the creative inventory to the hotbar is crashing my game. Otherwise feels good. I've been wishing we had this for a while

Image

careeoki (approved):
Yep, it works now.

IntegratedQuantum (approved):

PR #3365, opened 2026-07-14
Add player whitelist storage module

yel0h:
Adds the persistence layer for a server player whitelist, in preparation for #2566. This lays the groundwork for follow-up PRs where I will add a command and connection-time enforcement of the whitelist.

There are 2 unit tests in `whitelist.zig` covering add/contains/remove round-tripping and the enabled toggle.


Wunka (commented):
As the other whitelist pr was last updated a few months ago, I will let it go through, but please make a comment on the other whitelist pr: #2801

src/server/whitelist.zig
```diff
@@ -0,0 +1,112 @@
+const builtin = @import("builtin");
+const std = @import("std");
+
+const main = @import("main");
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const NeverFailingArenaAllocator = main.heap.NeverFailingArenaAllocator;
+const ZonElement = main.ZonElement;
+const sync = main.sync;
+
+pub var enabled: bool = false;
```
IntegratedQuantum: This should be part of the world config, not stored in the whitelist file

src/server/whitelist.zig
```diff
@@ -0,0 +1,112 @@
...
+
+var arena: NeverFailingArenaAllocator = undefined;
```
IntegratedQuantum: Any reason not to just use the worldArena? (the global arenas should also work fine during testing

src/server/whitelist.zig
```diff
@@ -0,0 +1,112 @@
...
+pub fn add(key: []const u8) bool {
+	sync.threadContext.assertCorrectContext(.server);
+	const result = entries.getOrPut(arena.allocator().allocator, key) catch unreachable;
+	if (result.found_existing) return false;
+	result.key_ptr.* = arena.allocator().dupe(u8, key);
+	save();
+	return true;
+}
+
+pub fn remove(key: []const u8) bool {
```
IntegratedQuantum: bool are ambiguous, ideally this and the add function should return an enum.

IntegratedQuantum (changes requested):
Also a design proposal:
The whitelist could also just be all the players that ever joined the server, then it wouldn't need to be separately maintained (the list of players already exists), and it would make it easy to setup the whitelist for an existing server.

yel0h:
Okay, I split the whitelist entries into explicit allowed and blocked sets, and when a key is in neither, it falls back to a `playerDatabase` membership check. + the other fixes

IntegratedQuantum:
But wouldn't it be easier to just store the blocked/allowed in the player files directly?
When you block someone you could set a blocked bool in their file.
When you whitelist someone you could add a player file for them.

yel0h:
Now the player files have a `blocked` field and I cache the blocked players when loading the database. I'm not sure if it's exactly easier, as for example a preauthorized player's file needs to have its entity position explicitly set so they don't spawn at {0, 0, 0}, but yeah, no new files saved.

Wunka:
> as for example a preauthorized player's file needs to have its entity position explicitly set so they don't spawn at {0, 0, 0}, but yeah, no new files saved.

This sounds like a bug. if the position is not set it should be set to the world spawn. please fix that in Entity.zig instead of in your whitelist system. Else there is a wierd behavior when this happens: player "A" is whitelisted, global spawn is changed, "A" joines the world and spawns at the old world spawn

src/server/whitelist.zig
```diff
@@ -0,0 +1,155 @@
+const builtin = @import("builtin");
+const std = @import("std");
+
+const main = @import("main");
+const ZonElement = main.ZonElement;
+const sync = main.sync;
+const Vec3d = main.vec.Vec3d;
+
+pub const AddResult = enum { added, alreadyAllowed };
+pub const RemoveResult = enum { blocked, alreadyBlocked };
```
Wunka:
```suggestion
const AddResult = enum { added, alreadyAllowed };
const RemoveResult = enum { blocked, alreadyBlocked };
```

src/server/whitelist.zig
```diff
@@ -0,0 +1,155 @@
...
+	const world = main.server.world.?;
+
+	if (value) {
+		world.blockedPlayers.put(main.worldArena.allocator, index, {}) catch unreachable;
+	} else {
+		_ = world.blockedPlayers.remove(index);
+	}
+
+	if (!builtin.is_test) {
+		const path = std.fmt.allocPrint(main.stackAllocator.allocator, "saves/{s}/players/{}.zon", .{world.path, index}) catch unreachable;
```
Wunka: the allocator has now also a `print` function to avoid the `catch unreachable` here please use that

src/server/whitelist.zig
```diff
@@ -0,0 +1,155 @@
...
+		defer main.stackAllocator.free(path);
+
+		var zon: ZonElement = main.files.cubyzDir().readToZon(main.stackAllocator, path) catch .null;
+		defer zon.deinit(main.stackAllocator);
+		if (zon != .object) {
+			zon.deinit(main.stackAllocator);
+			zon = ZonElement.initObject(main.stackAllocator);
```
Wunka:
```suggestion
			zon = .initObject(main.stackAllocator);
```

src/server/whitelist.zig
```diff
@@ -0,0 +1,155 @@
...
+	sync.threadContext.assertCorrectContext(.server);
+	const world = main.server.world.?;
+
+	if (value) {
+		world.blockedPlayers.put(main.worldArena.allocator, index, {}) catch unreachable;
+	} else {
+		_ = world.blockedPlayers.remove(index);
+	}
+
+	if (!builtin.is_test) {
```
Wunka:
```suggestion
	if (builtin.is_test) return
```
please use early returns when possible

src/server/whitelist.zig
```diff
@@ -0,0 +1,155 @@
...
+		main.files.cubyzDir().writeZon(path, zon) catch |err| {
+			std.log.err("Couldn't update blocked state for player {}: {t}", .{index, err});
+		};
+	}
+}
+
+pub fn add(key: []const u8) AddResult {
+	const result = ensurePlayerRecord(key);
+	const wasBlocked = main.server.world.?.blockedPlayers.contains(result.index);
+	if (wasBlocked) setBlocked(result.index, false);
```
Wunka: why not just run `setBlocked(result.index, false)` here? 
With my other comment this would also not create unnecessary file edits

src/server/whitelist.zig
```diff
@@ -0,0 +1,155 @@
...
+}
+
+fn setBlocked(index: usize, value: bool) void {
+	sync.threadContext.assertCorrectContext(.server);
+	const world = main.server.world.?;
+
+	if (value) {
+		world.blockedPlayers.put(main.worldArena.allocator, index, {}) catch unreachable;
+	} else {
+		_ = world.blockedPlayers.remove(index);
```
Wunka: I would just return here if the player was already not blocked

src/server/whitelist.zig
```diff
@@ -0,0 +1,155 @@
...
+	const result = ensurePlayerRecord(key);
+	const wasBlocked = main.server.world.?.blockedPlayers.contains(result.index);
+	if (wasBlocked) setBlocked(result.index, false);
+	return if (result.wasNew or wasBlocked) .added else .alreadyAllowed;
+}
+
+pub fn remove(key: []const u8) RemoveResult {
+	const result = ensurePlayerRecord(key);
+	const wasBlocked = main.server.world.?.blockedPlayers.contains(result.index);
+	if (!wasBlocked) setBlocked(result.index, true);
```
Wunka: here the same as above. why check contains?

yel0h:
Fixed, now `loadFrom` takes a `defaultPos` parameter

Wunka (approved):
Looks good. thanks for doing this.

src/server/whitelist.zig
IntegratedQuantum, line 1: I think instead of having a separate file dedicated for just the whitelist logic, it would be better to have all the authentication and player selection logic (including the playerDatabase) in one place. Then you also wouldn't need to create a world in your tests.

(you can move the playerDatabase and related stuff in a separate PR if you prefer, but we should at least find a more generic name here already)

src/server/world.zig
```diff
@@ -438,6 +438,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 	doGameTimeCycle: bool = true,
 
 	tickSpeed: std.atomic.Value(u32) = .init(12),
+	whitelistEnabled: std.atomic.Value(bool) = .init(false),
```
IntegratedQuantum: I'm torn whether this should be a launchConfig or a per-world config.
Either way it should be a part of the WorldSettings struct and be configurable when creating a world in headless (#2251)

src/server/world.zig
```diff
@@ -453,6 +454,7 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 	regionUpdateQueue: main.utils.CircularBufferQueue(RegionUpdateRequest),
 
 	playerDatabase: std.StringHashMapUnmanaged(usize) = .{},
+	blockedPlayers: std.AutoHashMapUnmanaged(usize, void) = .{},
```
IntegratedQuantum: In my opinion they belong together and we shouldn't have a separate hashmap for it.
```suggestion
	playerDatabase: std.StringHashMapUnmanaged(struct {playerIndex: usize, blocked: bool}) = .{},
```

src/game.zig
```diff
@@ -116,7 +116,7 @@ pub const Player = struct { // MARK: Player
 	pub const jumpHeight = 1.25;
 
 	fn loadFrom(zon: ZonElement) !void {
-		try super.loadFrom(id, zon, .client);
+		try super.loadFrom(id, zon, .client, .{0, 0, 0});
```
IntegratedQuantum: This should not be allowed to happen on the client.
```suggestion
		try super.loadFrom(id, zon, .client, undefined);
```

src/server/whitelist.zig
```diff
@@ -0,0 +1,152 @@
...
+const EnsureResult = struct { index: usize, wasNew: bool };
+
+fn ensurePlayerRecord(key: []const u8) EnsureResult {
+	sync.threadContext.assertCorrectContext(.server);
+	const world = main.server.world.?;
+
+	if (world.playerDatabase.get(key)) |index| return .{.index = index, .wasNew = false};
+
+	const index = world.nextPlayerIndex.fetchAdd(1, .monotonic);
+	world.playerDatabase.put(main.worldArena.allocator, main.worldArena.dupe(u8, key), index) catch unreachable;
```
IntegratedQuantum: Please use getOrPut instead of get and then put. You can still dupe the key afterwards.

IntegratedQuantum (changes requested):

yel0h:
Done, I moved all the player tracking together with the whitelisting to `players.zig`.

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
...
+const main = @import("main");
+const ZonElement = main.ZonElement;
+const sync = main.sync;
+
+const PlayerRecord = struct { playerIndex: usize, blocked: bool = false };
+
+var playerDatabase: std.StringHashMapUnmanaged(PlayerRecord) = .{};
+var localPlayerIndex: usize = 0;
+var nextPlayerIndex: std.atomic.Value(usize) = .init(0);
+var worldPath: []const u8 = &.{};
```
IntegratedQuantum: Since you have an explicit init function, these should all be left `undefined`

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
...
+	nextPlayerIndex = .init(0);
+}
+
+pub fn deinit() void {
+	sync.threadContext.assertCorrectContext(.server);
+	playerDatabase = .{};
+	localPlayerIndex = 0;
+	nextPlayerIndex = .init(0);
+	worldPath = &.{};
+}
```
IntegratedQuantum: The nice part about using an arena is that you don't even need the deinit.

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
...
+const sync = main.sync;
+
+const PlayerRecord = struct { playerIndex: usize, blocked: bool = false };
+
+var playerDatabase: std.StringHashMapUnmanaged(PlayerRecord) = .{};
+var localPlayerIndex: usize = 0;
+var nextPlayerIndex: std.atomic.Value(usize) = .init(0);
+var worldPath: []const u8 = &.{};
+
+pub fn init(path: []const u8, loadedLocalPlayerIndex: usize) void {
```
IntegratedQuantum: As far as I can tell this is not used externally
```suggestion
fn init(path: []const u8, loadedLocalPlayerIndex: usize) void {
```

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
+const builtin = @import("builtin");
+const std = @import("std");
+
+const main = @import("main");
+const ZonElement = main.ZonElement;
+const sync = main.sync;
+
+const PlayerRecord = struct { playerIndex: usize, blocked: bool = false };
```
IntegratedQuantum: The default is never used
```suggestion
const PlayerRecord = struct { playerIndex: usize, blocked: bool };
```

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
...
+			const nameEntry = main.stackAllocator.print("name:{s}", .{oldNameFromFile orelse break :removeOld});
+			defer main.stackAllocator.free(nameEntry);
+			if (playerDatabase.fetchRemove(nameEntry)) |kv| blocked = kv.value.blocked;
+		}
+	}
+	playerDatabase.put(main.worldArena.allocator, main.worldArena.dupe(u8, newKey), .{.playerIndex = index, .blocked = blocked}) catch unreachable;
+}
+
+const AddResult = enum { added, alreadyAllowed };
+const RemoveResult = enum { blocked, alreadyBlocked };
```
IntegratedQuantum: Please move these next to their respective functions, to make it easier to find when navigating the code

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
...
+
+pub fn add(key: []const u8) AddResult {
+	const result = ensurePlayerRecord(key);
+	const wasBlocked = result.entry.blocked;
+	result.entry.blocked = false;
+	if (wasBlocked) persistBlocked(result.entry.playerIndex, false);
+	return if (result.wasNew or wasBlocked) .added else .alreadyAllowed;
+}
+
+pub fn remove(key: []const u8) RemoveResult {
```
IntegratedQuantum: Judging by what it does, this should probably be called `block`

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
...
+	zon.put("publicKey", key);
+
+	main.files.cubyzDir().writeZon(path, zon) catch |err| {
+		std.log.err("Couldn't create player file for pre-authorized key {s}: {t}", .{key, err});
+	};
+
+	return .{.entry = result.value_ptr, .wasNew = true};
+}
+
+fn persistBlocked(index: usize, value: bool) void {
```
IntegratedQuantum: persist doesn't really fit here in my opinion
```suggestion
fn saveBlocked(index: usize, value: bool) void {
```

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
...
+fn ensurePlayerRecord(key: []const u8) EnsureResult {
+	sync.threadContext.assertCorrectContext(.server);
+
+	const result = playerDatabase.getOrPut(main.worldArena.allocator, key) catch unreachable;
+	if (result.found_existing) return .{.entry = result.value_ptr, .wasNew = false};
+
+	result.key_ptr.* = main.worldArena.dupe(u8, key);
+	result.value_ptr.* = .{.playerIndex = nextPlayerIndex.fetchAdd(1, .monotonic), .blocked = false};
+
+	if (builtin.is_test) return .{.entry = result.value_ptr, .wasNew = true};
```
IntegratedQuantum: I don't like that this hidden in the middle of the function.

I'd suggest to move the non-tested stuff into a separate function like this:
```zig
if (!builtin.is_test) {
    saveNewPlayer(...);
}

return ...
```

src/server/world.zig
```diff
@@ -34,6 +35,7 @@ pub const Settings = struct {
 	defaultGamemode: Gamemode = .creative,
 	allowCheats: bool = true,
 	testingMode: bool = false,
+	whitelistEnabled: bool = false,
```
IntegratedQuantum: Since these are not yet functional, I think it would be best to leave it for the next PR.

src/server/players.zig
```diff
@@ -0,0 +1,233 @@
...
+
+pub fn remove(key: []const u8) RemoveResult {
+	const result = ensurePlayerRecord(key);
+	const wasBlocked = result.entry.blocked;
+	result.entry.blocked = true;
+	if (!wasBlocked) persistBlocked(result.entry.playerIndex, true);
+	return if (result.wasNew or !wasBlocked) .blocked else .alreadyBlocked;
+}
+
+pub fn contains(key: []const u8) bool {
```
IntegratedQuantum: I think this should be called `isAllowedToJoin`, all the (future) whitelist checks should also be handled inside this function in my opinion.

src/server/server.zig
```diff
@@ -247,29 +248,29 @@ pub const User = struct { // MARK: User
...
+			self.playerIndex = main.server.players.lookupIndex(keyWithType) orelse continue;
 			foundKey = true;
 			const keyType = std.meta.stringToEnum(main.network.authentication.KeyTypeEnum, keyTypeName).?;
 			if (keyType == self.key) break;
 			self.legacyKey = try .initFromBase64(keyBase64, keyType);
 			break;
 		}
 		if (!foundKey) {
-			if (world.?.playerDatabase.size == 0) { // Claim the local player
+			if (main.server.players.isEmpty()) { // Claim the local player
```
IntegratedQuantum, line 259: Some of this logic should probably also move into the new, but I'd prefer if you did this in a separate PR.

IntegratedQuantum (changes requested):

yel0h:
Ok, done

src/server/players.zig
```diff
@@ -0,0 +1,227 @@
...
+		removeOld: {
+			const nameEntry = main.stackAllocator.print("name:{s}", .{oldNameFromFile orelse break :removeOld});
+			defer main.stackAllocator.free(nameEntry);
+			if (playerDatabase.fetchRemove(nameEntry)) |kv| blocked = kv.value.blocked;
+		}
+	}
+	playerDatabase.put(main.worldArena.allocator, main.worldArena.dupe(u8, newKey), .{.playerIndex = index, .blocked = blocked}) catch unreachable;
+}
+
+const EnsureResult = struct { entry: *PlayerRecord, wasNew: bool };
```
IntegratedQuantum: Again the struct declaration should be next to the function that uses it (which has now moved)

IntegratedQuantum (changes requested):
Also one missed from last review: https://github.com/PixelGuys/Cubyz/pull/3365#discussion_r3692058560

yel0h:
My bad, misinterpreted that one

IntegratedQuantum (approved):

PR #3367, opened 2026-07-15
Small audio.zig refactor

RanPix:
preparation for #3281 to reduce line count


src/audio.zig
```diff
@@ -13,39 +40,41 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
+		const path1 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "assets/{s}/{s}/{s}.ogg", .{addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path1);
 		var err: c_int = 0;
 		if (c.stb_vorbis_open_filename(path1.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/music/{s}.ogg", .{main.files.cubyzDirStr(), addon, fileName}, 0) catch unreachable;
+		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/{s}/{s}.ogg", .{main.files.cubyzDirStr(), addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path2);
 		if (c.stb_vorbis_open_filename(path2.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		std.log.err("Couldn't find music with id \"{s}\". Searched path \"{s}\" and \"{s}\"", .{id, path1, path2});
+		std.log.err("Couldn't handle audio file. Error: {s}. ID: \"{s}\". Searched path: \"{s}\" and \"{s}\"", .{@errorName(getMaError(err)), id, path1, path2});
```
Wunka: If you didn't know, you can also do:

```suggestion
		std.log.err("Couldn't handle audio file. Error: {t}. ID: \"{s}\". Searched path: \"{s}\" and \"{s}\"", .{getMaError(err), id, path1, path2});
```
I like it more, so I will push for it :red_snale_evil (I know that doens't render on github, I don't care)

Wunka (changes requested):
This seems like an simple change, so will put in Easy to review

src/audio.zig
RanPix: it looks like the CI does not like this change even though it compiles on my computer

src/audio.zig
```diff
@@ -5,6 +5,33 @@ const utils = main.utils;
 
 const c = @import("c");
 
+fn getMaError(err: c_int) anyerror {
```
IntegratedQuantum: Why are you calling it MaError, when the error comes from stb vorbis?

src/audio.zig
```diff
@@ -5,6 +5,33 @@ const utils = main.utils;
...
+		32 => error.VORBIS_continued_packet_flag_invalid,
+		33 => error.VORBIS_incorrect_stream_serial_number,
+		34 => error.VORBIS_invalid_first_page,
+		35 => error.VORBIS_bad_packet_type,
+		36 => error.VORBIS_cant_find_last_page,
+		37 => error.VORBIS_seek_failed,
+		38 => error.VORBIS_ogg_skeleton_not_supported,
+		else => unreachable,
+	};
+}
```
IntegratedQuantum, line 40: I'd suggest to do it like I did in vulkan using an enum and enum conversion functions, the problem I have with the error approach is that the `no_error` case makes no sense as an error value.

src/audio.zig
```diff
@@ -13,39 +40,41 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
+		const path1 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "assets/{s}/{s}/{s}.ogg", .{addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path1);
 		var err: c_int = 0;
 		if (c.stb_vorbis_open_filename(path1.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/music/{s}.ogg", .{main.files.cubyzDirStr(), addon, fileName}, 0) catch unreachable;
+		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/{s}/{s}.ogg", .{main.files.cubyzDirStr(), addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path2);
 		if (c.stb_vorbis_open_filename(path2.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		std.log.err("Couldn't find music with id \"{s}\". Searched path \"{s}\" and \"{s}\"", .{id, path1, path2});
+		std.log.err("Couldn't handle audio file. Error: {t}. ID: \"{s}\". Searched path: \"{s}\" and \"{s}\"", .{getMaError(err), id, path1, path2});
```
IntegratedQuantum: I think this should print both errors, otherwise the user will in the worst case only see the file not found error and not the one that relates to their file.

src/audio.zig
```diff
@@ -13,39 +40,41 @@ fn handleError(miniaudioError: c.ma_result) !void {
 }
 
 const AudioData = struct {
-	musicId: []const u8,
+	audioId: []const u8,
 	data: []f32 = &.{},
+	isMono: bool = false,
```
IntegratedQuantum:
```suggestion
	type: enum {mono, stereo} = .stereo,
```

src/audio.zig
```diff
@@ -76,26 +105,26 @@ const AudioData = struct {
...
 	pub fn equals(self: *const AudioData, _other: ?*const AudioData) bool {
 		if (_other) |other| {
-			return std.mem.eql(u8, self.musicId, other.musicId);
+			return std.mem.eql(u8, self.audioId, other.audioId);
 		} else return false;
 	}
 };
 
-var activeTasks: main.List([]const u8) = .empty;
+var activeTasks: main.List([]const u8) = .empty; // MARK: MUSIC
```
IntegratedQuantum: We don't SCREAM
```suggestion
var activeTasks: main.List([]const u8) = .empty; // MARK: Music
```

IntegratedQuantum (changes requested):

RanPix:
done

src/audio.zig
```diff
@@ -5,31 +5,37 @@ const utils = main.utils;
...
+	ogg_skeleton_not_supported = 38,
+};
+
+pub fn getStbVorbisError(result: c_int) []const u8 {
+	const resultEnum = std.enums.fromInt(StbErrorEnum, result) orelse {
+		std.log.err("Encountered an STB Vorbis error with unknown error code {}", .{result});
+		return "unknown_error";
 	};
+
+	return @tagName(resultEnum);
```
Wunka: why are you converting that? the `{t}` works on enums and errors

Wunka (changes requested):

src/audio.zig
RanPix: then i will have to add `unknown_error` to the enum

Wunka: I mean, why not?

RanPix:
done

src/audio.zig
```diff
@@ -13,39 +47,45 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
+		const path1 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "assets/{s}/{s}/{s}.ogg", .{addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path1);
 		var err: c_int = 0;
 		if (c.stb_vorbis_open_filename(path1.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/music/{s}.ogg", .{main.files.cubyzDirStr(), addon, fileName}, 0) catch unreachable;
+		if (err != @intFromEnum(StbVorbisErrorEnum.no_error)) {
+			std.log.err("Couldn't handle audio file. Error: {t}. ID: \"{s}\". Path: \"{s}\"", .{getStbVorbisError(err), id, path1});
+			return null;
+		}
+		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/{s}/{s}.ogg", .{main.files.cubyzDirStr(), addon, subPath, fileName}, 0) catch unreachable;
```
IntegratedQuantum, line 66: This code is unreachable since you always return before it.

src/audio.zig
```diff
@@ -13,39 +47,45 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
 		defer main.stackAllocator.free(path2);
 		if (c.stb_vorbis_open_filename(path2.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		std.log.err("Couldn't find music with id \"{s}\". Searched path \"{s}\" and \"{s}\"", .{id, path1, path2});
+		std.log.err("Couldn't handle or find audio file. Error: {t}. ID: \"{s}\". Searched path: \"{s}\" and \"{s}\"", .{getStbVorbisError(err), id, path1, path2});
```
IntegratedQuantum: Please print both errors here.

Returning on the first error leads to unrachable code, printing the error without returning above will lead to unavoidable error messages when the audio file is in the second path.

IntegratedQuantum (changes requested):

src/audio.zig
RanPix: what second error?

IntegratedQuantum: The one on the first path

src/audio.zig
RanPix: it only returns if there was an error with reading the file itself, or should it still search for another? i dont think this would make sense

IntegratedQuantum: It also returns when there was no error (when the return value is non-null)

RanPix: now better?

RanPix:
done (i think)

src/audio.zig
```diff
@@ -13,39 +47,45 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
-		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/music/{s}.ogg", .{main.files.cubyzDirStr(), addon, fileName}, 0) catch unreachable;
+		const err1 = err;
+		if (err != @intFromEnum(StbVorbisErrorEnum.file_open_failure)) {
+			std.log.err("Couldn't handle audio file. Error: {any}. ID: \"{s}\". Path: \"{s}\"", .{getStbVorbisError(err), id, path1});
+		}
+		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/{s}/{s}.ogg", .{main.files.cubyzDirStr(), addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path2);
 		if (c.stb_vorbis_open_filename(path2.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		std.log.err("Couldn't find music with id \"{s}\". Searched path \"{s}\" and \"{s}\"", .{id, path1, path2});
+		std.log.err("Couldn't handle or find audio file. Current Error: {any}. Previous Error: {any}. ID: \"{s}\". Searched path: \"{s}\" and \"{s}\"", .{getStbVorbisError(err), getStbVorbisError(err1), id, path1, path2});
```
IntegratedQuantum: I'd suggest this to make it clearer which error belongs to which path
```suggestion
		std.log.err("Couldn't handle or find audio file. ID: \"{s}\". Searched path: \"{s}\" (error: {any}) and \"{s}\" (error: {any})", .{id, path1, getStbVorbisError(err), path2, getStbVorbisError(err1)});
```

src/audio.zig
```diff
@@ -13,39 +47,45 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
 		};
 		const addon = id[0..colonIndex];
 		const fileName = id[colonIndex + 1 ..];
-		const path1 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "assets/{s}/music/{s}.ogg", .{addon, fileName}, 0) catch unreachable;
+		const path1 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "assets/{s}/{s}/{s}.ogg", .{addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path1);
 		var err: c_int = 0;
 		if (c.stb_vorbis_open_filename(path1.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/music/{s}.ogg", .{main.files.cubyzDirStr(), addon, fileName}, 0) catch unreachable;
+		const err1 = err;
```
IntegratedQuantum: This naming is confusing. Please make separate variables for err1 and err2 matching the path name.

src/audio.zig
```diff
@@ -13,39 +47,45 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
+		if (err != @intFromEnum(StbVorbisErrorEnum.file_open_failure)) {
+			std.log.err("Couldn't handle audio file. Error: {any}. ID: \"{s}\". Path: \"{s}\"", .{getStbVorbisError(err), id, path1});
```
IntegratedQuantum: Not really necessary, I'm also not certain if file_open_failure is the only error that can happen when the file is not found. So if the second path succeeds I think we shouldn't print any error.

IntegratedQuantum (changes requested):

RanPix:
done

src/audio.zig
```diff
@@ -13,39 +47,44 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
 		};
 		const addon = id[0..colonIndex];
 		const fileName = id[colonIndex + 1 ..];
-		const path1 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "assets/{s}/music/{s}.ogg", .{addon, fileName}, 0) catch unreachable;
+		const path1 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "assets/{s}/{s}/{s}.ogg", .{addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path1);
 		var err: c_int = 0;
 		if (c.stb_vorbis_open_filename(path1.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/music/{s}.ogg", .{main.files.cubyzDirStr(), addon, fileName}, 0) catch unreachable;
+		const err1 = err;
```
IntegratedQuantum: Just name it `var err1: c_int` before passing it to the function. It makes no sense to have a second variable for this.

src/audio.zig
```diff
@@ -13,39 +47,44 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
+
+		const path2 = std.fmt.allocPrintSentinel(main.stackAllocator.allocator, "{s}/serverAssets/{s}/{s}/{s}.ogg", .{main.files.cubyzDirStr(), addon, subPath, fileName}, 0) catch unreachable;
 		defer main.stackAllocator.free(path2);
 		if (c.stb_vorbis_open_filename(path2.ptr, &err, null)) |ogg_stream| return ogg_stream;
-		std.log.err("Couldn't find music with id \"{s}\". Searched path \"{s}\" and \"{s}\"", .{id, path1, path2});
+		const err2 = err;
```
IntegratedQuantum: Just name it `var err2: c_int` before passing it to the function. It makes no sense to have a second variable for this.

src/audio.zig
```diff
@@ -13,39 +47,44 @@ fn handleError(miniaudioError: c.ma_result) !void {
...
+		std.log.err("Couldn't handle or find audio file. ID: \"{s}\". Searched path: \"{s}\" (error: {any}) and \"{s}\" (error: {any})", .{id, path1, getStbVorbisError(err2), path2, getStbVorbisError(err1)});
```
IntegratedQuantum: This looks wrong
```suggestion
		std.log.err("Couldn't handle or find audio file. ID: \"{s}\". Searched path: \"{s}\" (error: {any}) and \"{s}\" (error: {any})", .{id, path1, getStbVorbisError(err1), path2, getStbVorbisError(err2)});
```

IntegratedQuantum (changes requested):

src/audio.zig
IntegratedQuantum: done

src/audio.zig
IntegratedQuantum: done

src/audio.zig
IntegratedQuantum: done

IntegratedQuantum (approved):

PR #3369, opened 2026-07-15
Converter/autocrafter

Crepestrom:
note:only works when you are looking at it
some issue with saving the block entities last time seems to be the issue i think


ikabod-kee:
This logic is good for furnaces.
But we should not have auto-crafters.

IntegratedQuantum:
Note that I will not merge any new block entities until #2564 is finished

du82:
I do not like auto-crafters

Crepestrom:
ah shame no new block entities
also no worries this is likely never coming out of drafts

PR #3370, opened 2026-07-16
Recipe Expansion

Crepestrom:
This pr adds crafting tags to all recipies (by default they all have .handCraftable)

used for next pr making crafting tables search the recipie tags


Crepestrom:
also i fixed the leak
i just didnt understand when different allocators leak and what they look like when they leak

IntegratedQuantum:
Did you intend to mark this as a draft?

Crepestrom:
i will test it to make sure it work

Crepestrom:
nope still has a issue with the players crafting tags being stuck when using a transmutation table

PR #3377, opened 2026-07-17
Make map generation 12% faster.

calebstenberg:
Instead of guessing around until we find a biome in the layer, just store what biomes are in the layer as well and use that to sample.
This brought generating a new world seed 101 at max render from 142 sec to 125 sec.

`generate` used to take 28% of CPU time and now takes 15.8% according to AMD uProf

Edit:
Now only down to 130 seconds. And uses 16.89% of CPU time

Edit 2:
Uses only 14.5% of CPU time

### DIFF
--- a/src/server/terrain/cave_layers.zig
+++ b/src/server/terrain/cave_layers.zig
@@ -17,6 +17,7 @@ pub const CaveLayer = struct {
 
 	biomes: main.utils.AliasTable(*const Biome),
 	id: []const u8,
+	tags: []const Tag,
 
 	pub fn init(id: []const u8, zon: ZonElement) ?CaveLayer {
 		var result: CaveLayer = undefined;
@@ -31,8 +32,8 @@ pub const CaveLayer = struct {
 		result.caveDensity = zon.get(f32, "caveDensity") orelse 1.0/32.0;
 		result.id = main.worldArena.dupe(u8, id);
 
-		const tags = Tag.loadTagsFromZon(main.stackAllocator, zon.getChild("tags"));
-		defer main.stackAllocator.free(tags);
+		const tags = Tag.loadTagsFromZon(main.worldArena, zon.getChild("tags"));
+		result.tags = tags;
 		if (tags.len == 0) {
 			std.log.err("Cave layer with id {s} is missing tags. Skipping", .{id});
 			return null;
@@ -43,30 +44,6 @@ pub const CaveLayer = struct {
 				return null;
 			}
 		}
-		var biomes: main.List(*const Biome) = .empty;
-		defer biomes.deinit(main.stackAllocator);
-		outer: for (terrain.biomes.getCaveBiomes()) |*biome| {
-			for (tags) |tag| {
-				if (biome.hasTag(tag)) {
-					biomes.append(main.stackAllocator, biome);
-					continue :outer;
-				}
-			}
-		}
-
-		if (biomes.items.len == 0) {
-			std.log.err("Cave layer with id {s} has no biomes that match the provided tags. Skipping", .{id});
-			return null;
-		}
-		for (biomes.items) |biome| {
-			if (biome.minHeight == std.math.minInt(i32) and biome.maxHeight == std.math.maxInt(i32) and biome.chance != 0) break;
-		} else {
-			std.log.err("Cave layer with id {s} has no biomes with unbounded height. At least one biome with unbounded height must exist to ensure compatibility with other addons. Skipping", .{id});
-			return null;
-		}
-
-		result.biomes = .init(main.worldArena, main.worldArena.dupe(*const Biome, biomes.items));
-
 		return result;
 	}
 };
@@ -111,10 +88,11 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 	var newLayers: main.List(CaveLayer) = .empty;
 
 	for (caveLayers.items) |layer| {
-		const split = splitLayer(main.worldArena, layer);
-
-		for (split.items) |newLayer| {
-			newLayers.append(main.worldArena, newLayer);
+		const splitLayers = splitLayer(main.worldArena, layer);
+		if (splitLayers) |layers| {
+			for (layers.items) |newLayer| {
+				newLayers.append(main.worldArena, newLayer);
+			}
 		}
 	}
 
@@ -126,10 +104,32 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 	}
 }
 
-fn splitLayer(allocator: NeverFailingAllocator, layer: CaveLayer) main.List(CaveLayer) {
+fn splitLayer(allocator: NeverFailingAllocator, layer: CaveLayer) ?main.List(CaveLayer) {
+	var possibleBiomes: main.List(*const Biome) = .empty;
+	defer possibleBiomes.deinit(main.stackAllocator);
+	outer: for (terrain.biomes.getCaveBiomes()) |*biome| {
+		for (layer.tags) |tag| {
+			if (biome.hasTag(tag)) {
+				possibleBiomes.append(main.stackAllocator, biome);
+				continue :outer;
+			}
+		}
+	}
+
+	if (possibleBiomes.items.len == 0) {
+		std.log.err("Cave layer with id {s} has no biomes that match the provided tags. Skipping", .{layer.id});
+		return null;
+	}
+	for (possibleBiomes.items) |biome| {
+		if (biome.minHeight == std.math.minInt(i32) and biome.maxHeight == std.math.maxInt(i32) and biome.chance != 0) break;
+	} else {
+		std.log.err("Cave layer with id {s} has no biomes with unbounded height. At least one biome with unbounded height must exist to ensure compatibility with other addons. Skipping", .{layer.id});
+		return null;
+	}
+
 	var splitPoints: main.List(i32) = .empty;
 	defer splitPoints.deinit(main.stackAllocator);
-	for (layer.biomes.items) |biome| {
+	for (possibleBiomes.items) |biome| {
 		if (biome.maxHeight > layer.minHeight and biome.maxHeight < layer.maxHeight) {
 			splitPoints.append(main.stackAllocator, biome.maxHeight);
 		}
@@ -155,13 +155,14 @@ fn splitLayer(allocator: NeverFailingAllocator, layer: CaveLayer) main.List(Cave
 	while (i <= splitPoints.items.len) : (i += 1) {
 		const point = if (i < splitPoints.items.len) splitPoints.items[i] else layer.maxHeight;
 		var biomes: main.List(*const Biome) = .empty;
-		for (layer.biomes.items) |biome| {
+		for (possibleBiomes.items) |biome| {
 			if (biome.maxHeight >= point and biome.minHeight <= lastPoint) {
 				biomes.append(allocator, biome);
 			}
 		}
 
 		newLayers.append(allocator, CaveLayer{
+			.tags = layer.tags,
 			.biomes = main.utils.AliasTable(*const Biome).init(allocator, biomes.toOwnedSlice(allocator)),
 			.caveDensity = layer.caveDensity,
 			.id = layer.id,
--- a/src/server/terrain/cave_layers.zig
+++ b/src/server/terrain/cave_layers.zig
@@ -111,15 +111,14 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 	var newLayers: main.List(CaveLayer) = .empty;
 
 	for (caveLayers.items) |layer| {
-		var split = splitLayer(layer);
-		defer split.deinit(main.worldArena);
+		var split = splitLayer(main.stackAllocator, layer);
+		defer split.deinit(main.stackAllocator);
 
 		for (split.items) |newLayer| {
 			newLayers.append(main.worldArena, newLayer);
 		}
 	}
 
-	caveLayers.deinit(main.worldArena);
 	caveLayers = newLayers;
 
 	std.log.debug("Registered cave layers:", .{});
@@ -128,15 +127,15 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 	}
 }
 
-fn splitLayer(layer: CaveLayer) main.List(CaveLayer) {
+fn splitLayer(allocator: NeverFailingAllocator, layer: CaveLayer) main.List(CaveLayer) {
 	var splitPoints: main.List(i32) = .empty;
 	defer splitPoints.deinit(main.stackAllocator);
 	for (layer.biomes.items) |biome| {
-		if (biome.*.maxHeight > layer.minHeight and biome.*.maxHeight < layer.maxHeight) {
-			splitPoints.append(main.stackAllocator, biome.*.maxHeight);
+		if (biome.maxHeight > layer.minHeight and biome.maxHeight < layer.maxHeight) {
+			splitPoints.append(main.stackAllocator, biome.maxHeight);
 		}
-		if (biome.*.minHeight < layer.maxHeight and biome.*.minHeight > layer.minHeight) {
-			splitPoints.append(main.stackAllocator, biome.*.minHeight);
+		if (biome.minHeight < layer.maxHeight and biome.minHeight > layer.minHeight) {
+			splitPoints.append(main.stackAllocator, biome.minHeight);
 		}
 	}
 	std.mem.sort(i32, splitPoints.items, {}, comptime std.sort.asc(i32));
@@ -159,12 +158,12 @@ fn splitLayer(layer: CaveLayer) main.List(CaveLayer) {
 		var biomes: main.List(*const Biome) = .empty;
 		for (layer.biomes.items) |biome| {
 			if (biome.maxHeight >= point and biome.minHeight <= lastPoint) {
-				biomes.append(main.worldArena, biome);
+				biomes.append(main.stackAllocator, biome);
 			}
 		}
 
-		newLayers.append(main.worldArena, CaveLayer{
-			.biomes = main.utils.AliasTable(*const Biome).init(main.worldArena, biomes.items),
+		newLayers.append(allocator, CaveLayer{
+			.biomes = main.utils.AliasTable(*const Biome).init(allocator, biomes.items.toOwnedSlice(main.worldArena)),
 			.caveDensity = layer.caveDensity,
 			.id = layer.id,
 			.layerHeight = point - lastPoint,
--- a/src/server/terrain/cave_layers.zig
+++ b/src/server/terrain/cave_layers.zig
@@ -107,10 +107,77 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 		height -= caveLayers.items[i].layerHeight;
 		caveLayers.items[i].minHeight = height;
 	}
+
+	var newLayers: main.List(CaveLayer) = .empty;
+
+	for (caveLayers.items) |layer| {
+		var split = splitLayer(layer);
+		defer split.deinit(main.worldArena);
+
+		for (split.items) |newLayer| {
+			newLayers.append(main.worldArena, newLayer);
+		}
+	}
+
+	caveLayers.deinit(main.worldArena);
+	caveLayers = newLayers;
+
 	std.log.debug("Registered cave layers:", .{});
 	for (caveLayers.items) |caveLayer| {
 		std.log.debug("{s}: {} to {}", .{caveLayer.id, caveLayer.minHeight, caveLayer.maxHeight});
 	}
+}
+
+fn splitLayer(layer: CaveLayer) main.List(CaveLayer) {
+	var splitPoints: main.List(i32) = .empty;
+	defer splitPoints.deinit(main.stackAllocator);
+	for (layer.biomes.items) |biome| {
+		if (biome.*.maxHeight > layer.minHeight and biome.*.maxHeight < layer.maxHeight) {
+			splitPoints.append(main.stackAllocator, biome.*.maxHeight);
+		}
+		if (biome.*.minHeight < layer.maxHeight and biome.*.minHeight > layer.minHeight) {
+			splitPoints.append(main.stackAllocator, biome.*.minHeight);
+		}
+	}
+	std.mem.sort(i32, splitPoints.items, {}, comptime std.sort.asc(i32));
+	// Remove duplicates in place
+	var write: usize = 0;
+	for (splitPoints.items) |value| {
+		if (write == 0 or splitPoints.items[write - 1] != value) {
+			splitPoints.items[write] = value;
+			write += 1;
+		}
+	}
+	splitPoints.shrinkAndFree(main.stackAllocator, write);
+	splitPoints.items.len = write;
+
+	var newLayers: main.List(CaveLayer) = .empty;
+	var lastPoint: i32 = layer.minHeight;
+	var i: usize = 0;
+	while (i <= splitPoints.items.len) : (i += 1) {
+		const point = if (i < splitPoints.items.len) splitPoints.items[i] else layer.maxHeight;
+		var biomes: main.List(*const Biome) = .empty;
+		for (layer.biomes.items) |biome| {
+			if (biome.maxHeight > point and biome.minHeight < lastPoint) {
+				biomes.append(main.worldArena, biome);
+			}
+			if (biome.maxHeight == point and biome.minHeight == lastPoint) {
+				biomes.append(main.worldArena, biome);
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


calebstenberg:
The problem is that the margin doesn't work any more. I will try a couple things

calebstenberg:
> The problem is that the margin doesn't work any more. I will try a couple things

Fixed it

src/server/terrain/cavebiomegen/RandomBiomeDistribution.zig
```diff
@@ -39,14 +35,9 @@ pub fn generate(map: *CaveBiomeMapFragment, worldSeed: u64) void {
 					const offset: Vec3i = @splat(if (_map == 0) 0 else CaveBiomeMapFragment.caveBiomeSize/2);
 					const biomeWorldPos = CaveBiomeMapFragment.rotateInverse(pos + offset);
 					const caveLayer = terrain.cave_layers.getLayer(biomeWorldPos[2]);
-					while (true) {
-						const biome = caveLayer.biomes.sample(&seed).*;
-						if (biome.minHeight < biomeWorldPos[2] + CaveBiomeMapFragment.caveBiomeSize*marginMulPositive/marginDiv and biome.maxHeight > biomeWorldPos[2] + CaveBiomeMapFragment.caveBiomeSize*marginMulNegative/marginDiv) {
```
IntegratedQuantum, line 44: To me it looks like you are completely disabling this feature (cave biomes with restricted heights).
Beyond layers we do want more fine-grained control over biomes, such that e.g. crystal caves don't spawn too close to the surface.

To cover this use-case you'd need to split the layers along the height requirements of the biomes.

src/server/terrain/cave_layers.zig
```diff
@@ -54,6 +59,17 @@ pub const CaveLayer = struct {
 			}
 		}
 
+		var layerBiomes: main.List(*const Biome) = .empty;
+		defer layerBiomes.deinit(main.stackAllocator);
+
+		for (biomes.items) |biome| {
+			if (biome.maxHeight > result.minHeight + CaveBiomeMapFragment.caveBiomeSize*marginMulNegative/marginDiv and
+				biome.minHeight < result.maxHeight + CaveBiomeMapFragment.caveBiomeSize*marginMulPositive/marginDiv)
```
IntegratedQuantum: It makes no sense that it needs the margin.

IntegratedQuantum (changes requested):
This is really interesting. I didn't know the random sampling was so influential.

src/server/terrain/cave_layers.zig
calebstenberg: Is this a margin that increases the possible configurations or decreases it? It was a little hard to tell for me

IntegratedQuantum: Ideally it shouldn't be needed at all. I can't remember why it was added.

calebstenberg:
Even with a faster layer search it still takes 17% of cpu time, but it's still better than 28%

src/server/terrain/cavebiomegen/RandomBiomeDistribution.zig
calebstenberg: I put the while loop back, you would have to store more AliasTables in cave_layers, which maybe I could try actually.

IntegratedQuantum: My suggestion is to store more Cave layers, that have duplicate properties, but a different alias table. Then you would not need any extra code to select the right alias table, it would automatically be handled by layer selection.

IntegratedQuantum: I think without this you are not going to get much long-term advantage out of it, e.g. #3360 will completely foil this optimization since it basically does the same thing in assets by giving the surface cave layer a distinct tag.

src/server/terrain/cave_layers.zig
calebstenberg: Removed

src/server/terrain/cavebiomegen/RandomBiomeDistribution.zig
calebstenberg: Duplicate minus the min and max heights though right? Otherwise layer selection would need to change.

IntegratedQuantum: Yeah, the min and max heights need to be adjusted to match the biome heights that you are splitting the layers for.

calebstenberg:
Now it is only 14.5% of CPU time!

calebstenberg:
This effect is even more pronounced in layers with few biomes like the sky or void.

src/server/terrain/cave_layers.zig
```diff
@@ -133,6 +254,19 @@ pub fn getLayer(height: i32) CaveLayer {
 	return caveLayers.items[minIndex];
 }
 
+pub fn getLayerGuess(height: i32, i: *usize) CaveLayer {
```
IntegratedQuantum: I think it would be best to have this in a separate PR so we can evaluate the changes independently.
I also think there are possibly faster (and branchless) solutions here with some precalculations and the use of `@popCount`.

src/server/terrain/cave_layers.zig
```diff
@@ -16,7 +17,9 @@ pub const CaveLayer = struct {
 	caveDensity: f32,
 
 	biomes: main.utils.AliasTable(*const Biome),
+	layerBiomes: main.utils.AliasTable(*const Biome),
```
IntegratedQuantum: There is no need to have two alias tables

src/server/terrain/cave_layers.zig
```diff
@@ -66,9 +68,32 @@ pub const CaveLayer = struct {
 		}
 
 		result.biomes = .init(main.worldArena, main.worldArena.dupe(*const Biome, biomes.items));
```
IntegratedQuantum: The biome list should be determined in `registerCaveLayers` before you subdivide the layer.
Then you don't need to clone and avoid needless allocations.

src/server/terrain/cave_layers.zig
```diff
@@ -107,6 +124,106 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
+	var iterator = caveLayerMap.iterator();
+	while (iterator.next()) |entry| {
+		register(entry.key_ptr.*, entry.value_ptr.*);
+	}
+
+	sortAndFitLayers();
+
+	var i: usize = 0;
+	for (caveLayers.items[i..]) |*caveLayer| {
```
IntegratedQuantum: I'd suggest to merge this with the other loop, there is no reason to keep them separate.

Also I'd suggest to make a separate function for the splitting layers stuff.

src/server/terrain/cave_layers.zig
```diff
@@ -107,6 +124,106 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+
+		if (heights.items.len == 0) {
+			caveLayers.append(main.worldArena, baseLayer);
+			continue;
+		}
+
+		std.mem.sort(i32, heights.items, {}, intLessThan);
+		var layerStart = baseLayer.minHeight;
+		i = 0;
+		while (i <= heights.items.len) : (i += 1) {
```
IntegratedQuantum: Use a for-loop unless you really need a while
```suggestion
		for (0..heights.items.len) |i| {
```

src/server/terrain/cave_layers.zig
```diff
@@ -107,6 +124,106 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+
+			layerStart = layerEnd;
+		}
+		baseLayer.deinit();
+	}
+
+	std.debug.assert(!finishedLoading);
+	finishedLoading = true;
+
+	sortAndFitLayers();
```
IntegratedQuantum: They should already be sorted at this point, if you append them in the right order.

IntegratedQuantum (changes requested):

src/server/terrain/cavebiomegen/RandomBiomeDistribution.zig
```diff
@@ -38,15 +35,10 @@ pub fn generate(map: *CaveBiomeMapFragment, worldSeed: u64) void {
...
-					const caveLayer = terrain.cave_layers.getLayer(biomeWorldPos[2]);
-					while (true) {
-						const biome = caveLayer.biomes.sample(&seed).*;
-						if (biome.minHeight < biomeWorldPos[2] + CaveBiomeMapFragment.caveBiomeSize*marginMulPositive/marginDiv and biome.maxHeight > biomeWorldPos[2] + CaveBiomeMapFragment.caveBiomeSize*marginMulNegative/marginDiv) {
-							const index = CaveBiomeMapFragment.getIndex(x, y, z);
-							map.biomeMap[index][_map] = biome;
-							break;
-						}
-					}
+					caveLayer = terrain.cave_layers.getLayer(biomeWorldPos[2]);
```
IntegratedQuantum:
```suggestion
					const caveLayer = terrain.cave_layers.getLayer(biomeWorldPos[2]);
```

src/server/terrain/cavebiomegen/RandomBiomeDistribution.zig
```diff
@@ -23,11 +23,8 @@ pub fn init(parameters: ZonElement) void {
 }
 
 pub fn generate(map: *CaveBiomeMapFragment, worldSeed: u64) void {
-	const marginDiv = 1024;
-	const marginMulPositive: comptime_int = comptime CaveBiomeMapFragment.rotateInverse(.{marginDiv, 0, marginDiv})[2];
-	const marginMulNegative: comptime_int = comptime CaveBiomeMapFragment.rotateInverse(.{0, marginDiv, 0})[2];
-
 	var seed = random.initSeed3D(worldSeed, .{map.pos.wx, map.pos.wy, map.pos.wz});
+	var caveLayer = terrain.cave_layers.getLayer(map.pos.wz);
```
IntegratedQuantum:
```suggestion
```

src/server/terrain/cave_layers.zig
```diff
@@ -31,42 +32,17 @@ pub const CaveLayer = struct {
...
-		}
-
-		if (biomes.items.len == 0) {
-			std.log.err("Cave layer with id {s} has no biomes that match the provided tags. Skipping", .{id});
-			return null;
-		}
-		for (biomes.items) |biome| {
-			if (biome.minHeight == std.math.minInt(i32) and biome.maxHeight == std.math.maxInt(i32) and biome.chance != 0) break;
-		} else {
-			std.log.err("Cave layer with id {s} has no biomes with unbounded height. At least one biome with unbounded height must exist to ensure compatibility with other addons. Skipping", .{id});
```
IntegratedQuantum, line 64: Please readd the error messages in the other code.

src/server/terrain/cave_layers.zig
```diff
@@ -107,18 +81,107 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+	}
+
+	//Sort new by
+	std.mem.sort(CaveLayer, caveLayers.items, {}, lessThan);
+	//filter biomes
+	for (0..caveLayers.items.len) |i| {
+		var biomes: main.List(*const Biome) = .empty;
+		defer biomes.deinit(main.stackAllocator);
+		outer: for (terrain.biomes.getCaveBiomes()) |*biome| {
+			for (caveLayers.items[i].tags) |tag| {
```
IntegratedQuantum: You should not discover the biomes twice.

src/server/terrain/cave_layers.zig
```diff
@@ -107,18 +81,107 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 		height -= caveLayers.items[i].layerHeight;
 		caveLayers.items[i].minHeight = height;
 	}
+
+	splitCaveLayers();
```
IntegratedQuantum: When I asked you to make a function I was hoping for something like this:
```
const fullLayers = main.stackAllocator.dupe(CaveLayer, caveLayers.items);
defer main.stackAllocator.free(fullLayers);
caveLayers.clearRetainingCapacity();
std.log.debug("Registered cave layers:", .{});
for (fullLayers) |layer| {
	std.log.debug("{s}: {} to {}", .{caveLayer.id, caveLayer.minHeight, caveLayer.maxHeight});
	splitCaveLayer(layer);
}
```

The other function is still kind of messy, since it has two for loops and quite large indentation.

src/server/terrain/cave_layers.zig
```diff
@@ -107,18 +81,107 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+	//Remember where we split
+	var splitHeights: main.List(i32) = .empty;
+	defer splitHeights.deinit(main.stackAllocator);
+
+	//Add duplicate layers for when we split
+	const original_len = caveLayers.items.len;
+	for (0..original_len) |i| {
+		outer: for (terrain.biomes.getCaveBiomes()) |*biome| {
+			for (caveLayers.items[i].tags) |tag| {
+				if (biome.hasTag(tag)) {
```
IntegratedQuantum: Please separate out the biome discovery process by writing all biomes of the layer into a temporary list.

src/server/terrain/cave_layers.zig
```diff
@@ -107,18 +81,107 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+}
+
 fn lessThan(_: void, lhs: CaveLayer, rhs: CaveLayer) bool {
 	if (lhs.depthHint < rhs.depthHint) return true;
 	if (lhs.depthHint > rhs.depthHint) return false;
+	if (lhs.layerHeight < rhs.layerHeight) return true;
 	return std.ascii.orderIgnoreCase(lhs.id, rhs.id) == .gt;
 }
 
+fn intLessThan(_: void, a: i32, b: i32) bool {
```
IntegratedQuantum: Unused

src/server/terrain/cave_layers.zig
```diff
@@ -107,18 +81,107 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+		}
+		std.debug.assert(biomes.items.len > 0);
+		caveLayers.items[i].biomes = .init(main.worldArena, main.worldArena.dupe(*const Biome, biomes.items));
+	}
+}
+
 fn lessThan(_: void, lhs: CaveLayer, rhs: CaveLayer) bool {
 	if (lhs.depthHint < rhs.depthHint) return true;
 	if (lhs.depthHint > rhs.depthHint) return false;
+	if (lhs.layerHeight < rhs.layerHeight) return true;
```
IntegratedQuantum: Could you explain why you added this?

IntegratedQuantum (changes requested):

calebstenberg:
I'm on a business trip and won't be working on this for about a week just fyi

src/server/terrain/cave_layers.zig
```diff
@@ -107,12 +107,76 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 		height -= caveLayers.items[i].layerHeight;
 		caveLayers.items[i].minHeight = height;
 	}
+
+	var newLayers: main.List(CaveLayer) = .empty;
+
+	for (caveLayers.items) |layer| {
+		var split = splitLayer(layer);
+		defer split.deinit(main.worldArena);
```
IntegratedQuantum: All local allocations should use the stackAllocator

src/server/terrain/cave_layers.zig
```diff
@@ -107,12 +107,76 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 		height -= caveLayers.items[i].layerHeight;
 		caveLayers.items[i].minHeight = height;
 	}
+
+	var newLayers: main.List(CaveLayer) = .empty;
+
+	for (caveLayers.items) |layer| {
+		var split = splitLayer(layer);
```
IntegratedQuantum: All functions that return a result to be freed should take an allocator parameter

src/server/terrain/cave_layers.zig
```diff
@@ -107,12 +107,76 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+		defer split.deinit(main.worldArena);
+
+		for (split.items) |newLayer| {
+			newLayers.append(main.worldArena, newLayer);
+		}
+	}
+
+	caveLayers.deinit(main.worldArena);
```
IntegratedQuantum: This is a no-op. You cant really free from an arena.

src/server/terrain/cave_layers.zig
```diff
@@ -107,12 +107,76 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
 	for (caveLayers.items) |caveLayer| {
 		std.log.debug("{s}: {} to {}", .{caveLayer.id, caveLayer.minHeight, caveLayer.maxHeight});
 	}
 }
 
+fn splitLayer(layer: CaveLayer) main.List(CaveLayer) {
+	var splitPoints: main.List(i32) = .empty;
+	defer splitPoints.deinit(main.stackAllocator);
+	for (layer.biomes.items) |biome| {
+		if (biome.*.maxHeight > layer.minHeight and biome.*.maxHeight < layer.maxHeight) {
```
IntegratedQuantum: Please remove these unnecessary `.*` everywhere. There are not needed for pointer field accesses
```suggestion
		if (biome.maxHeight > layer.minHeight and biome.maxHeight < layer.maxHeight) {
```

src/server/terrain/cave_layers.zig
```diff
@@ -107,12 +107,76 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+		const point = if (i < splitPoints.items.len) splitPoints.items[i] else layer.maxHeight;
+		var biomes: main.List(*const Biome) = .empty;
+		for (layer.biomes.items) |biome| {
+			if (biome.maxHeight >= point and biome.minHeight <= lastPoint) {
+				biomes.append(main.worldArena, biome);
+			}
+		}
+
+		newLayers.append(main.worldArena, CaveLayer{
+			.biomes = main.utils.AliasTable(*const Biome).init(main.worldArena, biomes.items),
```
IntegratedQuantum: toOwnedSlice makes the intent clearer
```suggestion
			.biomes = main.utils.AliasTable(*const Biome).init(main.worldArena, biomes.items.toOwnedSlice(main.worldArena)),
```

IntegratedQuantum (changes requested):
Also https://github.com/PixelGuys/Cubyz/pull/3377/changes#r3607941472 still applies

calebstenberg:
> Also https://github.com/PixelGuys/Cubyz/pull/3377/changes#r3607941472 still applies

How? Do you want it to not call splitLayers at all? If so then where? asset.zig where registerLayers is called?
I also can't think of a way to have the layers initially and then rebuild them with a bigger list without just wasting an allocation

IntegratedQuantum:
No I just want you to move the code that determines the biomes of each layer into the splitLayers function so you don't need to allocate pointless copies of the biome lists into the worldArena.

src/server/terrain/cave_layers.zig
```diff
@@ -107,12 +84,98 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
 		height -= caveLayers.items[i].layerHeight;
 		caveLayers.items[i].minHeight = height;
 	}
+
+	var newLayers: main.List(CaveLayer) = .empty;
+
+	for (caveLayers.items) |layer| {
+		const splitLayers = splitLayer(main.worldArena, layer);
```
IntegratedQuantum: Please don't store this temporary list in the worldArena.
Maybe it would be easiest to just pass a pointer to `newLayers`, this would avoid the need for a temporary list here.

src/server/terrain/cave_layers.zig
```diff
@@ -107,12 +84,98 @@ pub fn registerCaveLayers(caveLayerMap: *Assets.ZonHashMap) !void {
...
+
+	if (possibleBiomes.items.len == 0) {
+		std.log.err("Cave layer with id {s} has no biomes that match the provided tags. Skipping", .{layer.id});
+		return null;
+	}
+	for (possibleBiomes.items) |biome| {
+		if (biome.minHeight == std.math.minInt(i32) and biome.maxHeight == std.math.maxInt(i32) and biome.chance != 0) break;
+	} else {
+		std.log.err("Cave layer with id {s} has no biomes with unbounded height. At least one biome with unbounded height must exist to ensure compatibility with other addons. Skipping", .{layer.id});
+		return null;
```
IntegratedQuantum: I guess it isn't as easy as I thought, sorry for that.
We cannot discard layers after sorting, that lead to gaps in the layer structure, which leads to unspecified behavior while sampling the layers.
Could you move this check back up top where it originally was? Without needing to allocate the biome list you can shorten it down to just
```
	outer: for (terrain.biomes.getCaveBiomes()) |*biome| {
		for (layer.tags) |tag| {
			if (biome.minHeight == std.math.minInt(i32) and biome.maxHeight == std.math.maxInt(i32) and biome.chance != 0) break :outer;
		}
	} else {
		std.log.err("Cave layer with id {s} has no biomes with unbounded height. At least one biome with unbounded height must exist to ensure compatibility with other addons. Skipping", .{layer.id});
		return null;
	}
```
And the check for `length == 0` isn't even needed, this second condition is much stronger.

IntegratedQuantum (changes requested):

src/server/terrain/cave_layers.zig
calebstenberg: > We cannot discard layers after sorting, that lead to gaps in the layer structure, which leads to unspecified behavior while sampling the layers.

It doesn't lead to unspecified behavior because it re-builds the full layer in pieces, so there are no gaps.
But I do see the benefit of keeping the existing layers, so I will try something.

calebstenberg: The problem is that CaveLayers is const so I can't modify the original to have the correct fields

calebstenberg: I would want to do 
```zig
    while (i <= splitPoints.items.len) : (i += 1) {
        const point = splitPoints.items[i];
        var biomes: main.List(*const Biome) = .empty;
        for (layer.biomes.items) |biome| {
            if (biome.maxHeight >= point and biome.minHeight <= lastPoint) {
                biomes.append(allocator, biome);
            }
        }
        if (i < splitPoints.items.len) {

            newLayers.append(allocator, CaveLayer{
                .biomes = main.utils.AliasTable(*const Biome).init(allocator, biomes.toOwnedSlice(allocator)),
                .caveDensity = layer.caveDensity,
                .id = layer.id,
                .layerHeight = point - lastPoint,
                .depthHint = layer.depthHint,
                .minHeight = lastPoint,
                .maxHeight = point,
            });
            lastPoint = point;
        } else {
//HERE
            layer.minHeight = lastPoint; 
            layer.layerHeight = layer.maxHeight - layer.minHeight;
            layer.biomes = main.utils.AliasTable(*const Biome).init(allocator, biomes.toOwnedSlice(allocator));
        }
    }
```
But I can't change layer.minHeight or layer.layerHeight or layer.biomes, so I would need to initialize it to the correct values, the problem is the I split the layers based on the original values

src/server/terrain/cave_layers.zig
calebstenberg: Problem is I need world alloc biomes and world alloc layers. Would you like me to pass both a pointer and an allocator? Then I can allocate biomes and update a single list.

src/server/terrain/cave_layers.zig
IntegratedQuantum: > It doesn't lead to unspecified behavior because it re-builds the full layer in pieces

What I mean by gaps is that maxHeight of one layer is not equal to minHeight of the next layer. So it would depend on the search algorithm which one is selected.

src/server/terrain/cave_layers.zig
IntegratedQuantum, line 66: Why is this back here again?
What I mean in https://github.com/PixelGuys/Cubyz/pull/3377#discussion_r3660019891 was only moving the checks, not the biome list generation. (That's why the code I gave you doesn't use the biome list)

src/server/terrain/cave_layers.zig
IntegratedQuantum: Yes. What I propose is just this:
```
for (caveLayers.items) |layer| {
	splitLayer(main.worldArena, &newLayers, layer);
}
```

IntegratedQuantum (changes requested):

calebstenberg:
I'm closing this because the performance is worse, it is only slightly better than normal when using a better layer search, and I can't understand the changes requested.

PR #3378, opened 2026-07-17
Flat modifiers

Crepestrom:
Allows for items to have flat bonuses as well as percentage bonuses
Usefull for addons and creating even more intersting materials

Also changes uranium to have less mass damage and hardness damage in exchange for a damage boost on the modifier
image

### DIFF
--- a/assets/cubyz/items/uranium_ingot.zig.zon
+++ b/assets/cubyz/items/uranium_ingot.zig.zon
@@ -3,12 +3,18 @@
 	.texture = "uranium_ingot.png",
 	.material = .{
 		.durability = 1500,
-		.massDamage = 10.0,
-		.hardnessDamage = 15.0,
+		.massDamage = 8.0,
+		.hardnessDamage = 12.0,
 		.swingSpeed = 0.5,
 		.textureRoughness = 0.3,
 		.colors = .{
 			0xff395533, 0xff466e45, 0xff5d8a5f, 0xff6da367, 0xff74c761,
+		},
+		.modifiers = .{
+			.{
+				.id = "powerful",
+				.flatStrength = 5,
+			},
 		},
 	},
 }
--- a/assets/cubyz/items/uranium_ingot.zig.zon
+++ b/assets/cubyz/items/uranium_ingot.zig.zon
@@ -13,7 +13,7 @@
 		.modifiers = .{
 			.{
 				.id = "powerful",
-				.flatStrength = 5,
+				.flatStrength = 2,
 			},
 		},
 	},
--- a/src/items.zig
+++ b/src/items.zig
@@ -157,6 +157,8 @@ const Modifier = struct {
 	pub const VTable = struct {
 		const Data = packed struct(u128) { pad: u128 };
 		combineModifiers: *const fn (data1: Data, data2: Data) ?Data,
+		changeProceduralItemParametersFlat: *const fn (proceduralItem: *ProceduralItem, data: Data) void,
+		changeProceduralItemParametersMult: *const fn (proceduralItem: *ProceduralItem, data: Data) void,
 		changeProceduralItemParameters: *const fn (proceduralItem: *ProceduralItem, data: Data) void,
 		changeBlockDamage: *const fn (damage: f32, block: Block, data: Data) f32,
 		printTooltip: *const fn (outString: *main.ListManaged(u8), data: Data) void,
@@ -165,6 +167,8 @@ const Modifier = struct {
 
 		const Defaults = struct {
 			pub fn changeProceduralItemParameters(_: *ProceduralItem, _: Data) void {}
+			pub fn changeProceduralItemParametersFlat(_: *ProceduralItem, _: Data) void {}
+			pub fn changeProceduralItemParametersMult(_: *ProceduralItem, _: Data) void {}
 			pub fn changeBlockDamage(damage: f32, _: Block, _: Data) f32 {
 				return damage;
 			}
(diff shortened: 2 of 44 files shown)


src/proceduralItem/modifiers/bad_at.zig
```diff
@@ -3,26 +3,34 @@ const std = @import("std");
...
 	for (block.tags()) |tag| {
-		if (tag == data.tag) return damage*(1 - data.strength);
+		if (tag == data.tag) return @max(damage - data.flatStrength, 0)*(1 - data.multStrength);
 	}
 	return damage;
 }
 
 pub fn printTooltip(outString: *main.ListManaged(u8), data: Data) void {
-	outString.print("#a00050**Bad at**#808080 *Decreases damage by **{d:.0}%** on \n***#a00050{s}#808080*** blocks", .{data.strength*100, data.tag.getName()});
+	outString.print("#a00050**Bad at**#808080 *Decreases damage by **{d:.0}%** and **-{d:.0}** on \n***#a00050{s}#808080*** blocks", .{data.multStrength*100, data.flatStrength, data.tag.getName()});
```
SpellDigger: the code should consider cases where the modifier has only multiplicative and flat strength

in my opinion the code should handle the cases and print a proper description fitting the modifier

assets/cubyz/items/uranium_ingot.zig.zon
```diff
@@ -3,12 +3,18 @@
...
 		.textureRoughness = 0.3,
 		.colors = .{
 			0xff395533, 0xff466e45, 0xff5d8a5f, 0xff6da367, 0xff74c761,
 		},
+		.modifiers = .{
+			.{
+				.id = "powerful",
+				.flatStrength = 5,
+			},
+		},
```
SpellDigger, line 18: i dont think we should make uranium grant extra flat damage from a modifier in exchange for making it deal less damage

even though it *could* benifit pickaxes that use 1 uranium it would make using uranium more than once even less usefull than it is now

SpellDigger (commented):
as a side note, the way modifiers are applied are out of order, if we are really going to do this you should find a way to put the modifiers on some order so flat bonuses get added before/after multiplicative bonuses

assets/cubyz/items/uranium_ingot.zig.zon
Crepestrom: i meant for it give people a reason to put uranium in other parts of the tool
like in the handle for more durability with a bit of flat damage on top

Crepestrom:
yeah though it technically does have an order
its powerful first then good_at
though there is no order between powerful and weak

SpellDigger:
> yeah though it technically does have an order its powerful first then good_at though there is no order between powerful and weak

it infact does not have an order, its just the way good_at works imitates an order becuase the effect is applied when mining not on the tool stats itself

Crepestrom:
ye

SpellDigger:
yea so im asking YOU to make sure theres some sort of order between flat and multiplicative property modifications

Crepestrom:
ye i am doing that

src/items.zig
```diff
@@ -501,6 +515,12 @@ const ProceduralItemPhysics = struct { // MARK: ProceduralItemPhysics
...
 		proceduralItem.modifiers = main.globalAllocator.dupe(Modifier, tempModifiers.items);
+		for (tempModifiers.items) |mod| {
+			mod.changeProceduralItemParametersFlat(proceduralItem);
+		}
+		for (tempModifiers.items) |mod| {
+			mod.changeProceduralItemParametersMult(proceduralItem);
+		}
 		for (tempModifiers.items) |mod| {
 			mod.changeProceduralItemParameters(proceduralItem);
 		}
```
SpellDigger, line 526: this is a very inefficient way to do it, change the implementation to only need a single for loop

src/proceduralItem/modifiers/heavy.zig
```diff
@@ -3,22 +3,35 @@ const std = @import("std");
...
 
-pub fn changeProceduralItemParameters(proceduralItem: *ProceduralItem, data: Data) void {
-	proceduralItem.setProperty(.swingSpeed, proceduralItem.getProperty(.swingSpeed)*(1 - data.strength));
+pub fn changeProceduralItemParametersFlat(proceduralItem: *ProceduralItem, data: Data) void {
+	proceduralItem.setProperty(.swingSpeed, proceduralItem.getProperty(.swingSpeed) - data.flatStrength);
+}
+
+pub fn changeProceduralItemParametersMult(proceduralItem: *ProceduralItem, data: Data) void {
+	proceduralItem.setProperty(.swingSpeed, proceduralItem.getProperty(.swingSpeed)*(1 - data.multStrength));
 }
```
SpellDigger, line 30: this is unecessary bloat, change the implementation in such a way that only changeProceduralItemParameters is needed

SpellDigger: ngl scrap that, u prob could get away with 2 functions instead of 3 mostly to account for cases like single_use having changeProceduralItemParameters override the properties

Wunka (changes requested):
You are changing a lot here.
1. please state an exact reason for this. Why should this be added?
2. please only touch the powerfull modifier and not rename strength to "multStrength", to have the least amount of changes. (I would even only keep the powerfull changes for testing and not to merge, as its another question which modifers should get this)
3. as with powerfull, the uranium change will only be for testing should not be merged at the end.

IntegratedQuantum:
I think we should try to reduce the number of modifiers, the current system was made in a time where we didn't have to consider armor and accessories yet. Duplicating this logic 1 million times is not very fun.

Crepestrom:
we could combine every single basic stat modifier into one modifier

PR #3382, opened 2026-07-18
refactor: optimize cave sphere shaping

BoringOrng:


IntegratedQuantum (changes requested):
For performance improvements please always make and provide performance measurement results (otherwise you are guessing and not optimizing).
Measuring performance is really simple. The simplest approach would be to just put the following at the start of the `generate` function.
```zig
const t1 = main.timestamp();
defer std.log.debug("{}", .{main.timestamp().nanoseconds - t1.nanoseconds});
```

To be honest a lot of your code looks like pointless refactoring. e.g. moving addRange out of the loop makes no difference since addTerrain is comptime, precalculating some properties upfront is likely something the compiler already does on its own (and renaming them in a way that is different from established conventions is also not good in my opinion), moving displacement calculations to a separate function makes no difference.
There are many pitfalls in performance optimizations and it is even possible to repeatedly measure a performance improvement despite not changing a single line of the measured code. So without knowing what you actually did to make it faster, I cannot assess whether it could be a mirage or in the best case provide ideas how to improve it further.

BoringOrng:
Alright, that's on me for not including benches, but here's what I injected for benching:

```Zig
var sum: i96 = 0;
var count: usize = 0;

...
	const t1 = main.timestamp();

	defer {
		count += 1;
		sum += main.timestamp().nanoseconds - t1.nanoseconds;

		if (count%10 == 0)
			std.log.debug("avg gen_time: {}", .{@as(f64, @floatFromInt(sum))/@as(f64, @floatFromInt(count))});
	}
```
End-to-end, after letting the world settle for a bit to get consistent load:
Before: `avg gen_time: 638.078...`
After: `avg gen_time: 557.478...`

That's about a 13% reduction in total time spent in `generateSphere_`. Earlier in the dev channed I'd mentioned a ~70% number from a session where I can't remember who profiled this with AMD uProf, that number was self-time in this function specifically, which mostly excludes the `addRange`/`removeRange` calls. So it's measuring how much faster the loops own arithmetic got, not the functions total cost, which is why it doesn't match the 13% here.

The actual speedup here involves how the z-loop was previously recomputing `distToCenterZ = (curZ - relZ) / radius` every iteration, and floating-point division is expensive on basically every CPU generation. I hoisted `invRadiusSq = 1.0 / (radius * radius)` out of the loop once, and turned the per-voxel check into a multiply instead.

Everything else in the diff is readability-only and isn't meant to be doing any performance-work.

IntegratedQuantum:
Oh interesting, since the main performance difference comes from floating point maths, I wonder if we can get even bigger performance improvements with `@setFloatMode(.optimized)`?

src/server/terrain/cavegen/FractalCaveGenerator.zig
```diff
@@ -58,72 +58,88 @@ pub fn generate(map: *CaveMapFragment, worldSeed: u64) void {
...
+		.min = @max(@as(i32, @trunc(center - extent)) - 1, 0),
+		.max = @min(
+			@as(i32, @trunc(center + extent)) + 1,
+			max_extent,
+		),
+	};
+}
+
 fn generateSphere_(seed: *u64, map: *CaveMapFragment, relPos: Vec3f, radius: f32, comptime addTerrain: bool) void {
+	const lodScale = map.pos.voxelSize;
```
IntegratedQuantum: Please stick to the existing naming convention
```suggestion
	const voxelSize = map.pos.voxelSize;
```

src/server/terrain/cavegen/FractalCaveGenerator.zig
```diff
@@ -58,72 +58,88 @@ pub fn generate(map: *CaveMapFragment, worldSeed: u64) void {
...
+	const scaledWidth = CaveMapFragment.width*lodScale;
+	const scaledHeight = CaveMapFragment.height*lodScale;
+
+	// Makes walls rough by adding a 1-in-roughnessChance chance that blocks
+	// remain unchanged.
+	const roughnessChance = 6;
+	const terrainShaper = if (addTerrain) CaveMapFragment.addRange else CaveMapFragment.removeRange;
```
IntegratedQuantum: I think function pointers are generally harder to read and argue about.
How about instead adding a helper function to `CaveMapFragment` called `modifyTerrain` that gets the same comptime parameter?

src/server/terrain/cavegen/FractalCaveGenerator.zig
```diff
@@ -58,72 +58,88 @@ pub fn generate(map: *CaveMapFragment, worldSeed: u64) void {
...
-	xMin = @max(xMin, 0);
-	var xMax = @as(i32, @trunc(relX + radius)) + 1;
-	xMax = @min(xMax, CaveMapFragment.width*map.pos.voxelSize);
-	var yMin = @as(i32, @trunc(relY - radius)) - 1;
-	yMin = @max(yMin, 0);
-	var yMax = @as(i32, @trunc(relY + radius)) + 1;
-	yMax = @min(yMax, CaveMapFragment.width*map.pos.voxelSize);
-	if (xMin >= xMax or yMin >= yMax or relZ - radius + 1 >= @as(f32, @floatFromInt(CaveMapFragment.height*map.pos.voxelSize)) or relZ + radius + 1 < 0) {
+
+	const xDist = displaceByHalfExtentBounded(relX, radius, scaledWidth);
```
IntegratedQuantum: If  you make `displaceByHalfExtentBounded` return just `struct {i32, i32}`, then you could use tuple destructuring here.
Furthermore nowadays we usually prefer vectors over seperate x/y/z variables, also I do think the name is a bit awkward.
Putting it all together
```suggestion
	const min, const max = getSphereBounds(.{relX, relY, relZ}, radius, scaledWidth);
```

src/server/terrain/cavegen/FractalCaveGenerator.zig
```diff
@@ -58,72 +58,88 @@ pub fn generate(map: *CaveMapFragment, worldSeed: u64) void {
 	}
 }
 
+fn displaceByHalfExtentBounded(center: f32, extent: f32, max_extent: i32) struct {
+	min: i32,
+	max: i32,
+} {
```
IntegratedQuantum: See the other comment (also name convention is camel case)
```suggestion
fn getSphereBounds(center: Vec3f, radius: f32, maxExtent: i32) struct { Vec3i, Vec3i } {
```

src/server/terrain/cavegen/FractalCaveGenerator.zig
```diff
@@ -58,72 +58,88 @@ pub fn generate(map: *CaveMapFragment, worldSeed: u64) void {
...
+	var curX = xDist.min;
+	while (curX < xDist.max) : (curX += lodScale) {
+		const dx = @as(f32, @floatFromInt(curX)) - relX;
+		const dxSq = dx*dx;
+
+		var curY = yDist.min;
+		while (curY < yDist.max) : (curY += lodScale) {
+			const dy = @as(f32, @floatFromInt(curY)) - relY;
+			const dySq = dy*dy;
+			const xySumSq = dxSq + dySq;
```
IntegratedQuantum: There should be no performance difference to do the squaring here, avoiding the helper constants
Also I'd prefer the full name, as abbreviations can be confusing to outsiders.
```suggestion
			const xySumSquare = dx*dx + dy*dy;
```

IntegratedQuantum (changes requested):
Some proposed improvements for the refactoring

src/server/terrain/cavegen/FractalCaveGenerator.zig
BoringOrng: I do like the idea of this, but I will challenge doing the *same* comptime-parameter, if it were up to me, I think a signature that looks more or less like this would be more explicit as to what is being done:

`modifyTerrain(self: *CaveMapFragment, modifier: enum { add, remove }, ...)`

If I do go this route, would you want me to make the `_relX: i32, _relY: i32` something like `_rel: Vec2i`, and same with the `_start: i32, _end: i32` parameters? Another thing is if you do wanna do the modifier-enum, should I make that the input to generateSphere_ too in place of `addTerrain`?

Also, slight sidenote, this function doesn't even benefit from using the comptime `addTerrain`, because the only time it's called is in `generateSphere` where the input to that parameter is essentially just `radius < 0`. Honestly, I could inline the `generateSphere_` function into the `generateSphere` function and make the appropriate changes there.

src/server/terrain/cavegen/FractalCaveGenerator.zig
BoringOrng: I'll expand out the `-Sq` suffix but I do think `dxSquare` should be it's own helper-constant, not because I don't trust the compiler to inline it anyway, but because it makes more sense (at least to me) to read it being squared right where it's calculated, and not within another loop.

src/server/terrain/cavegen/FractalCaveGenerator.zig
BoringOrng: I rewrote that *mostly* as follows (completely missed the `max_extent` I've been writing too much rust as of late), I do have a question as to whether or not you want `maxExtent` to be a `Vec3i` or a `Vec2i`, it can't be an `i32`, because `z` has it's own scaling based off of the `CaveFragment.height`, whereas `x` and `y` use the `width`. `Vec3i` implies that `x` and `y` *can* use different `maxExtent`s, whereas `Vec2i` cements in place that whatever `x` is scaled by, `y` must be scaled by.

src/server/terrain/cavegen/FractalCaveGenerator.zig
BoringOrng: I'm not exactly a huge fan on doing just `min` or `max`, thoughts on suffixing some `-Dist` or `Bound`? I think that would clarify the intent a bit better.

BoringOrng:
> Oh interesting, since the main performance difference comes from floating point maths, I wonder if we can get even bigger performance improvements with `@setFloatMode(.optimized)`?

After I run through the fixes you've pointed out, I plan on benching it, I'll do another bench with `@setFloatMode(.optimized)`. If it doesn't yield any sorta significant differences I'm inclined to omit it entirely, but what do you wanna do?

src/server/terrain/cavegen/FractalCaveGenerator.zig
IntegratedQuantum: Yes, an enum is always better, and I'd prefer to use the same enum in both places.

> If I do go this route, would you want me to make the _relX: i32, _relY: i32 something like _rel: Vec2i

I'd suggest not to do too many changes in this one PR. Also it is kind of annoying to convert a Vec3i to Vec2i currently due to lack of swizzling in the language.

> could inline the generateSphere_ function into the generateSphere function and make the appropriate changes there.

I think when inlined you'd have trouble turning `radius < 0` into a comptime value.

src/server/terrain/cavegen/FractalCaveGenerator.zig
IntegratedQuantum: Sure

src/server/terrain/cavegen/FractalCaveGenerator.zig
IntegratedQuantum: Then Vec3i would be better, Vec2i would be rather hard to read in my opinion. There is no reason to cement this property in a helper function that doesn't care about it.

src/server/terrain/cavegen/FractalCaveGenerator.zig
IntegratedQuantum: I would disagree. I prefer reading arithmetic `dx*dx` instead of separate variable names like `dxSquare`.
As a reader/debugger `dx*dx` is unambiguous, whereas `dxSquare` is something where I'd need to check the definition.

IntegratedQuantum:
Please also mention the new performance numbers, and also put the final numbers in the PR description (which will go into the commit message)

src/server/terrain/cavegen/FractalCaveGenerator.zig
```diff
@@ -58,82 +58,100 @@ pub fn generate(map: *CaveMapFragment, worldSeed: u64) void {
...
-				if (addTerrain) {
-					map.addRange(curX, curY, zMin, zMax); // Add the center range in a single call.
-				} else {
-					map.removeRange(curX, curY, zMin, zMax); // Remove the center range in a single call.
-				}
+				map.modifyTerrain(terrainModifier, curX, curY, zMin, zMax);
 			}
-			// Add some roughness at the upper cave walls:
+
+			// My rather poor attempt at explaining to whatever poor soul wants to
```
IntegratedQuantum, line 115: It seems odd to explain it here when the same math was used literally 4 lines above.

src/server/terrain/cavegen/FractalCaveGenerator.zig
```diff
@@ -58,82 +58,100 @@ pub fn generate(map: *CaveMapFragment, worldSeed: u64) void {
...
+			// understand this:
+			//
+			// (x - r_x)^2 + (y - r_y)^2 + (z - r_z)^2 = r^2
```
IntegratedQuantum, line 118: For math I generally suggest to use unicode symbols, it makes it a lot nicer to read.
```suggestion
			// (x - r_x)² + (y - r_y)² + (z - r_z)² = r²
```
Sadly there are no subscripts for y and z though, only for ₓ, and it would look odd to only use one subscript.

IntegratedQuantum (changes requested):

PR #3385, opened 2026-07-19
Range Modifiers

Crepestrom:
Adds a new procedural tool modifier that changes how far a players range is


Wunka:
I know there was already once a pr about something with range, I think it was a command (I cant find the pr) and quantum made a command there about how it is limited. Just so you know

src/proceduralItem/modifiers/long.zig
Argmaster, line 1: To me it screams "long", although I believe it was described as far reaching in the issue too. I would argue that we have a "powerful" modifier, not "powerful damaging".

Crepestrom: sure added

PR #3387, opened 2026-07-19
set cmd update

Wunka:
Progress towards #3073

### DIFF
--- a/src/server/command/worldedit/set.zig
+++ b/src/server/command/worldedit/set.zig
@@ -12,14 +12,23 @@ const Pattern = main.blueprint.Pattern;
 pub const description = "Set all blocks within selection to a block.";
 pub const usage = "/set <pattern>";
 
+const Args = union(enum) {
+	@"/set": struct { selection: []const u8 },
+};
+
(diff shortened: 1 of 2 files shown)


src/server/command/worldedit/set.zig
```diff
@@ -12,14 +12,23 @@ const Pattern = main.blueprint.Pattern;
 pub const description = "Set all blocks within selection to a block.";
 pub const usage = "/set <pattern>";
 
+const Args = union(enum) {
+	@"/set": struct { selection: []const u8 },
```
IntegratedQuantum: It's called pattern in the help text.
```suggestion
	@"/set": struct { pattern: []const u8 },
```

Also I think we should introduce a parsing wrapper struct for this in command.zig, as was done for some of the other worldedit data structures.

IntegratedQuantum (changes requested):

src/server/command/worldedit/set.zig
Wunka: done

IntegratedQuantum (approved):

Issue #3156, opened 2026-06-03
argparse should use arena for args allocations


Argmaster:
As entioned in https://github.com/PixelGuys/Cubyz/pull/3138#discussion_r3336251094
As #3138 shows, adding and argument with sanitiazation (one that requires a new allocation) creates an explosion of deinit functions that have to be added to correctly reclaim memory. It is an undesirable outcome, as it bloats code with boilerplate. At the same time it is very convenient to express sanitization via custom types, so we don't want to give up that feature. A way to get the best of both worlds would be to use arena that gets trashed after command execution is done. This approach also guarantees that lifetimes of arguments will be all the same.

Issue #3128, opened 2026-05-24
argparser allocator usage for the errorMessages is used inconsistently


IntegratedQuantum:
The interface states that it should use the stackAllocator
```zig
/// `errorMessage` - out parameter used to store and return errors, if any occur. Has to be allocated with stackAllocator.
```
But most of the code actually uses the passed-in `allocator` for it which only happens to be the stack allocator (for now), and these errors are also hard to catch in a review.

I think the best solution would be to use a `List` instead of `ListUnmanaged`, so there is no chance to use the wrong allocator. The reason for using the unmanaged list is #1181.

This issue should not block the migration effort that has already started.

Argmaster:
I just want to mention that this issue would not exist is we were to use Result struct instead of out parameter.

IntegratedQuantum:
But Result struct would require needlessly allocating and copying the error strings around all the time right?

Argmaster:
Would it? We could still use current approach internally, just return an Result object from the public function. I need to try to implement that to check.

IntegratedQuantum:
What about the parse function of custom structs though?

Argmaster:
Nothing, it gets pre-allocated list object and an appropriate allocator, so problem doesn't exist.
Problem exists only upon list creation.

IntegratedQuantum:
I think this can wait until after #3073 and after the usages of the ArgParser helper are consilidated outside the command to reduce the number of changes needed.

Argmaster:
A related issue: https://github.com/PixelGuys/Cubyz/pull/3112/changes#r3328982446

IntegratedQuantum:
Yes, many of the command PRs had this problem. I decided to not worry about it in reviews to allow us to progress faster, after #3073 it should be easier to resolve all the incorrect use cases.

Argmaster:
IMO we should just use `allocator` everywhere, as it clearly makes impression that it should be used. Additionally we have this very bad pattern of interleaving allocations from different scopes using `stackAllocator`. From perspective of `parse` using `stackAllocator` for `errorMessage` looks like explicitly using `stackAllocator` for resource that outlives the scope where `stackAllocator` is used.
That is a clear contradiction with our guidelines:

> - local lifetime, things that are freed at the end of the scope (including local data structures such as lists) → `main.stackAllocator`

It would be much more standard and much easier to explain 

> "errorMessage has same lifetime as Args and use same allocator"

[...]

PR #3388, opened 2026-07-19
replace cmd update

Wunka:
Progress towards #3073


Wunka:
Okay I needed to do more here than I thought. Because it would leak without doing errdefer on the fields, thats why I needed to add this wierd run deinit where possible solution

Wunka:
@Argmaster maybe you would know a better solution here? This code doesn't look good...

IntegratedQuantum:
#3156

Wunka:
> #3156

Ahh. that makes sense... Should have searched for such an issue... So this is blocked by that issue?

Argmaster:
> Ahh. that makes sense... Should have searched for such an issue... So this is blocked by that issue?

No, #3138 got merged anyway. We should finish one round of refactorings before we do do another one.

IntegratedQuantum:
Yeah, but on the other hand #3156 is blocked on the last step of #3073 and would be problematic because of #3128

I think the best way forward for now is to just ignore the leaks on parser errors, finish #3073 and then switch to the arena soon.

Wunka:
Okay. I removed my fix

src/server/command.zig
```diff
@@ -236,3 +237,18 @@ pub const MaskExpression = struct {
 		self.mask.deinit(allocator);
 	}
 };
+
+pub const PatternExpression = struct {
```
IntegratedQuantum, line 241: Please rebase, I'm not sure how github will merge this

Wunka: it actually handeld that perfectly

IntegratedQuantum (approved):

PR #3390, opened 2026-07-19
particles cmd update

Wunka:
Progress towards #3073

achieved by removing space support as suggested by quantum here: https://github.com/PixelGuys/Cubyz/pull/3137#issuecomment-4586291828

### DIFF
--- a/src/server/command/particles.zig
+++ b/src/server/command/particles.zig
@@ -13,71 +13,55 @@ pub const usage =
 	\\/particles <id> <x> <y> <z> <collides> <count> <spawnDataZon>
 	\\
 	\\tip: use "~" to apply current player position coordinate in <x> <y> <z> fields.
-	\\zon example:
+	\\zon example (currently no support for spaces in the zon):
 	\\.{
-	\\  .shape = .sphere,
-	\\  .radius = 5,
-	\\  .mode = .scatter,
-	\\  .speed = .{0.5, 10},
-	\\  .lifeTime = .{0.5, 10},
-	\\  .randomRotate = true,
+	\\  .shape=.sphere,
+	\\  .radius=5,
+	\\  .mode=.scatter,
+	\\  .speed=.{0.5,10},
+	\\  .lifeTime=.{0.5,10},
+	\\  .randomRotate=true,
 	\\}
 ;
 
+const Args = union(enum) {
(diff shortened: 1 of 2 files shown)


src/server/command/particles.zig
```diff
@@ -13,71 +13,55 @@ pub const usage =
...
-	\\  .mode = .scatter,
-	\\  .speed = .{0.5, 10},
-	\\  .lifeTime = .{0.5, 10},
-	\\  .randomRotate = true,
+	\\  .shape=.sphere,
+	\\  .radius=5,
+	\\  .mode=.scatter,
+	\\  .speed=.{0.5,10},
+	\\  .lifeTime=.{0.5,10},
+	\\  .randomRotate=true,
```
IntegratedQuantum: You can keep the formatting, in the example, it already contains spaces anyways (because surprise: Zig does not support tabs in multiline strings)

src/server/command/particles.zig
```diff
@@ -13,71 +13,55 @@ pub const usage =
...
-		return;
-	};
-}
+const Args = union(enum) {
+	@"/particles <id> <x> <y> <z> <collides> <count> <spawnDataZon>": struct {
+		id: []const u8,
+		x: command.Coordinate,
+		y: command.Coordinate,
+		z: command.Coordinate,
+		collides: ?enum { true, false },
```
IntegratedQuantum: Do we not support bools?

IntegratedQuantum (changes requested):

src/server/command/particles.zig
Wunka: nope. I tried to add support, but something was wierd. bool triggers:
```zig
inline else => |other| @compileError("Unsupported type " ++ @tagName(other)),
```
with this message: `Unsupported type void` which I thought was weird. Because why `void`?

IntegratedQuantum: That's odd. Could you investigate this further please?
It would be nice if we could support bools.

Wunka: Okay I still don't know where void came from, but it works now. there seems to be no bool parser so internally I used a enum again, so I can do meta.stringToEnum

src/server/command/particles.zig
Wunka: reverted

IntegratedQuantum (approved):

PR #3393, opened 2026-07-19
Finish #3073, finnally

### LINKED ISSUE #3073: Update commands and commands helper to the new parser pattern

### LINKED ISSUE #3128: argparser allocator usage for the errorMessages is used inconsistently

### LINKED ISSUE #3156: argparse should use arena for args allocations

Wunka:
Closes #3073
Closes #3156
Closes #3128

### DIFF
--- a/src/argparse.zig
+++ b/src/argparse.zig
@@ -2,7 +2,7 @@ const std = @import("std");
 
 const main = @import("main");
 const NeverFailingAllocator = main.heap.NeverFailingAllocator;
-const List = main.List;
+const ListManaged = main.ListManaged;
 const utils = main.utils;
 
 pub const Options = struct {
@@ -20,34 +20,34 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 		/// Result is returned from this function as a value of type `T`.
 		///
 		/// Arguments:
-		/// - `allocator` - will be used for dynamic allocations of the parsing result returned.
+		/// - `arena` - will be used for dynamic allocations of the parsing result returned.
 		/// - `args` - unprocessed string containing command arguments without command name.
 		/// - `errorMessage` - out parameter used to store and return errors, if any occur. Has to be allocated with stackAllocator.
-		pub fn parse(allocator: NeverFailingAllocator, args: []const u8, errorMessage: *List(u8)) error{ParseError}!*T {
-			const result = allocator.create(T);
-			errdefer allocator.destroy(result);
-			result.* = try resolve(ResolveMode.parse, allocator, args, errorMessage);
+		pub fn parse(arena: NeverFailingAllocator, args: []const u8, errorMessage: *ListManaged(u8)) error{ParseError}!*T {
+			const result = arena.create(T);
+			errdefer arena.destroy(result);
+			result.* = try resolve(ResolveMode.parse, arena, args, errorMessage);
 			return result;
 		}
 
 		pub fn resolve(
 			comptime mode: ResolveMode,
-			allocator: NeverFailingAllocator,
+			arena: NeverFailingAllocator,
 			args: []const u8,
-			errorMessage: *List(u8),
+			errorMessage: *ListManaged(u8),
 		) switch (mode) {
 			.autocomplete => AutocompleteResult,
 			.parse => error{ParseError}!T,
 		} {
 			switch (@typeInfo(T)) {
 				inline .@"struct" => |s| {
-					return resolveStruct(mode, s, allocator, args, errorMessage);
+					return resolveStruct(mode, s, arena, args, errorMessage);
 				},
 				inline .@"union" => |u| {
 					if (u.tag_type == null) @compileError("Union must have a tag type");
 					return switch (mode) {
-						.autocomplete => autocompleteUnion(u, allocator, args, errorMessage),
-						.parse => parseUnion(u, allocator, args, errorMessage),
+						.autocomplete => autocompleteUnion(u, arena, args, errorMessage),
+						.parse => parseUnion(u, arena, args, errorMessage),
 					};
 				},
 				else => @compileError("Only structs and unions are supported"),
@@ -57,9 +57,9 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 		fn resolveStruct(
 			comptime mode: ResolveMode,
 			comptime s: std.builtin.Type.Struct,
-			allocator: NeverFailingAllocator,
+			arena: NeverFailingAllocator,
 			args: []const u8,
-			errorMessage: *List(u8),
+			errorMessage: *ListManaged(u8),
 		) switch (mode) {
 			.autocomplete => AutocompleteResult,
 			.parse => error{ParseError}!T,
@@ -67,20 +67,20 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 			var result: T = undefined;
 			var tokens = std.mem.tokenizeScalar(u8, args, ' ');
 
-			var tempErrorMessage: List(u8) = .empty;
-			defer tempErrorMessage.deinit(allocator);
+			var tempErrorMessage: ListManaged(u8) = .init(main.stackAllocator);
+			defer tempErrorMessage.deinit();
 
 			var nextArgument: ?[]const u8 = tokens.next();
 
 			inline for (s.fields) |field| {
-				const value = resolveArgument(field.type, allocator, field.name[0..], nextArgument, &tempErrorMessage);
+				const value = resolveArgument(field.type, arena, field.name[0..], nextArgument, &tempErrorMessage);
 
 				if (value == error.ParseError) {
 					if (@typeInfo(field.type) == .optional) {
 						@field(result, field.name) = null;
 						tempErrorMessage.clearRetainingCapacity();
 					} else {
-						errorMessage.appendSlice(allocator, tempErrorMessage.items);
+						errorMessage.appendSlice(tempErrorMessage.items);
 						return error.ParseError;
 					}
 				} else {
@@ -91,24 +91,24 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 			}
 
 			if (nextArgument != null) {
-				errorMessage.print(allocator, "Too many arguments for command, expected {}", .{s.fields.len});
+				errorMessage.print("Too many arguments for command, expected {}", .{s.fields.len});
 				return error.ParseError;
 			}
 
 			return result;
 		}
 
-		fn resolveArgument(comptime Field: type, allocator: NeverFailingAllocator, name: []const u8, argument: ?[]const u8, errorMessage: *List(u8)) error{ParseError}!Field {
+		fn resolveArgument(comptime Field: type, arena: NeverFailingAllocator, name: []const u8, argument: ?[]const u8, errorMessage: *ListManaged(u8)) error{ParseError}!Field {
 			const fieldTypeInfo = @typeInfo(Field);
 			if (fieldTypeInfo == .optional) {
 				if (argument == null) return error.ParseError;
-				return resolveArgument(fieldTypeInfo.optional.child, allocator, name, argument, errorMessage) catch |err| {
+				return resolveArgument(fieldTypeInfo.optional.child, arena, name, argument, errorMessage) catch |err| {
 					return err;
 				};
 			}
 
 			const arg = argument orelse {
-				errorMessage.print(allocator, "Missing argument at position <{s}>", .{name});
+				errorMessage.print("Missing argument at position <{s}>", .{name});
 				return error.ParseError;
 			};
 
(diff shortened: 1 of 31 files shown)


src/server/command/worldedit/replace.zig
```diff
@@ -25,17 +25,8 @@ const Args = union(enum) {
...
-	var errorMessage: main.List(u8) = .empty;
-	defer errorMessage.deinit(main.stackAllocator);
-
-	const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
-		source.sendMessage("#ff0000{s}", .{errorMessage.items});
-		return;
-	};
-	defer result.deinit(main.stackAllocator);
+pub fn execute(args: *Args, source: *User) void {
+	defer args.deinit(main.stackAllocator);
```
IntegratedQuantum: This does not look right? They are supposed to be allocated with the arena.

src/argparse.zig
```diff
@@ -23,8 +23,11 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 		/// - `allocator` - will be used for dynamic allocations of the parsing result returned.
 		/// - `args` - unprocessed string containing command arguments without command name.
 		/// - `errorMessage` - out parameter used to store and return errors, if any occur. Has to be allocated with stackAllocator.
-		pub fn parse(allocator: NeverFailingAllocator, args: []const u8, errorMessage: *List(u8)) error{ParseError}!T {
-			return resolve(ResolveMode.parse, allocator, args, errorMessage);
+		pub fn parse(allocator: NeverFailingAllocator, args: []const u8, errorMessage: *List(u8)) error{ParseError}!*T {
```
IntegratedQuantum: Please also rename this parameter everywhere to ensure everyone knows that it's supposed to be an arena.
```suggestion
		pub fn parse(arena: NeverFailingAllocator, args: []const u8, errorMessage: *List(u8)) error{ParseError}!*T {
```

src/server/command.zig
```diff
@@ -46,7 +48,15 @@ pub fn execute(msg: []const u8, source: *User) void {
 			return;
 		}
 		source.sendMessage("#00ff00Executing Command /{s}", .{msg});
-		cmd.exec(msg[@min(end + 1, msg.len)..], source);
+		var arena: main.heap.NeverFailingArenaAllocator = .init(main.stackAllocator);
```
IntegratedQuantum:
```suggestion
		const arena = main.stackAllocator.createArena();
		defer main.stackAllocator.destroyArena(arena);
```

src/server/command.zig
```diff
@@ -46,7 +48,15 @@ pub fn execute(msg: []const u8, source: *User) void {
...
+		defer arena.deinit();
+		var errorMessage: main.List(u8) = .empty;
```
IntegratedQuantum: Please turn this into a ListManaged to fix #3128 (otherwise some of the commands will crash from allocator misuse)

src/server/command.zig
```diff
@@ -10,7 +10,8 @@ const User = main.server.User;
 pub const commandList = @import("command/_list.zig");
 
 pub const Command = struct {
-	exec: *const fn (args: []const u8, source: *User) void,
+	parse: *const fn (allocator: NeverFailingAllocator, args: []const u8, errorMessage: *List(u8)) error{ParseError}!*anyopaque,
+	exec: *const fn (args: *anyopaque, source: *User) void,
```
IntegratedQuantum: I'd suggest to make only one virtual function that calls both of these. Then you can put the `Args` object directly on the stack without a needless allocation (that had you touch all the tests)

src/argparse.zig
```diff
@@ -65,7 +68,7 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 			var tokens = std.mem.tokenizeScalar(u8, args, ' ');
 
 			var tempErrorMessage: List(u8) = .empty;
-			defer tempErrorMessage.deinit(main.stackAllocator);
+			defer tempErrorMessage.deinit(allocator);
```
IntegratedQuantum: local allocations should use the stackAllocator to not leak temporary memory into the arena.

src/server/command/worldedit/rotate.zig
```diff
@@ -10,28 +10,19 @@ pub const usage =
...
-
-pub fn execute(args: []const u8, source: *User) void {
-	var errorMessage: main.List(u8) = .empty;
-	defer errorMessage.deinit(main.stackAllocator);
-
-	const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
-		source.sendMessage("#ff0000{s}", .{errorMessage.items});
-		return;
-	};
+pub fn execute(result: *Args, source: *User) void {
```
IntegratedQuantum:
```suggestion
pub fn execute(args: *Args, source: *User) void {
```

src/server/command/worldedit/toggledecay.zig
```diff
@@ -21,24 +21,14 @@ const State = enum {
...
-pub fn execute(args: []const u8, source: *User) void {
-	var errorMessage: main.List(u8) = .empty;
-	defer errorMessage.deinit(main.stackAllocator);
-
-	const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
-		source.sendMessage("#ff0000{s}", .{errorMessage.items});
-		return;
-	};
-
+pub fn execute(result: *Args, source: *User) void {
```
IntegratedQuantum:
```suggestion
pub fn execute(args: *Args, source: *User) void {
```

src/server/command/worldedit/undo.zig
```diff
@@ -9,20 +9,12 @@ const Blueprint = main.blueprint.Blueprint;
...
-
-pub fn execute(args: []const u8, source: *User) void {
-	var errorMessage: main.List(u8) = .empty;
-	defer errorMessage.deinit(main.stackAllocator);
-
-	_ = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
-		source.sendMessage("#ff0000{s}", .{errorMessage.items});
-		return;
-	};
+pub fn execute(result: *Args, source: *User) void {
```
IntegratedQuantum:
```suggestion
pub fn execute(args: *Args, source: *User) void {
```

src/server/command/gamemode.zig
```diff
@@ -12,22 +12,12 @@ pub const usage =
...
-	var errorMessage: main.List(u8) = .empty;
-	defer errorMessage.deinit(main.stackAllocator);
-
-	const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
-		source.sendMessage("#ff0000{s}", .{errorMessage.items});
-		return;
-	};
-
-	switch (result) {
+pub fn execute(result: *Args, source: *User) void {
```
IntegratedQuantum:
```suggestion
pub fn execute(args: *Args, source: *User) void {
```

src/server/command/help.zig
```diff
@@ -9,27 +9,17 @@ const User = main.server.User;
...
-pub fn execute(args: []const u8, source: *User) void {
-	var errorMessage: main.List(u8) = .empty;
-	defer errorMessage.deinit(main.stackAllocator);
-
-	const result = ArgParser.parse(main.stackAllocator, args, &errorMessage) catch {
-		source.sendMessage("#ff0000{s}", .{errorMessage.items});
-		return;
-	};
-
+pub fn execute(result: *Args, source: *User) void {
```
IntegratedQuantum:
```suggestion
pub fn execute(args: *Args, source: *User) void {
```

src/meta.zig
```diff
@@ -70,13 +70,32 @@ fn CastFunctionReturnToOptionalAnyopaqueType(Fn: type) type {
...
+}
+
 /// Turns the return parameter into a *anyopaque
 pub fn castFunctionReturnToAnyopaque(function: anytype) *const CastFunctionReturnToAnyopaqueType(@TypeOf(function)) {
 	return @ptrCast(&function);
 }
 pub fn castFunctionReturnToOptionalAnyopaque(function: anytype) *const CastFunctionReturnToOptionalAnyopaqueType(@TypeOf(function)) {
 	return @ptrCast(&function);
 }
+pub fn castFunctionReturnToErrorAnyopaque(function: anytype, errorSet: type) *const CastFunctionReturnToErrorAnyopaqueType(@TypeOf(function), errorSet) {
```
IntegratedQuantum: You can extract the errorset from the return type of the function

IntegratedQuantum (changes requested):

src/server/command/worldedit/replace.zig
Wunka: ups forgot to remove that, removed

src/argparse.zig
Wunka: renamed

src/server/command.zig
Wunka: If I do that I get an error when I want to create the allocator

src/server/command.zig

src/server/command.zig
Wunka: Actually, through creating the one function I was able to completly remove any `anyopaque`.

src/argparse.zig
Wunka: done

src/server/command/worldedit/rotate.zig
Wunka: ups. that is from the start where I first started with result (to not need to rename the other mentions) but I then switched over. done for all where this

src/meta.zig
Wunka: this function is removed as not needed anymore

src/server/command.zig
IntegratedQuantum: arena already is the allocator

Wunka: no? I don't understand? to get an `NeverfailingAlloctor` from the arena I need to call `allocator()` which doens't like it if its a const

IntegratedQuantum: createArena returns a NeverFailingAllocator

Wunka: Ah so I just used the wrong one... changed

src/server/command/worldedit/blueprint.zig
```diff
@@ -169,7 +158,7 @@ const FilePath = struct {
 		allocator.free(self.path);
 	}
 
-	pub fn parse(allocator: NeverFailingAllocator, _: []const u8, arg: []const u8, _: *List(u8)) error{ParseError}!FilePath {
+	pub fn parse(allocator: NeverFailingAllocator, _: []const u8, arg: []const u8, _: *ListManaged(u8)) error{ParseError}!FilePath {
```
IntegratedQuantum: Please rename all arenas to `arena`
```suggestion
	pub fn parse(arena: NeverFailingAllocator, _: []const u8, arg: []const u8, _: *ListManaged(u8)) error{ParseError}!FilePath {
```

src/server/command.zig
```diff
@@ -56,18 +72,19 @@ pub const Coordinate = union(enum) {
 	relative: f64, // Relative coordinates are indicated by leading `~`.
 	absolute: f64,
 
-	pub fn parse(allocator: NeverFailingAllocator, name: []const u8, arg: []const u8, errorMessage: *List(u8)) error{ParseError}!Coordinate {
+	pub fn parse(allocator: NeverFailingAllocator, name: []const u8, arg: []const u8, errorMessage: *ListManaged(u8)) error{ParseError}!Coordinate {
+		_ = allocator; // autofix
```
IntegratedQuantum: Oh you use ZLS autofix?
Maybe we should add that to the linter then.

src/server/command/worldedit/blueprint.zig
IntegratedQuantum:
```suggestion
	fn ensureBlueprintExtension(arena: NeverFailingAllocator, fileName: []const u8) []const u8 {
```

src/server/command/worldedit/blueprint.zig
```diff
@@ -169,7 +158,7 @@ const FilePath = struct {
 		allocator.free(self.path);
 	}
```
IntegratedQuantum, line 159: Should be removed

src/argparse.zig
```diff
@@ -407,25 +407,25 @@ test "x or xy negative too many args" {
...
 }
 
 test "subCommands bar" {
-	var errors: List(u8) = .empty;
+	var errors: ListManaged(u8) = .init(main.stackAllocator);
 	defer errors.deinit(main.stackAllocator);
 
-	const resultOrError = Test.@"subCommands foo or bar".parse(main.stackAllocator, "bar 2.0 3.0", &errors);
+	const result = try Test.@"subCommands foo or bar".parse(main.stackAllocator, "bar 2.0 3.0", &errors);
+	defer main.stackAllocator.destroy(result);
```
IntegratedQuantum: No longer needed in the tests

IntegratedQuantum (changes requested):

src/server/command.zig
Wunka: yes. I should probably deactivate that...

src/server/command/worldedit/blueprint.zig
Wunka: ups, done

src/server/command/worldedit/blueprint.zig
Wunka: done

src/server/command/worldedit/blueprint.zig
Wunka: removed

src/argparse.zig
Wunka: fixed tests

src/argparse.zig
```diff
@@ -20,31 +20,31 @@ pub fn Parser(comptime T: type, comptime options: Options) type {
 		/// Result is returned from this function as a value of type `T`.
 		///
 		/// Arguments:
-		/// - `allocator` - will be used for dynamic allocations of the parsing result returned.
+		/// - `arena` - will be used for dynamic allocations of the parsing result returned.
 		/// - `args` - unprocessed string containing command arguments without command name.
 		/// - `errorMessage` - out parameter used to store and return errors, if any occur. Has to be allocated with stackAllocator.
```
IntegratedQuantum:
```suggestion
		/// - `errorMessage` - out parameter used to store and return errors, if any occur.
```

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #3398, opened 2026-07-20
For each restriction

Crepestrom:
Adds a new restriction
for each
it counts the ammount of items that a child restriction counted and increases its percentage by that much

note: we do not scale it liniarily but scale it like how every other modifier scales


Wunka:
Sorry that no one saw this, There is no good way (at least from what I know) to see if a pr is set from draft to ready review

PR #3399, opened 2026-07-20
Replace allocPrint with Allocator.print

### LINKED ISSUE #1595: Helper function for formatting
#920 but not specialized for paths 🙃

IntegratedQuantum:
Was added upstream in https://codeberg.org/ziglang/zig/pulls/35926 (doc comment was copied from there)

fixes #1595


Wunka:
Does this not also fix: #920 ?

PR #3403, opened 2026-07-20
Allow loading material colors from texture

yel0h:
As suggested in https://github.com/PixelGuys/Cubyz/pull/3366#pullrequestreview-4738250562, this allows materials to source their `colorPalette`, `outlineColorLight` and `outlineColorShadow` from a small PNG instead of hardcoded ARGB hex literals in the `.zig.zon` file.

Adds a new optional zon field, `.colorTexture = "name.png"`, a sibling of the existing `.texture` field. It's resolved through the exact same `items/textures/<mod>/` + `assets/<mod>/items/textures/` fallback convention already used for item textures. I've defined the format as an N x 2 image, where the top row is the gradient palette, dark to light, and the bottom row's two pixels are `outlineColorShadow` (x=0) and `outlineColorLight` (x=1). A fully transparent pixel there means the color is not specified, falling back to the existing computed `darken()` default. For backward compatibility, it uses the old hex fields if `.colorTexture` is absent or fails to load.

image

I've migrated `iron_ingot` as an example from its hex values and dropped the old `colors` and `outlineColor*` fields from `iron_ingot.zig.zon` in favor of `.colorTexture = "iron_ingot_palette.png"`. I can do so for all the other materials via a script once the specifics of this solution are accepted.

Also includes a small fix in `graphics.zig`: `Image.readFromFile`'s `.asIs` orientation option wasn't actually resetting the vertical-flip flag, so once one image was flipped, all images loaded after would also get flipped, regardless of the selected orientation. It now resets the flag.


src/graphics.zig
```diff
@@ -2146,7 +2146,7 @@ pub const Image = struct { // MARK: Image
 		const nullTerminatedPath = main.stackAllocator.dupeZ(u8, path); // TODO: Find a more zig-friendly image loading library.
 		errdefer main.stackAllocator.free(nullTerminatedPath);
 		switch (options.orientation) {
-			.asIs => {},
+			.asIs => c.stbi_set_flip_vertically_on_load(0),
```
IntegratedQuantum, line 2149: Why is this needed now?

assets/cubyz/materials/iron_ingot_palette.png
IntegratedQuantum, line 1: I think we should make a materials folder, so it doesn't clutter with the other item textures.
Though I guess @careeoki should probably make the decision here.

src/items.zig
```diff
@@ -66,20 +93,22 @@ const Material = struct { // MARK: Material
...
+			self.colorPalette = allocator.alloc(Color, colors.toSlice().len);
+			for (colors.toSlice(), self.colorPalette) |item, *color| {
+				color.* = Color.fromArgb(item.as(u32) orelse 0xff000000);
+			}
+			if (zon.get(u32, "outlineColorLight")) |colorInt| {
+				self.outlineColorLight = Color.fromArgb(colorInt);
+			} else {
+				self.outlineColorLight = darken(self.colorPalette[self.colorPalette.len - 1], 0.7);
+			}
+			if (zon.get(u32, "outlineColorShadow")) |colorInt| {
```
IntegratedQuantum: Please remove these again, you removed their single use-case already, all new development should use the texture, eventually the other `colors` attribute can also be removed.

src/items.zig
```diff
@@ -48,7 +48,34 @@ const Material = struct { // MARK: Material
 		};
 	}
 
-	pub fn init(self: *Material, allocator: NeverFailingAllocator, zon: ZonElement) void {
+	fn loadColorsFromTexture(self: *Material, allocator: NeverFailingAllocator, colorTexturePath: []const u8, colorReplacementTexturePath: []const u8) bool {
```
IntegratedQuantum: Please don't use bools in function arguments/return values. An enum is more readable.

src/items.zig
```diff
@@ -48,7 +48,34 @@ const Material = struct { // MARK: Material
...
+				std.log.err("Error while reading material color texture '{s}': {s}", .{colorTexturePath, @errorName(err)});
+			}
+			break :blk graphics.Image.readFromFile(main.stackAllocator, colorReplacementTexturePath, .{.orientation = .asIs}) catch |err2| {
+				std.log.err("Error while reading material color texture. Tried '{s}' and '{s}': {s}", .{colorTexturePath, colorReplacementTexturePath, @errorName(err2)});
+				return false;
+			};
+		};
+		defer image.deinit(main.stackAllocator);
+		if (image.width == 0 or image.height < 2) {
+			std.log.err("Material color texture '{s}' must be at least 1x2 pixels (got {}x{}).", .{colorTexturePath, image.width, image.height});
```
IntegratedQuantum: Doesn't it need to be at least 2×2 so you have the two outline colors?

src/items.zig
```diff
@@ -48,7 +48,34 @@ const Material = struct { // MARK: Material
...
+			return false;
+		}
+		self.colorPalette = allocator.alloc(Color, image.width);
+		for (0..image.width) |x| {
+			self.colorPalette[x] = image.getRGB(x, 0);
+		}
+		const shadow = image.getRGB(0, 1);
+		self.outlineColorShadow = if (shadow.a != 0) shadow else darken(self.colorPalette[0], 0.5);
```
IntegratedQuantum: Instead of automatic behavior, I think it is often better to output an error.
Furthermore this should reject any alpha less than 100%.

IntegratedQuantum (changes requested):

assets/cubyz/materials/iron_ingot_palette.png
careeoki: Yes please, a separate folder.

src/graphics.zig
yel0h: Well I think these textures shouldn't be flipped since they're not sent to the GPU, and without this fix they would get flipped regardless of the chosen orientation

yel0h:
Ok, now the color textures are loaded from `assets/<mod>/materials/`, and the outline color pixels are no longer optional.

src/graphics.zig
IntegratedQuantum: Is this actual behavior that you observed?

IntegratedQuantum:
@careeoki are you otherwise happy with how the image is structured?

careeoki:
Yes, it looks good.

careeoki (approved):

src/graphics.zig
yel0h: Yes, when testing with the iron ingot texture this is what it looked like without the fix

image

So basically the texture was being processed upside down

IntegratedQuantum (approved):
Thanks for working on this, I think this is going to make adding new or tweaking existing materials a lot easier.

> I can do so for all the other materials via a script once the specifics of this solution are accepted.

I have also made an issue for this #3451 to also track how to deal with the current fallback logic once that's done.

PR #3405, opened 2026-07-21
Nicer dry grass palette

careeoki:
Currently, dry grass is a desaturated light green with a slight yellow tint. For a while I have felt it is kinda unpleasant looking.

image

**So here is a new palette:**
It is a bit more saturated, and has a stronger yellow tint. It should look a little more "golden"
It's subtle in screenshots but I think ingame it makes dry biomes look so much nicer.

image
image
image


du82:
I like it

IntegratedQuantum (approved):

PR #3413, opened 2026-07-23
Add rotation to tp command

### LINKED ISSUE #2403: add client camera rotation command
### Description

Be able to set the direction a player is looking at with a command

Mabeeck:
Adds optional yaw and pitch parameters to the /tp command.  

```
/tp <playerId> <x> <y> <z> <yaw> <pitch>
```

Yaw and pitch are expected to be supplied in degrees as that is the unit most users can be expected to be familiar with and will automatically be converted to radians internally.  
Yaw and pitch can be relative: `/tp ~ ~ ~ ~ ~10`

Closes #2403

### DIFF
--- a/src/network/protocols.zig
+++ b/src/network/protocols.zig
@@ -606,7 +606,6 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 		biome = 4,
 		particles = 5,
 		clear = 6,
-		setRotation = 7,
 	};
 
 	const WorldEditPosition = enum(u2) {
@@ -626,13 +625,6 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 			},
 			.teleport => {
 				game.Player.setPosBlocking(try reader.readVec(Vec3d));
-			},
-			.setRotation => {
-				var rot: Vec3f = try reader.readVec(Vec3f);
-				const bound = std.math.pi/2.0 - 0.001;
-				rot[0] = std.math.clamp(rot[0], -bound, bound);
-				game.camera.rotation[0] = rot[0];
-				game.camera.rotation[2] = rot[2];
 			},
 			.worldEditPos => {
 				const typ = try reader.readEnum(WorldEditPosition);
@@ -712,7 +704,7 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 
 	fn serverReceive(conn: *Connection, reader: *utils.BinaryReader) !void {
 		switch (try reader.readEnum(UpdateType)) {
-			.gamemode, .teleport, .setRotation, .time, .biome, .particles, .clear => return error.InvalidSide,
+			.gamemode, .teleport, .time, .biome, .particles, .clear => return error.InvalidSide,
 			.worldEditPos => {
 				const typ = try reader.readEnum(WorldEditPosition);
 				const pos: ?Vec3i = switch (typ) {
@@ -741,16 +733,6 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 
 		writer.writeEnum(UpdateType, .teleport);
 		writer.writeVec(Vec3d, pos);
-
-		conn.send(.secure, id, writer.data.items);
-	}
-
-	pub fn sendTPRotation(conn: *Connection, rot: Vec3f) void {
-		var writer = utils.BinaryWriter.initCapacity(main.stackAllocator, 20);
-		defer writer.deinit();
-
-		writer.writeEnum(UpdateType, .setRotation);
-		writer.writeVec(Vec3f, rot);
 
 		conn.send(.secure, id, writer.data.items);
 	}
--- a/src/network/protocols.zig
+++ b/src/network/protocols.zig
@@ -606,6 +606,7 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 		biome = 4,
 		particles = 5,
 		clear = 6,
+		setRotation = 7,
 	};
 
 	const WorldEditPosition = enum(u2) {
@@ -625,6 +626,13 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 			},
 			.teleport => {
 				game.Player.setPosBlocking(try reader.readVec(Vec3d));
+			},
+			.setRotation => {
+				var rot: Vec3f = try reader.readVec(Vec3f);
+				const bound = std.math.pi/2.0 - 0.001;
+				rot[0] = std.math.clamp(rot[0], -bound, bound);
+				game.camera.rotation[0] = rot[0];
+				game.camera.rotation[2] = rot[2];
 			},
 			.worldEditPos => {
 				const typ = try reader.readEnum(WorldEditPosition);
@@ -704,7 +712,7 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 
 	fn serverReceive(conn: *Connection, reader: *utils.BinaryReader) !void {
 		switch (try reader.readEnum(UpdateType)) {
-			.gamemode, .teleport, .time, .biome, .particles, .clear => return error.InvalidSide,
+			.gamemode, .teleport, .setRotation, .time, .biome, .particles, .clear => return error.InvalidSide,
 			.worldEditPos => {
 				const typ = try reader.readEnum(WorldEditPosition);
 				const pos: ?Vec3i = switch (typ) {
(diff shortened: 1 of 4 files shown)


Mabeeck:
Hmm, I don't really see any large batches of commits all working on the same issue in the commit-history.  
Should I squash these commits or does GitHub have a feature for that?

Wunka:
> Hmm, I don't really see any large batches of commits all working on the same issue in the commit-history.  
> Should I squash these commits or does GitHub have a feature for that?

The pr commits are squashed on merge automatically

src/server/command.zig
```diff
@@ -118,6 +139,19 @@ pub fn resolveCoordinates(x: Coordinate, y: Coordinate, z: Coordinate, source: S
 	};
 }
 
+pub fn resolveRotation(x: Rotation, z: Rotation, source: Source) error{InvalidArg}!main.vec.Vec3f {
```
Wunka:
```suggestion
pub fn resolveRotation(yaw: Rotation, pitch: Rotation, source: Source) error{InvalidArg}!main.vec.Vec3f {
```

src/network/protocols.zig
```diff
@@ -738,6 +746,16 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 		conn.send(.secure, id, writer.data.items);
 	}
 
+	pub fn sendTPRotation(conn: *Connection, rot: Vec3f) void {
+		var writer = utils.BinaryWriter.initCapacity(main.stackAllocator, 25);
```
Wunka: you use a `vec3f` and not a `vec3d` so it only needs 20
```suggestion
		var writer = utils.BinaryWriter.initCapacity(main.stackAllocator, 20);
```

Wunka (changes requested):
Else this looks really good. But please rename your pr to something more meaningfull like: "Add rotation to tp command"

Wunka (approved):
Seems good to me. As a lot of code is pretty much the same to other code I put you into easy to review. Sorry also that you got hit by my command rewrite pr.

Mabeeck:
> Sorry also that you got hit by my command rewrite pr.

No problem, it wasn't that hard to fix my PR. Always making sure never to interfere with anything isn't possible anyways.  
I am happy to know that a volunteer as dedicated as you is working on this project.

src/network/protocols.zig
```diff
@@ -627,6 +628,13 @@ pub const genericUpdate = struct { // MARK: genericUpdate
 			.teleport => {
 				game.Player.setPosBlocking(try reader.readVec(Vec3d));
 			},
+			.setRotation => {
+				var rot: Vec3f = try reader.readVec(Vec3f);
+				const bound = std.math.pi/2.0 - 0.001;
+				rot[0] = std.math.clamp(rot[0], -bound, bound);
+				game.camera.rotation[0] = rot[0];
+				game.camera.rotation[2] = rot[2];
```
IntegratedQuantum: This is a race condition, you are setting it from the network thread, while it may also be changed on the render thread.

IntegratedQuantum (changes requested):
I think this should wait until after #2414 so we don't need to introduce yet another needless network protocol for random command data.

Mabeeck:
> I think this should wait until after https://github.com/PixelGuys/Cubyz/issues/2414 so we don't need to introduce yet another needless network protocol for random command data.

I think I have the race condition fixed now, as for the issue of network protocols; `/tp` currently cannot be used to teleport other players, which I find surprising. Since it cannot be used to teleport other players the rotation parameters could be handled locally, however if in the future you intend `/tp` to teleport other players to you or to a specified location (which I expect to be the case), then the yaw/pitch parameters not working would be counterintuitive and the protocol would need to be added anyways. Alternatively I could bundle the rotation into the existing teleport packet, if you are wondering why I used a separate packet in the first place then you can look at the giant walls of text that are my commit messages.

Besides that I am not sure why you are opposed to this additional UpdateType. It may only be used by one command, but that command is useful and can only be expected to become more useful as Cubyz obtains more content (like command blocks for example) and gains more popularity and modded servers may in the future want to set a players rotation when teleporting into an area (like spawn or a lobby of sorts) as well (imagine if hypixel gave you a random yaw and pitch every time you entered the lobby) and the most significant bloat added by it is the camera mutex, which might have been needed anyways depending on how the clientside implementation would/will go. Again, I can bundle the setRotation packet into the teleport one. The mutex would still have to stay though.

So that was my opinion and the arguments for it. I hope I convinced you that some method for the server to set player rotation is worth it.  
If not, well then I'll be forced to close this PR, though I might have a go at a clientside setrotation command once the necessary dependencies exist.

IntegratedQuantum:
Yeah, right that is a good point. But it still should not be a random network command (which inherintly are not synchronized properly which you already discovered as a race condition).
Also, more importantly, if the server sets your direction to a new value, then the client needs to take care of that in its client-side prediction. e.g. if you get teleported while you try to hit someone, the hit should now be at the new location and in the new direction.

Instead you should use a SyncOperation in sync.zig (see also #1124).

> The mutex would still have to stay though.

It would not be needed, since the sync system already has its own mutex, and in the future will handle synchronization by putting everything onto the render thread.

Mabeeck:
> Instead you should use a SyncOperation in sync.zig

It's done now.

src/server/command/tp.zig
```diff
@@ -32,6 +32,14 @@ pub const Args = union(enum) {
 		sourcePlayerIndex: command.PlayerIndex,
 		destinationPlayerIndex: command.PlayerIndex,
 	},
+	@"/tp <sourcePlayerIndex> <x> <y> <z> <yaw> <pitch>": struct {
```
IntegratedQuantum, line 37: Please also adjust the help string above

src/server/command/tp.zig
```diff
@@ -99,10 +107,15 @@ pub fn execute(args: Args, source: Source) void {
 		.@"/tp <sourcePlayerIndex> <x> <y> <z>" => |pos| {
 			break :blk command.resolveCoordinates(pos.x, pos.y, pos.z, source) catch return;
 		},
+		.@"/tp <sourcePlayerIndex> <x> <y> <z> <yaw> <pitch>" => |pos| {
+			main.sync.server.executeCommand(.{.setRotation = .{.target = target.user.id, .rotation = command.resolveRotation(pos.yaw, pos.pitch, source) catch return}}, null);
```
IntegratedQuantum: Using a sync.Command adds a ton of unecessary boilerplate. Instead we should have a helper function that allows sending a sync operation to clients directly.

Note that beyond the boilerplate, this is also a vulnerability: Any client can send sync.Commands and the server will execute them. So if I figure out someone else's entity id (which is not difficult at all), I can send such a Command from my client, causing them to change their rotation.

IntegratedQuantum (changes requested):

src/server/command/tp.zig
Mabeeck: What is there to adjust?  
It seems to fit in just fine with all the others.

IntegratedQuantum: Add it to the help string, so /help shows it

src/sync.zig
```diff
@@ -139,6 +139,19 @@ pub const server = struct { // MARK: server
 		threadContext = .other;
 	}
 
+	pub fn sendSyncOperation(op: Command.SyncOperation, source: ?*main.server.User) void {
```
Mabeeck: > Instead we should have a helper function that allows sending a sync operation to clients directly.

You mean like this?

Mabeeck:
That's everything, I think.

src/server/command.zig
```diff
@@ -121,6 +142,19 @@ pub fn resolveCoordinates(x: Coordinate, y: Coordinate, z: Coordinate, source: S
...
 }
 
+pub fn resolveRotation(yaw: Rotation, pitch: Rotation, source: Source) error{InvalidArg}!main.vec.Vec3f {
+	if (source != .user and (yaw == .relative or pitch == .relative)) {
+		source.sendMessage("Command was run without a user; unable to interpret relative rotation.", .{});
+		return error.InvalidArg;
+	}
+	const bound = std.math.pi/2.0 - 0.001;
+	return .{
+		std.math.clamp(if (yaw == .relative) source.user.player().rot[0] + yaw.relative*std.math.pi/180 else yaw.absolute*std.math.pi/180, -bound, bound),
```
IntegratedQuantum: Please do the conversion from degrees during parsing, also there is a helper function in std.math for this, which is more readable.

src/sync.zig
```diff
@@ -139,6 +139,19 @@ pub const server = struct { // MARK: server
...
+		const syncData = op.serialize(main.stackAllocator);
+		defer main.stackAllocator.free(syncData);
+
+		const users = op.getUsers(main.stackAllocator);
+		defer main.stackAllocator.free(users);
+
+		for (users) |user| {
```
IntegratedQuantum: Ideally this should only be sent to one user, you are sending it to all users, so everyone's rotation will get changed when executed.

I would suggest to not try to deduplicate code here, this function should just take a single user and only send it to that single user.

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #3417, opened 2026-07-24
fix pemission system

Wunka:
Changes:
- add helpers for adding / removing permissions
- forgotten ! in permission check is added again
- move root permission giving -> allows users to open a world locally they were already a part of to get root permissions (not having to go through another acount just to get the permissions)


IntegratedQuantum (approved):
We should ban the person who reviewed this code

PR #3425, opened 2026-07-25
commands per console (except windows)

Wunka:
No windows support. this needs to be rewritten in a way I don't know exactly. ziggit: https://ziggit.dev/t/poll-from-stdin-but-windows-is-windows/16871

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -31,6 +31,7 @@ pub const physics = @import("physics.zig");
 pub const random = @import("random.zig");
 pub const renderer = @import("renderer.zig");
 pub const rotation = @import("rotation.zig");
+pub const stdin_handler = @import("stdin_handler.zig");
 pub const settings = @import("settings.zig");
 pub const sync = @import("sync.zig");
 pub const particles = @import("particles.zig");
--- a/src/main.zig
+++ b/src/main.zig
@@ -31,7 +31,6 @@ pub const physics = @import("physics.zig");
 pub const random = @import("random.zig");
 pub const renderer = @import("renderer.zig");
 pub const rotation = @import("rotation.zig");
-pub const stdin_handler = @import("stdin_handler.zig");
 pub const settings = @import("settings.zig");
 pub const sync = @import("sync.zig");
 pub const particles = @import("particles.zig");
--- a/src/server/stdin_handler.zig
+++ b/src/server/stdin_handler.zig
@@ -21,10 +21,14 @@ pub fn update() void {
 		while (readFromStdin() != 0) {}
 		return;
 	}
-	if (readBuffer[0] == '/') {
-		main.server.command.execute(readBuffer[1 .. result - 1], .server);
+	const msg = blk: {
+		if (readBuffer[result - 1] == '\n') break :blk readBuffer[0 .. result - 1];
+		break :blk readBuffer[0..result];
+	};
+	if (msg[0] == '/') {
+		main.server.command.execute(msg[1..], .server);
 	} else {
-		main.server.sendMessage("<Server> {s}", .{readBuffer[0 .. result - 1]});
+		main.server.sendMessage("<Server> {s}", .{msg});
 	}
 }
 
--- a/src/server/stdin_handler.zig
+++ b/src/server/stdin_handler.zig
@@ -3,7 +3,7 @@ const builtin = @import("builtin");
 
 const main = @import("main");
 
-var readBuffer: [10000]u8 = undefined;
+var readBuffer: [100_000]u8 = undefined;
 
 var running: bool = false;
 
--- a/src/server/stdin_handler.zig
+++ b/src/server/stdin_handler.zig
@@ -25,7 +25,7 @@ pub fn update() void {
 	const result = readFromStdin();
 	if (result == 0) return;
 	if (result == readBuffer.len) {
-		std.log.warn("Input exceeded 10000 character limit", .{});
+		std.log.warn("Input exceeded {} character limit", .{readBuffer.len});
 		while (readFromStdin() != 0) {}
 		return;
 	}
--- a/src/server/stdin_handler.zig
+++ b/src/server/stdin_handler.zig
@@ -5,23 +5,15 @@ const main = @import("main");
 
 var readBuffer: [100_000]u8 = undefined;
 
-var running: bool = false;
-
-pub fn init() void {
-	if (builtin.os.tag == .windows) {
-		std.log.warn("Console per stdin is currently not supported on windows", .{});
-		running = false;
-	} else {
-		running = true;
-	}
-}
-
-pub fn deinit() void {
-	running = false;
-}
+var running: bool = true;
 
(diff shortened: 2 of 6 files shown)


src/stdin_handler.zig
```diff
@@ -0,0 +1,45 @@
...
+
+const main = @import("main");
+
+var readBuffer: [1024]u8 = undefined;
+
+var running: bool = false;
+
+pub fn init() void {
+	if (builtin.os.tag == .windows) {
+		std.log.warn("Console per stdin is currently not supported on windows");
```
IntegratedQuantum:
```suggestion
		std.log.warn("Console per stdin is currently not supported on windows", .{});
```

IntegratedQuantum (changes requested):

src/main.zig
```diff
@@ -31,6 +31,7 @@ pub const physics = @import("physics.zig");
 pub const random = @import("random.zig");
 pub const renderer = @import("renderer.zig");
 pub const rotation = @import("rotation.zig");
+pub const stdin_handler = @import("stdin_handler.zig");
```
IntegratedQuantum: Since this is only used by the server, I think it should be in server.zig

src/stdin_handler.zig
```diff
@@ -0,0 +1,45 @@
+const std = @import("std");
+const builtin = @import("builtin");
+
+const main = @import("main");
+
+var readBuffer: [1024]u8 = undefined;
```
IntegratedQuantum: What happens when the message is longer than this? I don't think it would be far fetched to have commands that are larger than this.

IntegratedQuantum (changes requested):

src/stdin_handler.zig
Wunka: It reads until the buffer is full and reads the rest on the next read.
To fix this I did two things:
- `readBuffer` has now a len of 10000 (same as chat from my understanding)
- if the buffer is exceeded, a warning is shown and the rest of the message is dropped (by reading until nothing is there)

src/server/stdin_handler.zig
```diff
@@ -0,0 +1,54 @@
...
+pub fn deinit() void {
+	running = false;
+}
+
+pub fn update() void {
+	if (!running) return;
+	const result = readFromStdin();
+	if (result == 0) return;
+	if (result == readBuffer.len) {
+		std.log.warn("Input exceeded 10000 character limit", .{});
```
IntegratedQuantum:
```suggestion
		std.log.warn("Input exceeded {} character limit", .{readBuffer.len});
```

src/stdin_handler.zig
IntegratedQuantum: Chat technically is only limited when accessed from the client. I think we can definitely afford another order of magnitude here, just to be sure.

IntegratedQuantum (changes requested):

src/stdin_handler.zig
Wunka: done. Also tested 100_000 len message. Such long messages are not shown in the console: 
Output: 
```sh
Truncated long log message.
{s}{s}{s}
```
But they do though show in chat. 
Also tested 10_000 len message. They do show in chat and console.

src/server/stdin_handler.zig
```diff
@@ -0,0 +1,54 @@
...
+	if (result == 0) return;
+	if (result == readBuffer.len) {
+		std.log.warn("Input exceeded {} character limit", .{readBuffer.len});
+		while (readFromStdin() != 0) {}
+		return;
+	}
+	if (readBuffer[0] == '/') {
+		main.server.command.execute(readBuffer[1 .. result - 1], .server);
+	} else {
+		main.server.sendMessage("<Server> {s}", .{readBuffer[0 .. result - 1]});
```
IntegratedQuantum: Why are you removing the last character everywhere?

src/server/stdin_handler.zig
```diff
@@ -0,0 +1,54 @@
...
+pub fn init() void {
+	if (builtin.os.tag == .windows) {
+		std.log.warn("Console per stdin is currently not supported on windows", .{});
+		running = false;
+	} else {
+		running = true;
+	}
+}
+
+pub fn deinit() void {
```
IntegratedQuantum: Is global init/deinit really necessary?
How about a simple check at the beginning of the update function?
```
if (!running) return;
if (builtin.os.tag == .windows) {
	std.log.warn("Console per stdin is currently not supported on windows", .{});
	running = false;
	return;
}
```
Then it's only going to print an error/warning once, not every time you rejoin a world.

IntegratedQuantum (changes requested):

src/server/stdin_handler.zig
Wunka: that is nice. I think I had at the beginning a global init / deinit for something else

src/server/stdin_handler.zig
Wunka: From my testing I need to do it, else have a trailing `\n`

IntegratedQuantum (changes requested):
I played around a bit with automatism, and I noticed that it doesn't split commands on tool-generated newlines. Furthermore The trailing newline doesn't seem to exist when a tool doesn't add it:
```
mint@mint:~/Downloads/Wunka$ printf "/server restart\n/server stop" | zig build run
...
server: Executing Command /server restart
/server sto
server: ---
/server <action>
Expected one of stop/restart for <action>, found "restart
/server"
---
Error while reading from stdin: EndOfStream
```

IntegratedQuantum:
Also I think it would make sense to special case EndOfStream to end without printing an error.

Wunka:
Yeah I thought about these thing too.
About \n:
I thought about splitting at \n but we allow \n in chat and you can type in chat per console. So I don't know how to handle that exactly. Maybe a command that can chain commands :laughing: We could maybe say that if it beginns with / we can split. But what if there are some other messages which should go to chat? Check each "line"? But what if someone wanted to send:
```
Here is how you get free diamonds:
/server stop
```
It would then see the first line as a chat and the second as a command. At the end I am not sure how to handle that. 

For the not always \n at the end. I use std.mem.trim now instead.

About `EndOfStream`:
I can special case it but after it happens you can't use the console anymore. So I think it should be shown. With `echo /server stop | zig build run` you don't get an EndOfStream as the stdin_handler is not run anymore, so that one is save from it. For everything else it would show that you used the stdin "wrong" or atleast brought it to a state where it is not usable anymore.

IntegratedQuantum:
It's odd, I can't seem to be able to trigger it from the terminal with manual command typing (even after adding a generous timer). So I guess it's maybe just an issue with the tooling then, and hopefully it will be better once we are entering things line by line instead of all in a big blob

IntegratedQuantum (approved):
I guess we'll just go with it for now. We'll see how well it works with multiple commands when we get to actually developing the testing framework.

PR #3427, opened 2026-07-26
Add biome-specific ocean height and relative ocean height

MrPiedPiper:
### Changes:
- Add new biome fields `oceanHeight`, `liquidBlock`, `isOceanRelative`, `relativeOceanGap`, and `relativeOceanOffset`.
- Adjust surface structures to biome ocean height (relative ocean surface structures are not yet supported)

### Ocean features:
- Oceans can be made of any block, at any height. 
- Oceans can instead be relative to topmost blocks. Distance between ocean levels in the biome is chosen using `relativeOceanGap`, with offsets available using `relativeOceanOffset`. This is best used for small pools, as happening to aligning with biome height for deeper oceans is not likely.
- Walls form to contain oceans. If neighboring oceans are compatible (based on ocean height and liquid block), no wall will form.
- Caves are now supported via `isOceanRelative`

### Performance:
I tested performance on master vs this update in 5 different newly generated worlds.
I started each world, waiting until generation finished, and noted the average time.

## Most recent implementation
Featuring 6 test seeds instead of 5!

Master:
Seeds 0, 1, 2, 3, 4, and 5 average time: 479196.862656112

This update:
Seeds 0, 1, 2, 3, 4, and 5 average time: 479150.0882511663
-0.01% added time to generation (probably just randomness at work, honestly.)

This update with more extreme settings applied to all biomes for a stress-test (`oceanHeight` 20, `isOceanRelative` true, `relativeOceanGap` 3, `relativeOceanOffset` 7)
Seeds 0, 1, 2, 3, 4, and 5 average time: 876749.7865698625
82.97% added time to generation

This update with more tame settings applied to all biomes as a comparison point (I forgot to note the settings, I'll run tests again and update this post. I think it was 3 ocean height applied to all biomes, forgot the other settings 😕)
Seeds 0, 1, 2, 3, 4, and 5 average time: 651547.8004061362
35.97% added time to generation

## Original implementation
Master:
Seeds 0, 1, 2, 3, and 4 average time: 468451.1438742311

This update:
Seeds 0, 1, 2, 3 and 4 average time: 472425.7381765770
0.85% added time to generation

This update with more extreme settings applied to all biomes for a stress-test (`oceanHeight` 20, `isOceanRelative` true, `relativeOceanGap` 3, `relativeOceanOffset` 7)
Seeds 0, 1, 2, 3 and 4 average time: 542972.4462500415
15.91% added time to generation

Code used for benchmarks [copied from BoringOrng here](https://github.com/PixelGuys/Cubyz/pull/3382#issuecomment-5052279908):
```
var sum: i96 = 0;
var count: usize = 0;

...
	const t1 = main.timestamp();

	defer {
		count += 1;
		sum += main.timestamp().nanoseconds - t1.nanoseconds;

		if (count%10 == 0)
			std.log.debug("avg gen_time: {}", .{@as(f64, @floatFromInt(sum))/@as(f64, @floatFromInt(count))});
	}
```
This was placed into the top of `TerrainGenerator.zig`'s `generate` function.


MrPiedPiper:
Known issue: if an ocean is 128+ blocks above a surface block the ocean can be cut-off by a cave biome (e.g. sky). I figure I can/should make that a separate change.

MrPiedPiper:
Should close #570

MrPiedPiper:
Vanilla biomes with small modifications as an example:

`cubyz:grasslands` with
```
	.oceanHeight = 3,
	.relativeOceanGap = 8,
	.relativeOceanOffset = -3,
	.isOceanRelative = true,
```

grassland

`cubyz:volcano/peak` with
```
	.isOceanRelative = true,
	.liquidBlock = "cubyz:lava",
	.oceanHeight = 20,
	.relativeOceanGap = 64,
	.relativeOceanOffset = 15,
```

volcano

A completely different volcano because I couldn't find the first one again

volcano2

`cubyz:cave/mantle/lava_pool` with
```
	.liquidBlock = "cubyz:lava",
	.oceanHeight = 3,
	.relativeOceanGap = 6,
	.isOceanRelative = true,
	.ground_structure = .{
		"1 to 2 cubyz:magma",
	},
``` 
(note I removed the lava from the ground_structure)

mantle

MrPiedPiper:
Apparently that did _not_ in-fact fix walls not forming right... I'll get that fixed up.

IntegratedQuantum:
Let's say I want to fill a `cubyz:decorative/stone_pit` biome with lava, how would I go about this?
I tried a few things based on the volcano example you made, but I think the current system is too limited for it (note that stone pits can spawn at any height).
Screenshot at 2026-07-30 20-13-18
I think we'd need a mode that allows you to set it up relative to the biome's base height, but that is not currently accessible anywhere.

MrPiedPiper:
Unfortunately, in this PR's current form, it's best suited for small pools or preset heights. The best I can think of for a `cubyz:decorative/stone_pit` would be something like this (I forgot to change it to lava):
stone_pit
It's the default stone pit with these added
```
	.oceanHeight = 4,
	.relativeOceanGap = 3,
	.isOceanRelative = true,
```

I agree, having access to the base biome height would be a real game-changer. I assume that'd mean attaching that information to another pre-existing system. I could look into adding that 👍

Is that sort of information something that would belong in MapFragment?

IntegratedQuantum:
> Is that sort of information something that would belong in MapFragment?

Yes. Ideally all points on the map, instead of indexing the biome directly should index a palette of biomes, which could contain the base height of the biome and its biome seed (#431).

If you want to go for this, I'd suggest to make a separate PR for this.
In the mean time I'd like to get some artist opinion on the current system and parameters.

MrPiedPiper:
I've updated performance info in the original post 👍
Will update the missing settings bit soon.

careeoki (changes requested):
Played around with this for a couple minutes. The current parameters are nice and intuitive to work with. This would be quite useful, especially for caves.
I think it would be nice to be able to change the thickness of the walls. Other than that I'm not sure what else I would want, and I'm not desperate for it.

It seems to create floating water in surface caves. At Z:0 the water just stops, and the walls that would be there usually are not.
image

MrPiedPiper:
Wall thickness should be an easy addition, no idea what's going on with that cave. I'll see what I can find out 👍

As an aside note, I looked at codemob's work on this and saw that the biome settings for the oceans were all contained inside a struct.
```zig
pub const WaterProperties = struct {
	waterLevel: i32,
	waterBlock: main.blocks.Block,
};
```
https://github.com/codemob-dev/Cubyz/blob/d72b74b8f4c261e39e3efbde76da7eb6dd67632c/src/server/terrain/biomes.zig

I believe if I did this the properties would be set like so:
```zig
.climate = .{.mountain, .cold, .wet},
// This right here
.waterProperties = .{
	.height = 0, // tempted to call this "level" now
	liqiudBlock = "cubyz:water",
	isRelative = true,
}
.tags = .{.pine, .snowy},
.minHeight = 80,
.maxHeight = 256,
```
If something like this would be preferable I could move stuff around.

careeoki:
> If something like this would be preferable I could move stuff around.

I think that looks cleaner, yes.

H41ogen:
> Let's say I want to fill a `cubyz:decorative/stone_pit` biome with lava, how would I go about this? I tried a few things based on the volcano example you made, but I think the current system is too limited for it (note that stone pits can spawn at any height). Screenshot at 2026-07-30 20-13-18 I think we'd need a mode that allows you to set it up relative to the biome's base height, but that is not currently accessible anywhere.

it would probably help if the way stone pits were implemented wasnt super hacky

IntegratedQuantum:
> it would probably help if the way stone pits were implemented wasnt super hacky

which would require #614

PR #3429, opened 2026-07-26
fix item hand shader uniforms

### LINKED ISSUE #3428: Item in hand is not displayed.

This is since https://github.com/PixelGuys/Cubyz/commit/63e2b6c5fa1b09b5e5355d5fe86edd84373c6698
Thanks to @iNiKKo for noticing
on commit https://github.com/PixelGuys/Cubyz/commit/a4016d00b931e5b02ac265d8af02fe28054137dc

Image

on commit https://github.com/PixelGuys/Cubyz/commit/63e2b6c5fa1b09b5e5355d5fe86edd84373c6698

Image
@IntegratedQuantum I am pinging you as its your commit and because you mostly work on vulkan rewrite and know what is going on

IntegratedQuantum:
The mistake was that I thought the matrix could be statically initialized, which is not true since it depends on window aspect ratio.

broken by #3380

fixes #3428


Argmaster (approved):
...and they held items happily ever after. 

Image

PR #3440, opened 2026-07-30
Added disable option to TextInput

1-fr33:


src/gui/components/TextInput.zig
```diff
@@ -34,9 +34,10 @@ maxWidth: f32,
 maxHeight: f32,
 textSize: Vec2f = undefined,
 scrollBar: *ScrollBar,
-callbacks: Callbacks,
+options: Options,
 lastBlinkTime: std.Io.Timestamp = .fromNanoseconds(0),
-showCusor: bool = true,
+showCursor: bool = true,
+disabled: bool = false,
```
Wunka: you already have it in the options. you can access it from there, no need to have a copy here

Wunka (changes requested):

PR #3444, opened 2026-07-31
Bag UX changes prototype

Argmaster:
So, according to our content guidelines, to propose a change, you should make a prototype, thus here is a prototype. I had this in mind for some time and finally got around to making the changes.

image

image

image

image


ikabod-kee:
The idea is good, but the execution is not very good. 
image
There is like, no contrast between the items and it looks very visually cluttered.

Though the execution of a double-tall slot is very well-done. I love that aspect of it.

My recommendation would be to color the items to the color of the slot's background rather than adding transparency.
And also to make the gradient stronger, so that it seems like the bottom block is almost fully gone.
The most important thing is to make sure that the top item is visible.
Oh, and I suppose the bag icon on the bottom left should render above the items.

ikabod-kee:
image

ikabod-kee:
image
Maybe this would be a better compromise?

du82:
I like it!

Issue #1896, opened 2025-10-05
[WE] Add palette compression to `Blueprint` struct.


Argmaster:
Add palette compression to `Blueprint` struct.
Requires implementing `PaletteCompressedRegion` equivalent with size determined at runtime.

Related-to: #1894

PR #3446, opened 2026-07-31
Add `/count` command

### LINKED ISSUE #3043: `/count` command (worldedit)

Argmaster:
Adds trivial `/count` command implementation. 

image

Currently based on hash map, when we swap to palette storage for blueprints (#1896) we can reduce it to just printing palette entries.
It could be useful to get also entity counts and differentiate blocks with different data values, all out of scope, material for future improvements.

Resolves: #3043

### DIFF
--- a/src/server/command/worldedit/count.zig
+++ b/src/server/command/worldedit/count.zig
@@ -7,14 +7,17 @@ const Source = command.Source;
 const Block = main.blocks.Block;
 const Blueprint = main.blueprint.Blueprint;
 
-pub const description = "Count blocks in selection.";
-pub const usage = "/count";
+pub const description = "Count block(s) appearance(s) in selection.";
+pub const usage =
+	\\/count <blockId>
+	\\/count
+;
 
 pub const Args = union(enum) {
-	@"/count": struct {},
+	@"/count": struct { block: ?command.BlockId },
 };
 
-pub fn execute(_: Args, source: Source) void {
+pub fn execute(args: Args, source: Source) void {
 	if (source != .user) {
 		source.sendMessage("Command cannot be run without a user", .{});
 		return;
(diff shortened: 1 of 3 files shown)


Wunka:
Its still a draft, but would it maybe make sense to give it an optional value to say for which block you want to know the count?

Argmaster:
Yeah its doable, tho since this is primarily aimed at humans, I think its rare they will be looking for very specific block and they will care to write down the Id.

src/server/command/worldedit/count.zig
```diff
@@ -0,0 +1,64 @@
...
+		source.sendMessage("Command cannot be run without a user", .{});
+		return;
+	}
+	const user = source.user;
+	const selection = command.getCurrentSelection(user) catch return;
+
+	var result = Blueprint.capture(main.stackAllocator, selection);
+
+	switch (result) {
+		.success => |*success| {
```
IntegratedQuantum:
```suggestion
		.success => |*blueprint| {
```

src/server/command/worldedit/count.zig
```diff
@@ -0,0 +1,64 @@
...
+
+			success.apply(&context, countBlocks);
+
+			if (args.@"/count".block) |block| {
+				const count = context.get(block.block.typ) orelse 0;
+				user.sendMessage("#ffff00{s} #ffffff{d}", .{block.block.id(), count});
+			} else {
+				var iterator = context.iterator();
+				while (iterator.next()) |next| {
+					user.sendMessage("#ffff00{s} #ffffff{d}", .{(Block{.typ = next.key_ptr.*, .data = 0}).id(), next.value_ptr.*});
```
IntegratedQuantum: Could you print them sorted by count, descending?

IntegratedQuantum (changes requested):
Very useful, thanks for implementing

src/server/command/worldedit/count.zig
Argmaster: image

I also added padding on count and swapped ids with count, since ids tend to get quite long

IntegratedQuantum: Oh that looks so much nicer!

src/server/command/worldedit/count.zig
```diff
@@ -0,0 +1,75 @@
...
+			defer context.deinit(main.stackAllocator.allocator);
+
+			blueprint.apply(&context, countBlocks);
+
+			if (args.@"/count".block) |block| {
+				const count = context.get(block.block.typ) orelse 0;
+				user.sendMessage("#ffff00{s} #ffffff{d}", .{block.block.id(), count});
+			} else {
+				const TypAndCount = struct { typ: u16, count: u32 };
+				var items: main.List(TypAndCount) = .empty;
```
IntegratedQuantum, line 44: This is leaking

src/server/command/worldedit/count.zig
```diff
@@ -0,0 +1,75 @@
...
+			user.sendMessage("#ff0000Error while capturing block {}: {s}", .{e.pos, e.message});
+			std.log.warn("Error while capturing block {}: {s}", .{e.pos, e.message});
+		},
+	}
+}
+
+fn countBlocks(context: *std.AutoHashMapUnmanaged(u16, u32), current: Block) Block {
+	const result = context.getOrPut(main.stackAllocator.allocator, current.typ) catch unreachable;
+	if (result.found_existing) {
+		result.value_ptr.* = result.value_ptr.* + 1;
```
IntegratedQuantum:
```suggestion
		result.value_ptr.* += 1;
```

IntegratedQuantum (changes requested):

src/server/command/worldedit/count.zig
Argmaster: Added defer deinit.

IntegratedQuantum (approved):

Issue #3336, opened 2026-07-07
Permissions based on command arguments


Wunka:
This is a goal I had since the beginning of the permission system.
Here I want finally make an issue out of it and make this the place to discuss how that should look like.

What do I exactly mean with this?
Take for example the spawn command. 
What if we wanted everyone to be able to get their spawnpoint but from no one else (and not being able to set it)

through the argparser I think we can achive this.

For example the permissionPath could be: 
(whitelist)`/command/spawn/playerIndex=null` (allow running /spawn where playerIndex is null => the player themself
(blacklist)`/command/spawn/playerIndex=null/x` => disallows x parameter => you cannot set the spawnpoint (could also do `/command/spawn/playerIndex=null/x/y/z` but with x is shorter

We could also go by the `Args` union and get permissionPaths through that:
(whitelist)`/command/spawn/@"/spawn <playerIndex>"` (how do we set here that the player should only be able to do it for them self?) With `<playerIndex=null>`? or maybe we append the arguments after that like in my first example? so `/command/spawn/@"/spawn <playerIndex>"/playerIndex=null`?

(needs to integrate nicely with #3125 to not make different syntaxes, thats also where I took the `playerIndex=null` idea from)

Argmaster:
Overall I am not to found of this idea, the main reason is that not every command needs same granularity of control resulting in either boilerplate or insufficient control. I would prefer if permission checks were done explicitly, manually and through simple API - does user have permission to do that? No? Then GTFO. Yes? Great, let's continue.
Are the paths arbitrary or hard coded?
Imo it would be best of they were arbitrary, described in command documentation.
The most glaring case of privilege granularity being critical is `/sbb` and `/blueprint` commands - they are very complicated in terms of what they can access (or will be able to access in the future). As I recently described in discord message https://discord.com/channels/443805812390100992/702167480654757991/1524022936988811424

> So, I have ran into `assets.zig` pissing me off again. I guess after I hard resetted my last attempt on 
> refactoring it I am damned to revisit it.
> This time I wanted to finally enable loading blueprints from assets, but that's kind of hard, even if I were to brute force it by hardcode search logic. Basically, I need to be able to re-do the proper asset search process with up to date filesystem data. 
> Also, I have my previous wild idea of adding more abstraction to where server/client differentiation occurs.
> For requirements, just `/sbb` (still doesnt exist) and `/blueprint` suggest following inputs would be very, very important:
> - absolute path
> - relative path
> - id
[...]

PR #3447, opened 2026-07-31
Introduce CommandError Set

Wunka:
Does what was proposed in this review https://github.com/PixelGuys/Cubyz/pull/3223#discussion_r3650460160

### DIFF
--- a/src/server/command.zig
+++ b/src/server/command.zig
@@ -28,8 +28,12 @@ pub const Source = union(enum) {
 	}
 };
 
+pub const CommandError = error{
+	InvalidSource,
+};
+
 pub const Command = struct {
-	exec: *const fn (args: []const u8, source: Source) void,
+	exec: *const fn (args: []const u8, source: Source) CommandError!void,
 	name: []const u8,
 	description: []const u8,
 	usage: []const u8,
@@ -38,18 +42,18 @@ pub const Command = struct {
 
 pub var commands: std.StringHashMap(Command) = undefined;
 
-fn initExecutionFn(comptime name: []const u8) *const fn (args: []const u8, source: Source) void {
+fn initExecutionFn(comptime name: []const u8) *const fn (args: []const u8, source: Source) CommandError!void {
 	const ArgPaser = main.argparse.Parser(@field(commandList, name).Args, .{.commandName = name});
 	return struct {
-		fn exec(msg: []const u8, source: Source) void {
+		fn exec(msg: []const u8, source: Source) CommandError!void {
 			const arena: main.heap.NeverFailingAllocator = .createArena(main.stackAllocator);
 			defer main.stackAllocator.destroyArena(arena);
 			var errorMessage: main.ListManaged(u8) = .init(arena);
-			const result = ArgPaser.parse(arena, msg, &errorMessage) catch {
+			const args = ArgPaser.parse(arena, msg, &errorMessage) catch {
 				source.sendMessage("#ff0000{s}", .{errorMessage.items});
 				return;
 			};
-			@field(commandList, name).execute(result, source);
+			return @field(commandList, name).execute(args, source);
 		}
 	}.exec;
 }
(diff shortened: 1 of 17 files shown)


Argmaster:
Would it be possible to remove the `if (source != .user) return error.InvalidSource;` completely by simply changing the signature of `execute` depending on what is allowed, using comptime metaprogramming?

Both server and user mode:
```
pub fn execute(args: Args, source: ?*User) !void {
```

Only user mode:

```
pub fn execute(args: Args, source: *User) !void {
```

Only server mode (I don't think there are any cases?)

```
pub fn execute(args: Args) !void {
```

Wunka:
> Would it be possible to remove the `if (source != .user) return error.InvalidSource;` completely by simply changing the signature of `execute` depending on what is allowed, using comptime metaprogramming?

Maybe its just me but I want a command to actually think about why they don't support console etc, so I like it more if its "harder" to be restrictive.

Also the Source union may not stay at what it currently is. I could see that for example some kind of Entity / Block Source is also added. 

Also, some commands may only support console on specifc parts. with the error they can (if entered a "wrong" path) just return that and a message is displayed. I am also thinking about adding a `PermissionDenied` error for commands who check more explicit permssionPaths etc

Argmaster:
> think about why they don't support console etc, so I like it more if its "harder" to be restrictive.

Excessive boilerplate is not a way to make API less error prone. 
_What's with Zig programmers wanting to write more code than needed? (Referring to Quantums preference of no auto-serialization)._
Approach I proposed on its own should be easy to use right - it requires explicit declaration of intent through function argument type.

> Also the Source union may not stay at what it currently is. I could see that for example some kind of Entity / Block Source is also added.

For future extensions, implementation can be adjusted accordingly, changes in the source code of command system will be required anyway, since union will have to change.
Moreover, we cannot prepare for all the funny things future will bring and being ready to switch behavior based on signature is more flexible than using Unions.

Honestly I am puzzled about push for explicit source checks compared to #3336 - so, should it be more automatic or more explicit? Security doesn't require explicit, fine grained control, but function arguments do?

Wunka:
One nice thing about the union is that it can have functions (like it currently already has 2) so that commands don't really have to change unless they don't support something. 
Also this explicit source check should be pretty rare. In my opinion  all commands (outside of the worldedit commands) should support console. And for them the code which is the least is if they have the union. 

I could see adding support so you can do either: `Source` or `*User` (here is though again the question, what if we add a third option? will we add all possible options of the union as options for `args`?) but I don't think we should go to things like `?*User`. And also one of my points still stands. What about commands that are only restricted on specific paths? 

At the end I can see that this is boilerplate for the Worldedit commands, and I am fine if we add the option to accept `*User` for args but I don't see the union going away again or this pr. (I still think for path related things or permission things this pr would be nice)

IntegratedQuantum:
I thought about this for a bit, and I tend to agree with Argmaster, letting the type system handle the source selection is much safer than relying on the command creator to check their sources correctly (which could lead to a server crash).
This is (going to be) a moddable interface, so we should make it easy for the user to not make mistakes. And in my opinion not crashing beats not having the command available for one possible source.

Extending on argmasters idea I propose the following system:

The command system switches over the source, for each case it looks for a function of matching type and name `executeUser(Args, *User)` or `executeServer(Args, *User)`, if it doesn't find one it defaults to a default `execute(Args)` which has no additional parameter. This extra default would make it easy to default to all possibilities, if no requirements for the source exist.

Though honestly I'm not 100% sure if this is the best approach, it seems problematic with all the commands that take an optional user.

----

Another idea could be to put the source inside the parser struct, e.g.
```

pub const Args = union(enum) {
	@"/tp <biome>": struct { biome: command.BiomeId, source: *User },
	@"/tp <x> <y> <z>": struct {
		x: command.Coordinate,
		y: command.Coordinate,
		z: command.Coordinate,
                source: *User,
	},
	@"/tp <playerIndex>": struct { playerIndex: command.PlayerIndex, source: *User },
	@"/tp <playerIndex>": struct { sourcePlyerIndex: command.PlayerIndex, playerIndex: command.PlayerIndex },
	...
};
```

Wunka:
I will close this, as #3504 also suggests a change in the source system

PR #3449, opened 2026-07-31
Add disable option to TextInput

1-fr33:
Mistakenly delated the branch previously, so had to create a new pull request.

### DIFF
--- a/src/gui/components/TextInput.zig
+++ b/src/gui/components/TextInput.zig
@@ -494,7 +494,7 @@ pub fn newline(self: *TextInput, mods: main.Window.Key.Modifiers) void {
 }
 
 fn ensureCursorVisibility(self: *TextInput) void {
-	self.showCursor = !self.options.disabled;
+	if (self.options.disabled) return;
 	self.lastBlinkTime = main.timestamp();
 	if (self.textSize[1] > self.maxHeight - 2*border) {
 		var y: f32 = 0;
--- a/src/gui/components/TextInput.zig
+++ b/src/gui/components/TextInput.zig
@@ -464,6 +464,7 @@ pub fn copy(self: *TextInput, mods: main.Window.Key.Modifiers) void {
 }
 
 pub fn paste(self: *TextInput, mods: main.Window.Key.Modifiers) void {
+	if (self.options.disabled) return;
 	if (mods.control) {
 		const string = main.Window.getClipboardString();
 		self.deleteSelection();
--- a/src/gui/components/TextInput.zig
+++ b/src/gui/components/TextInput.zig
@@ -494,7 +494,7 @@ pub fn newline(self: *TextInput, mods: main.Window.Key.Modifiers) void {
 }
 
 fn ensureCursorVisibility(self: *TextInput) void {
-	self.cursor = true;
+	self.showCursor = true;
 	if (self.options.disabled) return;
 	self.lastBlinkTime = main.timestamp();
 	if (self.textSize[1] > self.maxHeight - 2*border) {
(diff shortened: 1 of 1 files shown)


src/gui/components/TextInput.zig
```diff
@@ -432,6 +436,7 @@ pub fn inputCharacter(self: *TextInput, character: u21) void {
 }
 
 pub fn setString(self: *TextInput, utf8EncodedString: []const u8) void {
+	if (self.options.disabled) return;
```
IntegratedQuantum: setString is a public function, how else would someone from the outside change the string of a disabled TextInput?

src/gui/components/TextInput.zig
IntegratedQuantum, line 466: paste should be disabled

IntegratedQuantum:
Could you explain the point of a disabled text input?
I think it would make more sense to add the desired functionality (selecting, copying) to the label, instead of manually disabling all function that edit its contents (and requiring every future contributor of TextInput to know and apply this check)

1-fr33:
> Could you explain the point of a disabled text input? I think it would make more sense to add the desired functionality (selecting, copying) to the label, instead of manually disabling all function that edit its contents (and requiring every future contributor of TextInput to know and apply this check)

The reason for having to add a disabled option is to tell the player that the content of the field cannot be edited. When making the edit_world window, I was thinking of showing the seed value as a Label but it didnt look good. The TextInput changes work without changing any previous code.

IntegratedQuantum:
So it's mainly the visual difference between a label and a text field? So the background color and (optionally) the scrollbar?
Wouldn't it then be more useful to allow vertical/HorizontalLists to overwrite the background color?

1-fr33:
Yes, that can be done but It will help us in the long run, I have used a few UI libraries that give us an option to disable the text field, so I think that maybe it is better than changing the background color of of a list.

IntegratedQuantum:
right, that makes sense.

src/gui/components/TextInput.zig
```diff
@@ -555,12 +562,14 @@ pub fn render(self: *TextInput, mousePosition: Vec2f) void {
...
-			self.showCusor = !self.showCusor;
+		if (!self.options.disabled) {
+			if (self.lastBlinkTime.durationTo(currentTime).nanoseconds > blinkDuration.nanoseconds) {
+				self.lastBlinkTime = currentTime;
+				self.showCursor = !self.showCursor;
+			}
 		}
 
-		if (self.showCusor) {
+		if (self.showCursor) {
```
IntegratedQuantum: Instead of overwriting all behaviors of setCursor, why not just add `and !self.disabled` here?

IntegratedQuantum (changes requested):

src/gui/components/TextInput.zig
1-fr33: I added that in the start but removed it later, thought that there will be too many if statements

src/gui/components/TextInput.zig
1-fr33: Yup, I didnt think that we might update the TextInput later on, will fix this.

src/gui/components/TextInput.zig
```diff
@@ -481,16 +485,16 @@ pub fn cut(self: *TextInput, mods: main.Window.Key.Modifiers) void {
...
+	if (!mods.shift and self.options.onNewline.inner != null and !self.options.disabled) {
+		self.options.onNewline.run();
 		return;
 	}
 	self.inputCharacter('\n');
 	self.ensureCursorVisibility();
 }
 
 fn ensureCursorVisibility(self: *TextInput) void {
-	self.showCusor = true;
```
IntegratedQuantum, line 493: unintended removal

IntegratedQuantum (changes requested):

IntegratedQuantum (approved):

PR #3452, opened 2026-08-01
save permission groups in a list

Wunka:


src/server/permission.zig
```diff
@@ -111,25 +111,16 @@ pub const Permissions = struct { // MARK: Permissions
 
 pub const Group = struct { // MARK: Group
 	permissions: Permissions,
-	// Each group must have a unique ID to avoid stale membership issues.
-	// Example scenario:
-	// - User1 joins Group1
-	// - Group1 is deleted while User1 is offline (so their data isn’t updated)
-	// - A new Group1 is created
-	// - When User1 reconnects, they are incorrectly treated as a member of the new Group1
```
IntegratedQuantum, line 119: Maybe this comment should be put elsewhere?

src/server/permission.zig
```diff
@@ -200,54 +190,53 @@ pub const Group = struct { // MARK: Group
 	}
 };
 
-var groups: std.StringHashMapUnmanaged(*Group) = .{};
+var groups: main.ListManaged(?*Group) = undefined;
+var groupNameToIdMap: std.StringHashMapUnmanaged(usize) = .{};
```
IntegratedQuantum: I'd suggest to stay with `u32`, with each entry costing 8 bytes best case (when deleted), we will have severe memory issues long before we get to 4 billion. (And we can easily increase the limit later when we actually do need it, decreasing on the other hand is more difficult)
```suggestion
var groupNameToIdMap: std.StringHashMapUnmanaged(u32) = .{};
```

IntegratedQuantum (changes requested):

src/server/permission.zig
Wunka: Added it before the groups definition

src/server/permission.zig
Wunka: yeah I just did it because arrays are indexed by usize. So now I need to `@truncate` at a few points

IntegratedQuantum (approved):

PR #3458, opened 2026-08-02
allow player to player in tp command

Wunka:


src/server/command/tp.zig
```diff
@@ -7,8 +7,8 @@ const Source = command.Source;
 pub const description = "Teleport to location.";
 pub const usage =
 	\\/tp <biome>
-	\\/tp <x> <y> <z>
-	\\/tp @<playerIndex>
+	\\/tp @<playerIndex> <x> <y> <z>
+	\\/tp @<playerIndex> @<playerIndex>
```
IntegratedQuantum: This is very unhelpful. Please specify which one is source and destination (same in the code below)

IntegratedQuantum (changes requested):

src/server/command/tp.zig
Wunka: yeah, I was not sure what to name them. Now I have something

src/server/command/tp.zig
```diff
@@ -7,8 +7,8 @@ const Source = command.Source;
 pub const description = "Teleport to location.";
 pub const usage =
 	\\/tp <biome>
-	\\/tp <x> <y> <z>
-	\\/tp @<playerIndex>
+	\\/tp @<playerIndex> <x> <y> <z>
```
IntegratedQuantum: Please also make this one

src/server/command/tp.zig
```diff
@@ -7,8 +7,8 @@ const Source = command.Source;
 pub const description = "Teleport to location.";
 pub const usage =
 	\\/tp <biome>
```
IntegratedQuantum, line 9: You should be able to teleport other players to biomes.

IntegratedQuantum (changes requested):
Also I think this is making it a lot more verbose to use on yourself (e.g. I now need to know my own player index), I'd suggest to make the source player optional.

src/server/command/tp.zig
Wunka: If this refers to the optional, I had before a version where it just was with optional playerIndexes at the front of each sub command. but for `/tp ?playerIndex playerIndex` if you typed: `/tp @0` it would instead of teleporting to `@0` show an error for not enough arguments. thats why I switched to a version where I have a copy of each with the playerIndex infront. 

> Also I think this is making it a lot more verbose to use on yourself (e.g. I now need to know my own player index), I'd suggest to make the source player optional

In regards to this. No this never happens. you can still teleport to another player by just doing `/tp @<playerIndex>` this will infer your own playerIndex

IntegratedQuantum: It's referring to naming it with source as well

Wunka: Okay I added a bit more to the command usage and added biome support. Also named everything.

src/server/command/tp.zig
```diff
@@ -7,31 +7,53 @@ const Source = command.Source;
...
+	\\/tp @<playerIndex> <biome>
```
IntegratedQuantum: Again, please always say in text that this is the source, same with coordinates

src/server/command/tp.zig
```diff
@@ -7,31 +7,53 @@ const Source = command.Source;
...
+		biome: command.BiomeId,
+	},
 	@"/tp <x> <y> <z>": struct {
 		x: command.Coordinate,
 		y: command.Coordinate,
 		z: command.Coordinate,
 	},
-	@"/tp <playerIndex>": struct { playerIndex: command.PlayerIndex },
+	@"/tp <sourcePlayerIndex> <x> <y> <z>": struct {
+		sourcePlayerIndex: command.PlayerIndex,
```
IntegratedQuantum: You could make it optional for biome and coordinates though right?

IntegratedQuantum (changes requested):

src/server/command/tp.zig
Wunka: yes, done. also changed the last description errors

IntegratedQuantum (approved):

Issue #3451, opened 2026-08-01
Convert all item color lists into textures


IntegratedQuantum:
As added in #3403

Also the fallback code related to loading from zon should be removed and an error should be produced when the image is not specified.

PR #3460, opened 2026-08-02
Ingots - Palette Textures part 1

careeoki:
progress towards #3451 

Doing this across multiple PRs so it is more manageable for myself
image


IntegratedQuantum:
Before you do too much labor here, note that @yel0h said in #3403 that they could write a script for this

IntegratedQuantum:
(of course you'd probably still end up tweaking the results, but then you wouldn't need to manually create and link the files)

careeoki:
Most the work here is tweaking the colors, but sure, if they want to make a script that could be helpful. I haven't done any other materials.

IntegratedQuantum (approved):

Issue #3241, opened 2026-06-15
Integrate more stuff into ECS


IntegratedQuantum:
- [ ] `main.game.Player.super` should be just an entity id, consequently server.Entity switching on Side can be removed
- [ ] interpolation should be done using a component and a system
- [ ] gamemode and other states like flying should be components, this requires integrating ECS with the sync system
- [ ] The inventory (and the hand inventory, and other future types) should be components
- [ ] Basically all other variables in `game.Player` should be part of the ECS
- [ ] Same for `server.User`, it should also be split into components
- [ ] There should be a (or multiple) system for all the player-related logic in `game.update`
- [ ] `server.Entity` and `client.Entity` should be merged into a single type of all the data common to all entities, everything else (e.g. health, hunger, interpolation data, name) should be split off into individual components.

PR #3462, opened 2026-08-02
Respawn Block

Crepestrom:
A block that lets you set your spawnpoint


du82:
This is what beds are for

IntegratedQuantum:
This is what totems are for (#1468)

BoySanic:
> This is what beds are for

What about those of us who sleep in hammocks?

Crepestrom:
i wanted to test their functioality here before making all the assets for totems

IntegratedQuantum (changes requested):
Please first move the respawn point into an ECS component in a separate PR. see #3241

Crepestrom:
that will wait on #3562 for client to server communication of ecs

PR #3463, opened 2026-08-02
Convert remaining item color lists into palette textures

### LINKED ISSUE #3451: Convert all item color lists into textures

yel0h:
Finishes the migration started in #3403 and #3460.

Removed the legacy `.colors` parsing fallback in `Material.init` and helpers. `loadColorsFromTexture` now always resolves a color source instead of returning a `.notLoaded` status for the caller to fall back on. Missing `.colorTexture` field, failed load, or invalid image (too small or with non-opaque outline pixels) all log an error and fall back to `graphics.Image.defaultImage`.

Closes #3451


src/items.zig
```diff
@@ -39,50 +39,52 @@ const Material = struct { // MARK: Material
...
-		}
+		defer if (loadedImage) |img| img.deinit(main.stackAllocator);
+
+		const image: graphics.Image = if (loadedImage) |img| validate: {
+			if (img.width < 2 or img.height < 2) {
+				std.log.err("Material color texture '{s}' must be at least 2x2 pixels (got {}x{}).", .{colorTexturePath, img.width, img.height});
+				break :validate graphics.Image.defaultImage;
+			}
+			break :validate img;
+		} else graphics.Image.defaultImage;
```
IntegratedQuantum: Instead of all this complex logic for handling the default image and error printing, I'd suggest to use errors to simplify the process:
1. Change the return value to `!void`
2. at the callsite handle the error by setting some default colors:
```
catch |err| {
    std.log.err("Could not load material colors for item {s} from paths {s} and {s}: {s}", .{self.id, path1hhawehu, path2dwdhihawdhuw, @errorName(err)});
    self.outlineColorLight = 0xff00ff;
    self.outlineColorDark = 0x000000;
    self.colorPalette = &.{0xff00ff, 0x000000, 0xff00ff, 0x000000};
};
```
3. all error cases in here can be replaced with a simple `return error.@"MissingAttribute: colorTexture"` or `return error.@"Color texture must be at least 2×2 pixels"` or `return error.@"Colors must be fully opaque"`

IntegratedQuantum (changes requested):
Also for completeness, would you mind sharing the script you used for conversion? It would be handy if someone searches for something similar in the future.

yel0h:
Sure, here's the script: https://gist.github.com/yel0h/d9ccbdf51c59e3915feceff490024872

IntegratedQuantum:
Oh thanks, looking at the script, it seems that you totally missed items that do not have a block (e.g. metal ingots which carrie already converted, so nevermind)

IntegratedQuantum (approved):
Looks good now, except from a small conversion issue for amber (which mistakenly used 9 hex digits in one of the colors, tricking your regex), but not a big deal I'm sure @careeoki will fix it once she goes through them to add better shadow colors.

IntegratedQuantum:
By the way if you want to keep working on improving procedural item textures, there is also the idea to add custom small (4×4) textures for gems and stuff.

PR #3464, opened 2026-08-02
Fix handshake panic

### LINKED ISSUE #3477: Crash when hosting world on master (on linux)

With a fresh world. Apparently not an issue on windows, my brother was able to host a world on master just fine
Image

yel0h:
My PR #3365 introduced a bug which crashes the game on debug during connection to a server. 3 of the functions I added assert server context, but they're actually called on the network thread (`handShake.serverReceive` calls `User.identifyFromKeysAndName`). This removes those asserts.

fixes #3477

### DIFF
--- a/src/server/players.zig
+++ b/src/server/players.zig
@@ -83,8 +83,6 @@ pub fn isEmpty() bool {
 }
 
 pub fn allocateNewIndex() usize {
-	mutex.lock();
-	defer mutex.unlock();
 	return nextPlayerIndex.fetchAdd(1, .monotonic);
 }
 
@@ -128,6 +126,7 @@ const EnsureResult = struct { entry: *PlayerRecord, wasNew: bool };
 
 fn ensurePlayerRecord(key: []const u8) EnsureResult {
 	sync.threadContext.assertCorrectContext(.server);
+	mutex.assertLocked();
 
 	const result = playerDatabase.getOrPut(main.worldArena.allocator, key) catch unreachable;
 	if (result.found_existing) return .{.entry = result.value_ptr, .wasNew = false};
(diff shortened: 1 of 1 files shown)


IntegratedQuantum (changes requested):
HashMaps are not thread safe!
If playerDatabase is accessed from multiple threads, then all accesses to it (except from loading, which happens before players can join) need to lock a mutex (or assert that it's locked by the caller).

Also is this responsible for #3477?

yel0h:
I'm aware, but originally, prior to my PR, no locks were used for this either. But yes, it does look like this is what caused #3477, and it no longer crashes with the locks added.

src/server/players.zig
```diff
@@ -68,23 +70,28 @@ pub fn getLocalPlayerIndex() usize {
...
+	mutex.lock();
+	defer mutex.unlock();
 	return playerDatabase.size == 0;
 }
 
 pub fn allocateNewIndex() usize {
-	sync.threadContext.assertCorrectContext(.server);
+	mutex.lock();
+	defer mutex.unlock();
 	return nextPlayerIndex.fetchAdd(1, .monotonic);
```
IntegratedQuantum, line 86: mutex is unnecessary here, access is already synchronized by the atomic

src/server/players.zig
IntegratedQuantum, line 131: This function also accesses the HashMap, so please add mutex.assertLocked() to the start of the function. This is part of the pattern we have for making it easy to spot mistakes in concurrency.

IntegratedQuantum (changes requested):

yel0h:
Done

IntegratedQuantum (approved):

PR #3467, opened 2026-08-03
Enforce whitelist and add command

### LINKED ISSUE #2566: Player whitelist

yel0h:
Enforces the player whitelist during the connection handshake and adds a `/whitelist` command to manage it.

After a connecting player's key is resolved, the server checks `players.isAllowedToJoin` when `world.settings.whitelistEnabled` is set, rejecting the connection with `error.NotWhitelisted`.

Command syntax: `/whitelist <add/block> <keyType>:<base64Key>` or `/whitelist <add/block> @<playerIndex>`

Also removes error logging for `error.NotWhitelisted` since it's expected to reject and already logs the reason right after rejection.

Closes #2566

### DIFF
--- a/src/network.zig
+++ b/src/network.zig
@@ -1737,16 +1737,11 @@ pub const Connection = struct { // MARK: Connection
 
 	pub fn receive(self: *Connection, data: []const u8) void {
 		self.tryReceive(data) catch |err| {
-			switch (err) {
-				error.NotWhitelisted => {},
-				else => {
-					std.log.err("Got error while processing received network data: {s}", .{@errorName(err)});
-					if (@errorReturnTrace()) |trace| {
-						std.log.info("{f}", .{main.fmt.FormatErrorTrace{.stackTrace = trace.*}});
-					}
-					std.log.debug("Packet data: {any}", .{data});
-				},
+			std.log.warn("Got error while processing received network data: {s}", .{@errorName(err)});
+			if (@errorReturnTrace()) |trace| {
+				std.log.info("{f}", .{main.fmt.FormatErrorTrace{.stackTrace = trace.*}});
 			}
+			std.log.debug("Packet data: {any}", .{data});
 			self.disconnect();
 		};
 	}
--- a/src/network.zig
+++ b/src/network.zig
@@ -1737,11 +1737,16 @@ pub const Connection = struct { // MARK: Connection
 
 	pub fn receive(self: *Connection, data: []const u8) void {
 		self.tryReceive(data) catch |err| {
-			std.log.err("Got error while processing received network data: {s}", .{@errorName(err)});
-			if (@errorReturnTrace()) |trace| {
-				std.log.info("{f}", .{main.fmt.FormatErrorTrace{.stackTrace = trace.*}});
+			switch (err) {
+				error.NotWhitelisted => {},
+				else => {
+					std.log.err("Got error while processing received network data: {s}", .{@errorName(err)});
+					if (@errorReturnTrace()) |trace| {
+						std.log.info("{f}", .{main.fmt.FormatErrorTrace{.stackTrace = trace.*}});
+					}
+					std.log.debug("Packet data: {any}", .{data});
+				},
 			}
-			std.log.debug("Packet data: {any}", .{data});
 			self.disconnect();
 		};
 	}
--- a/src/network/protocols.zig
+++ b/src/network/protocols.zig
@@ -171,7 +171,7 @@ pub const handShake = struct { // MARK: handShake
 						const keys = zon.getChild("keys");
 						try conn.user.?.identifyFromKeysAndName(name, keys);
 
-						if (!main.server.players.isAllowedToJoin(conn.user.?.newKeyString.?, main.server.world.?.whitelistEnabled.load(.monotonic))) {
+						if (!main.server.players.isAllowedToJoin(conn.user.?.newKeyString.?, main.server.world.?.settings.whitelistEnabled.load(.monotonic))) {
 							std.log.info("Rejected connection from '{s}' ({s})", .{name, conn.user.?.newKeyString.?});
 							return error.NotWhitelisted;
 						}
--- a/src/network/protocols.zig
+++ b/src/network/protocols.zig
@@ -171,11 +171,6 @@ pub const handShake = struct { // MARK: handShake
 						const keys = zon.getChild("keys");
 						try conn.user.?.identifyFromKeysAndName(name, keys, main.server.world.?.settings.whitelistEnabled.load(.monotonic));
 
-						if (!conn.user.?.allowedToJoin) {
-							std.log.info("Rejected connection from '{s}' ({s})", .{name, conn.user.?.newKeyString.?});
-							return error.NotWhitelisted;
-						}
-
 						var writer: utils.BinaryWriter = .init(main.stackAllocator);
 						defer writer.deinit();
 						writer.writeEnum(Connection.HandShakeState, .signatureRequest);
--- a/src/network/protocols.zig
+++ b/src/network/protocols.zig
@@ -169,9 +169,9 @@ pub const handShake = struct { // MARK: handShake
 
 					if (main.server.world.?.mode != .singleplayer) {
 						const keys = zon.getChild("keys");
-						try conn.user.?.identifyFromKeysAndName(name, keys);
+						try conn.user.?.identifyFromKeysAndName(name, keys, main.server.world.?.settings.whitelistEnabled.load(.monotonic));
 
-						if (!main.server.players.isAllowedToJoin(conn.user.?.newKeyString.?, main.server.world.?.settings.whitelistEnabled.load(.monotonic))) {
+						if (!conn.user.?.allowedToJoin) {
 							std.log.info("Rejected connection from '{s}' ({s})", .{name, conn.user.?.newKeyString.?});
 							return error.NotWhitelisted;
 						}
--- a/src/network/protocols.zig
+++ b/src/network/protocols.zig
@@ -171,16 +171,9 @@ pub const handShake = struct { // MARK: handShake
 						const keys = zon.getChild("keys");
 						try conn.user.?.identifyFromKeysAndName(name, keys);
 
-						switch (main.server.players.isAllowedToJoin(conn.user.?.newKeyString.?)) {
-							.allowed => {},
-							.blocked => {
-								std.log.info("Rejected connection from '{s}': blocked", .{name});
-								return error.NotWhitelisted;
-							},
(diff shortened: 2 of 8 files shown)


src/network/protocols.zig
```diff
@@ -171,6 +171,11 @@ pub const handShake = struct { // MARK: handShake
 						const keys = zon.getChild("keys");
 						try conn.user.?.identifyFromKeysAndName(name, keys);
 
+						if (main.server.world.?.settings.whitelistEnabled and !main.server.players.isAllowedToJoin(conn.user.?.newKeyString.?)) {
```
Wunka: I don't like this. In my mind this can perfectly work in sync with a ban system (blocked = banned)
I would suggest to go in the direction I went with permissions. Instead of returning in `isAllowedToJoin` a `bool` return an `enum: {allowed, neutral, blocked}` neutral would then be the case where you need to look if whitelist is enabled

src/server/command/whitelist.zig
```diff
@@ -0,0 +1,64 @@
...
+}
+
+fn applyAction(source: Source, action: Action, key: []const u8) void {
+	switch (action) {
+		.add => switch (players.add(key)) {
+			.added => source.sendMessage("#00ff00Added {s}§#00ff00 to the whitelist", .{key}),
+			.alreadyAllowed => source.sendMessage("#ff0000{s}§#ff0000 is already on the whitelist", .{key}),
+		},
+		.block => switch (players.block(key)) {
+			.blocked => source.sendMessage("#00ff00Blocked {s}§#00ff00 from connecting", .{key}),
```
Wunka: If we go by the principle that blocked = ban then I would suggest to also kick the player here

src/server/command/whitelist.zig
```diff
@@ -0,0 +1,64 @@
...
+			.alreadyBlocked => source.sendMessage("#ff0000{s}§#ff0000 is already blocked", .{key}),
+		},
+	}
+}
+
+const KeyString = struct {
```
Wunka: In general we put these kinds of helper in `command.zig` I can already imagine there being a command to for example update the key where this would help

Wunka (changes requested):
Thanks for continuing with this!

yel0h:
True, it does make more sense to treat an explicit block as a ban

src/network/protocols.zig
```diff
@@ -171,6 +171,18 @@ pub const handShake = struct { // MARK: handShake
 						const keys = zon.getChild("keys");
 						try conn.user.?.identifyFromKeysAndName(name, keys);
 
+						switch (main.server.players.isAllowedToJoin(conn.user.?.newKeyString.?)) {
+							.allowed => {},
+							.blocked => {
+								std.log.info("Rejected connection from '{s}': blocked", .{name});
```
IntegratedQuantum: The name alone is rather unhelpful, it would make sense to also print the public key string

src/server/command.zig
```diff
@@ -169,6 +169,22 @@ pub const PlayerIndex = struct {
 	}
 };
 
+pub const KeyString = struct {
+	key: []const u8,
+
+	pub fn parse(_: NeverFailingAllocator, name: []const u8, arg: []const u8, errorMessage: *ListManaged(u8)) error{ParseError}!KeyString {
+		const colonIndex = std.mem.indexOfScalar(u8, arg, ':') orelse {
+			errorMessage.print("Expected a public key of the form \"<keyType>:<base64>\" for <{s}>, found \"{s}\"", .{name, arg});
```
IntegratedQuantum, line 177: Instead of adding your own validation logic, please use `authentication.PublicKey.initFromBase64`

src/network/protocols.zig
```diff
@@ -171,6 +171,18 @@ pub const handShake = struct { // MARK: handShake
 						const keys = zon.getChild("keys");
 						try conn.user.?.identifyFromKeysAndName(name, keys);
 
+						switch (main.server.players.isAllowedToJoin(conn.user.?.newKeyString.?)) {
```
IntegratedQuantum: I'd prefer to have this switch and the check for `whitelistEnabled` inside the isAllowedToJoin function, also in my opinion the same generic message for both paths would be enough.

src/server/server.zig
```diff
@@ -891,3 +891,14 @@ pub fn getUserByIndex(index: PlayerIndex) ?*User {
 	}
 	return null;
 }
+
+pub fn getUserByKey(key: []const u8) ?*User {
```
IntegratedQuantum: This is a one-time use case, there is no need to pollute the namespace with it, everything used inside here is already public, so please inline it to the implementation site.

src/network.zig
```diff
@@ -1737,11 +1737,16 @@ pub const Connection = struct { // MARK: Connection
 
 	pub fn receive(self: *Connection, data: []const u8) void {
 		self.tryReceive(data) catch |err| {
-			std.log.err("Got error while processing received network data: {s}", .{@errorName(err)});
-			if (@errorReturnTrace()) |trace| {
-				std.log.info("{f}", .{main.fmt.FormatErrorTrace{.stackTrace = trace.*}});
+			switch (err) {
+				error.NotWhitelisted => {},
```
IntegratedQuantum: I don't think this code here should worry about the error sets of arbitrary network protocols. I assume you want to get rid of the error popup? I think the right solution here would be to change the error to a warning, there are many other cases where this code is run (e.g. version differences) that probably shouldn't have the error popup either.

src/server/command/whitelist.zig
```diff
@@ -0,0 +1,51 @@
...
+
+const main = @import("main");
+const command = main.server.command;
+const Source = command.Source;
+const players = main.server.players;
+
+pub const description = "Manages the connection whitelist";
+pub const usage =
+	\\/whitelist <add/block> <keyType>:<base64Key>
+	\\/whitelist <add/block> @<playerIndex>
```
IntegratedQuantum, line 11: Would it be possible to also have whitelist enable/disable commands?
To implement this in a thread-safe manner you can just make the bool atomic.

IntegratedQuantum (changes requested):

yel0h:
Sure, but if `whitelistEnabled` is supposed to be changed after world creation, and since commands can now be entered through the console for headless, I think it should be part of `ServerWorld` itself and not the `Settings` struct, whose values don't change after creation.

IntegratedQuantum:
WorldSettings will also include stuff like gamerules in the future, which I would also like to have configurable in-game.

Wunka:
> Sure, but if `whitelistEnabled` is supposed to be changed after world creation, and since commands can now be entered through the console for headless, I think it should be part of `ServerWorld` itself and not the `Settings` struct, whose values don't change after creation.

Actually they will change: #3499 (maybe not from within the world, but that can also change)

yel0h:
Ahh okay. Fine to save like this then?

src/server/command/whitelist.zig
```diff
@@ -0,0 +1,65 @@
...
+		.@"/whitelist <action> <playerIndex>" => |params| {
+			const target = command.Target.fromPlayerIndex(params.playerIndex, source) catch return;
+			const key = target.user.newKeyString orelse {
+				source.sendMessage("#ff0000Player {s}§#ff0000 has no public key to whitelist", .{target.user.name});
+				return;
+			};
+			applyAction(source, params.action, key);
+		},
+		.@"/whitelist <enable/disable>" => |params| {
+			main.server.world.?.settings.whitelistEnabled.store(params.toggle == .enable, .monotonic);
```
IntegratedQuantum, line 36: The change also needs to be stored to disk immediately to avoid losing it when the game doesn't close correctly. (`world.saveWorldConfig`)

src/network/protocols.zig
```diff
@@ -171,6 +171,11 @@ pub const handShake = struct { // MARK: handShake
 						const keys = zon.getChild("keys");
 						try conn.user.?.identifyFromKeysAndName(name, keys);
 
+						if (!main.server.players.isAllowedToJoin(conn.user.?.newKeyString.?, main.server.world.?.settings.whitelistEnabled.load(.monotonic))) {
```
IntegratedQuantum: There is a problem with this. There are multiple keys in order to allow switching to a new crypto algorithm when one is broken. So when the server switches this (for testing you can do this in the launchConfig), it will then reject the old player which is still in their list with one of the old keys.

To fix this I think the best solution would be to do this check together with `players.lookupIndex` in `User.identifyFromKeysAndName`, so that the old entries are found correctly.

IntegratedQuantum (changes requested):

yel0h:
Ok, I merged `players.isAllowedToJoin` into `lookupIndex`, works with algorithm switching now

src/server/server.zig
```diff
@@ -232,9 +233,10 @@ pub const User = struct { // MARK: User
 		self.jobQueue.deinit();
 	}
 
-	pub fn identifyFromKeysAndName(self: *User, name: []const u8, keys: main.ZonElement) !void {
+	pub fn identifyFromKeysAndName(self: *User, name: []const u8, keys: main.ZonElement, whitelistEnabled: bool) !void {
 		std.debug.assert(self.name.len == 0);
 		self.name = main.globalAllocator.dupe(u8, name);
+		self.allowedToJoin = !whitelistEnabled;
```
IntegratedQuantum: Why not just make this a local variable and return the error here? There is no need to waste extra runtime memory for something that is needed once and at the function callsite.

src/server/players.zig
```diff
@@ -69,11 +69,13 @@ pub fn getLocalPlayerIndex() usize {
 	return localPlayerIndex;
 }
 
-pub fn lookupIndex(key: []const u8) ?usize {
+const LookupResult = struct { playerIndex: usize, blocked: bool };
+
+pub fn lookupIndex(key: []const u8) ?LookupResult {
```
IntegratedQuantum:
```suggestion
pub fn lookupIndex(key: []const u8) ?PlayerRecord {
```

src/server/players.zig
```diff
@@ -69,11 +69,13 @@ pub fn getLocalPlayerIndex() usize {
...
 	mutex.lock();
 	defer mutex.unlock();
 	const entry = playerDatabase.get(key) orelse return null;
-	return entry.playerIndex;
+	return .{.playerIndex = entry.playerIndex, .blocked = entry.blocked};
```
IntegratedQuantum:
```suggestion
	return entry;
```

IntegratedQuantum (changes requested):

yel0h:
Fixed

IntegratedQuantum (approved):
Looks good, and appears to work, thanks for working on this, it's an important milestone for servers.

PR #3474, opened 2026-08-04
Arid Grass + Savannah Fields

careeoki:
Arid grass is a new grass type that grows on sand. It has a very dry, golden color.

image

Savannah fields  - A savannah variant with lots of arid grass.
image

Bushy savannah fields - Subbiome with less arid grass and some bushes and marigolds.
image

Arid savannah fields - Subbiome with less dry grass and fewer trees.
image

For an example build application, it can make some nice pathes:
image


TyrrelMeter:
Honestly for arid savannah fields I'd go with even less dry grass. It looks a bit too green. I think some is helpful but it really doesn't match the grass blocks.

careeoki:
reduced the amount of dry grass in arid savannah field

IntegratedQuantum:
arid grass and tussock do look pretty similar. Do we really need both?

careeoki:
> arid grass and tussock do look pretty similar. Do we really need both?

Originally I planned to turn tussock into arid grass. But the color I ended up with for arid grass was a good bit darker than tussock, so I kept them seperate. I don't think it's a bad thing to have some similar-ish plants for some variation.

Buuut just to be sure: I changed the tussock to be a bit brighter. You should be able to see right away that it's tussock and not arid grass.


image
image

IntegratedQuantum (approved):
The biome also makes me wish we had some desert flowers

PR #3479, opened 2026-08-06
Allow the time difference to jump when it notices that its consistently too far off from the true value

IntegratedQuantum:
This is detected by checking whether the last 5 seconds worth of samples were off in the same direction, at which point it should be obvious that the old estimate is wrong by a big margin.

This improves correction of an issue on startup where the initial time difference is overestimated, causing interpolation on the client to be incorrect.

(Note this does not fix the root issue, I'll make an issue for the root issue in a bit)

@iNiKKo Please confirm that this fixes the issue, it will still be present for 5 seconds, but then it should fix itself.

### DIFF
--- a/src/utils.zig
+++ b/src/utils.zig
@@ -1620,19 +1620,26 @@ pub fn GenericInterpolation(comptime elements: comptime_int) type { // MARK: Gen
 
 pub const TimeDifference = struct { // MARK: TimeDifference
 	difference: Atomic(i16) = .init(0),
-	firstValue: bool = true,
+	// Checks how many times in a row we are off in the same direction
+	biasCounter: Atomic(i16) = .init(std.math.maxInt(i16)),
 
 	pub fn addDataPoint(self: *TimeDifference, time: i16) void {
 		const currentTime: i16 = @truncate(main.timestamp().toMilliseconds());
 		const timeDifference = currentTime -% time;
-		if (self.firstValue) {
+		if (@abs(self.biasCounter.load(.monotonic)) > main.server.updatesPerSec*5) {
 			self.difference.store(timeDifference, .monotonic);
-			self.firstValue = false;
+			self.biasCounter.store(0, .monotonic);
 		}
 		if (timeDifference -% self.difference.load(.monotonic) > 0) {
 			_ = @atomicRmw(i16, &self.difference.raw, .Add, 1, .monotonic);
+			if (self.biasCounter.fetchAdd(1, .monotonic) < 0) {
+				self.biasCounter.store(0, .monotonic);
+			}
 		} else if (timeDifference -% self.difference.load(.monotonic) < 0) {
 			_ = @atomicRmw(i16, &self.difference.raw, .Add, -1, .monotonic);
+			if (self.biasCounter.fetchAdd(-1, .monotonic) > 0) {
+				self.biasCounter.store(0, .monotonic);
+			}
 		}
 	}
 };


iNiKKo:
image
In local multiplayer, I will test a bit more to make sure this is not caused by my setup.

IntegratedQuantum:
same account multiplayer is no longer supported as of #3434

iNiKKo:
image
When tried to "host" a world (without dev mode) tried on 2 different machines

IntegratedQuantum:
This is unrelated too #3477

iNiKKo:
I cannot host local due to same account (#3434) and I can't host multiplayer due to #3477 
So I cannot test this fix, unless I generate another account to test local (which I will try next)

IntegratedQuantum:
Can you not host local with different account?

iNiKKo:
I am not sure where my account key is, so changing account might make me lose my current one.

IntegratedQuantum:
Then set a different cubyz dir in the launch config

IntegratedQuantum:
or, even better, copy it from the settings file and store it somewhere secure this time

iNiKKo:
image

Local multiplayer with two accounts
error when joining via localhost
(I did the same thing as with same account but on same account i do actually join the server.

iNiKKo:
image

**These don't work**
1. Local Online multiplayer ( 2 pc's same network)
2. Local multiplayer with same account
3. Local multiplayer with different accounts

IntegratedQuantum:
#3434 should be fixed now and I rebased the branch, please try again.

iNiKKo:
image
**Image 1**
Two accounts on same PC

image
**Image 2**
Two accounts on two different PC's 
When i tried to join first time i got "disconnected by server" and on second time it crashed the server

IntegratedQuantum:
Please give me a debug trace of the second one. Also did you join with the same account? That's one reason why it would kick you.
Also again, first use case is no longer supported.

iNiKKo:
You said "same account multiplayer is no longer supported as of https://github.com/PixelGuys/Cubyz/pull/3434"
I tested on two different accounts on the same PC which should be possible.

```
thread 72261 panic: reached unreachable code
/home/nick/cubyztest/compiler/zig/lib/std/debug.zig:420:14: 0x15a6449 in assert (Cubyz)
    if (!ok) unreachable; // assertion failure
             ^
/home/nick/cubyztest/src/gui/gui.zig:175:19: 0x1369d49 in deinit (Cubyz)
                window.onCloseFn();
                  ^
/home/nick/cubyztest/src/main.zig:409:33: 0x128b846 in main (Cubyz)
        defer if (!headless) gui.deinit();
                                ^
/home/nick/cubyztest/compiler/zig/lib/std/start.zig:699:88: 0x127f06a in callMain (Cubyz)
                                                                                       ^
???:?:?: 0x7feebce27780 in ??? (/usr/lib/libc.so.6)
???:?:?: 0x7feebce278b8 in ??? (/usr/lib/libc.so.6)
???:?:?: 0x127ecd4 in ??? (???)
./debug_linux.sh: line 31: 72261 Aborted                    (core dumped) ./zig-out/bin/Cubyz
Press enter key to continue. (Or export NO_PAUSE=1 to skip this prompt.)
```


I will just have AI fix this tomorrow and you can have a look and see what was wrong as I can't make a PR out of it.

IntegratedQuantum:
That trace isn't particularly helpful, could you try getting a better trace using `gdb`?

IntegratedQuantum:
Or could you tell me what exactly you did, because I cannot reproduce this at all. (also it's again completely unrelated to this PR)

iNiKKo:
`./zig-out/bin/Cubyz` & .`/run_linux.sh` produce two different things

via the zig-out it works just fine but via ./run it causes the blank world on second client, or at least in my testing.

IntegratedQuantum:
this makes no sense.

iNiKKo:
Just tested it again and out of nowhere it work's just fine, I am actually confused why this is so unreliable.

EDIT: will test extra scenarios and let you know the results.

iNiKKo:
Okay I tested your fixes, and from my test's it looked fine but there is no "5 second jump" it just works from the get go but that might just be due to me playing local on two pc's on the same network, so for online users it may be different

I also added this fix to my custom client and it works also, (no gliding) but did notice that it takes maybe 2-3 seconds before the world loads but it's much better this way.

Tyrrel joined and we tested Teleporting and /spawn and other and they did not notice any gliding or issues.

PR #3481, opened 2026-08-06
Fix incorrect initial RTT estimate for localhost

IntegratedQuantum:
I forgot to set it when the connection has no RTT estimate yet.
This alone was responsible for about 1-2 seconds of debug mode startup time. oops
I found this while investigating the issue described in #3479, and in fact this also fixes it (somewhat).


Wunka (approved):
So thats where the pause at the start came from.

PR #3483, opened 2026-08-06
Add more info to the first Account setup screen and put save as file first

IntegratedQuantum:
Related discussion
https://discord.com/channels/443805812390100992/895731509972058194/1534991669106315495

Putting this in waiting for artistic review so Ikabod can take a look at it too.

Only the following two screens were changed:
First one has more text
Screenshot at 2026-08-06 20-28-10
Second one has the order changed to put save as file first
Screenshot at 2026-08-06 20-45-15


careeoki (approved):
I think this sounds better.
@ikabod-kee Do you have any thoughts here?

IntegratedQuantum:
I'll merge this now because Ikabod doesn't have time for a thorough review before the release.

Issue #2784, opened 2026-03-27
Add option to show local ip


Wunka:
### Description

This would make it easier for people who are on the same network, as currently you have to just enter the local IP when the game is showing your external IP, which is unintuitive and has caused some confusion for new players

This would mean showing the local ip + port (beside the public one or with a toggle option which to show)

ibadullahsyed:
Hope it gets added

H41ogen:
The challenge is making this make sense for non-computer people. Not many people know what an IP address even is, much less the difference between a local and external one.

BoySanic:
> The challenge is making this make sense for non-computer people. Not many people know what an IP address even is, much less the difference between a local and external one.

This at least would make  it more obvious what to do, because currently people who wanna play together on their LAN just enter the public IP and get confused.

Minecraft hasn't had an issue with IP addresses frankly in its many years. It has a similar feature to Cubyz' invite system but theirs only works on LAN. 

I think it'll be as simple as putting some text on the window telling the user to use one if they're inviting someone not on their local network, and use the other if they're inviting someone on their local network.

PR #3491, opened 2026-08-07
add `getLocalIpAdress` function

Wunka:
Progress towards #2784 

This is based on other code using the same function. (for example `getPort`)

### DIFF
--- a/src/network.zig
+++ b/src/network.zig
@@ -1954,6 +1954,33 @@ pub const Connection = struct { // MARK: Connection
 	}
 };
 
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
CI is failing, I'd suggest to add a test to prevent unused error. (Maybe just test that the result is not localhost)
Also did you have someone check if it works on Windows?

Wunka:
> Also did you have someone check if it works on Windows?

Its just a copy of `getPort` but I will go test it

PR #3497, opened 2026-08-07
Ignore non expected confirm of command on client side

Wunka:
Normally this shouldn't happen, but I also see no reason to crash a client because of this


src/sync.zig
```diff
@@ -64,7 +64,11 @@ pub const client = struct { // MARK: client
 	pub fn receiveConfirmation(reader: *BinaryReader) !void {
 		mutex.lock();
 		defer mutex.unlock();
-		try commands.popFront().?.finalize(main.globalAllocator, .client, reader);
+		if (commands.popFront()) |cmd| {
+			try cmd.finalize(main.globalAllocator, .client, reader);
+		} else {
+			std.log.err("Received unexpected confirmation sync", .{});
```
IntegratedQuantum, line 70: Should return an error. Clients that are out of sync with the server should disconnect.

IntegratedQuantum (changes requested):

src/sync.zig
Wunka: returning an error is currently not hooked up to a disconnect on client side as it seems. I tried `return error.Invalid` and it would just ignore it then instead of doing a disconnect

IntegratedQuantum: As far as I can tell it should be hooked up. Could you investigate this further?

Wunka: I don't know what I tested 3 weeks ago, but `return error.Invalid` now works as expected

IntegratedQuantum (approved):

PR #3498, opened 2026-08-07
Add basic coz integration

bluesillybeard:
In case you aren't aware, Coz is a native profiling tool. Coz has a number of benefits over a traditional profiler, you can look for those elsewhere, such as their github page: [link](https://github.com/plasma-umass/coz/tree/master). 

In the end, Coz will generate some graphics that look like this, which relate the speedup of a line of code with relation to how it should speed up progress points (such as the frame time measured by mainLoop or how quickly chunk meshes are generated:
image

This PR was created mainly just "for fun", I wanted to see how easy / difficult it would be to integrate Coz into an existing project written in Zig. In total it took me about 5 hours. Most of that time spent looking into how Coz works under-the-hood, and much of the time was 'wasted' waiting for Coz to collect data only to later realize it wasn't working.

If you want to try this out, make sure to install Coz from their releases page or by building it from source. The version of Coz packaged by your OS may or may not be up to date. Coz only supports Linux and MacOS, and this particular integration only supports Linux.

This is not complete per se, but I figured I would submit a PR regardless, in the event anyone is interested.


bluesillybeard:
aw dang, CI failed. I'll look into this at a later time/date.

IntegratedQuantum:
Oh yeah, I remember seeing the talk about this.
However I'm not sure if it's actually that useful in practice. Did you manage to actually get any interesting results here? To be honest I'm having a hard time interpreting the graphs you have posted, and I do not see any that is actually referencing code from Cubyz.

bluesillybeard:
> Oh yeah, I remember seeing the talk about this.

Nice! That's where I first heard about it too.

> However I'm not sure if it's actually that useful in practice. Did you manage to actually get any interesting results here?

I only ran it for 5 minutes, just to see if it was working in any meaningful way. One downside of Coz is it takes a *really* long time to collect anything useful - all it found in those 5 minutes is that making memcpy faster will make chunks load faster (duh, memcpy is used everywhere).

My goal for today is to iron out the rest of the integration (notably, some extra care is needed around locks/mutexes), tweak/add a few more progress points, then run the game for like 1-2 hours to see if there are any interesting/useful results.

bluesillybeard:
I believe the integration is "done". I've set my gaming PC to run a benchmark (flying at high altitude in a default world sprinting forward indefinitely) with Coz collecting data, I'll leave that for the next hour or two and we'll see what comes out.

bluesillybeard:
I ended up running it for a bit more than two hours while I did other things. I ran it with `-Doptimize=ReleaseSafe`. Attached below is the file it ultimately generated. I wrapped it into a .zip because the file gets quite large, make sure to unzip it before using `coz plot`.

[profile.jsonl.zip](https://github.com/user-attachments/files/30862452/profile.jsonl.zip)

You have to fully install Coz in order to use the viewer (how annoying), so I attached some highlights here as well. The horizontal axis of each graph reports how much faster that line was made (by slowing down all other threads), the vertical axis reports how much faster/slower the selected metric is expected to be. So a point on the graph with a line speedup of 50% and a program speedup of 10% predicts that if that line of code were sped up/optimized by 50%, the metric would improve by 10%.

image

As expected, there is nothing that can be done for swap_and_clear, although it's noteworthy that all of the graphs state swap_and_clear will *take longer* for basically any optimization we do, which probably just means I forgot to disable vsync (oops). In the event vsync is enabled, any optimization will just make it wait longer for the monitor refresh .

image

Apparently the audio callback is called more often if this `vtable.run(task.self)` line (utils.zig:962) is optimized. I don't believe there is anything that can be done about that.

image

It seems world generation is also responsible for audio lag, that isn't a surprise to me as the audio thread and world gen threads fight for who gets CPU time.

image


Speeding up generateInitializedFractalTerrain is supposed to improve the render time ? but has no effect on the overall framerate.

image

This one is really interesting, speeding up this line in the world generation is supposed to speed up rendering but slow down the overall framerate. I suspect this is an anomaly and not really useful, but it's still interesting nonetheless.

image

Surprise surprise, speeding up seemingly any part of world generation / lighting / meshing makes chunk meshes load faster.

I won't post a screenshot for this one, but the game updates/ticks are not sped up by anything since, like with swap_and_clear, it is set to run at a specific rate so there isn't any optimization that exists that will make it faster.

At some point I want to re-run this whole thing with vsync properly disabled, and maybe see if some of its optimization claims are correct - either by intentionally slowing part of the code down to see if Coz reports it, or by actually making something faster. But for now my own personal curiosity is satisfied.

I am still more than happy to adjust/fix anything related to this Coz integration, or answer any questions for anyone curious. So feel free to give feedback / ask questions.

IntegratedQuantum:
> This one is really interesting, speeding up this line in the world generation is supposed to speed up rendering but slow down the overall framerate. I suspect this is an anomaly and not really useful, but it's still interesting nonetheless.

In this line the entire chunk generation happens, if you speed it up, and assuming the test runs of a longer time, there will be more chunks available, causing framerate to drop.

Also in general I'd like to think that rendering is pretty isolated from the rest of the engine, beyond fighting for CPU resources (including hyperthreading) I don't think there really is anything that should slow it down or speed it up from the other threads.
 
 Exploring the data some more, a lot of things don't make sense, e.g. why would improving the BlockUpdate system by 10% have a 50% impact on mesh generation despite the systems being on completely different threads, and BlockUpdate only running on a single thread and not interacting with meshes at all?
Screenshot at 2026-08-09 08-25-01

Maybe a more stable test environment is needed? Though I doubt it's worth the effort.
Overall I'm also a bit skeptical if there is even any data that I also couldn't get with a sampling based profiler. e.g. with some trickery, like turning all mutexes into spinlocks, I've also been able to get a good view on congestion with a sampling based profiler.

bluesillybeard:
BlockUpdateSystem.zig line 27 is just `self.mutex.lock()`. I suspect that is an artifact, and I wonder if it's related to how Coz interacts with locks. Perhaps I've just made a mistake getting Coz integrated? That seems quite likely to me, considering I threw this together in a matter of hours with relatively minimal research.

The biggest thing I have learned so far is that Coz is really sensitive to the test environment / setup. Coz is certainly trickier to get working than I expected. The 'bang for buck' ratio is underwhelming at this point, where traditional profilers (at least in my experience) are much more plug-and-play (and are much easier to understand).

I might tinker with it more on my own time. I want to get a better idea of how Coz actually works. This is just a learning experience for me, not something I intended on necessarily being useful. Thanks for taking some of your time to give this a look.

bluesillybeard:
I've done a decent amount of investigation / testing, and I have determined the following:
- This integration of Coz correctly integrates Futex - it's fair to say Mutex likely works too
- Coz is most definitely useless to Cubyz, at least in its current form

Coz has a number of glaring issues, notably:
- Coz does not consider the caller of a line of code. In the toy example, it determines xorshift (the function 'doing work') is the bottleneck and completely ignores the blatant "actual" bottleneck. The bottleneck is detected correctly if xorshift is split so each caller gets their own copy.
  - This is an instant deal breaker for me personally. I will not be using or experimenting with Coz again until it is resolved.
- Coz takes *way too long* to collect any amount of meaningful data.
- This is admittedly more of a problem with Zig, but in debug mode Coz is unable to find any line number information and becomes absolutely useless as a result.

I am closing this PR now that this experiment is finished.

PR #3500, opened 2026-08-08
For localhost connections set the mtu to the localhost interface mtu

Wunka:
This should also work on mac. (I don't know why `SIOCGIFMTU` is in `std.os.linux` but it should work on mac)

This was made after I saw that I can see the `mtu` of the interface while working on #3491.
I can work on a windows implementation after this.


IntegratedQuantum:
In my opinion we shouldn't treat localhost too different. If localhost uses a different way of discovering MTUs then it is less likely that we find issues in the MTU discovery algorithm (e.g. automated tests, once we have some would require special setup to cover the real MTU discovery algorithm).
Also at the end of the day localhost is also the one place where network bandwidth is not an issue anyways, so it makes no sense to have extra code for it being a bit more efficient than dynamic MTU discovery.

Wunka:
Okay that is understandable. When I started this code, my actual goal was to get a better max mtu value (which would have been the mtu of the interface) but it seems this is not really possible that's why this pr only did this at the end.

PR #3507, opened 2026-08-09
Snalehead rare biome

TyrrelMeter:
Adds a rare biome with a bunch of large stone snale heads. There are warm, temperate, cold, and snowy variants.
Screenshot from 2026-08-09 15-27-12
Screenshot from 2026-08-09 15-11-44
Screenshot from 2026-08-09 15-11-08
Screenshot from 2026-08-09 15-10-43


TyrrelMeter:
I just realized there are no Snela statues. Should I change that?

Argmaster:
Sorry, but I will be brutal here: This seems to me like a low effort gimmick biome. With all the respect, but that head is all there is to it and it might as well just be a rare structure appearing on any biome.

IntegratedQuantum:
I do agree with Argmaster, the contribution here is not the biomes, but the heads.
So instead of copy-pasting multiple biomes (which notably need to be maintained), this would make more sense as a large mega-rare structure (not supported yet) that spawns a bunch of smaller structures in the surrounding area (not supported yet).

TyrrelMeter:
It could also make sense to make the snale head statue a rare structure like mushroom poles or triclops. I can see what you pointed out about the biome and will close the PR.

Issue #2551, opened 2026-02-14
Use the TPM to encrypt the player Account Code when choosing to store it locally


IntegratedQuantum:
This would make it less severe if a player accidentally shares their settings.zig.zon

follow-up on #2520

Argmaster:
What happens when I update my UEFI version? (I think TPM gets wiped in this case) Or the settings get corrupted? (So they can be red but the encrypted string is garbage) What is the intended behavior? Will the game re-prompt me for my pass phase?

IntegratedQuantum:
In that case you just have to login again. I'll make sure to use checksums such that it won't corrupt (same happens also if you switch PC and take over your settings)

Mabeeck:
I am not sure on the intent behind this issue/enhancement.  
- If you want to protect the key such that it can only be retrieved with the password or by the program that called the syscall for protecting the key, then I am fairly sure that's not possible.  
- If you want to protect the key such that it can only be retrieved with the password or by an application running as the user that protected the key then [CryptProtectData](https://learn.microsoft.com/en-us/windows/win32/api/Dpapi/nf-dpapi-cryptprotectdata) on Windows seems appropriate and equivalents exist on linux (you could create a userwide key in the [linux keystore](https://docs.kernel.org/security/keys/core.htm) and extract the payload whenever you please). This would prevent leakage of the key when sharing cubyz-related files. I believe a secure implementation is possible where the CNG saves the key and performs all the crypto operations, refusing to ever release the key to anyone thus preventing copying the key by an unauthorized app running as the same user, but that would require a large amount of effort and the support for cryptographic algorithms is bad (the same applies for linux) so we might need to fall back to the less secure option anyway.
- If the motivation is specifically only to make it less severe if a player accidentally shares their `settings.zig.zon`, then it would be sufficient to simply move the key into a file that is not `settings.zig.zon`, preferably something with a filename that makes it obvious that it contains sensitive information and should not be shared. Alternatively it would also be possible to encrypt the key with a different automatically generated key that is stored in a different file, such that the accidental sharing of any single file cannot reveal the key.

IntegratedQuantum:
> If the motivation is specifically only to make it less severe if a player accidentally shares their settings.zig.zon, then it would be sufficient to simply move the key into a file that is not settings.zig.zon

[...]

PR #3508, opened 2026-08-11
Protect locally stored Acount Code on disk with system api

Mabeeck:
This PR adds the `main.protect.protect` and `main.protect.unprotect` functions and integrates them into the `PasswordEncodedAccountCode` class.  

`protect` Takes an allocator and a slice of bytes as arguments. The function returns a different slice of bytes that has been allocated with the provided allocator and can be passed to `unprotect` to get back the original slice of bytes. The function can fail, if the platform Cubyz is currently running on does not have an implementation yet (in which case the error will be `error.Unsupported`) or the syscall fails for some undisclosed reason (Windows), in which case the error will be `error.syserr`.  
`unprotect` Takes an allocator and a slice of bytes that has been previously generated by `protect` as arguments. The function returns a different slice of bytes that has been allocated with the provided allocator and is equivalent in value to the slice of bytes that was passed to `protect` in order to produce the provided slice. The function can return `error.Invalid` if the provided input was protected on a different device; can no longer be unprotected for some reason; the current platform does not have an implementation. If something unexpected happened the function will fail with `error.syserr`.  
`canProtect` Takes no arguments and returns a boolean indicating weather the protection functions have an implementation on the current platform.

Currently the protection functions only support Windows. They can later be easily expanded to support Linux as well.

Adds a `protected` attribute to `PasswordEncodedAccountCode` that indicates weather a call to `unprotect` is needed, before the `AccountCode` can be decrypted.  
Functions to initialize `PasswordEncodedAccountCode` now take a `shouldProtect` boolean argument that when set to true will protect the function with the native system api, **if available**. Setting it to false will prevent usage of the protection api.

Adds a checkbox to Account Code saving menu for opt-out.
image

Contributes to #2551


src/protect.zig
```diff
@@ -0,0 +1,90 @@
...
+
+pub fn unprotect(allocator: NeverFailingAllocator, data: []u8) error{ syserr, Invalid }![]u8 {
+	if (builtin.os.tag == .windows) {
+		var plainblob: c.DATA_BLOB = undefined;
+		var cipherblob: c.DATA_BLOB = undefined;
+		cipherblob.cbData = @intCast(data.len);
+		cipherblob.pbData = @as([*c]u8, data.ptr);
+		if (c.CryptUnprotectData(&cipherblob, @as([*c][*c]c_ushort, null), @as([*c]c.DATA_BLOB, null), null, @as([*c]c.CRYPTPROTECT_PROMPTSTRUCT, null), @as(c_ulong, 0), &plainblob) == 0) {
+			const err = c.GetLastError();
+			if (err == 13 or err == 87) {
```
Wunka: where did you get these numbers from? Is there no better way to map the error codes? It would be better if you make a function thats maps the error number to an error set. see: https://github.com/PixelGuys/Cubyz/blob/8c7520012f822f59e3b847c8bc8e94b59b1e91b3/src/network.zig#L28-L47

(if possible)
At least you should not have any magic numbers

src/protect.zig
```diff
@@ -0,0 +1,90 @@
...
+		@memcpy(out, plainblob.pbData);
+		return out;
+	} else {
+		return error.Invalid;
+	}
+}
+
+test "slice==unprotect(protect(slice))" {
+	const slice: []u8 = @as([]u8, @constCast("Test"));
+	if (canProtect()) {
```
Wunka: maybe instead of just passing, skip tests on unsopported platforms, as in the future it should work? 
So:

```suggestion
	if (!canProtect()) return error.SkipZigTest;
```

Wunka (changes requested):

src/protect.zig
Mabeeck: I got the numbers from the error messages produced by the test on line 89. I have now found [the proper documentation](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-) and replaced the numbers with fields of `c` for better readability and replaced the if statement with a switch statement.  
I have decided against adding a function, as that function would end up used only once and incredibly small. I consider it better to directly use the switch statement here.

Mabeeck: Do you approve of these changes?

If yes, there's a button for that. If not I'd like to hear further improvement ideas from you.

src/protect.zig
IntegratedQuantum, line 1: Should not be at the root. `main.protect` is quite ambiguous.
I'd suggest to put it in `main.network.autentication.protect`, to emphasize that it belongs thematically to the authentication system.

src/protect.zig
```diff
@@ -0,0 +1,98 @@
+const std = @import("std");
+const builtin = @import("builtin");
+
+const main = @import("main");
+const NeverFailingAllocator = main.heap.NeverFailingAllocator;
+const EncodingType = main.network.authentication.EncodingType;
+const c = @import("c");
+
+pub inline fn canProtect() bool {
+	switch (builtin.os.tag) {
```
IntegratedQuantum: For system-dependent code I prefer the style that used to be present in e.g. zig's mutex implementation, where you have one shallow wrapper over a platform dependent `Impl` struct. This keeps code for each platform together, making it easier to argue about.
You can find an example for this in file_monitor.zig.

src/protect.zig
```diff
@@ -0,0 +1,98 @@
...
+		else => return false,
+	}
+}
+
+pub fn protect(allocator: NeverFailingAllocator, data: []u8) error{ syserr, Unsupported }![]u8 {
+	if (builtin.os.tag == .windows) {
+		var plainblob: c.DATA_BLOB = undefined;
+		var cipherblob: c.DATA_BLOB = undefined;
+		plainblob.cbData = @intCast(data.len);
+		plainblob.pbData = @as([*c]u8, data.ptr); // Does this need to be secureZeroed?
```
IntegratedQuantum: No, only if you make a copy of the data, here you only store the pointer.

src/protect.zig
```diff
@@ -0,0 +1,98 @@
...
+pub inline fn canProtect() bool {
+	switch (builtin.os.tag) {
+		.windows => return true,
+		else => return false,
+	}
+}
+
+pub fn protect(allocator: NeverFailingAllocator, data: []u8) error{ syserr, Unsupported }![]u8 {
+	if (builtin.os.tag == .windows) {
+		var plainblob: c.DATA_BLOB = undefined;
```
IntegratedQuantum: Should be directly initialized instead of initializing all fields
```zig
		var plainblob: c.DATA_BLOB = .{
		    .cbData = ...,
		    .pbData = ...,
		};
```

src/protect.zig
```diff
@@ -0,0 +1,98 @@
...
+		var cipherblob: c.DATA_BLOB = undefined;
+		plainblob.cbData = @intCast(data.len);
+		plainblob.pbData = @as([*c]u8, data.ptr); // Does this need to be secureZeroed?
+		if (c.CryptProtectData(&plainblob, @as([*c]const c_ushort, null), @as([*c]c.DATA_BLOB, null), null, @as([*c]c.CRYPTPROTECT_PROMPTSTRUCT, null), @as(c_ulong, 0), &cipherblob) == 0) {
```
IntegratedQuantum: You shouldn't need `@as` here

src/protect.zig
```diff
@@ -0,0 +1,98 @@
...
+			std.log.err("CryptProtectData syscall failed. Errorcode: {}. This should never happen. Please report it to the maintainers.", .{c.GetLastError()});
+			return error.syserr;
```
IntegratedQuantum: Please follow the existing naming convention, CapitalCamelCase, for errors.

src/protect.zig
```diff
@@ -0,0 +1,98 @@
...
+		}
+		defer if (c.LocalFree(cipherblob.pbData) != null) std.log.err("LocalFree syscall failed to free previously allocated memory. Errorcode: {}. This should never happen. Please report it to the maintainers.", .{c.GetLastError()});
+		const out: []u8 = allocator.alloc(u8, @intCast(cipherblob.cbData));
+		@memcpy(out, cipherblob.pbData);
```
IntegratedQuantum: Use allocator.dupe with cipherblob.pbData[0..cipherblob.cbData]

src/network/authentication.zig
```diff
@@ -338,8 +361,17 @@ pub const PasswordEncodedAccountCode = struct {
 	}
 
 	pub fn decryptFromPassword(self: PasswordEncodedAccountCode, password: []const u8, failureText: *main.ListManaged(u8)) !AccountCode {
+		if (self.protected and !protect.canProtect()) return error.Invalid;
 		if (self.typ == .none) {
-			return AccountCode.initFromUserInput(self.data, failureText);
+			var data = self.data;
+			if (self.protected) {
+				data = try protect.unprotect(main.stackAllocator, data);
```
IntegratedQuantum: I'd prefer to just duplicate the one call in here, to avoid the complex logic on defer:
```suggestion
				const unprotectedData = try protect.unprotect(main.stackAllocator, data);
				defer ...
				return AccountCode.init(...);
```
Or even better, instead of repeating this in every encryption path, how about just calling the function again?
```zig
if (self.protected) {
   ... // unprotect
  var copy = self;
  copy.protected = false;
  copy.data = unprotectedData;
  return decryptFromPassword(copy, password, failureText);
}
```

src/gui/windows/authentication/encrypt_with_password.zig
```diff
@@ -40,10 +41,10 @@ pub fn setAccountCode(accountCode_: main.network.authentication.AccountCode) voi
 fn confirm() void {
 	if (encryptAccountCode) {
 		settings.storedAccount.deinit(main.globalAllocator);
-		settings.storedAccount = .initFromPassword(main.globalAllocator, accountCode, passwordTextField.currentString.items);
+		settings.storedAccount = PasswordEncodedAccountCode.initFromPassword(main.globalAllocator, accountCode, passwordTextField.currentString.items, true) catch PasswordEncodedAccountCode.initFromPassword(main.globalAllocator, accountCode, passwordTextField.currentString.items, false) catch unreachable;
```
IntegratedQuantum: Any fallback logic should be handled internally. Also the user should be given a checkbox whether they want to do this (defaulting to true), the checkbox should be disabled if the platform doesn't support protection, then protection failure can just print an error.

src/protect.zig
```diff
@@ -0,0 +1,98 @@
...
+		try std.testing.expectError(error.Unsupported, protect(main.stackAllocator, slice));
+		try std.testing.expectError(error.Invalid, unprotect(main.stackAllocator, slice));
+	} else {
+		return error.SkipZigTest;
+	}
+}
+
+test "Unprotect fails when supplied with garbage" {
+	if (canProtect()) {
+		const slices: [5][]u8 = .{@as([]u8, @constCast("TestdwadadÖOUWHdöouHIOSUdhöoUHNWLJDKNOÖPAHUIwdoöJKNSdlkjöwuHOÖIhso8zpo9IKj")), @as([]u8, @constCast("Test")), @as([]u8, @constCast("Testd")), @as([]u8, @constCast("")), @as([]u8, @constCast("WIJDp8iU)(du098UÜ=JHd0ü8hz=Ü(HJ0isidjowi8h=(Z\"ß08IJUISdhd0w98hdoi8uoIWUJDoikjsoIKHJOwiuhdOISHNdo9i8H(UIHNASUJhdnbiuJBWGiudjhbIAKUJHnbsiudjkhiWUAHNIUDshjliuAHELIUHFILUHNIUJBDIUHwiuHushoujhdiiuwhIUHsouhdUHwiuhdUAHLsuidhlHU)"))};
```
IntegratedQuantum: constCast should only be used when required by a C interface, and in that case it should be used when calling the C interface to avoid poisoning the rest of the codebase with this.

IntegratedQuantum (changes requested):

src/protect.zig
Mabeeck: Crap. I forgot to remove that comment.

Mabeeck:
Done

src/protect.zig
IntegratedQuantum: I think it should still be put into a separate file, since it will likely grow as we add support for the other operating systems.

src/network/authentication.zig
```diff
@@ -284,18 +287,130 @@ pub const AccountCode = struct {
...
-const EncodingType = enum { none, argon2_aes_gcm };
+pub const protection = struct {
+	const Impl = switch (builtin.os.tag) {
+		.windows => WindowsImpl,
+		else => NoImpl,
+	};
+
+	pub inline fn canProtect() bool {
+		return Impl.canProtect;
+	}
```
IntegratedQuantum:
```suggestion
	pub const canProtect: bool = Impl.canProtect;
```

src/network/authentication.zig
```diff
@@ -284,18 +287,130 @@ pub const AccountCode = struct {
...
+
+	pub fn protect(allocator: NeverFailingAllocator, data: []const u8) error{ SystemError, Unsupported }![]u8 {
+		return Impl.protect(allocator, data);
+	}
+
+	pub fn unprotect(allocator: NeverFailingAllocator, data: []const u8) error{ SystemError, Invalid }![]u8 {
+		return Impl.unprotect(allocator, data);
+	}
+
+	const NoImpl = struct {
```
IntegratedQuantum: technically these are namespaces (no fields), so they should follow our naming scheme for them: no_impl windows_impl impl

src/network/authentication.zig
```diff
@@ -284,18 +287,130 @@ pub const AccountCode = struct {
...
+				const unprotected = try unprotect(main.stackAllocator, protected);
+				defer main.stackAllocator.free(unprotected);
+				try std.testing.expectEqualSlices(u8, slice, unprotected);
+			}
+		} else {
+			return error.SkipZigTest;
+		}
+	}
+
+	test "Protect fails on unsupported platforms" {
```
IntegratedQuantum: Note that these currently do not run in the CI. I made an issue for this: #3535

IntegratedQuantum (changes requested):

src/protect.zig
Wunka: sorry that I didn't look at this again, Yeah that is better. I will approve this and let quantum do the rest of reviewing.

Wunka (approved):

src/network/authentication.zig
Mabeeck: The linter complains about that. Does this warrant a change to the linter?

src/network/authentication.zig
Mabeeck: I have taken notice.

src/network/authentication.zig
Mabeeck: I do feel the need to note that it took me a bit of time to understand what you meant here.  
I always took "field" to mean "a named entry in a data structure" and they do have those and so I look for a definition and Wikipedia has one and by Wikipedia´s definition the canProtect constant is a field and then I noticed that taking the definition of "a namespace is a struct that has no fields" doesn't get me anywhere as I have started to wonder if the functions in there could be fields too and so I just search up the Wikipedia definition of a namespace and now I am even more confused as that claims all classes and objects to be namespaces and then I got the brilliant idea that I should maybe think less about the meaning of your words and more about what you actually mean by namespace considering not only the literally named definition, but the context in which it was given and after I did that I concluded that you (probably) consider a namespace to be a struct that does not have any variable fields, but after reviewing this comment before pressing `Comment` I realized that by namespace you probably mean a namespace in the syntax sense which both makes more and less sense, because now I need to ask myself why not every other struct like those in file_monitor.zig uses snake_case too, because those are accessed using dot syntax too, making them namespaces, but then I realized that that was probably because those can be instantiated and become objects, while a namespace is more of a way to group together values and functions serving the same purpose and then I thought that that's just what objects are, but objects are still a bit different and then I realized that this whole debate with I, me and myself is getting nowhere or at least not fast enough and this whole thing was fun at first with the philosophical thinking and writing this long and horrendously unreadable comment all in one sentence was funny and a good way to organize my thoughts, but now I am tired of it and I already did all the changes anyway, so debating about it is completely pointless.

Speaking of organizing ones thoughts, this is exactly how...  
Oh no, I'm an LLM.

Mabeeck: I posted this pointless long comment knowing perfectly well that I am wasting the time of everybody involved including myself all for a bad joke, but I still can't stop myself from reading that whole thing and retroactively correcting mistakes. This took me an hour and this isn't even the right place to crack a joke and somehow I do it anyways.

src/network/authentication.zig
IntegratedQuantum: Should be fixed by #3544, please rebase to get the changes

Mabeeck:
Done

src/network/authentication.zig
IntegratedQuantum: A field in Zig has a pretty well-defined meaning. I'd suggest to next time look at the Zig documentation first, when you are unsure what a term means.

src/gui/windows/authentication/encrypt_with_password.zig
```diff
@@ -76,6 +91,10 @@ pub fn onOpen() void {
 	const width = 480;
 	list.add(Label.init(.{0, 0}, width, "Your Account Code will be stored in your settings to allow you to stay logged in. Please decide how we should store it:", .left));
 	innerList = VerticalList.init(.{0, 0}, 100, 16);
+	if (main.network.authentication.protection.canProtect) {
+		protectCheckbox = CheckBox.init(.{0, 0}, width, "Protect with system api (recommended)", protectAccountCode, &protectAccountCodeCallback);
```
IntegratedQuantum: I would suggest to add more details on what this means, also please try to write it for someone who does not know what a "system api" is.

src/network/authentication.zig
```diff
@@ -284,18 +286,19 @@ pub const AccountCode = struct {
 	}
 };
 
-const EncodingType = enum { none, argon2_aes_gcm };
+pub const EncodingType = enum { none, argon2_aes_gcm };
```
IntegratedQuantum: Why is this public now?

src/network/authentication.zig
```diff
@@ -311,21 +314,42 @@ pub const PasswordEncodedAccountCode = struct {
...
 		};
 	}
 
-	pub fn initUnencoded(allocator: NeverFailingAllocator, accountCode: AccountCode) PasswordEncodedAccountCode {
+	pub fn initUnencoded(allocator: NeverFailingAllocator, accountCode: AccountCode, shouldProtect: bool) error{SystemError}!PasswordEncodedAccountCode {
+		const protected = shouldProtect and protection.canProtect;
+		var data: []u8 = undefined;
+		if (protected) {
+			data = protection.protect(allocator, accountCode.text) catch |err| {
+				if (err == error.SystemError) return error.SystemError else unreachable;
```
IntegratedQuantum: If you want to discard errors with `unreachable`, please use a `switch` on the error to prevent mistakes when the error set changes.

src/network/authentication.zig
```diff
@@ -311,21 +314,42 @@ pub const PasswordEncodedAccountCode = struct {
 		main.io.random(&nonce);
 		std.crypto.aead.aes_gcm.Aes256Gcm.encrypt(encryptedBuffer, &authenticationTag, accountCode.text, &.{}, nonce, key);
 
+		const protected = shouldProtect and protection.canProtect;
+		var data: []u8 = undefined;
+		if (protected) {
+			data = protection.protect(allocator, encryptedBuffer) catch |err| {
+				if (err == error.SystemError) return error.SystemError else unreachable;
+			};
+			defer allocator.free(encryptedBuffer); // Deferred, because that way even if a SystemError is thrown it will still get freed
```
IntegratedQuantum: This defer statement runs immediately, since you have a closing `}` afterwards.

src/network/protection.zig
```diff
@@ -0,0 +1,114 @@
...
+	else => no_impl,
+};
+
+pub const canProtect: bool = impl.canProtect;
+
+pub fn protect(allocator: NeverFailingAllocator, data: []const u8) error{ SystemError, Unsupported }![]u8 {
+	return impl.protect(allocator, data);
+}
+
+pub fn unprotect(allocator: NeverFailingAllocator, data: []const u8) error{ SystemError, Invalid }![]u8 {
```
IntegratedQuantum, line 20: After seeing you `unreachable` some of the errors, maybe it would make sense to rethink their purpose.
Errors are kind of useless if they are not being used at the callsite.

IntegratedQuantum (changes requested):

src/network/authentication.zig
Argmaster: Uh that was harsh
To clean up he nomenclature: 
- *fields* are on *instances*, alternatively called *attributes* or *members* defined as `foo: typ`. 
- `const x` and `var x` are part of a *scope* (local - in function, or static - in definition of struct), not of an instance. 
- Every struct is a namespace in a technical sense, but when using the term *namespace* in writing or speech we mean a special case of struct that does **not** have *fields* and thus makes no sense to instantiate. 
- mind that in zig a file is also technically a struct, so it can have fields and can be instantiated.

src/network/authentication.zig
Mabeeck: That is because the very first commit in this pull request added more EncodingTypes instead of adding a protected attribute. That implementation offered a function to spit out the correct protected EncodingType for the platform and that function was in protect.zig and not in authentication.zig, thus the need for it to be public. When I later undid those changes in favor of the current approach I forgot to remove the pub keyword.

src/network/protection.zig
Mabeeck: I have removed error.Unsupported, now this PR does not include any mention of `unreachable` anymore.

Mabeeck:
Done

src/network/authentication.zig
```diff
@@ -311,21 +314,38 @@ pub const PasswordEncodedAccountCode = struct {
 		main.io.random(&nonce);
 		std.crypto.aead.aes_gcm.Aes256Gcm.encrypt(encryptedBuffer, &authenticationTag, accountCode.text, &.{}, nonce, key);
 
+		const protected = shouldProtect and protection.canProtect;
+		var data: []u8 = undefined;
+		if (protected) {
+			defer allocator.free(encryptedBuffer);
```
IntegratedQuantum: I think this is error-prone design. `defer` should always be right under the thing that creates the resource.

Furthermore all local allocations should be done with the stackAllocator.

I propose to add a helper function for the encryption, then you can do it like this:
```zig
const data, const authenticationTag, const nonce = blk: {
	if (protected) {
		const encryptedData, const authenticationTag, const nonce = encryptWithPassword(main.stackAllocator, ...);
		defer main.stackAllocator.free(encryptedData);
		break :blk .{try protection.protect(allocator, encryptedData), authenticationTag, nonce};
	} else {
		break :blk encryptWithPassword(allocator, ...);
};
```

src/network/protection.zig
```diff
@@ -0,0 +1,113 @@
...
+	const canProtect = true;
+
+	fn protect(allocator: NeverFailingAllocator, data: []const u8) error{SystemError}![]u8 {
+		var plainblob: c.DATA_BLOB = .{
+			.cbData = @intCast(data.len),
+			.pbData = @constCast(data.ptr),
+		};
+		var cipherblob: c.DATA_BLOB = undefined;
+		if (c.CryptProtectData(&plainblob, null, null, null, null, 0, &cipherblob) == 0) {
+			std.log.err("CryptProtectData syscall failed. Errorcode: {}. This should never happen. Please report it to the maintainers.", .{c.GetLastError()});
```
IntegratedQuantum: Reporting it to the maintainers is true for all errors, no need to reiterate it here.
```suggestion
			std.log.err("CryptProtectData syscall failed. Errorcode: {}.", .{c.GetLastError()});
```

src/gui/windows/authentication/encrypt_with_password.zig
```diff
@@ -76,6 +91,10 @@ pub fn onOpen() void {
 	const width = 480;
 	list.add(Label.init(.{0, 0}, width, "Your Account Code will be stored in your settings to allow you to stay logged in. Please decide how we should store it:", .left));
 	innerList = VerticalList.init(.{0, 0}, 100, 16);
+	if (main.network.authentication.protection.canProtect) {
+		protectCheckbox = CheckBox.init(.{0, 0}, width, "Protect from theft (recommended)\nForces re-authentication when device changes", protectAccountCode, &protectAccountCodeCallback);
```
IntegratedQuantum: I think this is too generic. Protection from theft can mean a lot of things.

I'd suggest to find a non-dev on the discord server to work together on this. Or maybe @ikabod-kee could help here.

IntegratedQuantum (changes requested):
Please also include a screenshot (since I cannot test this locally), also please adjust the window height to avoid a scrollbar.

src/gui/windows/authentication/encrypt_with_password.zig
Mabeeck: image

Does this work? Or maybe "Prevent loading on different devices"? I have also asked on discord. Will change once I have further improvement ideas.

src/network/authentication.zig
Mabeeck: I think the better solution is to just make the encryptedBuffer always be copied into a different buffer, that way the defer can be directly below the encryptedBuffer, as it will never be needed to persist beyond the scope of the function that way.

Mabeeck:
I think that's all of it.

IntegratedQuantum:
Please rebase after #3538 to run the tests in the CI

Mabeeck:
Done

src/c.h
```diff
@@ -55,6 +55,13 @@
 #include <errno.h>
 #endif
 
+// Used for platform-speciffic keystorage
+#ifdef _WIN32
+#pragma comment(lib, "crypt32.lib")
```
IntegratedQuantum: Does translate-c even support pragmas? Does this even make a difference?

src/network/protection.zig
```diff
@@ -0,0 +1,113 @@
...
+					return error.SystemError;
+				},
+			}
+		}
+		var pbDataSlice: []u8 = undefined;
+		pbDataSlice.len = plainblob.cbData;
+		pbDataSlice.ptr = plainblob.pbData;
+		defer {
+			std.crypto.secureZero(u8, pbDataSlice);
+			if (c.LocalFree(plainblob.pbData) != null) std.log.err("LocalFree syscall failed to free previously allocated memory. Errorcode: {}. This should never happen.", .{c.GetLastError()});
```
IntegratedQuantum: defers should always be directly after the thing you are trying to defer, i.e. the defer for LocalFree should be directly after the if (c.CryptUnprotect...) {...} block.

src/network/protection.zig
```diff
@@ -0,0 +1,113 @@
...
+	const canProtect = true;
+
+	fn protect(allocator: NeverFailingAllocator, data: []const u8) error{SystemError}![]u8 {
+		var plainblob: c.DATA_BLOB = .{
+			.cbData = @intCast(data.len),
+			.pbData = @constCast(data.ptr),
+		};
+		var cipherblob: c.DATA_BLOB = undefined;
+		if (c.CryptProtectData(&plainblob, null, null, null, null, 0, &cipherblob) == 0) {
+			std.log.err("CryptProtectData syscall failed. Errorcode: {}. This should never happen.", .{c.GetLastError()});
```
IntegratedQuantum: Unless the API contract states that it can't happen. We must assume that it does.
Presenting the user with a random number and "This should never happen" is not good user experience.
Please remove the last part, and use `std.os.windows.GetLastError()` and print the error as a string.

Anything where you can be sure that it cannot happen should be a panic instead.

src/gui/windows/authentication/encrypt_with_password.zig
```diff
@@ -75,7 +90,11 @@ pub fn onOpen() void {
 	const list = VerticalList.init(.{padding, 16 + padding}, 320, 8);
 	const width = 480;
 	list.add(Label.init(.{0, 0}, width, "Your Account Code will be stored in your settings to allow you to stay logged in. Please decide how we should store it:", .left));
-	innerList = VerticalList.init(.{0, 0}, 100, 16);
+	innerList = VerticalList.init(.{0, 0}, 120, 16);
+	if (main.network.authentication.protection.canProtect) {
+		protectCheckbox = CheckBox.init(.{0, 0}, width, "Force auto-logout when device changes (recommended)", protectAccountCode, &protectAccountCodeCallback);
```
IntegratedQuantum, line 95: I think we should also mention what this does. In it current form it only mentions downsides, so why would anyone check it?

src/network/authentication.zig
```diff
@@ -6,6 +6,8 @@ const BinaryReader = main.utils.BinaryReader;
 const NeverFailingAllocator = main.heap.NeverFailingAllocator;
 const ZonElement = main.ZonElement;
 
+pub const protection = @import("protection.zig");
```
IntegratedQuantum: The folder structure should match the imports, so please move this into a `authentication` subfolder.

src/network/authentication.zig
```diff
@@ -338,6 +358,18 @@ pub const PasswordEncodedAccountCode = struct { // MARK: PasswordEncodedAccountC
 	}
 
 	pub fn decryptFromPassword(self: PasswordEncodedAccountCode, password: []const u8, failureText: *main.ListManaged(u8)) !AccountCode {
```
IntegratedQuantum, line 360: I forgot to add tests for these. Would you mind adding them, including one for protection if available?
Some basic roundtrips (encrypting→decrypting) should suffice.

IntegratedQuantum (changes requested):

src/c.h
Mabeeck: Your local copy of the PR is (slightly) outdated. I already removed that line.

src/network/protection.zig
Mabeeck: Well, the thing is that I shouldn't free it without secure-zeroing it first and for that I need to make a slice out of it.  
I can't think of a way for the defer to come any earlier.

src/network/authentication.zig
Mabeeck: I did add them, but they end up blocking eternally on ubuntu in the CI.

PR #3510, opened 2026-08-12
Allocate and upload data to the GPU

IntegratedQuantum:
Upload happens using a staging buffer on a separate command buffer which gets submitted (and synchronized) before the other command buffers.

I also had to add a garbage collection thing for triple buffering, and I also hardcoded the number of frames to 2, any more than that and we have too much latency anyways.

- [x] Update cubyz-libs after https://github.com/PixelGuys/Cubyz-libs/pull/20

progress towards #102


src/graphics/vulkan.zig
```diff
@@ -772,6 +792,109 @@ pub const command_pool = struct { // MARK: command_pool
...
+		const vkFunctions: c.VmaVulkanFunctions = .{
+			.glad_vkGetInstanceProcAddr = c.glad_vkGetInstanceProcAddr,
+			.glad_vkGetDeviceProcAddr = c.glad_vkGetDeviceProcAddr,
+			.glad_vkCreateImage = c.glad_vkCreateImage,
+		};
+		const allocatorCreateInfo: c.VmaAllocatorCreateInfo = .{
+			.flags = 0,
+			.physicalDevice = physicalDevice,
+			.device = device,
+			.preferredLargeHeapBlockSize = 128 << 9, // TODO: What does this do?
```
Argmaster: Why this blizzard way of saying 65536?

Argmaster: Based on my extended research, I think this controls the size of actual allocation (size of object rounded up) when allocation crosses the threshold that classifies it as large allocation. 
Or more precisely, the overshoot of that allocation when allocating above 1GB, eg. If I alloc 1.1GB, with defaults it gets rounded to 1.25GB (block of 0.25GB)

IntegratedQuantum: But why allocate more than needed?

Argmaster: For same reason GPA does it. I have no idea.
I think this eases the consequences of deallocaitng and allocating again with just a bit different size (or reallocating or whatever).
If you perfect fit it, you can only use it for identical or smaller object.
And allocation calls are bad too, so I would assume they preallocate blocks instead of just calling driver for every object.

PR #3532, opened 2026-08-18
Refactor natural block dropping

DellieDelta:
Progress towards #2385

Related to #3533

Is about `decay` and the `check_support_block` callbacks

### DIFF
--- a/src/callbacks/block/server/decay.zig
+++ b/src/callbacks/block/server/decay.zig
@@ -137,7 +137,7 @@ pub fn run(self: *@This(), params: main.callbacks.ServerBlockCallback.Params) ma
 			if (world.cmpxchgBlock(wx, wy, wz, leaf, self.decayReplacement) == null) {
 				const modelIndex = params.block.mode().model(params.block);
 				for (self.blockDrops) |drop| {
-					drop.tryDropNaturally(modelIndex, .{wx, wy, wz});
+					drop.dropNaturally(modelIndex, .{wx, wy, wz});
 				}
 				return .handled;
 			}
--- a/src/callbacks/block/server/decay.zig
+++ b/src/callbacks/block/server/decay.zig
@@ -135,10 +135,17 @@ pub fn run(self: *@This(), params: main.callbacks.ServerBlockCallback.Params) ma
 
 			// no, there is no log in proximity
 			if (world.cmpxchgBlock(wx, wy, wz, leaf, self.decayReplacement) == null) {
-				const modelIndex = params.block.mode().model(params.block);
-				for (self.blockDrops) |drop| {
-					drop.dropNaturally(modelIndex, .{wx, wy, wz});
-				}
+				const model = params.block.mode().model(params.block).model();
+				const location = BlockDrop.Location{
+					.normalDir = main.vec.normalize(main.random.nextFloatVectorSigned(3, &main.seed)),
+					.min = model.min,
+					.max = model.max,
+				};
+				const dropCtx = BlockDrop.Context{
+					.oldBlock = params.block,
+					.newBlock = self.decayReplacement,
+				};
+				dropCtx.drop(location, .{wx, wy, wz});
 				return .handled;
 			}
 		}
--- a/src/callbacks/block/server/decay.zig
+++ b/src/callbacks/block/server/decay.zig
@@ -136,7 +136,7 @@ pub fn run(self: *@This(), params: main.callbacks.ServerBlockCallback.Params) ma
 			// no, there is no log in proximity
 			if (world.cmpxchgBlock(wx, wy, wz, leaf, self.decayReplacement) == null) {
 				for (self.blockDrops) |drop| {
-					drop.dropRandomly(params.block, wx, wy, wz);
+					drop.tryDropFromBlock(params.block, .{wx, wy, wz});
 				}
 				return .handled;
 			}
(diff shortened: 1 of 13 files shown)


src/server/BlockDrop.zig
```diff
@@ -28,13 +29,34 @@ pub fn isDroppedWhenBrokenWithItem(self: @This(), item: Item) bool {
...
+		const model = modelIndex.model();
+		for (self.itemStacks) |itemStack| {
+			var randomDir = main.vec.normalize(main.random.nextFloatVectorSigned(3, &main.seed));
+			// Bias upwards
+			randomDir[2] += main.random.nextFloat(&main.seed)*4.0;
+			const randomPos = Vec3f{
+				@as(f32, @floatFromInt(pos[0])) + model.min[0] + main.random.nextFloat(&main.seed)*(model.max[0] - model.min[0]),
+				@as(f32, @floatFromInt(pos[1])) + model.min[1] + main.random.nextFloat(&main.seed)*(model.max[1] - model.min[1]),
+				@as(f32, @floatFromInt(pos[2])) + model.min[2] + main.random.nextFloat(&main.seed)*(model.max[2] - model.min[2]),
+			};
```
IntegratedQuantum, line 50: Could we maybe include this as another case into the `Location`?
I think it would be nicer to only have a single drop function, with everything else being configured through its parameters.

IntegratedQuantum (changes requested):

src/server/BlockDrop.zig
DellieDelta: Ohh I hadn't made that connection, I'll try it out

DellieDelta:
Before I continue, I need your feedback on a direction I want to take this in. See the last commit

Now that we have allowed/forbidden tool tags on drops, we can get rid of the `dropNaturally` function too on the `BlockDrop` by letting decay also use `Context`

Instead of passing a list of block drops directly to the decay callback, we can instead rely completely on the drops defined on the block itself. That would mean you'd always treat decay as if you're breaking the block by hand. Currently, any block that can decay already works like that

It makes sense to me, but I'm wondering if you still see any value in allowing the caller to define custom drops in the callback

src/callbacks/block/server/decay.zig
```diff
@@ -135,22 +135,17 @@ pub fn run(self: *@This(), params: main.callbacks.ServerBlockCallback.Params) ma
...
-								@as(f32, @floatFromInt(wy)) + model.min[1] + main.random.nextFloat(&main.seed)*(model.max[1] - model.min[1]),
-								@as(f32, @floatFromInt(wz)) + model.min[2] + main.random.nextFloat(&main.seed)*(model.max[2] - model.min[2]),
-							};
-							main.server.world.?.drop(stack.clone(), pos, dir, 1);
-						}
-					}
-				}
+				const model = params.block.mode().model(params.block).model();
+				const location = BlockDrop.Location{
+					.normalDir = main.vec.normalize(main.random.nextFloatVectorSigned(3, &main.seed)),
```
IntegratedQuantum: Ideally all item stacks should have a randomized direction, not one direction for all of them.

IntegratedQuantum (commented):
> It makes sense to me, but I'm wondering if you still see any value in allowing the caller to define custom drops in the callback

As far as I remember this was mainly a decision because we didn't have any other tools to restrict block drops.

In the future, if we ever ed this capability again, I think it would make more sense to add a mechanism that makes use of the existing tags instead.

> Now that we have allowed/forbidden tool tags on drops, we can get rid of the dropNaturally function too on the BlockDrop by letting decay also use Context

Yes, please, but do note that it should still give each stack a new random poisition.

src/server/BlockDrop.zig
```diff
@@ -27,14 +27,30 @@ pub fn isDroppedWhenBrokenWithItem(self: @This(), item: Item) bool {
 	return true;
 }
 
-pub fn drop(self: @This(), pos: Vec3d, dir: Vec3f, velocity: f32) void {
+pub fn drop(self: @This(), pos: Vec3d, dir: Vec3f) void {
 	if (self.chance == 1 or main.random.nextFloat(&main.seed) < self.chance) {
 		for (self.itemStacks) |itemStack| {
```
IntegratedQuantum, line 32: Not just each drop, but each itemStack should also get a random position, I think the best solution would be to pass a location here as well.

IntegratedQuantum (changes requested):

DellieDelta:
I also want to move `pos` into the `Location` struct. But I think that's better for a follow-up PR because it'll make the change bigger

src/server/BlockDrop.zig
```diff
@@ -27,14 +27,26 @@ pub fn isDroppedWhenBrokenWithItem(self: @This(), item: Item) bool {
 	return true;
 }
 
-pub fn drop(self: @This(), pos: Vec3d, dir: Vec3f, velocity: f32) void {
+pub fn drop(self: @This(), loc: Location, pos: Vec3i, spread: Spread) void {
```
IntegratedQuantum: Have you considered putting the spread into the Location?

IntegratedQuantum (changes requested):

src/server/BlockDrop.zig
DellieDelta: That works! I can't combine the parameters though

IntegratedQuantum: Why not?

DellieDelta: Because which spread it is only gets decided in the Context drop function, after the location has already been initialized. But maybe I'm missing something. Maybe you could somehow add the spread to the location afterwards, but not sure how to do that in a neat way

IntegratedQuantum: Ok, then let's keep it like this for now. Though I do think that in the future we may need to revise some of this again (when we have more drop sources, like entities).
Either way it's a step in the right direction.

IntegratedQuantum (approved):

PR #3533, opened 2026-08-18
Refactor block dropping when breaking block with a tool

DellieDelta:
Related to #3532 

Progress towards #2385

### DIFF
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -563,6 +563,15 @@ pub const Block = packed struct(u32) { // MARK: Block
 	pub inline fn isSelectableByItem(self: Block, item: Item) bool {
 		return self.selectionCapabilities().allowsSelectionByItem(self, item);
 	}
+
+	pub fn dropFromPosition(self: Block, newBlock: Block, item: items.Item, pos: Vec3d, dir: Vec3f, velocity: f32) void {
+		const dropAmount = self.mode().itemDropsOnChange(self, newBlock);
+		for (0..dropAmount) |_| {
+			for (self.blockDrops()) |drop| {
+				drop.dropFromPosition(item, pos, dir, velocity);
+			}
+		}
+	}
 };
 
 pub const meshes = struct { // MARK: meshes
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -567,7 +567,7 @@ pub const Block = packed struct(u32) { // MARK: Block
 		return self.selectionCapabilities().allowsSelectionByItem(self, item);
 	}
 
-	pub fn tryDropWithItem(self: Block, newBlock: Block, item: items.Item, ctx: BlockDrop.BlockDropLocation.DropContext) void {
+	pub fn tryDropWithItem(self: Block, newBlock: Block, item: items.Item, ctx: BlockDrop.Location.DropContext) void {
 		const dropAmount = self.mode().itemDropsOnChange(self, newBlock);
 		for (0..dropAmount) |_| {
 			for (self.blockDrops()) |drop| {
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -29,6 +29,9 @@ const sbb = main.server.terrain.sbb;
 const blueprint = main.blueprint;
 const Assets = main.assets.Assets;
 const BlockDrop = main.server.BlockDrop;
+const vec = main.vec;
+const Vec3d = vec.Vec3d;
+const Vec3f = vec.Vec3f;
 
 const c = @import("c");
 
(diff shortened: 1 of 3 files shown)


IntegratedQuantum (changes requested):
I don't think this is the right place for this functionality. blocks.zig is already a big file and so far it has mostly definitions of block properties and loading/access thereof, but little functionality (and all functions on `Block` are retrieving properties, none are actually doing things).

In the issue I proposed to move them to the server world, which is also maybe a bit too big right now.

So maybe the BlockDrop struct deserves its own file (and in server, since blockdrops are not done on the client). And all the helper structs and functions (including #3532) could be defined there.

DellieDelta:
I agree. I'll make a new PR for moving `BlockDrop` to its own file then because it applies to both changes

src/blocks.zig
```diff
@@ -563,6 +563,15 @@ pub const Block = packed struct(u32) { // MARK: Block
 	pub inline fn isSelectableByItem(self: Block, item: Item) bool {
 		return self.selectionCapabilities().allowsSelectionByItem(self, item);
 	}
+
+	pub fn tryDropWhenBrokenWithItem(self: Block, newBlock: Block, item: Item, ctx: BlockDrop.Location.Context) void {
```
IntegratedQuantum: Again, I don't think we should have any actions in the Block interface. A block doesn't do things on it's own, it's a collection of properties.

So I suggest the following:
Make a BlockDrop.Context that contains all the relevant information and has the actual drop function which takes a position and a BlockDrop.Location (note that the BlockDrop.Location.Context isn't needed, you can store the relevant data in the BlockDrop.Context).

IntegratedQuantum (changes requested):

src/blocks.zig
DellieDelta: Oh right, missed it from your last comment. That's a good approach

src/server/BlockDrop.zig
```diff
@@ -20,3 +26,95 @@ pub fn isDroppedWhenBrokenWithItem(self: @This(), item: Item) bool {
...
+pub const Context = struct {
+	oldBlock: Block,
+	newBlock: Block,
+	item: ?Item,
+
+	pub fn drop(self: Context, location: Location, pos: Vec3i) void {
+		const dropAmount = self.oldBlock.mode().itemDropsOnChange(self.oldBlock, self.newBlock);
+		if (dropAmount == 0) return;
+
+		if (self.item) |item| {
```
IntegratedQuantum: This now only drops anything if it was broken by an item (i.e. no apples)

src/server/BlockDrop.zig
```diff
@@ -20,3 +26,95 @@ pub fn isDroppedWhenBrokenWithItem(self: @This(), item: Item) bool {
 
 	return true;
 }
+
+pub fn tryDropWhenBrokenWithItem(self: @This(), item: Item, pos: Vec3d, dir: Vec3f, velocity: f32) void {
```
IntegratedQuantum: I think this name is rather awkward, also in my option the item should be checked at the callsite instead.

IntegratedQuantum (changes requested):

src/server/BlockDrop.zig
```diff
@@ -20,3 +26,93 @@ pub fn isDroppedWhenBrokenWithItem(self: @This(), item: Item) bool {
...
+		const dropAmount = self.oldBlock.mode().itemDropsOnChange(self.oldBlock, self.newBlock);
+		if (dropAmount == 0) return;
+
+		const dropPos = if (self.newBlock.collide()) location.outsidePos(pos) else location.insidePos(pos);
+		const dropDir = location.dropDir();
+		const dropVelocity = location.dropVelocity();
+
+		for (0..dropAmount) |_| {
+			for (self.oldBlock.blockDrops()) |blockDrop| {
+				if (self.item == null or blockDrop.isDroppedWhenBrokenWithItem(self.item.?)) {
```
IntegratedQuantum: The function already handles the null case internally.
```suggestion
				if (blockDrop.isDroppedWhenBrokenWithItem(self.item)) {
```

IntegratedQuantum (changes requested):

src/server/BlockDrop.zig
DellieDelta: I also had to change the optional `?Item` field to an `Item` field with a default value to match the type expected in the function, while still keeping the field optional for callers

IntegratedQuantum (approved):

PR #3538, opened 2026-08-21
Run tests in CI on both Windows and Linux

### LINKED ISSUE #3535: Run the tests in the CI both on Windows and on Linux
#3508 adds tests that currently only work when ran on a Windows operating system, we should test them (and future OS-dependent tests) in the CI.

(only the tests should be moved, all the formatting and compilation can stay)

yel0h:
Split the CI workflow's `zig build test` step out of the `compile` job into its own `test` job that runs on a matrix, so unit tests execute on both operating systems.

Closes #3535

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
   test:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     name: Unit Tests (${{ matrix.os }})
     defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,5 +38,21 @@ jobs:
       - run: zig build -Dtarget=x86_64-windows-gnu
       - run: zig build -Dtarget=aarch64-linux-gnu
       - run: zig build -Dtarget=aarch64-windows-gnu
-      - run: zig build test
       - run: 'find assets/cubyz/shaders -type f ! -name "*.glsl" | xargs -L1 glslangValidator -G100 -P"#extension GL_GOOGLE_include_directive : enable" -Iassets/cubyz/shaders/include -Dgl_VertexIndex=gl_VertexID -DOPEN_GL'
+
+  test:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    name: Unit Tests (${{ matrix.os }})
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v3
+      - run: echo "VERSION=$(cat .zigversion)" >> $GITHUB_ENV
+      - uses: mlugg/setup-zig@v2
+        with:
+          version: ${{ env.VERSION }}
+      - run: zig build test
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
Would you mind also checking if we can get it running on macos as well?

yel0h:
We sure can, however currently Mac fails to compile due to #3541

IntegratedQuantum:
Alright, then lets do it without macos first, it's good to know that it's at least possible in principle

IntegratedQuantum (changes requested):
Also, after seeing that each of them takes several minutes again (whereas in the main CI job it took only a few seconds to compile and run the tests), I think we should do the platform specific compilation in the same CI unit as the tests, to avoid duplicated dependency downloads.

The main CI unit would then only do the format checking and shader compilation

yel0h:
Done

.github/workflows/ci.yml
IntegratedQuantum, line 18: This does not need the zig setup anymore

IntegratedQuantum (changes requested):

.github/workflows/ci.yml
```diff
@@ -34,9 +34,26 @@ jobs:
...
+        with:
+          version: ${{ env.VERSION }}
       - run: zig build
-      - run: zig build -Dtarget=x86_64-windows-gnu
-      - run: zig build -Dtarget=aarch64-linux-gnu
-      - run: zig build -Dtarget=aarch64-windows-gnu
+      - if: matrix.os == 'ubuntu-latest'
+        run: zig build -Dtarget=aarch64-linux-gnu
+      - if: matrix.os == 'windows-latest'
+        run: zig build -Dtarget=aarch64-windows-gnu
```
Wunka: can you not just use:

```suggestion
        run: zig build -Dtarget=aarch64-native-gnu
```

Wunka (changes requested):

.github/workflows/ci.yml
IntegratedQuantum: If that's possible we should also change the default one to `-Dtarget=x86_64-native`

.github/workflows/ci.yml
yel0h: it still runs `zig ast-check` though

yel0h:
Ok, replaced the ifs with a native target

Wunka (approved):

.github/workflows/ci.yml
```diff
@@ -34,9 +34,23 @@ jobs:
...
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v3
+      - run: echo "VERSION=$(cat .zigversion)" >> $GITHUB_ENV
+      - uses: mlugg/setup-zig@v2
+        with:
+          version: ${{ env.VERSION }}
+      - run: zig build -Dtarget=x86_64-native
+      - run: zig build -Dtarget=aarch64-native-gnu
```
IntegratedQuantum, line 55: gnu shouldn't be needed here. It also would be incorrect when we want to extend this for macos.

IntegratedQuantum (changes requested):

yel0h:
Unfortunately it seems gnu needs to say, leaving it out [tries to use the native x86_64 headers](https://github.com/PixelGuys/Cubyz/actions/runs/32664255742/job/97255095119)...

IntegratedQuantum:
Ah I see, I guess we will need to go back to ifs once we implement macos then. Either way it's probably fine for now.

Though it seems that the CI is failing now, not sure why. I also tried restarting it with the same results. Please investigate.

IntegratedQuantum (approved):
I restarted the CI, it looks like it was only a sporadic issue.

PR #3540, opened 2026-08-21
Fix file selection UI problems

### LINKED ISSUE #3489: File selection UI problems
- The button in the file picker says "Save", but the actual code for Cubyz only saves the file after an additional check.
I can't see an option to change the button text, so maybe instead we need to save it right away.
- On windows the path shown on screen contains backslashes, which are escape characters and do not show up. Once again we are discovering how poorly designed Windows is.

yel0h:
The account code is now written once a path is chosen, and the second button, now labelled "Return to login", stays disabled until the write succeeds. The label now shows a normalized copy of the path with / separators.

file_selection

Fixes #3489

### DIFF
--- a/src/gui/windows/authentication/create_account_account_code.zig
+++ b/src/gui/windows/authentication/create_account_account_code.zig
@@ -29,6 +29,7 @@ var accountCodeLabel: *Label = undefined;
 var accountCode: ?main.network.authentication.AccountCode = null;
 var fileNameEntry: *Label = undefined;
 var fileName: []const u8 = undefined;
+var fileSaved: bool = undefined;
 
 pub const StorageMethod = enum(usize) {
 	file = 0,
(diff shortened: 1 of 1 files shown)


IntegratedQuantum:
Instead of only displaying the path, please add some text in front of it, to make it clear that it was already saved there.

yel0h:
Added

image

src/gui/windows/authentication/create_account_account_code.zig
```diff
@@ -29,6 +29,7 @@ var accountCodeLabel: *Label = undefined;
 var accountCode: ?main.network.authentication.AccountCode = null;
 var fileNameEntry: *Label = undefined;
 var fileName: []const u8 = undefined;
+var fileSaved: bool = undefined;
```
IntegratedQuantum: These can be local variables.
The one case where fileSaved is used can be moved into the same function too.

IntegratedQuantum (changes requested):

yel0h:
Fixed

IntegratedQuantum (approved):

PR #3550, opened 2026-08-22
Fix player spawn sometimes chosen midair above cavern

### LINKED ISSUE #2097: It is possible for the player spawn position to be mid-air

 And it seems it will also be above a giant cavern, causing a death loop.

https://discord.com/channels/443805812390100992/1425321371348766720
https://discord.com/channels/443805812390100992/1430297314685227170

yel0h:
`isValidSpawnLocation` now also queries the `CaveMap` at the candidate column to confirm there's solid ground at the surface height and clear air above it to stand in, rejecting the candidate and letting the spiral search keep looking otherwise. Also made the 2D height calculation run when no valid location was found.

Example test seed with this bug: 9071844392573517448

Fixes #2097

### DIFF
--- a/src/server/world.zig
+++ b/src/server/world.zig
@@ -884,9 +884,9 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 					}
 				}
 				std.log.err("Found no valid spawn location", .{});
+				const map = terrain.SurfaceMap.getOrGenerateFragment(self.spawn[0], self.spawn[1], 1);
+				self.spawn[2] = map.getHeight(self.spawn[0], self.spawn[1]) + 1;
 			}
-			const map = terrain.SurfaceMap.getOrGenerateFragment(self.spawn[0], self.spawn[1], 1);
-			self.spawn[2] = map.getHeight(self.spawn[0], self.spawn[1]) + 1;
 		}
 		const newBiomeCheckSum: i64 = @bitCast(terrain.biomes.getBiomeCheckSum(self.settings.seed));
 		if (newBiomeCheckSum != self.biomeChecksum) {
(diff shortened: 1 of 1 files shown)


IntegratedQuantum (changes requested):
Ideally you should check the actual chunk(s) it's in, not just the CaveMap. Otherwise there may still be problem with structure generation (e.g. you might spawn in a tree, or in a lava lake or whatever)

yel0h:
Okay, now also checks whether the actual generated blocks are safe

src/server/world.zig
```diff
@@ -1045,9 +1045,36 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		return ch.getBlock(relPos.x, relPos.y, relPos.z);
+	}
+
+	fn isValidSpawnLocation(self: *ServerWorld, wx: i32, wy: i32) bool {
 		const map = terrain.SurfaceMap.getOrGenerateFragment(wx, wy, 1);
-		return map.getBiome(wx, wy).isValidPlayerSpawn;
+		if (!map.getBiome(wx, wy).isValidPlayerSpawn) return false;
+		const height = map.getHeight(wx, wy);
+
+		const caveMap = terrain.CaveMap.CaveMapView.init(main.stackAllocator, .{.wx = wx, .wy = wy, .wz = height, .voxelSize = 1}, 0, 1);
```
IntegratedQuantum, line 1063: There is no need to check both the cavemap and the actual blocks.

src/server/world.zig
```diff
@@ -1045,9 +1045,36 @@ pub const ServerWorld = struct { // MARK: ServerWorld
...
+		defer caveMap.deinit(main.stackAllocator);
+		if (!caveMap.isSolid(0, 0, -1)) return false;
+		if (caveMap.isSolid(0, 0, 0) or caveMap.isSolid(0, 0, 1)) return false;
+
+		const groundBlock = getGeneratedBlock(wx, wy, height - 1);
+		if (!groundBlock.collide() or !groundBlock.onTouch().isNoop()) return false;
+		const feetBlock = getGeneratedBlock(wx, wy, height);
+		const headBlock = getGeneratedBlock(wx, wy, height + 1);
+		inline for (.{feetBlock, headBlock}) |block| {
```
IntegratedQuantum, line 1072: If we use inline for anyways, why not go one step further?
```suggestion
		inline for (.{height, height + 1}) |h| {
				const block = getGeneratedBlock(wx, wy, h);
```

src/server/world.zig
```diff
@@ -1045,9 +1045,36 @@ pub const ServerWorld = struct { // MARK: ServerWorld
 		try files.cubyzDir().write(itemsPath, itemDropData.data.items);
 	}
 
-	fn isValidSpawnLocation(_: *ServerWorld, wx: i32, wy: i32) bool {
+	fn getGeneratedBlock(wx: i32, wy: i32, wz: i32) Block {
```
IntegratedQuantum, line 1048: I think we should follow the existing getOrGenerate scheme.
```suggestion
	fn getOrGenerateBlock(wx: i32, wy: i32, wz: i32) Block {
```

IntegratedQuantum (changes requested):

yel0h:
Ok, done

IntegratedQuantum (approved):

PR #3552, opened 2026-08-22
#3228 grey out unusable items when using workbench

### LINKED ISSUE #3228: Grey out unusable items in workbench UI
### Description

Grey out items that cannot be used in crafting when using a workbench. This should help tell the player what can and cannot be used as a material.

LukePearcy:
closes #3228

### DIFF
--- a/src/gui/components/ItemSlot.zig
+++ b/src/gui/components/ItemSlot.zig
@@ -21,8 +21,12 @@ pub const sizeWithBorder = 32 + 2*border;
 const Mode = enum {
 	normal,
 	takeOnly,
-	unfocused,
 	immutable,
+};
+
+const Style = enum {
+	normal,
+	unfocused,
 };
 
 pos: Vec2f,
@@ -37,6 +41,7 @@ pressed: bool = false,
 renderFrame: bool = true,
 texture: ?Texture,
 mode: Mode,
+style: Style,
 
 var defaultTexture: Texture = undefined;
 var immutableTexture: Texture = undefined;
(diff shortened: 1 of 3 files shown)


src/gui/components/ItemSlot.zig
```diff
@@ -21,6 +21,7 @@ pub const sizeWithBorder = 32 + 2*border;
 const Mode = enum {
 	normal,
 	takeOnly,
+	unfocused,
 	immutable,
 };
```
Wunka, line 25: `normal`, `takeOnly`, and `immutable` actually represent logic behavior changes, whereas `unfocused` only affects rendering behavior. So I don't think `unfocused` fits here.

One example where this can shoot you in the foot is that the item slots in the creative window are `takeOnly`, and therefore aren't affected by your logic. Even though it would be nice if they were.

Wunka (changes requested):
Welcome to Cubyz!

if you don't know: I am am not someone who can merge your code, so you will still receive a review by quantum.

src/gui/components/ItemSlot.zig
LukePearcy: That's a good point, it now applies the logic by inventory source instead of item slot mode. Though maybe the creative inventory should have it's own inventory source type instead of using .other?

IntegratedQuantum:
This still requires more design discussions, see my comment on the issue: https://github.com/PixelGuys/Cubyz/issues/3228#issuecomment-5387799203

PR #3555, opened 2026-08-23
Show loading overlay on inventory slots until their contents arrive

### LINKED ISSUE #2700: No way to know whether a chest is empty or just hasn't finished loading
### Description

Many times on Ashframe, I've thought a chest was empty, only for it to turn out my internet was just slow. Some kind of indicator that the contents of a chest aren't yet loaded would be extremely nice.

yel0h:
The server now also sends the number of items to expect, and the client counts them down as they arrive. Slots for an inventory that isn't loaded yet render with a dark overlay. All `.serverShared` inventories are tracked.

image

Fixes #2700

### DIFF
--- a/src/Inventory.zig
+++ b/src/Inventory.zig
@@ -26,15 +26,18 @@ pub const client = struct { // MARK: client
 	var maxId: InventoryId = @enumFromInt(0);
 	var freeIdList: main.List(InventoryId) = .empty;
 	var serverToClientMap: std.AutoHashMap(InventoryId, Inventory) = undefined;
+	var loadingInventories: std.AutoHashMap(InventoryId, u32) = undefined;
 
 	pub fn init() void {
 		serverToClientMap = .init(main.globalAllocator.allocator);
+		loadingInventories = .init(main.globalAllocator.allocator);
 	}
 
 	pub fn deinit() void {
 		std.debug.assert(freeIdList.items.len == @intFromEnum(maxId)); // leak
 		freeIdList.clearAndFree(main.globalAllocator);
 		serverToClientMap.deinit();
+		loadingInventories.deinit();
 	}
 
 	fn nextId() InventoryId {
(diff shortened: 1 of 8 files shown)


src/Inventory.zig
```diff
@@ -75,6 +79,34 @@ pub const client = struct { // MARK: client
...
+	fn startLoadTracking(clientId: InventoryId) void {
+		main.sync.client.mutex.lock();
+		defer main.sync.client.mutex.unlock();
+		loadingInventories.put(clientId, 1) catch unreachable;
+	}
+
+	pub fn setExpectedItemCount(clientId: InventoryId, count: u32) void {
+		main.sync.client.mutex.assertLocked();
+		if (count == 0) {
+			_ = loadingInventories.remove(clientId);
```
Wunka: this should never happen right? 
(the remove not the `== 0`, so I would either do an assert that nothing was removed or just do an early return

Wunka (changes requested):

src/Inventory.zig
yel0h: Well if the client receives that the inventory is empty, I immediately remove it from the inventories being loaded since there's nothing to wait for

IntegratedQuantum:
I think the best solution is to make chest opening server initiated.
Then opening the GUI and receiving its contents happens at once, and there is no gap in the timeline.

yel0h:
Ok, I added a `chestOpen` protocol; now the client sends a request and opens the GUI once a response is received

IntegratedQuantum:
I kind of would like to have a more generic interface. It goes beyond just the chest, and it goes beyond just inventories.
I don't know yet how the specifics would look like, I actually tried working on it a bit in #2564, but I wasn't too happy with my attempt there either.

All I know is that we need a generic way for the server to tell the client to open a specific window, including the data relevant for it. And on the site of the client, we need a way to share block interactions with the server through the sync system.

All in all this probably requires more discussion and design work. The only reason this issue landed in the 0.4.0 milestone is because I was confident I could get it done in #2564, but after introduction of the ECS I have paused my efforts on it to see how the ECS would evolve first.

PR #3557, opened 2026-08-24
Status effects

Crepestrom:
Adds status effects #624
Status effects have 3 parts to their Zon
Type - (what type of status effect)
DecayStyle - (how stacks are removed when Timeleft reaches 0 eg; (-1 every removal or /2 every removal)) [not implimented yet]
OnUpdate - (A callback that applies some effect) [currently only for hurting player]

When Applied they also have these 2 parts:
Stacks - (how many stacks of effects are applied)
Timeleft - (how much time until a stack is decayed) [not implimented yet]

Something Applying a status effect will have to provide these things:
Type - (what type of status effect)
Stacks - (how many at once)
Stacklimit - (the max limit that this source will give)
Merge type - (the way that this will add onto more of the same type of status effect) [not implimented yet]
RefreshTime - (What time the status is set to when the status goes down by a stack)

Merge types come in different varieties:
Overwrite - (resets TimeLeft and sets the stack count directly)
Add - (resets TimeLeft and adds onto the stacks)
AddNoRefresh - (does NOT resets TimeLeft and adds onto the stacks)


IntegratedQuantum:
Please explain why you decided to make a new system instead of just using the ECS at 1 component per effect, as mentioned when you asked during the play session.

Crepestrom:
It would give a more unified system for handleing effects
also it would mean status effects could be more easily handled by addons

IntegratedQuantum:
But can you achieve the same variety of effects?
To give some simple examples: How would you implement something like Noita's stains that decay when moving instead of being time based? Or effects that add temporary shields, like minecraft's absorption?

Furthermore how would you integrate this into the mod system, what if a mod adds a new player property, can your effect system add effects that change this new property?

Crepestrom:
well the mod can add a new callback that interacts with that property
for the movement based one instead of using deltatime as deltatime we use how much a player has moved in a frame as the deltatime

Crepestrom:
though that would require implementing a seperate callback that is called with that parameter

Crepestrom:
when a player takes damage it is modified by all of the players status effects through a callback
a status would likely need a data struct for a absorption effect to track hp which it couldnt currently (kinda like tool modifiers)

when a player takes damage
goes through status effects to see what changes
Absorption reduces its hp and any overflow is passed onto the next status

though this would need a few callbacks for damage processing
one for effects that reduce damage
and one for bonus health

Crepestrom:
i think i need to implement a more generalized callback function in the status effect component
cause there is going to be a lot of the exact same callback structure but called in different places and times

IntegratedQuantum:
> i think i need to implement a more generalized callback function in the ecs that goes through the status effects

What makes status effects special? Shouldn't all components be accounted for when it comes to damage? e.g. maybe you have an item in your inventory that breaks and prevents death.

This is touching many topics that we haven't decided on yet.

And even if we want to allow addons to add their own status effects, I'm not sure if this is the right approach, if we want this for effects, then maybe we also want this for other types of components too.

I think it would be best to first finish migrating the existing stuff to ECS before opening new adjacent topics. We can't keep creating new systems before finalizing the old ones. It makes design decisions and discussion so much more difficult since we haven't got enough data to figure out what the right approach is.

Crepestrom:
> > i think i need to implement a more generalized callback function in the ecs that goes through the status effects
> 
> What makes status effects special? Shouldn't all components be accounted for when it comes to damage? e.g. maybe you have an item in your inventory that breaks and prevents death.
> 
> This is touching many topics that we haven't decided on yet.
> 
> And even if we want to allow addons to add their own status effects, I'm not sure if this is the right approach, if we want this for effects, then maybe we also want this for other types of components too.
> 
> I think it would be best to first finish migrating the existing stuff to ECS before opening new adjacent topics. We can't keep creating new systems before finalizing the old ones. It makes design decisions and discussion so much more difficult since we haven't got enough data to figure out what the right approach is.

i mean that for optimization cause i didnt wanna write 50 functions that all did the same thing with different flavors
that trigger different callbacks
in the status effect component
my bad for the confusion

PR #3562, opened 2026-08-30
Health Ecs Component

Crepestrom:
Healthbar Ecs component
a component storing
max health
health

Also adds a method for the client to send arbritary binary data to the server and vice versa to be interpreted by components

Next to implement:
proper death
death messages

### DIFF
--- a/src/entityComponent/health.zig
+++ b/src/entityComponent/health.zig
@@ -27,16 +27,18 @@ const ItemStack = items.ItemStack;
 const random = main.random;
 
 const c = @import("c");
+const Self = @This();
 
 pub var entityComponentID: main.entity.EntityComponentId = undefined;
 pub const entityComponentVersion = 0;
 
-const playerBagSizeLimit = 120;
+var playerBagSizeLimit = 120;
 
 // ############################# Client only stuff ################################
 pub const client = struct {
 	const Component = struct {
-		bag: items.Inventory.BagInventory,
+		health: f32,
+		maxHealth: f32,
 	};
 	pub var components: main.utils.SparseSet(Component, Entity) = .{};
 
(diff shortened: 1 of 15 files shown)


IntegratedQuantum:
please don't bypass the sync system, component updates should be synced properly.

Crepestrom:
should i also remove transmit change?
its a legacy function that overwrites the clients copy of the component with the servers when (currently only used by model)
the modifyComponent function can do all the same functionality

Crepestrom:
Removing it will also remove the entity component update protocall and move it over to sync

src/entityComponent/health.zig
```diff
@@ -0,0 +1,156 @@
...
+	pub fn unload(entity: Entity) void {
+		components.remove(entity) catch {};
+	}
+
+	pub fn modifyComponent(entity: Entity, reader: *utils.BinaryReader) void {
+		const addedHealth = reader.readFloat(f32) catch return;
+		addHealth(entity, addedHealth);
+	}
+
+	pub fn addHealth(entity: Entity, healthChange: f32) void {
```
Argmaster, line 151: So what stops the user code from just using get() method and accessing the member directly, hence bypassing the synchronization? This API is unfortunately very very easy to use wrong and requires quite a lot of boilerplate.
I do not have a good answer to that issue unfortunately (I'm still thinking), but please be aware of that flaw.

Argmaster: Oh never mind, you can't modify the component because get() doesn't return a reference, well, so only issue is the amount of boilerplate.

src/entityComponent/health.zig
```diff
@@ -0,0 +1,156 @@
...
+		components = .{};
+	}
+	pub fn deinit() void {
+		components.deinit(main.globalAllocator);
+	}
+
+	pub fn get(entity: Entity) ?Component {
+		return (components.get(entity) orelse return null).*;
+	}
+	pub fn getHealth(entity: Entity) ?f32 {
```
Argmaster, line 120: What's of the point of this method if you already have get()?

src/entityComponent/health.zig
```diff
@@ -0,0 +1,156 @@
...
+		std.log.debug("modifed component {}", .{health});
+
+		if (health.* <= 0) {
+			die(entity);
+		}
+
+		main.entity.server.transmitChange(Self, entity);
+	}
+
+	fn die(entity: Entity) void {
```
Argmaster, line 162: Considering content of this function, resetHealth() seems more appropriate?

src/entityComponent/health.zig
```diff
@@ -0,0 +1,156 @@
...
+		const component = components.get(entity) orelse return;
+		const health = &component.health;
+		const maxHealth = &component.maxHealth;
+		health.* = maxHealth.*;
```
Argmaster: What stops you from just doing

```suggestion
		component.health = component.maxHealth;
```

Crepestrom:
Components are now synced

src/entityComponent/health.zig
Crepestrom: i plan to add the kill thing to it so the player is teleported

src/entityComponent/health.zig
Crepestrom: gives a number instead of a component

Crepestrom:
thanks for reviewing argmaster

src/sync.zig
```diff
@@ -854,6 +809,60 @@ pub const Command = struct { // MARK: Command
...
+					}
+				} else {
+					std.log.debug("hello right here", .{});
+					var binaryReader = BinaryReader.init(info.modification);
+					const entityId: main.entity.Entity = @enumFromInt(binaryReader.readVarInt(u32) catch return);
+					const componentId = info.componentId;
+					const actionType: main.entity.ComponentActionType = binaryReader.readEnum(main.entity.ComponentActionType) catch return;
+
+					inline for (@typeInfo(main.entity.components).@"struct".decls) |decl| {
+						if (@field(main.entity.components, decl.name).server.get(entityId)) |component| {
```
IntegratedQuantum, line 864: you cannot access the server side component data when on the client

Crepestrom: oops i made that save the wrong side that is fixed

Crepestrom:
im splitting this up into multiple prs

Issue #3485, opened 2026-08-06
Visual feedback for inviting IPs in the "invite players" screen


careeoki:
### Description

When you press the "invite" button, there is no feedback if it actually did anything.
Maybe a temporary popup saying "Invite Sent" or just clearing the text box so you don't invite the same IP again.

<img width="646" height="132" alt="Image" src="https://github.com/user-attachments/assets/c6a343e8-7b8d-4a0f-9d4b-9d711a5bc7d6" />

Wunka:
We could also maybe add a comment about that you can cancel the invite in players window

Argmaster:
I would suggest to just list invited IPs below the textbox, but that that would look better if the window would resize to account for that only when there is anything to display.

PR #3566, opened 2026-09-01
Clear IP address textbox after clicking invite

oridbot:
babies first zig line
attempt at fixing #3485


Wunka:
that's some very complicated code.... I don't know if even quantum can understand that... I think you will have to split that into 100 parts \s

src/gui/windows/invite.zig
```diff
@@ -50,6 +50,7 @@ fn invite() void {
 		if (err != error.AlreadyConnected) {
 			std.log.err("Cannot connect user: {s}", .{@errorName(err)});
 		}
+		ipAddressEntry.clear();
```
Wunka: this is not the correct place. THis would only give feedback if the user is already connected or another error appeard. The issue is describing that there should always be visual feedback

Wunka (changes requested):

src/gui/windows/invite.zig
oridbot: ok, i put it in the front of the function, i think this is the right place but honestly I mainly have no clue what that "_thread" thing is

oridbot: ok i was informed (by MrPiedPiper) this is not the way to go as we do actually need to use ipadressentry and i have to store it

src/gui/windows/invite.zig
```diff
@@ -42,6 +42,8 @@ fn discoverIpAddressFromNewThread() void {
 }
 
 fn invite() void {
+	ipAddress = ipAddressEntry.currentString.items;
```
IntegratedQuantum: This makes no sense

src/gui/windows/invite.zig
IntegratedQuantum: `ipAddressEntry.currentString.items` is now empty because you cleared it.

Please actually test the invite functionality and check in the player list if it actually invited the right IP address.

IntegratedQuantum (changes requested):

src/gui/windows/invite.zig
oridbot: sorry, meant to put ipAddress there, i currently just test it out on the client and then copy the code to the github website so i should probably figure out how to push a change from vscode. im gonna look at the code some more to see if ipAddress is ok to use to store the entry here.

src/gui/windows/invite.zig
```diff
@@ -42,11 +42,13 @@ fn discoverIpAddressFromNewThread() void {
 }
 
 fn invite() void {
+	const storeIpAddressEntry = ipAddressEntry.currentString;
+	ipAddressEntry.clear();
```
Wunka, line 46: please read into what you are actually doing here.
As quantum said, it doens't really make sense and makes very wierd behavior possible.

Wunka (changes requested):
> ok im decently confident its good now

what does confident mean? I don't want to sound harsh, but have you actually now tested the code?

src/gui/windows/invite.zig
IntegratedQuantum: This is still the same issue, you are clearing the memory, yet you are still using a reference to it (`const storeIpAddressEntry = ipAddressEntry.currentString;` is making a copy of the pointer, not the underlying memory).
This is working in your test because in reality clear doesn't change the memory, but this is an implementation detail and you are just lucky that it worked.

Since you are just starting out, I'd recommend to maybe first do something like [ziglings](https://codeberg.org/ziglings/exercises/)  so you understand all the basics before contributing to Cubyz and stumbling over how Zig handles memory differently from other languages.

oridbot: this is old code, I properly allocate the memory now, i changed it to a draft since I'm not sure where to free it yet. but ill figure it out, maybe I should do some ziglings

PR #3569, opened 2026-09-04
Fix bug that you can't place fog against fog

DellieDelta:
I think I also accidentally introduced this bug, lol. But only ran into it today

Fog is set to be never selectable, but that means it immediately returns before getting to the special cases. So the special cases need to come before the never selectable check

### DIFF
--- a/src/blocks.zig
+++ b/src/blocks.zig
@@ -57,7 +57,7 @@ const SelectionCapabilities = union(enum) {
 		toolEffective: bool = false,
 
 		pub fn allowsSelectionByItem(self: @This(), block: Block, item: Item) bool {
-			// Hardcoded cases
+			// Hardcoded cases should come first
 			if (item == .baseItem) {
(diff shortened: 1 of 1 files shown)


src/blocks.zig
```diff
@@ -57,14 +57,7 @@ const SelectionCapabilities = union(enum) {
...
 		pub fn allowsSelectionByItem(self: @This(), block: Block, item: Item) bool {
-			if (self == @This(){}) return false;
-
-			if (self.toolEffective) {
-				if (item == .proceduralItem and item.proceduralItem.isEffectiveOn(block)) {
-					return true;
-				}
-			}
-
+			// Hardcoded cases should come first
```
IntegratedQuantum, line 60: *hardcoded cases should not exist :P
but that's a different issue

IntegratedQuantum (approved):

PR #3576, opened 2026-09-06
Stones - Palette textures 5

careeoki:
enough procrastinating

image


IntegratedQuantum (approved):
What else is left after this?

careeoki:
Just the non-wood branches I think.

PR #3579, opened 2026-09-06
fix tabs for mods

### LINKED ISSUE #3578: Generated mod files do not follow our formatter
Just found this while running the formatter locally:
```
~/Cubyz$ cubyz-fmt
mods/rotations.zig
src/game.zig
```
probable related to #3565 @Wunka could you fix this and investigate why the CI didn't catch this?

Wunka:
Fixes: #3578
Don't ask why it looks like this. This is the best I could do with the formatter

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -158,8 +158,8 @@ pub fn makeModFeature(io: std.Io, step: *std.Build.Step, name: []const u8) !void
 		\\}
 	;
 	try featureList.appendSlice(step.owner.allocator, try std.mem.replaceOwned(u8, step.owner.allocator, testTextSpaces, "    ", "\t"));
-	
-        const file_path = step.owner.fmt("mods/{s}.zig", .{name});
+
+	const file_path = step.owner.fmt("mods/{s}.zig", .{name});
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -153,8 +153,12 @@ pub fn makeModFeature(io: std.Io, step: *std.Build.Step, name: []const u8) !void
...
 		\\const main = @import("main");
 		\\test "abc" {
-		\\  @setEvalBranchQuota(1000000);
-		\\  main.refAllDeclsRecursiveExceptCImports(@This());
+		\\
+	++ "\t" ++
+		\\@setEvalBranchQuota(1000000);
+		\\
+	++ "\t" ++
+		\\main.refAllDeclsRecursiveExceptCImports(@This());
```
IntegratedQuantum: This is awful, I propose to use 4 spaces and then pass it through std.mem.replace to replace 4 spaces with tabs again.

Also mandatory mention of https://github.com/ziglang/zig-spec/issues/38

IntegratedQuantum (changes requested):

build.zig
Wunka: done

IntegratedQuantum (approved):

PR #3582, opened 2026-09-07
Vulkan text support

Wunka:


src/graphics.zig
```diff
@@ -1220,13 +1263,31 @@ const TextRendering = struct { // MARK: TextRendering
 		fontSize: c_int,
 	} = undefined;
 
+	const CompleteUniforms = extern struct {
+		textureRect: [4]f32 align(16),
+		offset: [2]f32 align(8),
+		fontEffects: c_int,
+		textureBounds: [4]f32 align(16),
```
```suggestion
		glyphData: GlyphUniforms,
```

src/graphics/vulkan.zig
```diff
@@ -1012,6 +1012,69 @@ pub const Image = struct { // MARK: Image
 			},
 		}});
 	}
+
+	pub fn uploadGlyphs(self: Image, offset: i32, width: i32, height: i32, pitch: i32, data: []const u8) void {
```
IntegratedQuantum: Why is this here? vulkan.zig is meant for low-level interfaces, glyphs definitely do not qualify

IntegratedQuantum (changes requested):

src/graphics/vulkan.zig
Wunka: .... yeah a lot of my things were not even needed.... I added now a config struct to UploadData, where we can add things if we want to configure more.

src/graphics/vulkan.zig
```diff
@@ -963,7 +963,12 @@ pub const Image = struct { // MARK: Image
 		gpu_garbage_collection.deferredFree(.{.image = self});
 	}
 
-	pub fn uploadData(self: Image, offset: usize, data: []const u8) void {
+	const UploadDataConfig = struct {
+		bufferOffset: usize = 0,
```
IntegratedQuantum: pointless, please remove

src/graphics.zig
```diff
@@ -1265,51 +1317,88 @@ const TextRendering = struct { // MARK: TextRendering
...
-		c.glBindTexture(c.GL_TEXTURE_2D, glyphTexture[0]);
-		c.glTexImage2D(c.GL_TEXTURE_2D, 0, c.GL_R8, newWidth, textureHeight, 0, c.GL_RED, c.GL_UNSIGNED_BYTE, null);
-		c.glCopyImageSubData(glyphTexture[1], c.GL_TEXTURE_2D, 0, 0, 0, 0, glyphTexture[0], c.GL_TEXTURE_2D, 0, 0, 0, 0, textureOffset, textureHeight, 1);
-		pipeline.bind(draw.getScissor());
-		c.glUniform2f(uniforms.fontSize, @floatFromInt(textureWidth), @floatFromInt(textureHeight));
+		if (main.settings.launchConfig.vulkanTestingMode) {
+			glyphTexture[0].vulkanImage = vulkan.Image.init(.{textureWidth, textureHeight, 1}, .{
+				.format = c.VK_FORMAT_R8_UNORM,
+				.usage = c.VK_IMAGE_USAGE_TRANSFER_DST_BIT | c.VK_IMAGE_USAGE_SAMPLED_BIT,
+			});
```
IntegratedQuantum, line 1370: I think you need to copy the data from the old image first.
Also old image is never deinited.

src/graphics.zig
```diff
@@ -1344,16 +1433,41 @@ const TextRendering = struct { // MARK: TextRendering
...
+					.fontEffects = fontEffects,
+					.textureBounds = .{@floatFromInt(glyph.textureX), 0, @floatFromInt(glyph.size[0]), @floatFromInt(glyph.size[1])},
+				});
+			} else {
+				c.glUniform1i(uniforms.fontEffects, fontEffects);
+				c.glUniform4f(uniforms.textureBounds, @floatFromInt(glyph.textureX), 0, @floatFromInt(glyph.size[0]), @floatFromInt(glyph.size[1]));
+				c.glUniform2f(uniforms.offset, @as(f32, @floatFromInt(glyph.bearing[0]))*draw.scale + x, @as(f32, @floatFromInt(glyph.bearing[1]))*draw.scale + y);
+				c.glUniform4f(uniforms.textureRect, @floatFromInt(glyph.textureX), 0, @floatFromInt(glyph.size[0]), @floatFromInt(glyph.size[1]));
+			}
+		}
```
IntegratedQuantum, line 1445: This is too big, please do
```
var uniforms: GlyphUniforms = .{
   ....
};
if (bold) {
 uniforms.xyz = 01043825857;
} else {
 ...
}
// Then the if on platform
```

IntegratedQuantum (changes requested):

src/graphics.zig
Wunka: done

src/graphics/vulkan.zig
```diff
@@ -1016,6 +1016,29 @@ pub const Image = struct { // MARK: Image
 			},
 		}});
 	}
+
+	pub fn uploadImage(self: Image, other: Image) void {
+		currentFrame.uploadCommands.copyImageToImage(self, c.VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, other, c.VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL, &.{
```
IntegratedQuantum: You also need to add pipeline barriers (like the other code) to ensure that both images are in the right layout and to return the dest image to `VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL`

src/graphics.zig
```diff
@@ -1293,23 +1365,39 @@ const TextRendering = struct { // MARK: TextRendering
...
-		c.glCopyImageSubData(glyphTexture[1], c.GL_TEXTURE_2D, 0, 0, 0, 0, glyphTexture[0], c.GL_TEXTURE_2D, 0, 0, 0, 0, textureOffset, textureHeight, 1);
-		pipeline.bind(draw.getScissor());
-		c.glUniform2f(uniforms.fontSize, @floatFromInt(textureWidth), @floatFromInt(textureHeight));
+		if (main.settings.launchConfig.vulkanTestingMode) {
+			glyphTexture[0].vulkanImage = vulkan.Image.init(.{textureWidth, textureHeight, 1}, .{
+				.format = c.VK_FORMAT_R8_UNORM,
+				.usage = c.VK_IMAGE_USAGE_TRANSFER_DST_BIT | c.VK_IMAGE_USAGE_SAMPLED_BIT | c.VK_IMAGE_USAGE_TRANSFER_SRC_BIT,
+			});
+			glyphTexture[0].vulkanImage.?.uploadImage(glyphTexture[1].vulkanImage.?);
+			glyphTexture[1].vulkanImage.?.deferredDeinit();
```
IntegratedQuantum: I'd suggest to restrict the swap to the OpenGL code, and instead just do the following. We don't need the swap here, since the deiniting is deferred until it's no longer used. It probably wasn't necessary in OpenGL either, but we shouldn't to too much cleanup there anymore.
```zig
const old = glyphTexture[0].vulkanImage;
glyphTexture[0].vulkanImage = .inti(...);
old.deferredDeinit();
```
(current approach is also a double free on program exit)

src/graphics.zig
```diff
@@ -1393,9 +1393,7 @@ const TextRendering = struct { // MARK: TextRendering
 		}
 		if (main.settings.launchConfig.vulkanTestingMode) {
 			glyphTexture[0].vulkanImage.?.size = .{width, height, 1};
-			glyphTexture[0].vulkanImage.?.uploadData(buffer[0..@intCast(pitch*height)], .{
-				.imageOffset = .{.x = textureOffset},
-			});
+			glyphTexture[0].vulkanImage.?.uploadData(buffer[0..@intCast(pitch*height)], .{.imageOffset = .{.x = textureOffset}, .imageExtent = .{.width = @intCast(width), .height = @intCast(height), .depth = 1}});
```
IntegratedQuantum: Could you change the code, so that the intCasts are only needed in the OpenGL code?
Currently used bitCast for conversion from freetype also looks suspicious (it should have been intCast probably).

src/graphics/vulkan.zig
```diff
@@ -1012,6 +1017,29 @@ pub const Image = struct { // MARK: Image
...
+				},
+				.srcOffset = .{},
+				.dstSubresource = .{
+					.aspectMask = c.VK_IMAGE_ASPECT_COLOR_BIT,
+					.mipLevel = 0,
+					.baseArrayLayer = 0,
+					.layerCount = 1,
+				},
+				.dstOffset = .{},
+				.extent = .{.width = @intCast(dest.size[0]), .height = @intCast(dest.size[1]), .depth = @intCast(dest.size[2])},
```
IntegratedQuantum: this is still wrong

src/graphics.zig
```diff
@@ -1293,23 +1365,39 @@ const TextRendering = struct { // MARK: TextRendering
...
 		const height: i32 = @bitCast(bitmap.rows);
+		const pitch: i32 = @bitCast(bitmap.pitch);
 		const buffer = bitmap.buffer orelse return;
 		if (textureOffset + width > textureWidth) {
 			resizeTexture(textureWidth*2);
 		}
-		c.glPixelStorei(c.GL_UNPACK_ALIGNMENT, 1);
-		c.glTexSubImage2D(c.GL_TEXTURE_2D, 0, textureOffset, 0, width, height, c.GL_RED, c.GL_UNSIGNED_BYTE, buffer);
+		if (main.settings.launchConfig.vulkanTestingMode) {
+			glyphTexture[0].vulkanImage.?.size = .{width, height, 1};
```
IntegratedQuantum: This is still wrong (see other comment)

IntegratedQuantum (changes requested):

src/graphics/vulkan.zig
Wunka: all done, I think

src/graphics.zig
```diff
@@ -1265,52 +1317,88 @@ const TextRendering = struct { // MARK: TextRendering
...
+			pipeline.bind(draw.getScissor());
+			c.glUniform2f(uniforms.fontSize, @floatFromInt(textureWidth), @floatFromInt(textureHeight));
+		}
 	}
 
 	fn uploadData(bitmap: c.FT_Bitmap) void {
-		const width: i32 = @bitCast(bitmap.width);
-		const height: i32 = @bitCast(bitmap.rows);
+		const width: u32 = @intCast(bitmap.width);
+		const height: u32 = @intCast(bitmap.rows);
```
IntegratedQuantum: These are already c_uint, no need to cast

src/graphics/vulkan.zig
```diff
@@ -1012,6 +1017,81 @@ pub const Image = struct { // MARK: Image
...
+				.srcAccessMask = c.VK_ACCESS_2_NONE,
+				.dstStageMask = c.VK_PIPELINE_STAGE_2_TRANSFER_BIT,
+				.dstAccessMask = c.VK_ACCESS_2_TRANSFER_WRITE_BIT,
+				.oldLayout = c.VK_IMAGE_LAYOUT_UNDEFINED,
+				.newLayout = c.VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
+				.image = dest.handle,
+				.subresourceRange = .{.aspectMask = c.VK_IMAGE_ASPECT_COLOR_BIT, .levelCount = dest.mipLevels, .layerCount = 1},
+			},
+		}});
+		currentFrame.uploadCommands.pipelineBarrier(.{.imageMemoryBarriers = &.{
```
IntegratedQuantum: This takes a slice, so please only do one command with both barriers

src/graphics/vulkan.zig
```diff
@@ -1012,6 +1017,81 @@ pub const Image = struct { // MARK: Image
...
+			.{
+				.sType = c.VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER_2,
+				.srcStageMask = c.VK_PIPELINE_STAGE_2_NONE,
+				.srcAccessMask = c.VK_ACCESS_2_NONE,
+				.dstStageMask = c.VK_PIPELINE_STAGE_2_TRANSFER_BIT,
+				.dstAccessMask = c.VK_ACCESS_2_TRANSFER_WRITE_BIT,
```
IntegratedQuantum: should be read

src/graphics/vulkan.zig
```diff
@@ -1012,6 +1017,81 @@ pub const Image = struct { // MARK: Image
...
+				.oldLayout = c.VK_IMAGE_LAYOUT_UNDEFINED,
+				.newLayout = c.VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
+				.image = source.handle,
+				.subresourceRange = .{.aspectMask = c.VK_IMAGE_ASPECT_COLOR_BIT, .levelCount = dest.mipLevels, .layerCount = 1},
```
IntegratedQuantum: Should be source.mipLevels

src/graphics/vulkan.zig
```diff
@@ -1012,6 +1017,81 @@ pub const Image = struct { // MARK: Image
...
+				.newLayout = c.VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL,
+				.image = dest.handle,
+				.subresourceRange = .{.aspectMask = c.VK_IMAGE_ASPECT_COLOR_BIT, .levelCount = dest.mipLevels, .layerCount = 1},
+			},
+		}});
+		currentFrame.uploadCommands.pipelineBarrier(.{.imageMemoryBarriers = &.{
+			.{
+				.sType = c.VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER_2,
+				.srcStageMask = c.VK_PIPELINE_STAGE_2_TRANSFER_BIT,
+				.srcAccessMask = c.VK_ACCESS_2_TRANSFER_WRITE_BIT,
```
IntegratedQuantum: read

src/graphics/vulkan.zig
```diff
@@ -1012,6 +1017,81 @@ pub const Image = struct { // MARK: Image
...
+				.dstStageMask = c.VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT,
+				.dstAccessMask = c.VK_ACCESS_SHADER_READ_BIT,
+				.oldLayout = c.VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
+				.newLayout = c.VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL,
+				.image = source.handle,
+				.subresourceRange = .{.aspectMask = c.VK_IMAGE_ASPECT_COLOR_BIT, .levelCount = dest.mipLevels, .layerCount = 1},
```
IntegratedQuantum: sourcce.mipLevels

src/graphics/vulkan.zig
```diff
@@ -1012,6 +1017,81 @@ pub const Image = struct { // MARK: Image
...
+				.srcAccessMask = c.VK_ACCESS_2_TRANSFER_WRITE_BIT,
+				.dstStageMask = c.VK_PIPELINE_STAGE_2_ALL_GRAPHICS_BIT,
+				.dstAccessMask = c.VK_ACCESS_SHADER_READ_BIT,
+				.oldLayout = c.VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
+				.newLayout = c.VK_IMAGE_LAYOUT_READ_ONLY_OPTIMAL,
+				.image = dest.handle,
+				.subresourceRange = .{.aspectMask = c.VK_IMAGE_ASPECT_COLOR_BIT, .levelCount = dest.mipLevels, .layerCount = 1},
+			},
+		}});
+		currentFrame.uploadCommands.pipelineBarrier(.{.imageMemoryBarriers = &.{
```
IntegratedQuantum: Can be done in one command

IntegratedQuantum (changes requested):

src/graphics/vulkan.zig
Wunka: ups... all done

IntegratedQuantum (approved):

PR #3584, opened 2026-09-08
Item Frames

Crepestrom:
image
image


Wunka (changes requested):
There is still a lot of leftover naming from the sign code you copied. 
This makes bugs possible thorugh for example your callback where you lock the signs mutex and not the item frames one

Crepestrom:
oops didnt catch that
updated a bunch of the names

src/block_entity.zig
```diff
@@ -535,6 +538,242 @@ pub const BlockEntityTypes = struct { // MARK: BlockEntityTypes
...
+		pub fn getClientToServerData(pos: Vec3i, chunk: *Chunk, writer: *BinaryWriter) void {
+			StorageClient.mutex.lock();
+			defer StorageClient.mutex.unlock();
+
+			const data = StorageClient.get(pos, chunk) orelse return;
+			if (data.item == .null) return;
+			main.items.Item.toBytes(data.item, writer);
+		}
+
+		pub fn updateTextFromClient(pos: Vec3i, newText: main.items.Item) void {
```
Wunka: still naming from the sign

Wunka (changes requested):

careeoki (changes requested):
Attempting to put items in frames freezes the game for me, with and without debug mode. 
It doesn't crash so I can't show a log.

IntegratedQuantum:
Please wait until the block entity rewrite is finished, to avoid having to migrate the save data.
In the future block entities will probably use the ECS too.

Crepestrom:
ok will wait

du82:
I like it! Although I think more colors than just green should be possible

TyrrelMeter:
Item frames in many colours would be really cool, I agree. Might be a fair bit of work for Crepestrom when Carrie will probably redo the texture anyway.

PR #3448, opened 2026-07-31
Show a clear error when autoEnterWorld names a missing world

arimu1:
## Summary
Fixes #3436

When `launchConfig.autoEnterWorld` points at a world that does not exist, Cubyz used to:
1. create an empty `saves/<name>/` directory (because `openDir` creates missing paths),
2. fail reading `world.zig.zon` with a raw `FileNotFound` stacktrace,
3. hit `unloadAssets` from an `errdefer` that ran before assets were loaded, which asserted and printed a second stacktrace.

## Changes
- Check for `saves/<name>/world.zig.zon` before opening the save directory.
- Log a clear error that includes the world name and the expected path.
- Return `error.WorldNotFound` and exit cleanly from the server instead of panicking for that case.
- Move the `unloadAssets` `errdefer` to after a successful `loadWorldAssets`, so failed world loads no longer assert during cleanup.

## Test plan
- [x] `zig build -Doptimize=Debug` succeeds
- [ ] Set `headlessServer = true` and `autoEnterWorld = "does_not_exist"` in `launchConfig.zon`; expect a short console error naming the missing world and a non-zero exit, with no assert stacktrace and no empty `saves/does_not_exist/` folder left behind
- [ ] Open an existing world via `autoEnterWorld` still works as before

PR #3585, opened 2026-09-08
Error when you leave the name field of .autoEnterWorld empty

Crepestrom:


Wunka:
This is not the correct fix. This only solves the problem for if you enter no world name, but what if you type the wrong name? Also please link the issue this tries fix #3448

PR #3587, opened 2026-09-08
Modify component Pt 2

Crepestrom:
Adds a new sync command that allows sending small actions from client to server or server to client

Goal:
- a system allows the client/server to send smaller messages to the server/client rather (apart from sending and overwriting the entire component data)

Method:
- In each ECS component is defined a modifyComponent function
- if a component wants to send a modify component command to the server (a taking damage command) the client will compress the message into a binary writer and send it with a syncCommand
- the sync command checks what side it is on and interprets its messages (server side ignores load and unload commands because the client should not overwrite server data while the server should be able to override client data)
- if the sync command is undone it will revert that component back to a previously saved state before the sync command was issued
- if a component receives a modifyComponent Command from sync (Eg: server to client) then the side that recieves it will run modifyComponent (client will run it)
- modifyComponent only allows what the component allows to be changed by it (eg: you cant change max health when you take damage)

Use case:
- A inventory component (use a enum to determine the operation and then use the rest of the binary data to finish it)
- A health component
- Status effect component

### DIFF
--- a/src/sync.zig
+++ b/src/sync.zig
@@ -277,7 +277,7 @@ pub const Command = struct { // MARK: Command
 		updateBlock: UpdateBlock,
 		addHealth: AddHealth,
 		chatCommand: ChatCommand,
-		modifyComponent: ModifyComponent
+		modifyComponent: ModifyComponent,
 	};
 
 	const BaseOperationType = enum(u8) {
@@ -903,7 +903,7 @@ pub const Command = struct { // MARK: Command
 					std.log.debug("side {} actiontype {}", .{side, actionType});
 
 					inline for (@typeInfo(main.entity.components).@"struct".decls) |decl| {
-						if (@field(main.entity.components, decl.name).server.get(entityId)) |component| {
+						if (@field(main.entity.components, decl.name).client.get(entityId)) |component| {
 							var binaryWriter = main.utils.BinaryWriter.init(main.stackAllocator);
 							defer binaryWriter.deinit();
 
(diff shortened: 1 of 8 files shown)


src/sync.zig
```diff
@@ -854,6 +871,60 @@ pub const Command = struct { // MARK: Command
...
+					}
+				} else {
+					var binaryReader = BinaryReader.init(info.modification);
+					const entityId: main.entity.Entity = @enumFromInt(binaryReader.readVarInt(u32) catch return);
+					const componentId = info.componentId;
+					const actionType: main.entity.ComponentActionType = binaryReader.readEnum(main.entity.ComponentActionType) catch return;
+					std.log.debug("side {} actiontype {}", .{side, actionType});
+
+					inline for (@typeInfo(main.entity.components).@"struct".decls) |decl| {
+						if (@field(main.entity.components, decl.name).server.get(entityId)) |component| {
```
IntegratedQuantum: it still accesses the server on the client

Crepestrom:
made a write up for the system and its goals

Crepestrom:
Goal:
- a system allows the client/server to send smaller messages to the server/client rather (apart from sending and overwriting the entire component data)

Method:
- In each ECS component is defined a modifyComponent function
- if a component wants to send a modify component command to the server (a taking damage command) the client will compress the message into a binary writer and send it with a syncCommand
- the sync command checks what side it is on and interprets its messages (server side ignores load and unload commands because the client should not overwrite server data while the server should be able to override client data)
- if the sync command is undone it will revert that component back to a previously saved state before the sync command was issued
- if a component receives a modifyComponent Command from sync (Eg: server to client) then the side that recieves it will run modifyComponent (client will run it)
- modifyComponent only allows what the component allows to be changed by it (eg: you cant change max health when you take damage)

Use case:
- A inventory component (use a enum to determine the operation and then use the rest of the binary data to finish it)
- A health component
- Status effect component

PR #3588, opened 2026-09-08
Ecs Modify Component Pt 1

Crepestrom:
adds the basic functions required for the Modify Component Ecs pt 2

meant for sending signals from the server to the client or the client to the server
modifyComponent
is meant for smaller signals that dont overwrite a entire component. (eg: sending how much damage a player takes, or sending a inventory command)

### DIFF
--- a/src/entityComponent/permissions.zig
+++ b/src/entityComponent/permissions.zig
@@ -23,6 +23,10 @@ pub const client = struct {
 	pub fn init() void {}
 	pub fn deinit() void {}
 	pub fn clear() void {}
+	pub fn get(entity: Entity) null {
+		_ = entity;
+		return null;
+	}
 
 	pub fn modifyComponent(entity: Entity, reader: *utils.BinaryReader) void {
 		_ = entity;
--- a/src/entityComponent/permissions.zig
+++ b/src/entityComponent/permissions.zig
@@ -23,22 +23,20 @@ pub const client = struct {
 	pub fn init() void {}
 	pub fn deinit() void {}
 	pub fn clear() void {}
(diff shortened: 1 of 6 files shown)


src/entityComponent/permissions.zig
```diff
@@ -89,39 +86,25 @@ pub const server = struct {
...
 	}
 
-	pub fn addToGroup(entity: Entity, group: main.server.permission.Group) void {
-		(getPermissionGroups(entity) orelse return).put(main.globalAllocator.allocator, group, {}) catch unreachable;
-	}
-
-	pub fn removeFromGroup(entity: Entity, group: main.server.permission.Group) bool {
-		return getPermissionGroups(entity).?.remove(group);
-	}
-
```
Wunka, line 99: you removed the changes from #3332 ?
Please fix this. Also as you already know the ECS is currently in flux. 
So while I will put this into low priority (another reason is that we soon reaching the 0.4.0 release, so big changes are not that much accepted now), please make a comment about your plan with for example your part 2 so we can correctly evaluate if this the direction we want to go.

Wunka (changes requested):

src/entityComponent/permissions.zig
Crepestrom: Oops
Also that’s fine if I have to wait for the next version
I will write up a plan later

Crepestrom: wrote the thing on it

Crepestrom: also i fixed this thing now

PR #3590, opened 2026-09-09
Crafting table

Crepestrom:
When you open the crafting menu it will check if it has the right tags to show a recipe to the player
It also allows blocks to call open the crafting menu with its own tags (allows for gating recipes behind certain crafting tables)


careeoki:
used this to quickly whip together a table saw (#1734) 
image

src/items/recipes.zig
```diff
@@ -190,19 +190,23 @@ fn generateItemCombos(allocator: NeverFailingAllocator, recipe: []const ZonEleme
...
-pub fn addRecipe(itemCombo: []const ItemWithAmount, list: *main.ListManaged(Recipe)) void {
+pub fn addRecipe(itemCombo: []const ItemWithAmount, list: *main.ListManaged(Recipe), craftingTags: []Tag) void {
 	const inputs = itemCombo[0 .. itemCombo.len - 1];
 	const output = itemCombo[itemCombo.len - 1];
 	const recipe = Recipe{
 		.sourceItems = main.worldArena.alloc(BaseItemIndex, inputs.len),
 		.sourceAmounts = main.worldArena.alloc(u16, inputs.len),
 		.resultItem = output.item,
 		.resultAmount = output.amount,
+		.craftingTags = main.worldArena.alloc(Tag, craftingTags.len),
```
Wunka:
```suggestion
		.craftingTags = main.worldArena.dupe(Tag, craftingTags),
```
Then you don't need to copy them later

src/items/recipes.zig
```diff
@@ -218,11 +222,16 @@ pub fn parseRecipe(zon: ZonElement, list: *main.ListManaged(Recipe)) !void {
 		return error.InvalidReversibleRecipe;
 	}
 
+	var defaultCraftingTags = arena.alloc(main.Tag, 1);
+	defaultCraftingTags[0] = main.Tag.handCraftable;
+	const foundCraftingTags = Tag.loadTagsFromZon(arena, zon.getChild("craftingTags"));
+	const craftingTags = if (foundCraftingTags.len != 0) foundCraftingTags else defaultCraftingTags;
```
Wunka: If you decide 4 lines later if you even need your allocation, then only do it if you need it

src/items.zig
```diff
@@ -1361,7 +1349,15 @@ pub const Recipe = struct { // MARK: Recipe
 			sourceAmounts.append(main.stackAllocator, try reader.readVarInt(u16));
 		}
 
-		return getValidRecipe(.{.sourceItems = sourceItems.items, .sourceAmounts = sourceAmounts.items, .resultItem = resultItem, .resultAmount = resultAmount});
+		const tagCount = try reader.readVarInt(usize);
+		var craftingTagTypes: main.List(Tag) = .initCapacity(main.stackAllocator, @min(256, tagCount));
```
Wunka: no magic numbers please, where does the 256 come from? where do you have this limitation? If its by a type, then use `std.math.maxInt` for example

Wunka (changes requested):

src/items.zig
Crepestrom: i dont actually know

src/items/recipes.zig
Crepestrom: ah i see now
i copied the other type of code but this doesnt need that implementaiton

Crepestrom: swapped that over

src/items/recipes.zig
Crepestrom: moved over so we dont unessicarily duplicate the work

src/items.zig
Crepestrom: its gone now

src/items/recipes.zig
```diff
@@ -218,11 +219,18 @@ pub fn parseRecipe(zon: ZonElement, list: *main.ListManaged(Recipe)) !void {
 		return error.InvalidReversibleRecipe;
 	}
 
+	const foundCraftingTags = Tag.loadTagsFromZon(arena, zon.getChild("craftingTags"));
+	const craftingTags = if (foundCraftingTags.len != 0) foundCraftingTags else blk: {
+		var defaultCraftingTags = arena.alloc(main.Tag, 1);
+		defaultCraftingTags[0] = main.Tag.handCraftable;
+		break :blk defaultCraftingTags;
+	};
```
Wunka: I just remembered, because you don't use an arena, you should also just be able to do:
```suggestion
	const craftingTags = if (foundCraftingTags.len != 0) foundCraftingTags else &.{main.Tag.handCraftable}
```

Wunka (changes requested):

src/items/recipes.zig
Crepestrom: oh yeah  i did something like that

Wunka: It never leaves the stack, so please don't dupe it.

Crepestrom: ah ok

PR #3592, opened 2026-09-09
Vulkan coordinate System

Wunka:
This is not ready... first every thing has to be tested to still work. The nice thing is, for this can even a normal window be used, without `vulkanTestingMode`, as this changes also in general how opengl things need to be renderd.

- [x] projection matrix
- [x] change shaders where an obvious switch can be made
- [x] fix clipping ( `c.glClipControl(c.GL_UPPER_LEFT, c.GL_ZERO_TO_ONE)` does not apply this setting to everything. clipping for example still works on the old coordinate system.)
- [x] fix wierd image moving behavior (only in vulkan window)
- [x] fix item drops vanishing (the problem is item_drop.frag, the FragDepth calculation seems to be doing something not correct anymore) (done, by not doing the depth conversion)
- [x] problems with light sources, can be good seen on torches.
- [x] fix problems with fog (can be good seen in the rainbow_forest) (done, by not doing the depth conversion)
- [x] fix reflections
- [x] fix stars
- [x] find more bugs and fix them (hopefully there are no more...)

image

### DIFF
--- a/assets/cubyz/shaders/bloom/color_extractor_downsample.vert
+++ b/assets/cubyz/shaders/bloom/color_extractor_downsample.vert
@@ -12,7 +12,7 @@ layout(location = 0) uniform mat4 invViewMatrix;
 layout(location = 1) uniform vec2 tanXY;
 
 void main() {
-	vec2 position = inTexCoords*2 - vec2(1, 1);
+	vec2 position = inTexCoords*2+vec2(-1, -1);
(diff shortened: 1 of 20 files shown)


Crepestrom:
good job with all this work

Wunka:
As talked with quantum, while this should be looked at in general to find bugs, it's probably better to only merge this after 0.4.0 release to not get any bugs in the release

PR #3593, opened 2026-09-10
Fixed time help info

codingbanana21:
Added dusk and dawn to the help info and added info about starting and stopping the server time.


Wunka:
Just for your information, your git config is probably not correctly set. Your git name is "=", probably because you typed `git config user.name = [NAME]` instead of `git config user.name [NAME]`. This doesn't mean this can't be merged, just wanted to make you aware

IntegratedQuantum (approved):

PR #3596, opened 2026-09-13
remove deprecated things in preperation to zig 0.17.0

Wunka:
The rest of the changes I have in my 0.17.0 branch, can't sadly be already brought over I think.


src/utils/heap.zig
```diff
@@ -483,7 +483,7 @@ pub const NeverFailingAllocator = struct { // MARK: NeverFailingAllocator
 
 	/// Copies `m` to newly allocated memory, with a null-terminated element. Caller owns the memory.
 	pub fn dupeZ(self: NeverFailingAllocator, comptime T: type, m: []const T) [:0]T {
```
IntegratedQuantum: We want to keep our interface equivalent to the std one, so please also rename this function and adjust the signature accordingly.

IntegratedQuantum (changes requested):

src/utils/heap.zig
Wunka: done

IntegratedQuantum (approved):

PR #3608, opened 2026-09-17
Other branches & Misc - Palette textures 6

careeoki:
The grand finale. 
image
Remember how resin is a thing?? that exists???????? in this game??????????????


IntegratedQuantum (approved):
> Remember how resin is a thing?? that exists???????? in this game??????????????

Oh yeah, I wonder when that will be obtainable... Maybe after we make furnaces work...

PR #3612, opened 2026-09-19
makes things pub

### LINKED ISSUE #3609: `@hasDecl` with zig 0.17.0 needs the field to be public
upstream: https://codeberg.org/ziglang/zig/pulls/36621

I discovered this when I wanted to test my 0.17.0 branch. 
Affected are at least our protocols as I discovered this by them not working anymore.

Wunka:
Fixes: #3609 I think. At least this works on my 0.17.0 branch


src/sync.zig
```diff
@@ -716,7 +716,7 @@ pub const Command = struct { // MARK: Command
 		}
 	}
 
-	fn confirmationData(self: *Command, allocator: NeverFailingAllocator) []const u8 {
+	pub fn confirmationData(self: *Command, allocator: NeverFailingAllocator) []const u8 {
```
IntegratedQuantum: The global one doesn't need to be public

src/sync.zig
```diff
@@ -686,7 +686,7 @@ pub const Command = struct { // MARK: Command
 		}
 	}
 
-	fn finalize(self: Command, allocator: NeverFailingAllocator, side: Side, reader: *BinaryReader) !void {
+	pub fn finalize(self: Command, allocator: NeverFailingAllocator, side: Side, reader: *BinaryReader) !void {
```
IntegratedQuantum: The global one doesn't need to be public

IntegratedQuantum (changes requested):

src/sync.zig
Wunka: both fixed.

IntegratedQuantum (approved):

PR #3613, opened 2026-09-19
default permission group

### LINKED ISSUE #3143: remove temporary permission
delete ``user.permissions.addPermission(.white, "/command/avatar");`` from the codebase in server.zig

Wunka:
The next task after this would be to change how the settings work.
Fixes: #3143

Changes:
- `allowCheats` now just means that the host has root permissions.
- Everyone gets added to default permission on join, removing them is possible, but on rejoin they get added again. Server Moderators should instead just blacklist the users permission.


IntegratedQuantum (changes requested):
> removing them is possible

I think we should have some protection here. Permission command should not allow deleting `default` or removing it from any player.

> allowCheats now just means that the host has root permissions

The local user should have the privilege to kick/whitelist people and configure their permissions. Maybe we should introduce a moderator group for this purpose?

Wunka:
> > removing them is possible
> 
> I think we should have some protection here. Permission command should not allow deleting `default` or removing it from any player.

Mhhh yeah. That we could do.

> > allowCheats now just means that the host has root permissions
> 
> The local user should have the privilege to kick/whitelist people and configure their permissions. Maybe we should introduce a moderator group for this purpose?

Yeah, when I wrote this code I also realized that we probably need a second "moderator" or whatever group. I also now remembered that if you have permission to use `/perm` you kinda have permission to do anything (you can give yourself `/` permissions). I do want to fix that.

IntegratedQuantum:
> I also now remembered that if you have permission to use /perm you kinda have permission to do anything (you can give yourself / permissions). I do want to fix that.

Yeah, I think you should only be allowed to modify the permission paths/groups that you also own. This would also be important for the protector block #81 (but that's not important for this PR)

Wunka:
Changes:
- moderator group added, local player gets added to this, problem: how to remove the local player again. Or should we even? If you host the world once, you are now a moderator for life?
- if `cheatsEnabled`, the local player also gets root permissions
- perm command now only works on commands you also have permissions for.

IntegratedQuantum:
> Or should we even? If you host the world once, you are now a moderator for life?

If you trust someone else to host your world, then they can change anything they want anyways. No need to add special handling for this.

src/server/command/permission/perm.zig
```diff
@@ -36,6 +36,10 @@ pub fn execute(args: Args, source: Source) void {
 				.blacklist => .black,
 			};
 
+			if (!source.hasPermission(params.permissionPath.path)) {
```
IntegratedQuantum, line 39: Oh this is actually easier than I thought it would be xD

Wunka: This does not fix the problem with, what groups are you allowed to manipulate. But this is a problem for the group command pr. I think I will there go for special paths, like:
`/command/group` -> everything, like creation etc.
`/command/group/{id}` -> delete, invite and remove for that group alone.

src/server/permission.zig
```diff
@@ -323,18 +327,34 @@ pub fn loadGroups(dir: main.files.Dir) !void {
...
 	metadataZon.put("currentId", @as(u32, @truncate(groups.items.len)));
 	try main.files.cubyzDir().writeZon(metadatPath, metadataZon);
 }
 
+fn createDefaultPermissionGroups() void {
+	blk: {
+		const group = Group.createGroup(defaultGroupName) catch break :blk;
+
+		group.addPermission(.white, "/command/avatar") catch break :blk;
+		group.addPermission(.white, "/command/help") catch break :blk;
```
IntegratedQuantum: The error on addPermission shouldn't happen → `catch unreachable`

src/server/permission.zig
```diff
@@ -197,6 +197,9 @@ var groups: main.ListManaged(?*GroupInstance) = undefined;
 var groupNameToIdMap: std.StringHashMapUnmanaged(Group) = .{};
 
 var groupsArena: NeverFailingArenaAllocator = undefined;
+var groupsPath: []const u8 = undefined;
+pub const defaultGroupName = "default";
+pub const moderatorGroupName = "moderator";
```
IntegratedQuantum: If the name is mainly used to access these groups to e.g. add to a player, why not just store the `Group` here directly to avoid needing to repull them by name every time?

src/server/server.zig
```diff
@@ -310,11 +310,18 @@ pub const User = struct { // MARK: User
 		}
 		if (main.entity.components.@"cubyz:permissions".server.get(self.id) == null) {
 			main.entity.components.@"cubyz:permissions".server.loadEmpty(self.id);
-			main.entity.components.@"cubyz:permissions".server.addPermission(self.id, .white, "/command/avatar");
-			main.entity.components.@"cubyz:permissions".server.addPermission(self.id, .white, "/command/help");
 		}
+		if (permission.Group.getByName(permission.defaultGroupName)) |group| {
+			main.entity.components.@"cubyz:permissions".server.addToGroup(self.id, group);
+		} else |_| {}
```
IntegratedQuantum: I think these cases deserve an error log.

src/server/command/permission/perm.zig
Wunka: Actually.... my group command idea doesn't work... mhhh I have to thing about this.

src/server/permission.zig
```diff
@@ -323,18 +327,34 @@ pub fn loadGroups(dir: main.files.Dir) !void {
...
+		const group = Group.createGroup(defaultGroupName) catch break :blk;
+
+		group.addPermission(.white, "/command/avatar") catch break :blk;
+		group.addPermission(.white, "/command/help") catch break :blk;
+	}
+	blk: {
+		const group = Group.createGroup(moderatorGroupName) catch break :blk;
+
+		group.addPermission(.white, "/command/perm") catch break :blk;
+		group.addPermission(.white, "/command/whitelist") catch break :blk;
```
IntegratedQuantum: `/invite` and `/server` and `/kick` also belong in this category in my opinion.

IntegratedQuantum (changes requested):

src/server/server.zig
Wunka: Everything done and these else cases don't exist anymore with your other suggestion

src/server/permission.zig
```diff
@@ -323,18 +328,37 @@ pub fn loadGroups(dir: main.files.Dir) !void {
...
+		defaultGroup.addPermission(.white, "/command/help") catch unreachable;
+	}
+	blk: {
+		moderatorGroup = Group.createGroup("moderator") catch break :blk;
+
+		moderatorGroup.addPermission(.white, "/command/perm") catch unreachable;
+		moderatorGroup.addPermission(.white, "/command/whitelist") catch unreachable;
+		moderatorGroup.addPermission(.white, "/command/invite") catch unreachable;
+		moderatorGroup.addPermission(.white, "/command/server") catch unreachable;
+		moderatorGroup.addPermission(.white, "/command/kick") catch unreachable;
```
IntegratedQuantum: Could you sort this alphabetically?

src/server/server.zig
```diff
@@ -310,11 +310,14 @@ pub const User = struct { // MARK: User
 		}
 		if (main.entity.components.@"cubyz:permissions".server.get(self.id) == null) {
 			main.entity.components.@"cubyz:permissions".server.loadEmpty(self.id);
-			main.entity.components.@"cubyz:permissions".server.addPermission(self.id, .white, "/command/avatar");
-			main.entity.components.@"cubyz:permissions".server.addPermission(self.id, .white, "/command/help");
 		}
+		main.entity.components.@"cubyz:permissions".server.addToGroup(self.id, permission.defaultGroup);
```
IntegratedQuantum: Ok crazy idea, but what if you could just do

```suggestion
		main.entity.components.@"cubyz:permissions".server.addToGroup(self.id, .default);
```

This would be possible by moving the constants inside the `Group` struct which allows using them as decl literals.

IntegratedQuantum (changes requested):

src/server/server.zig
Wunka: that's a crazy idea.

Wunka: image
:(

IntegratedQuantum: Sad

Wunka: but still moved, and the permissions paths are now sorted

IntegratedQuantum (approved):

IntegratedQuantum:
Thanks for still making this before the release. Even without the group command this should make server management a lot easier.

PR #3615, opened 2026-09-19
convert texture* from i32 to u31 in TextRendering

### LINKED ISSUE #3594: TextRendering.textureWidth/-Height/-Offset should be unsigned integers
extracted from #3582

realugi:
fix #3594 

I chose to use `u31` instead of `u32`, because `u31` is implicitly convertable to `i32` and therefore `c_int`. Also saw precedence in `main.graphics.Image`.

Also removed the now unneeded cast at `L1390`, because both `bitmap.width` and `textureOffset` are now unsigned integers.


IntegratedQuantum (changes requested):
Is there a reason to support c_int here? Will it even be needed once we get rid of OpenGL after #102?

realugi:
So the reason to support `c_int` in OpenGL case is essentially so that we do not need to do any conversions when passing our values to the OpenGL API.

The same applies to Vulkan (in some cases):
- for the `TextRendering.init` vulkan path, the `vulkan.Image.init` method accepts `main.vec.Vec3i` (which contains `i32`s), but the actual vulkan api accepts `VkExtent2D`, which contains `u32`s. So in this case, `u32`s for texture{Width|Height} would suffice. Though we'd need to adjust the `vulkan.Image.init` method params if we wanna do this cleanly.
- for the `TextRendering.uploadData` vulkan path, `vulkan.Image.uploadData` method accepts a `VkOffset3D` (through `vulkan.Image.UploadDataConfig`'s `.imageOffset`), which contains `i32`s. So in this case, making textureOffset a `u32` would require a conversion.

So if we want `texture{Width|Height|Offset}` types to fit the vulkan api, then:
- `texture{Width|Height}` should be `u32`, w/ conversions for opengl, w/o conversions for vulkan, though I'd need to adjust the `vulkan.Image.init`'s `size: main.vec.Vec3i` parameter, which also maybe necessitates changin the `size` field of `vulkan.Image`, which could become quite an extensive change actually
- `textureOffset` stays `u31`

PR #3616, opened 2026-09-19
fix default groups

Wunka:
... yeah this bug shouldn't have happend


IntegratedQuantum (approved):
Right, we need a way to update them on updates.

I was about to say that this makes changing permission groups more difficult, but then you can still just use the blacklist, so I guess it's fine.

Wunka:
this was more of a fix, because the groups would stay `undefined` on joining a world the second time...

IntegratedQuantum:
Oh, right that's an even bigger issue

PR #3621, opened 2026-09-19
Combined all of the basic stats modifier

### LINKED ISSUE #3611: Combine all Basic Stat Modifiers
Should we combine every basic stat modifier into one big one (eg: damage, swing speed, durability)
We can allow the modifier zon to include what we want the name of it and the color of that name to be when it shows up so that we keep all of the old names.

Crepestrom:
Fixes: #3611  
increases maintainability of the code as you do not need to modify 8 basic stat modifiers in the same exact way
Breaks many addons that use the old system of modifiers
Also negative and positive modifiers are combined instead of being applied seperatly
eg: a +50% durability and -50% durability modifier will cancel out to +0% durability instead of -25% durability (75% of total durability)

Before:
image
After:
image


Crepestrom:
Which is better
1
image
or
2
image

calebstenberg:
> Which is better

I almost like not seeing Fragile and Powerful and just having the stat change

Crepestrom:
good point i will remove the increases and decreases
was also drawing away the eye from the number

PR #3626, opened 2026-09-20
Sound Effects (pt 1 continueation)

Crepestrom:
working from the pr by [RanPix](https://github.com/RanPix)
Continueing from: #3281


Crepestrom:
i have fixed all the previous issues
exept converting into mono

Crepestrom:
if its fine can i seperate the pr for loading sound effects from assets into a different one?

Crepestrom:
how can i convert stereo sound into mono?

RanPix:
> how can i convert stereo sound into mono?

you can either use ffmpeg to convetr the sound files themselves or in audio struct sound loading it has conversions from mono to stereo, i think you can find it somewhere iirc

Crepestrom:
i think it could be useful to have stereo sounds for horror mods
but its probably fine

PR #3627, opened 2026-09-21
New slate texture

careeoki:
**Why change it?**

I've been wanting to update this texture for a while. Being 2 years old, it is falling behind in terms of quality compared to other textures. I have attempted changing it a few times already, but haven't liked the results until this one.
 The main problems with the old texture are:
- it is quite dense with detail, and since you very often see this block in large quantities, it can get noisy looking.
- colors are distributed very evenly, which makes it look quite flat. at a distance it turns into basically a uniform grey. 
- to the average person it can look very similar to m*necraft's stone texture (not that the new one is perfect in that regard, but it at least is better)

I totally understand if you feel it does not look enough like slate. However I have already tried to make more slate-like versions, but never liked the results. It runs into similar problems as above, and often just does not tile well and is too repetitive. 
I think it is more important to consider the context in appears in ingame (this is easily the most common and plentiful block in the world, the one you see the most) rather than how much it looks like a real-life stone.
Also the original texture predates it being called "slate" so any similarities to slate are coincidental
slate before after-export

anyway essay over. TLDR: the new one is better


SpellDigger:
- what detail? i always seen the texture as basic noise
- new one looks way flatter to me even though ur stating it being quite flat as an issue
- why is minecraft the center of the world that we need to care about some randoms comparing cubyz to it

either way idc about the change but the reasons just look nitpicked to me atp u could have just called it texture update

DellieDelta:
Screenshot_2026-09-22_09-32-50
I preferred the original one, but after seeing it in a build I really like the new texture

Fabrovio:
screen-2026-09-23-15-28-45
please merge

